PR #480, opened 2014-09-15
Feature/java api improvements

Jurrie:
Add support for updateDmxData RPC callback by using a separate Thread to read RPC messages.


peternewman:
@Jurrie did you ever get chance to look into this further?

Jurrie:
@peternewman You're right, I should work this out. I want to use OLA for a project, but I have other hobby-projects piling up :) And without a project to test the OLA-Java-API, it's hard to keep supporting this. So when I pick up my OLA hobby-project again, I'll put some effort in. But for now, I'm afraid not.

PR #588, opened 2015-01-03
Address some more defects.

nomis52:


common/rdm/AckTimerResponder.cpp
```diff
@@ -220,6 +220,7 @@ const RDMResponse *AckTimerResponder::ResponseFromQueuedMessage(
           queued_response->ParamDataSize());
       break;
     case RDMCommand::SET_COMMAND_RESPONSE:
+      // coverity(SWAPPED_ARGUMENTS)
       return new RDMSetResponse(
```
peternewman: I was looking at this the other day. Shouldn't we create an RDMSetResponse constructor that takes a SetRequest and does half the leg work for you?

peternewman:
Fine aside from the comment.

common/rdm/AckTimerResponder.cpp
nomis52: Yeah, that would be reasonable.

peternewman: Well volunteered :)

peternewman: Done! :-1:

PR #589, opened 2015-01-03
Fix some more defects and some minor tidying

peternewman:


plugins/ftdidmx/FtdiDmxPlugin.cpp
```diff
@@ -139,8 +138,9 @@ bool FtdiDmxPlugin::SetDefaultPreferences() {
 int unsigned FtdiDmxPlugin::GetFrequency() {
   unsigned int frequency;
 
-  if (!StringToInt(m_preferences->GetValue(K_FREQUENCY), &frequency))
-    StringToInt(DEFAULT_FREQUENCY, &frequency);
+  if (!StringToInt(m_preferences->GetValue(K_FREQUENCY), &frequency)) {
```
nomis52: I'd invert this, I think it's easier to read :+1: 

if (StringToInt(m_preferences->GetValue(K_FREQUENCY), &frequency)) {
  return frequency;
}
return DEFAULT_FREQUENCY;

plugins/usbpro/UsbSerialPlugin.cpp
```diff
@@ -361,8 +359,9 @@ string UsbSerialPlugin::GetDeviceName(
 unsigned int UsbSerialPlugin::GetProFrameLimit() {
   unsigned int fps_limit;
   if (!StringToInt(m_preferences->GetValue(USB_PRO_FPS_LIMIT_KEY) ,
-                   &fps_limit))
-    StringToInt(DEFAULT_PRO_FPS_LIMIT, &fps_limit);
+                   &fps_limit)) {
```
nomis52: Same here

plugins/ftdidmx/FtdiDmxPlugin.cpp
peternewman: I'm actually thinking I might create a StringToIntOrDefault which wraps it all in one function.

peternewman: Switched.

plugins/usbpro/UsbSerialPlugin.cpp
peternewman: Switched.

PR #590, opened 2015-01-03
Add and use StringToIntOrDefault

peternewman:


nomis52:
LGTM

peternewman:
I'll skip enabling strict for now I think.

plugins/ftdidmx/FtdiDmxPlugin.cpp
```diff
@@ -133,15 +133,12 @@ bool FtdiDmxPlugin::SetDefaultPreferences() {
 
 
 /**
- * Return the frequency as specified in the config file.
+ * @brief Return the frequency as specified in the config file or the default if
+ * invalid
  */
 int unsigned FtdiDmxPlugin::GetFrequency() {
```
nomis52: Does this still need to be a separate method now?

peternewman: No I guess not.

peternewman: Done. Should even provide a minor optimisation.

common/protocol/Makefile.mk
```diff
@@ -21,8 +21,8 @@ common_protocol_libolaproto_la_CXXFLAGS = $(COMMON_CXXFLAGS_ONLY_WARNINGS)
 
 common_libolacommon_la_LIBADD += common/protocol/libolaproto.la
 
-common/protocol/Ola.pb.cc common/protocol/Ola.pb.h: common/protocol/Ola.proto
+common/protocol/Ola.pb.cc common/protocol/Ola.pb.h: common/protocol/Makefile.mk common/protocol/Ola.proto
```
nomis52: I seem to remember removing this because it was causing a lot of things to get rebuilt on each pass through autoconf. That doesn't make sense though as the Makefile.mk should be static.

Can you just confirm?

include/ola/StringUtils.h
```diff
@@ -331,6 +331,78 @@ bool StringToInt(const std::string &value,
...
 
 /**
+ * @brief Convert a string to a unsigned int or return a default if it failed.
+ * @param value the string to convert
+ * @param alternative the default value to return if conversion failed.
+ * @param[in] strict this controls if trailing characters produce an error.
+ * @returns the value if it converted successfully or the default if the string
+ * was not an int or the value was too large / small for the type.
+ */
+unsigned int StringToIntOrDefault(const std::string &value,
```
nomis52: Make this a template function. Then you don't have to repeat it 6 times.

common/protocol/Makefile.mk
peternewman: I started investigating this because include/ola/plugin_id.h kept getting rebuilt while tweaking the doxygen files stuff.

I can confirm it only gets rebuilt if the files listed are touched and not if Makefile is.

include/ola/StringUtils.h
peternewman: Done.

nomis52:
LGTM

PR #591, opened 2015-01-03
Provide a means to build the files normally built during the build, so t...

peternewman:
...hey are present for Doxygen to run against


peternewman:
When this gets merged I need to update the buildbot config too to use it. Unless we want to and can override the dependencies of make doxygen-doc, which I guess other people may not like/want.

peternewman:
PR for the builtbot change: https://github.com/OpenLightingProject/buildbot/pull/27

PR #593, opened 2015-01-06
Build a protoc plugin instead of wrapping protoc

mbeards:


peternewman:
Presumably we can delete protoc/ola-protoc.cpp now? Can you do the honours?

peternewman:
Thanks for the Pull Request, a few minor comments from me; @nomis52 may have more bits.

nomis52:
LGTM

PR #594, opened 2015-01-10
Fix the build on Windows

peternewman:
Unless @lukase or @nightrune think trunk builds (and make check compiles) fine on Windows for them, in which case it may be a local config issue at my end.


common/web/Makefile.mk
```diff
@@ -40,6 +40,11 @@ common_web_libolaweb_la_SOURCES = \
     common/web/SchemaParser.cpp \
     common/web/SchemaParser.h
 
+if USING_WIN32
+#Work around limitations with Windows library linking
+common_web_libolaweb_la_LIBADD = common/libolacommon.la
```
peternewman: Should this actually be here all the time, it is for plugins, although I guess they aren't within it, or is libolaweb supposed to be linked into libolacommon and it's just the Windows limitations that break it?

nomis52: I think this is correct. Windows is a bit funny about libraries from what I can tell and you have to declare the dependencies up front.  Linux allows you to put everything together at the end

nomis52:
LGTM, but this seems to have all the protoc plugin changes as well.

common/web/Makefile.mk
peternewman: Good, it just felt a bit hacky given nothing else seemed to need it.

PR #595, opened 2015-01-11
Clean up Doxygen configuration. 

nomis52:


peternewman:
LGTM and seems like the version hackery should stil work.

PR #614, opened 2015-01-29
Fix finding ola_protoc_plugin when cross-compiling

simark:
It seems like the recent changes concerning ola_protoc did not take into
account cross-compiling.


nomis52:
LGTM

config/ola.m4
```diff
@@ -71,17 +71,17 @@ elif test -n "$1" ; then
...
+  [],[with_ola_protoc_plugin=no])
 
-OLA_PROTOC="\$(PROTOC) --plugin=protoc-gen-cppservice=\$(top_builddir)/protoc/ola_protoc_plugin${EXEEXT}";
 
-if test "$with_ola_protoc" != "no"; then
-  OLA_PROTOC=$with_ola_protoc;
-  echo "set ola_protoc to $with_ola_protoc"
+if test "$with_ola_protoc_plugin" != "no"; then
+  OLA_PROTOC="\$(PROTOC) --plugin=protoc-gen-cppservice=${with_ola_protoc_plugin}${EXEEXT}";
+  echo "set ola_protoc to $with_ola_protoc_plugin"
```
peternewman: The variable here should probably be $OLA_PROTOC, so it shows the full command being used now.

config/ola.m4
```diff
@@ -71,17 +71,17 @@ elif test -n "$1" ; then
...
+    [use the given ola_protoc_plugin instead of building one (useful for cross-compiling)])],
+  [],[with_ola_protoc_plugin=no])
 
-OLA_PROTOC="\$(PROTOC) --plugin=protoc-gen-cppservice=\$(top_builddir)/protoc/ola_protoc_plugin${EXEEXT}";
 
-if test "$with_ola_protoc" != "no"; then
-  OLA_PROTOC=$with_ola_protoc;
-  echo "set ola_protoc to $with_ola_protoc"
+if test "$with_ola_protoc_plugin" != "no"; then
+  OLA_PROTOC="\$(PROTOC) --plugin=protoc-gen-cppservice=${with_ola_protoc_plugin}${EXEEXT}";
```
peternewman: You probably don't want the EXEEXT on the end (although you could work around it), looking at how it worked before and I guess what used to be passed in.

PR #624, opened 2015-02-07
Improve some test coverage

peternewman:


nomis52:
+10%. Impressive

nomis52:
LGTM. What % do we need to get a green badge? I guess we'll find out.

peternewman:
The 10% is only due to excluding loads of files. :) I think green may be 80% possibly.

PR #625, opened 2015-02-08
Remove some duplicate code

peternewman:


peternewman:
Currently building on Windows...

peternewman:
SelectServerTester.exe and TCPConnectorTester.exe fail on Windows, but I don't think that's anything new.

PR #626, opened 2015-02-08
Fix a test failure on Windows

peternewman:


common/io/SelectServerTest.cpp
```diff
@@ -582,7 +582,7 @@ void SelectServerTest::testTimeout() {
       10,
       ola::NewSingleCallback(this, &SelectServerTest::ReentrantTimeout, m_ss));
   m_ss->RegisterSingleTimeout(
-      20,
+      40,
```
nomis52: What's the failure with the original line?

peternewman: It only logs one timeout in m_timeout_counter not 2. I'm assuming the terminate timeout is triggering before the Reentrant one?

nomis52: Do you mind adding some LOGs to see where the delay is. What should happen is this:

T= 0, ReentrantTimeout added, Terminate timeout added
T= 10ms, ReentrantTimeout fires, adds two timeouts
T= 10ms, first of the timeouts fires, increments the counter
T = 15ms, second of the timeouts fires, increments the counter
T = 20ms, Terminate timeout runs

If we're delaying by more than 5ms it's likely that health checked connections in E1.33 won't work correctly so it would be good to know now.

peternewman: With the logging changes below in, and my timeout, it logs the following:
common/io/SelectServerTest.cpp:135: Re-entrant timeout called, adding two single
 increment timeouts
common/io/SelectServerTest.cpp:129: Single increment timeout called
common/io/SelectServerTest.cpp:131: Timeout counter is now 1
common/io/SelectServerTest.cpp:129: Single increment timeout called
common/io/SelectServerTest.cpp:131: Timeout counter is now 2
common/io/SelectServerTest.cpp:124: Terminate called

Which looks like it's working correctly.

With that time knocked down to 35ms I get:
common/io/SelectServerTest.cpp:135: Re-entrant timeout called, adding two single
 increment timeouts
common/io/SelectServerTest.cpp:124: Terminate called
common/io/SelectServerTest.cpp:129: Single increment timeout called
common/io/SelectServerTest.cpp:131: Timeout counter is now 1
common/io/SelectServerTest.cpp:596: Timeout counter is now 1

peternewman:
@lukase feel free to look too if you want.

common/io/SelectServerTest.cpp
nomis52: Are you running in a VM? That may explain it.

peternewman: Nope, just an old XP box. I think @lukase or someone mentioned that has a lower timer resolution than later OSes.

nomis52: Hmm, ok. Can you add a comment that references this PR so we know why it's so high.

peternewman: Done.

peternewman:
I think this is done now @nomis52 . Merge if you're happy.

common/io/SelectServerTest.cpp
```diff
@@ -121,14 +121,19 @@ class SelectServerTest: public CppUnit::TestFixture {
   }
 
   void Terminate() {
+    OLA_WARN << "Terminate called";
```
nomis52: Can you remove all the warnings please.

nomis52:
LG, just remove the extra logging.

common/io/SelectServerTest.cpp
peternewman: I'd left it in deliberately; if the order that executes in is important, don't we want to ensure it's working like that if there's another failure?

peternewman: Done.

PR #628, opened 2015-02-10
Tidying and Coverity tweaks

peternewman:


peternewman:
As Coverity doesn't run on PRs, we need to check the Travis build that runs just after we've merged to confirm it actually worked.

PR #629, opened 2015-02-11
Fix the Ja Rule Makefile

nomis52:


.travis-ci.sh
```diff
@@ -53,5 +53,5 @@ elif [[ $TASK = 'coverity' ]]; then
   fi;
 else
   # Otherwise compile and check as normal
-  autoreconf -i && ./configure && make distcheck
+  autoreconf -i && ./configure --enable-ja-rule && make distcheck
```
peternewman: And coverity, coveralls, doxygen etc.

peternewman:
You also don't currently have any safety mechanism to not build ja-rule if libusb isn't present.

PR #615, opened 2015-01-30
Web rework

daveol:
don't merge yet. The web interface is incomplete but i wanted to see what you had to say about it and what you think

PR #630, opened 2015-02-11
Update rdm_patcher.js to use correct "identify" command

Goober68:
The identify command string is "identify_device" not "identify".  This fix will allow identify checkbox to work in the patcher popup for a particular device in the patch panel

### DIFF
--- a/javascript/ola/full/rdm_patcher.js
+++ b/javascript/ola/full/rdm_patcher.js
@@ -971,7 +971,7 @@ ola.RDMPatcher.prototype.toggleIdentify_ = function(e) {
   server.rdmSetSectionInfo(
       this.universe_id,
       this.active_device.uid,
-      'identify',
+      'identify_device',
       '',
       'bool=' + (e.target.isChecked() ? '1' : '0'),
       function(e) { patcher.identifyComplete_(e); });


peternewman:
Thanks and good spot @Goober68 it looks like I broke this some time ago here: https://github.com/OpenLightingProject/ola/commit/c753949d8b51e3959a60d0572966622130ac7e09 .

Can you regenerate ola.js and mobile.js and add them to the PR too (instructions at https://github.com/OpenLightingProject/ola/blob/master/javascript/README ), otherwise your fix doesn't actually appear in the UI for other people.

If you're interested in the web stuff, you may want to help daveol who has a PR open here https://github.com/OpenLightingProject/ola/pull/615 to update to Angular. When the initial PR is merged in, there will be lots of seperate sections of RDM support people can work on.

Goober68:
I tried those instructions, but closure builder has moved, I’ll poke around and find it.

peternewman:
Ah yes @Goober68 it looks like the repo is now here:
https://github.com/google/closure-library

With the specific file here within it:
https://github.com/google/closure-library/blob/master/closure/bin/build/closurebuilder.py

Would you mind updating that README too please and adding it to your PR.

peternewman:
How are you getting on @Goober68 ? We're planning for a release this weekend and we'd really love to get this fix in.

Goober68:
I’ll make time for it tonight then, would be great to have this in.

 

Separate issue: Do you know of a way to test/develop the GUI without having to run it through closure?  I’d like to be able to extend some of the features of the OLA gui, but it’s a pain debugging through the giant compiled ola.js.  I searched for info on doing this on wiki but couldn’t find anything.  Seems like it might be possible to move the original .js files into olad/www, overwrite ola.js and it might just work – what do you guys do when you’re developing?

 

-N

nomis52:
You can use:

trunk/closure/bin/build/closurebuilder.py --root=trunk/  --root=ola --namespace="ola.Setup" --output_mode=compiled  --compiler_jar=compiler.jar  > ../olad/www/ola.js

to avoid mini-flying the JS. That's what I used when I want to debug it.

I wouldn't spend too much time on this version of the UI (besides getting the fix in). As Peter mentioned, @daveol  is rewriting it in Bootstrap / Angular so it should be much easier to work with (and look nicer).

Simon

Goober68:
After running closurebuilder with latest sources, I get these errors in browser console:

Uncaught AssertionError: Assertion failed: Event target is not initialized. Did you call the superclass (goog.events.EventTarget) constructor?

ola.js:1310 Uncaught TypeError: Cannot set property 'innerHTML' of null

 

I made sure I had latest sources from git.  Any ideas?

 

-N

peternewman:
Are you not getting those errors before you rebuilt @Goober68 ? I guess it may be a bug that existed before and the new version shows up, or some quirk we avoided before. I'll try and do some tests with my older version of Closure tomorrow and see how I get on.

The AngularJS version also has the ability to generate a source map to make debugging easier.

Goober68:
No errors before rebuild, I had installed from apt-get on Ubuntu, and from git source and they both worked fine, but I don’t think the standard install/make/ldconfig does anything with ola.js.  Before I tried to closure-compile I think I was using the prebuilt ola.js that is in git.

 

-N

nomis52:
I get the same problem with the HEAD of closure and an older compiler.jar . 

At this point I'm inclined to say submit as is and then I'll regenerate the compiled js on my system.

This makes the new UI even more important. It  sounds like the old one is on it's last legs.

peternewman:
@Goober68 you're right, make/install doesn't touch ola.js, so we don't force all users to have those dependancies too.

@nomis52 would your issue not be a mismatch of closure and compiler versions?

So which browsers and versions are people seeing this in, and doing anything in particular or just browsing to the homepage?

nomis52:
The compiler and the library are (or at least are supposed to be) independent. The fact that it doesn't work with either the new or old compiler indicates the library has changed.

The failure is as soon as you load the web page.

Goober68:
I’ve already submitted a PR for the rdm_patcher.js with the identify fix, is there anything else I should do at this point?  Do I have to resubmit the PR?

nomis52:
I've rebuilt the js files and updated the repo. 

Thanks for the fix, please consider helping out with the Angular code as well.

Goober68:
I would love to.  I currently write websites for a living using backbone/marionette, so Angular isn’t that far of a stretch.  How do I get involved?

daveol:
well the new sources are in javascript/new-src. there is a [grunt](http://gruntjs.com/) script in there that is used for minifying the source. the libary's included are: [jQuery](http://jquery.com), [bootstrap](http://getbootstrap.com), [angularjs](https://angularjs.org/) and [angular-router](https://code.angularjs.org/1.2.28/docs/api/ngRoute). If you run into any issues with grunt or you don't want to install it i can minify for you.

daveol:
as for things to do:
-implement rdm functions
-implement keypad
-ux from #635

nomis52:
And there is plenty more after that!  Once the UI has been improved, the rdm.openlighting.org site could do with a refresh as well.

Simon

PR #639, opened 2015-02-17
Fix small ola_protoc configure issues

simark:
This fixes issues raised in #614.

- In the configure output, report the full OLA_PROTOC command used, not
  just the plugin path.
- Do not append EXEEXT at the end of the passed plugin.

For the configure output, I changed "\$(PROTOC)" to "$PROTOC", so it's
replaced immediately and the full path is printed. Normally, PROTOC
should always be defined at this point (it's done higher in the same
file).


nomis52:
LGTM. I'll let Peter comment as well.

peternewman:
Thanks @simark

PR #647, opened 2015-02-25
Ja-rule tweaks

peternewman:


tools/ja-rule/Makefile.mk
```diff
@@ -11,6 +13,6 @@ tools_ja_rule_ja_rule_SOURCES = \
     tools/ja-rule/USBDeviceManager.h \
     tools/ja-rule/ja-rule.cpp
 tools_ja_rule_ja_rule_CXXFLAGS = $(COMMON_CXXFLAGS) $(libusb_CFLAGS)
-tools_ja_rule_ja_rule_LDADD = $(libusb_CFLAGS) \
+tools_ja_rule_ja_rule_LDADD = $(libusb_LIBS) \
```
peternewman: I assume this is correct @nomis52 as it matches what we do in plugins/usbdmx

nomis52: Yup.

peternewman: Ah good.

peternewman:
@nomis52 is the rest okay? I think Travis was only stuck incorrectly on Doxygen, but they're in maintenance now so I can't poke it.

peternewman:
Probably nice to get into https://github.com/OpenLightingProject/ola/milestones/0.9.5 but not essential.

PR #652, opened 2015-02-28
Fix the build on OpenBSD

peternewman:


peternewman:
Another retrospective @nomis52

nomis52:
LGTM

PR #659, opened 2015-03-07
attempt to close #635 & update angular

### LINKED ISSUE #635: UX feed back
Here are my comments after playing with the new UI for a bit. 

- Can you set the page <title> to the instance name & IP address. Doing so helps differentiate different instances of OLA

- It wasn't immediately obvious how to get back to the original page once I was adding a new universe. Can you add a 'System' / 'Home' item to the menu. See http://picpaste.com/UI-mocks-3w9syD1d.png 

- From looking at UX like Gmail & github. They both use two different buttons. Blue / Red / Green buttons are used to draw the user's attention, e.g. "Compose" and "Submit new issue". Grey buttons are used for less frequently used features like "Remove label" and "Fork".

- Following that I'd change the "Shutdown" button on the first page to a grey one. Since I don't expect it to be used often.

- On that note, I'd add a colored "Add new universe" button to the front page. So a user arriving at the UI for the first time has a hint of what to do to get started. See http://picpaste.com/UI-mocks-3w9syD1d.png

- When the shutdown button is pressed. Add a note that OLA may not restart automatically. This is important since the user may not have ssh access and may have to power cycle the host if they shut down olad.

- On the plugins page, it's not immediately clear that the plugin names / rows are clickable. It may be worth adding an "Info icon" (I) to hint to the user.

- On the plugins page, can you add a 'reload' plugins button at the top. It should be grey not colored.

- I'd make the universe overview page what you get if you click directly on the 'Universes' link at the top of the page, and remove it from the drop down menu

- I'd also move 'Add Universe' to the top of the drop down. 

That's it for now. I'll go over the universe settings one next.

daveol:
- attempts to close #635
- updates angular from 1.2.27 to 1.3.14
- make all the grunt tasks work


javascript/new-src/src/js/app.js
```diff
@@ -328,14 +328,15 @@ angular
     if (typeof $scope.Data.id === 'number' &&
         $scope.Data.add_ports !== '' &&
         $scope.Universes.indexOf($scope.Data.id) === -1) {
-     if ($scope.Data.name === '') {
+     if ($scope.Data.name === '' || $scope.Data.name === undefined) {
```
peternewman: These ought to be the other way round I'd say, otherwise you're doing a string comparison on a null.

javascript/new-src/src/js/app.js
```diff
@@ -328,14 +328,15 @@ angular
...
       $scope.Data.name = 'Universe ' + $scope.Data.id;
      }
      $ola.post.AddUniverse($scope.Data);
      $location.path('/universe/' + $scope.Data.id);
     } else if ($scope.Universes.indexOf($scope.Data.id) !== -1) {
      $ola.error.modal('Universe Id already exists.');
-    } else if ($scope.Data.add_ports === '') {
+    } else if ($scope.Data.add_ports === '' ||
+               $scope.Data.add_ports === undefined) {
```
peternewman: Likewise.

peternewman:
LGTM when the comment is addressed.

javascript/new-src/Gruntfile.js
```diff
@@ -57,7 +55,7 @@ module.exports = function (grunt) {
  grunt.loadNpmTasks('grunt-contrib-watch');
  grunt.loadNpmTasks('grunt-contrib-cssmin');
  grunt.loadNpmTasks('grunt-bower-task');
- grunt.file.setBase('../../');
+ //grunt.file.setBase('../../');
```
peternewman: Do you want to delete this if its not required anymore?

olad/www/new/views/universe-add.html
```diff
@@ -5,7 +5,7 @@
  </div>
  <div class="form-group">
   <label for="UniverseName" class="control-label">Name:</label>
-  <input placeholder="downstage-spots" required id="UniverseName" class="form-control" type="text" ng-model="Data.name"/>
+  <input placeholder="dimmer" required id="UniverseName" class="form-control" type="text" ng-model="Data.name"/>
```
peternewman: Can we call it dimmers as the example text please, rather than the singular form.

peternewman:
Just a few comments.

javascript/new-src/Gruntfile.js
daveol: sure

olad/www/new/views/universe-add.html
daveol: Don't know what i was thinking here. A full universe for a singular dimmer?

peternewman:
LGTM

PR #662, opened 2015-03-09
Python client: Invoke callbacks if RPC call fails

simark:
Currently, for most RPC calls, the callback is not invoked if the
call fails. This is a problem, because the client should know that
its call failed, so it can do something about it.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,7 +4,7 @@ x/y/2015 ola-0.9.6
  * Default to the async libusb implementation in the usbdmx plugin.
 
  API:
- * 
+ * Python: callbacks are now also systematically called when the RPC calls fail.
 
  RDM Tests:
  * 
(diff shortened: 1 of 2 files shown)


peternewman:
Looks good. @nomis52

nomis52:
LGTM, Please update the NEWS file to document the change.

simark:
Done.

NEWS
```diff
@@ -4,7 +4,7 @@ x/y/2015 ola-0.9.6
  * Default to the async libusb implementation in the usbdmx plugin.
 
  API:
- * 
+ * Python: callbacks are now systematically called when the RPC calls fail.
```
peternewman: Perhaps "now also systematically".

peternewman:
Can you resync so GitHub lets us merge automatically please.

simark:
Ok done (I think I had made a git fart).

PR #663, opened 2015-03-09
Python 3 compatibility

simark:
Here are a few patches improving the compatibility with Python 3.

The sad thing is that protobuf does not support Python 3 yet. Hopefully it will come soon. We can at least make OLA ready for when that happens. In the mean time, I am testing/using protobuf from this PR:
https://github.com/google/protobuf/pull/166

One of the patches (Python client: Replace __cmp__ with __lt__/__eq__) requires Python 2.7, but if we want to support 2.6 it would be possible to do things differently (but not as nicely).

I couldn't test everything. Hopefully the testsuite will validate a bit that I did not break Python 2, although there doesn't seem to be much testing of the Python OlaClient.

### DIFF
--- a/python/examples/ola_artnet_params.py
+++ b/python/examples/ola_artnet_params.py
@@ -26,9 +26,9 @@ from ola import ArtNetConfigMessages_pb2
 def ArtNetConfigureReply(state, response):
   reply = ArtNetConfigMessages_pb2.Reply()
   reply.ParseFromString(response)
-  print 'Short Name: %s' % reply.options.short_name
-  print 'Long Name: %s' % reply.options.long_name
-  print 'Subnet: %d' % reply.options.subnet
+  print('Short Name: %s' % reply.options.short_name)
+  print('Long Name: %s' % reply.options.long_name)
+  print('Subnet: %d' % reply.options.subnet)
   wrapper.Stop()
 
 
--- a/python/ola/OlaClient.py
+++ b/python/ola/OlaClient.py
@@ -23,6 +23,7 @@ import array
 import logging
 import socket
 import struct
+from functools import total_ordering
 from ola.rpc.StreamRpcChannel import StreamRpcChannel
 from ola.rpc.SimpleRpcController import SimpleRpcController
 from ola import Ola_pb2
@@ -41,7 +42,7 @@ class Error(Exception):
 class OLADNotRunningException(Error):
   """Thrown if we try to connect and olad isn't running."""
 
-
+@total_ordering
 class Plugin(object):
   """Represents a plugin.
 
@@ -73,9 +74,6 @@ class Plugin(object):
   def enabled(self):
     return self._enabled
 
-  def __cmp__(self, other):
-    return cmp(self._id, other._id)
-
   def __repr__(self):
     s = 'Plugin(id={id}, name="{name}", active={active}, enabled={enabled})'
     return s.format(id=self.id,
@@ -83,12 +81,19 @@ class Plugin(object):
                     active=self.active,
                     enabled=self.enabled)
 
+  def __lt__(self, other):
+      return self._id < other._id
+
+  def __eq__(self, other):
+      return self._id == other._id
+
 
 # Populate the Plugin class attributes from the protobuf
 for value in Ola_pb2._PLUGINIDS.values:
   setattr(Plugin, value.name, value.number)
 
 
+@total_ordering
 class Device(object):
   """Represents a device.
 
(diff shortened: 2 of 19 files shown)


python/examples/ola_artnet_params.py
```diff
@@ -26,9 +26,9 @@
 def ArtNetConfigureReply(state, response):
   reply = ArtNetConfigMessages_pb2.Reply()
   reply.ParseFromString(response)
-  print 'Short Name: %s' % reply.options.short_name
-  print 'Long Name: %s' % reply.options.long_name
-  print 'Subnet: %d' % reply.options.subnet
+  print(('Short Name: %s' % reply.options.short_name))
```
nomis52: Why the double ( ?

python/ola/PidStore.py
```diff
@@ -120,8 +122,8 @@ def GetRequest(self, command_class):
     return self._requests.get(command_class)
 
   def GetRequestField(self, command_class, field_name):
-    fields = filter(lambda field: field.name == field_name,
```
nomis52: Was filter removed?

nomis52:
A couple of questions, mostly from my lack of familiarity of Python 3

python/examples/ola_artnet_params.py
simark: As mentioned in the corresponding commit message, this was done automatically using the 2to3 tool. Going through the whole patch to remove the superfluous parentheses would be time consuming, but could be done. (any idea on how to do it programmatically?)

I think the tool is being overly safe, adding parentheses to wrap the % operation and some for the print.

nomis52: Try this:

sed -i '' "s/print((\(.*\)))/print(\1)/" *.py

You may have to fix the multi-line ones manually.

python/ola/PidStore.py
simark: The filter is now as part of the list comprehension, the ```if f.name == field_name``` part.

The "why" of that change is explained in the commit message.

Note that I split the line because it was a bit long, but it could be written directly as:
```python
fields = [f for f in self.GetRequest(command_class).GetAtoms() if f.name == field_name]
```

nomis52:
Also Proto3 (https://github.com/google/protobuf/releases ) supports Python3. We haven't updated though.

python/examples/ola_artnet_params.py
simark: Good idea, I'll try something like that.

simark:
I might be missing something, but Python 3 can't work since they still have occurences of the old-style exception syntax, which doesn't work with Python 3. At least, when I try to import the module in Python 3.4, it doesn't accept it. But indeed, they claimed Python 3 support back in 2.6.0, so you see me confused.

Example: https://github.com/google/protobuf/blob/master/python/google/protobuf/text_format.py#L583

simark:
All right, extra parentheses taken care of.

python/ola/OlaClient.py
```diff
@@ -23,6 +23,7 @@
 import logging
 import socket
 import struct
+from functools import total_ordering
```
nomis52: This is breaking 2.6

nomis52:
So it looks like 2.6 is still working apart from the "from functools import total_ordering".

I think it's worth removing this bit, announcing that we're planning to deprecate support for 2.6 and then adding it back in say 3-6 months.

Simon

simark:
A simple way to make it 2.6 compatible would be to copy the `total_ordering` function from Python directly.

https://hg.python.org/cpython/file/2.7/Lib/functools.py#l53

It could be easily replaced by the appropriate import later. The only thing I am not sure would be the IP-cleanliness. I believe that it would be permitted by the Python License [1], which is GPL-compatible, but we would have to make it clear that this function is under a different licence, ship the Python Licence text in the tarball, etc. It seems more complicated than defining the damn operators by hand.

[1] https://www.python.org/download/releases/2.7/license/

nomis52:
Yeah, I'd rather not have this for 3-6 months rather than deal with IP headaches.

simark:
Ok, I added the missing operators by hand. I rebased the series to current master, since it was conflicting.

nomis52:
LGTM.

FYI: If you find yourself with some spare time, improving the tests for the Python client would be really good. As you can see it doesn't have much test coverage right now.

simark:
This is on my TODO list. Both unit tests and tests with olad running (not sure what to call that, functional tests?).

You can have a glimpse here although it's still very much in draft form.
https://github.com/simark/ola/commit/b441d6b30ca9718628a9f5e02602773124d95cd0

nomis52:
Awesome, thanks for the contributions!

hunternet93:
With a bit of work, you can get OLA to compile for Python 3.

First, install the latest version of Protobuf by following the instructions at https://github.com/google/protobuf. When installing the Python module, replace lines 55-56 of setup.py with:

          d = {}
          exec(version_file.read())
          return __version__

Then, as root, run:

    $ python3 setup.py install

Once protobuf is installed, get the latest version of OLA and install like normal, except pass the following arguments to configure:

    PYTHON="/usr/bin/python3" ./configure --enable-python-libs

Then finish the build and install normally.

PR #664, opened 2015-03-09
Python client: Add __str__ to Port, Universe and Device

simark:
This is useful for logging and debugging.

If you like the format, I could change the other objects that already
have a __str__ so the format is consistent.


peternewman:
Again looks fine, @nomis52 ?

nomis52:
LGTM. I'm fine if you change the other __str__ methods.

simark:
Updated. I made minor edits to the existing hunks (code style).

simark:
Updated, I switched from __str__ to __repr__ for reasons outlined in the commit message.

PR #665, opened 2015-03-09
Python client: check if universe field is present when deserializing a Port

simark:
Right now, there is no way to distinguish a Port bound to universe 0
from a Port not bound to any unverse (its universe field will also be
0). This is because protobuf returns 0 by default if you ask for an
optional field that happens not to be present.

With this patch, an unbound Port will have universe=None.


peternewman:
LGTM and answers a question I was pondering about protobuf optional fields. I'll let @nomis52 take a look too.

peternewman:
@simark we should probably update the bugs section of the NEWS file to note this too.

PR #675, opened 2015-03-15
Expose the plugin enabled/disabled state via /json/universe_plugin_list, closes #617

### LINKED ISSUE #617: Export a list of plugins, the enabled & active state via the HTTP API
Dave would like this for the new web API.

peternewman:


common/protocol/Ola.proto
```diff
@@ -166,6 +166,7 @@ message PluginInfo {
   required int32 plugin_id = 1;
   required string name = 2;
   required bool active = 3;
+  optional bool enabled = 4;
```
peternewman: Hopefully this should make it backwards compatible?

olad/PluginManager.cpp
```diff
@@ -73,15 +72,14 @@ void PluginManager::LoadAll() {
...
     }
   }
 
   // The second pass checks for conflicts and starts each plugin
-  vector<AbstractPlugin*>::iterator plugin_iter = enabled_plugins.begin();
-  for (; plugin_iter != enabled_plugins.end(); ++plugin_iter) {
-    AbstractPlugin *plugin = *plugin_iter;
+  PluginMap::iterator plugin_iter = m_enabled_plugins.begin();
+  for (; plugin_iter != m_enabled_plugins.end(); ++plugin_iter) {
+    AbstractPlugin *plugin = &(*plugin_iter->second);
```
peternewman: I don't know if you want AbstractPlugin* changed to a straight AbstractPlugin and the necessary changes in the rest of this code?

peternewman:
I don't know if you'd rather if this was exposed via a new URL rather than expanding /json/universe_plugin_list ?

peternewman:
@daveol hopefully this does what you want?

daveol:
LGTM

common/protocol/Ola.proto
nomis52: Correct.

include/ola/client/ClientTypes.h
```diff
@@ -42,10 +42,14 @@ namespace client {
  */
 class OlaPlugin {
  public:
-  OlaPlugin(unsigned int id, const std::string &name, bool active)
+  OlaPlugin(unsigned int id,
+            const std::string &name,
+            bool active,
+            bool enabled = true)
```
nomis52: I think you can avoid the default here, since the only thing that should be creating plugin objects is the client itself.

python/ola/OlaClient.py
```diff
@@ -48,10 +48,14 @@ class Plugin(object):
   Attributes:
     id: the id of this plugin
     name: the name of this plugin
+    active: whether this plugin is active
+    enabled: whether this plugin is enabled
   """
-  def __init__(self, plugin_id, name):
+  def __init__(self, plugin_id, name, active=True, enabled=True):
```
nomis52: Same comment here, no need for defaults.

olad/PluginManager.cpp
nomis52: This is fine.

nomis52:
LGTM

nomis52:
Oh just note this doesn't fully close #617 . We should also add a method of changing the state of a plugin.

common/protocol/Ola.proto
peternewman: Great.

olad/PluginManager.cpp
peternewman: Okay.

include/ola/client/ClientTypes.h
peternewman: Fixed. I wasn't sure which bits needed to be backwards compatible like that.

python/ola/OlaClient.py
peternewman: Fixed.

peternewman:
Yes, I saw the other comments in #617 after I'd named the commit. Did you see the comment about URLs, you're happy to just extend /json/universe_plugin_list rather than add a new URL and hope everyone is parsing the JSON properly? TBH I'm not really sure what the universe part of that URL is for or does the old UI use both bits for one page?

nomis52:
Yup, this is fine. I'm not quite sure what you mean by "parsing the JSON properly" . Adding new properties to the JSON is a non-breaking API change.

The ola UI fetches both universes and plugins in a single call and uses the result to populate the left hand side menu.

If you want to clean up the API, it would be good to separate these out into separate calls.

nomis52:
LGTM anyway.

peternewman:
You've clearly only seen nice code then @nomis52 . If the JSON is parsed with some hacky regex for example, this will probably break it. Although perhaps thats a good thing to encourage them to do it properly.

I'll merge this bit in as is, but we can always do a seperate URL for just plugins in future.

nomis52:
Er sure, but we have to standardize on something. If a client doesn't handle JSON property then it's outside the spec and deserves to break.

PR #676, opened 2015-03-18
use change made in #675

daveol:
the plugin page now uses the values given in /json/universe_plugin_list instead of polling the api for each plugin.


daveol:
whoops wrong branch copied

Issue #679, opened 2015-03-19
Minor fixes for the new UI


nomis52:
On the Faders page:

- Can you add a checkbox 'activate console' that is unchecked by default. The problem we have right now for both the old and new UI is that switching to the faders tab generates a universe of 0s, which will cause flickering if the universe is set to LTP merge mode. Forcing the user to take an explicit 'activate' action would avoid this.
- When I click the forward / back buttons in the console my browser highlights them. I think they are getting focus. See screenshot.
-  I think making the blue band more of a toolbar and making the buttons in grey would help stand out. 
- If we had such a toolbar, it would be nice to have a numeric input as well. e.g. Channel X and Y . 

On the Universe settings page:
 - If the only option is 'static', I think it should be grey text, rather than something that looks like a text input field
 - Right now there is no feedback if a port patch fails, for example, trying to patch an E1.33 port to universe 0 silently fails. The CLI suffers from the same problem. We should return an error message back in the RPC and display this in the UI.
![screen shot 2015-03-14 at 2 06 40 pm](https://cloud.githubusercontent.com/assets/4355332/6739677/9ba9f290-ce50-11e4-8708-fb6aabc24312.png)

daveol:
> - Can you add a checkbox 'activate console' that is unchecked by default. The problem we have right now for both the old and new UI is that switching to the faders tab generates a universe of 0s, which will cause flickering if the universe is set to LTP merge mode. Forcing the user to take an explicit 'activate' action would avoid this.

Actually the new ui only sends data if you click one of the faders and it cuts of any zero's that are not needed so I'm not sure what you are talking about. if you are talking about the zero's you can see i generate those because I first fill the array to the limit of what the api gives me  and the rest are all zero's to make the faders not freak out and say NaN.

> - I think making the blue band more of a toolbar and making the buttons in grey would help stand out.

I will play around with this and look at what is best.

> - If we had such a toolbar, it would be nice to have a numeric input as well. e.g. Channel X and Y .

For the pages? It's not really clear what you mean.

nomis52:
Ah ok, maybe this isn't a problem with the new console.

I've attached a mock up of what I was suggesting:

![screen shot 2015-03-22 at 10 27 48 am](https://cloud.githubusercontent.com/assets/4355332/6769790/25c83ec0-d07f-11e4-8849-67b5bed9e89d.png)

daveol:
okay, I will comment on each of the points of the original list to get the status of the issue:

[...]

PR #685, opened 2015-03-26
resolve some of #679, some comments and add more files to .gitignore

daveol:
resolve some of #679, some comments and add more files to .gitignore


javascript/new-src/README.md
```diff
@@ -18,7 +18,7 @@ sudo npm install --global grunt-cli
 to give you a nicer way to run the grunt task runner from the command-line
 
 ```bash
-grunt build --force
+grunt build
 ```
 grunt is not happy with moving outside of its parent directory but it works with --force
```
nomis52: Can you remove this comment now?

PR #686, opened 2015-03-28
Use std::string::c_str instead of ::data where appropriate

simark:
I was looking for something else and this caught my attention.

std::string::data does not guarantee to return a null-terminated
string. That's what std::string::c_str does. Even though they will
likely always return the same thing, it would be safer to use the right
one.


nomis52:
Good catch.

PR #687, opened 2015-03-29
Add patched ports in Universe protobuf

simark:
This series in to add details about the patched ports in the Universe protobuf. Currently, it contains only the number of ports, which is not that useful.

### DIFF
--- a/common/protocol/Ola.proto
+++ b/common/protocol/Ola.proto
@@ -266,6 +266,8 @@ message UniverseInfo {
   required int32 input_port_count = 4;
   required int32 output_port_count = 5;
   required int32 rdm_devices = 6;
+  repeated PortInfo input_ports = 7;
+  repeated PortInfo output_ports = 8;
 }
 
 message UniverseInfoReply {
--- a/include/ola/client/ClientTypes.h
+++ b/include/ola/client/ClientTypes.h
@@ -29,6 +29,8 @@
 #include <string>
 #include <vector>
 
+#include "common/protocol/Ola.pb.h"
+
 /**
  * @file
  * @brief Types used as return values from the OLA Client.
@@ -77,6 +79,8 @@ class OlaPlugin {
   bool operator<(const OlaPlugin &other) const {
     return m_id < other.m_id;
   }
+
+  static OlaPlugin FromProtobuf(ola::proto::PluginInfo plugin_info);
 
  private:
   unsigned int m_id;  // id of this plugin
@@ -191,6 +195,8 @@ class OlaInputPort: public OlaPort {
       OlaPort(port_id, universe, active, description,
               capability, mode, priority, supports_rdm) {
   }
+
+  static OlaInputPort FromProtobuf(ola::proto::PortInfo port_info);
 };
 
 /**
@@ -209,6 +215,8 @@ class OlaOutputPort: public OlaPort {
       OlaPort(port_id, universe, active, description,
               capability, mode, priority, supports_rdm) {
   }
+
+  static OlaOutputPort FromProtobuf(ola::proto::PortInfo port_info);
 };
 
 
@@ -246,6 +254,8 @@ class OlaDevice {
   bool operator<(const OlaDevice &other) const {
     return m_alias < other.m_alias;
   }
+
+  static OlaDevice FromProtobuf(ola::proto::DeviceInfo device_info);
 
  private:
   std::string m_id;            // device id
@@ -287,6 +297,8 @@ class OlaUniverse {
   unsigned int InputPortCount() const { return m_input_port_count; }
   unsigned int OutputPortCount() const { return m_output_port_count; }
   unsigned int RDMDeviceCount() const { return m_rdm_device_count; }
+
+  static OlaUniverse FromProtobuf(ola::proto::UniverseInfo universe_info);
 
  private:
   unsigned int m_id;
--- a/include/ola/client/ClientTypes.h
+++ b/include/ola/client/ClientTypes.h
@@ -29,8 +29,6 @@
 #include <string>
 #include <vector>
 
-#include "common/protocol/Ola.pb.h"
-
 /**
  * @file
  * @brief Types used as return values from the OLA Client.
@@ -79,8 +77,6 @@ class OlaPlugin {
   bool operator<(const OlaPlugin &other) const {
     return m_id < other.m_id;
   }
-
-  static OlaPlugin FromProtobuf(ola::proto::PluginInfo plugin_info);
 
  private:
   unsigned int m_id;  // id of this plugin
@@ -195,8 +191,6 @@ class OlaInputPort: public OlaPort {
       OlaPort(port_id, universe, active, description,
               capability, mode, priority, supports_rdm) {
   }
-
-  static OlaInputPort FromProtobuf(ola::proto::PortInfo port_info);
 };
 
 /**
@@ -215,8 +209,6 @@ class OlaOutputPort: public OlaPort {
       OlaPort(port_id, universe, active, description,
               capability, mode, priority, supports_rdm) {
   }
-
-  static OlaOutputPort FromProtobuf(ola::proto::PortInfo port_info);
 };
 
 
@@ -255,8 +247,6 @@ class OlaDevice {
     return m_alias < other.m_alias;
   }
 
-  static OlaDevice FromProtobuf(ola::proto::DeviceInfo device_info);
-
  private:
   std::string m_id;            // device id
   unsigned int m_alias;   // device alias
@@ -280,8 +270,8 @@ class OlaUniverse {
   OlaUniverse(unsigned int id,
               merge_mode m,
               const std::string &name,
-              std::vector<OlaInputPort> input_ports,
-              std::vector<OlaOutputPort> output_ports,
+              const std::vector<OlaInputPort> &input_ports,
+              const std::vector<OlaOutputPort> &output_ports,
               unsigned int rdm_device_count):
     m_id(id),
     m_merge_mode(m),
@@ -304,8 +294,6 @@ class OlaUniverse {
   const std::vector<OlaOutputPort> &OutputPorts() const {
     return m_output_ports;
   }
-
-  static OlaUniverse FromProtobuf(ola::proto::UniverseInfo universe_info);
 
  private:
   unsigned int m_id;
--- a/include/ola/client/ClientTypes.h
+++ b/include/ola/client/ClientTypes.h
@@ -280,23 +280,30 @@ class OlaUniverse {
   OlaUniverse(unsigned int id,
               merge_mode m,
               const std::string &name,
-              unsigned int input_port_count,
-              unsigned int output_port_count,
+              std::vector<OlaInputPort> input_ports,
+              std::vector<OlaOutputPort> output_ports,
               unsigned int rdm_device_count):
     m_id(id),
     m_merge_mode(m),
     m_name(name),
-    m_input_port_count(input_port_count),
-    m_output_port_count(output_port_count),
+    m_input_ports(input_ports),
+    m_output_ports(output_ports),
     m_rdm_device_count(rdm_device_count) {}
   ~OlaUniverse() {}
 
   unsigned int Id() const { return m_id;}
   merge_mode MergeMode() const { return m_merge_mode; }
   const std::string& Name() const { return m_name;}
-  unsigned int InputPortCount() const { return m_input_port_count; }
-  unsigned int OutputPortCount() const { return m_output_port_count; }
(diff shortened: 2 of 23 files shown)


nomis52:
Overall looks good. Do you mind updating the C++ client as well (in /ola ).
There are a couple of lint and doxygen warnings that are preventing the checks from passing.

simark:
Any tips on running cpplint.py? When running this, I don't get the outputs seen in Travis:

    ./cpplint.py --filter=-legal/copyright,-readability/streams,-runtime/arrays ./olad/OlaServerServiceImpl.cpp

simark:
Is it fine if I do a change similar to what I did in the Python client, adding some FromProtobuf static methods to the C++ client data classes (OlaUniverse, OlaDevice, OlaPort, OlaInputPort, OlaOutputPort and OlaPlugin)? The decoding of the protobuf to a OlaUniverse is currently done at two places, but I'd change the others as well to keep everything consistent.

Or do you see a better way to do it?

simark:
The C++ client part builds, but it would need some more testing.

nomis52:
Make sure you have the newest version of cpplint. The travis runs pull from head.

include/ola/client/ClientTypes.h
```diff
@@ -29,6 +29,8 @@
 #include <string>
 #include <vector>
 
+#include "common/protocol/Ola.pb.h"
```
nomis52: Please don't include this here as it exposes the internal rpc messages to the callers.

I'm fine if you want to make a

ObjectFactory {
  static OlaPlugin FromProtobuf(ola::proto::PluginInfo plugin_info);
  static OlaInputPort FromProtobuf(ola::proto::PortInfo port_info);
  etc. 
}

and keep ObjectFactory.h out of the install path.

include/ola/client/ClientTypes.h
```diff
@@ -270,30 +280,39 @@ class OlaUniverse {
   OlaUniverse(unsigned int id,
               merge_mode m,
               const std::string &name,
-              unsigned int input_port_count,
-              unsigned int output_port_count,
+              std::vector<OlaInputPort> input_ports,
```
nomis52: Pass vectors as const references, that avoids the copy. e.g.

const std::vector<OlaInputPort> &input_ports.

You could even pass in pointers and have the object take ownership. That would avoid a second copy:

std::auto_ptr<std::vector<OlaInputPort>> m_input_ports;

OlaUniverse(unsigned int id,
                     const std::vector<OlaInputPort> *input_ports)
    : m_input_ports(input_ports) {
}

include/ola/client/ClientTypes.h
```diff
@@ -270,30 +280,39 @@ class OlaUniverse {
...
     m_rdm_device_count(rdm_device_count) {}
   ~OlaUniverse() {}
 
   unsigned int Id() const { return m_id;}
   merge_mode MergeMode() const { return m_merge_mode; }
   const std::string& Name() const { return m_name;}
-  unsigned int InputPortCount() const { return m_input_port_count; }
-  unsigned int OutputPortCount() const { return m_output_port_count; }
+  unsigned int InputPortCount() const { return m_input_ports.size(); }
+  unsigned int OutputPortCount() const { return m_output_ports.size(); }
```
nomis52: With auto_ptr, this would then become m_input_ports->size();

include/ola/client/ClientTypes.h
```diff
@@ -270,30 +280,39 @@ class OlaUniverse {
...
   unsigned int RDMDeviceCount() const { return m_rdm_device_count; }
 
+  const std::vector<OlaInputPort> &InputPorts() const {
+    return m_input_ports;
```
nomis52: And:

return *m_input_ports;

ola/ClientTypes.cpp
```diff
@@ -0,0 +1,125 @@
...
+/*
+ * Create an OlaUniverse object from a protobuf.
+ */
+OlaUniverse OlaUniverse::FromProtobuf(
+    ola::proto::UniverseInfo universe_info) {
+  OlaUniverse::merge_mode merge_mode =
+    universe_info.merge_mode() == ola::proto::HTP ?
+    OlaUniverse::MERGE_HTP: OlaUniverse::MERGE_LTP;
+
+  vector<OlaInputPort> input_ports;
```
nomis52: If you transfer ownership ,  you can make these pointers:

vector<OlaInputPort> *input_ports = new vector<OlaInputPort>();

nomis52:
Generally looks good, I made some C++ suggestions.

include/ola/client/ClientTypes.h
simark: I am doing this at the moment. The only problem is that

    static OlaInputPort FromProtobuf(ola::proto::PortInfo port_info);

and

    static OlaInputPort FromProtobuf(ola::proto::PortInfo port_info);

can't coexist.

simark: For now I'll do 

    static OlaInputPort FromProtobufIn(ola::proto::PortInfo port_info);
    static OlaOutputPort FromProtobufOut(ola::proto::PortInfo port_info);

simark:
It seems like if I Implement pointer ownership transfer like this, I need to define a copy constructor for OlaUniverse. Probably because it contains an auto_ptr member. When an OlaUniverse is return by value (is it the right term?) by FromProtobuf, its auto_ptr members can't be simply copied, because that would lead to a bad situation where two auto_ptr point the the same object, which would get deallocated twice. I suppose that we need to define an explicit copy constructor that does a copy of the vectors. Does it make sense? Is it worth the extra complexity?

For reference, here is the compilation error I get if I don't define the copy constructor.

    ola/ClientTypesFactory.cpp: In static member function ‘static ola::client::OlaUniverse ola::client::ClientTypesFactory::FromProtobuf(const ola::proto::UniverseInfo&)’:
    ola/ClientTypesFactory.cpp:125:49: error: no matching function for call to ‘ola::client::OlaUniverse::OlaUniverse(ola::client::OlaUniverse)’
                          universe_info.rdm_devices());
                                                     ^
    ola/ClientTypesFactory.cpp:125:49: note: candidates are:
    In file included from ola/ClientTypesFactory.h:25:0,
                     from ola/ClientTypesFactory.cpp:22:
    ./include/ola/client/ClientTypes.h:271:3: note: ola::client::OlaUniverse::OlaUniverse(unsigned int, ola::client::OlaUniverse::merge_mode, const string&, std::vector<ola::client::OlaInputPort>*, std::vector<ola::client::OlaOutputPort>*, unsigned int)
       OlaUniverse(unsigned int id,
       ^
    ./include/ola/client/ClientTypes.h:271:3: note:   candidate expects 6 arguments, 1 provided
    ./include/ola/client/ClientTypes.h:264:7: note: ola::client::OlaUniverse::OlaUniverse(ola::client::OlaUniverse&)
     class OlaUniverse {
           ^
    ./include/ola/client/ClientTypes.h:264:7: note:   no known conversion for argument 1 from ‘ola::client::OlaUniverse’ to ‘ola::client::OlaUniverse&’
    ola/ClientTypesFactory.cpp:126:1: error: control reaches end of non-void function [-Werror=return-type]
     }
     ^

nomis52:
Ah yes, you're correct. The API has the contract that client types are copy-able. Without implementing reference counting we have to incur a copy anyway.

Just go back to what you had, but pass by const reference rather than const pointers. 

Sorry for the extra work.

ola/ClientTypesFactory.cpp
```diff
@@ -0,0 +1,129 @@
...
+  return OlaPlugin(plugin_info.plugin_id(),
+                   plugin_info.name(),
+                   plugin_info.active(),
+                   plugin_info.enabled());
+}
+
+/*
+ * Create an OlaInputPort object from a protobuf.
+ */
+OlaInputPort ClientTypesFactory::FromProtobufIn(
```
nomis52: I'd call these:

PluginFromProto(..)
OutputPortFromProto(...)
DeviceFromProto(...)

etc.

simark:
Ok, I am completely puzzled. In ClientTypesFactory.cpp, I tried:

    #include "ClientTypesFactory.h"
    #include <ola/ClientTypesFactory.h>
    #include "ola/ClientTypesFactory.h"

all of them build on my machine, but don't on Travis. They all give

    ../ola/ClientTypesFactory.cpp:22:10: fatal error: 'ClientTypesFactory.h' file

And it's not that I forgot to add the file to git...

ola/ClientTypesFactory.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ *
+ * ClientTypes.cpp
+ * Copyright (C) 2015 Simon Marchi
+ */
+
+#include <vector>
+
+#include "ClientTypesFactory.h"
```
nomis52: This should be #include "ola/ClientTypesFactory.h"

ola/Makefile.mk
```diff
@@ -18,6 +18,7 @@ lib_LTLIBRARIES += ola/libola.la
 ola_libola_la_SOURCES = \
     ola/AutoStart.cpp \
     ola/ClientRDMAPIShim.cpp \
+    ola/ClientTypesFactory.cpp \
```
nomis52: You need to add the .h file here as well. Otherwise it won't be included in the tarball, which is what travis tries to build.

ola/ClientTypesFactory.h
```diff
@@ -0,0 +1,50 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ *
+ * OlaClientWrapper.h
```
nomis52: Please update the comment & copyright to match the file

nomis52:
Looks good, just one comment about fixing the comment.

simark:
Ahhh, doh. Done.

nomis52:
LGTM.

common/protocol/Ola.proto
```diff
@@ -266,6 +266,8 @@ message UniverseInfo {
   required int32 input_port_count = 4;
   required int32 output_port_count = 5;
   required int32 rdm_devices = 6;
+  repeated PortInfo input_ports = 7;
```
peternewman: Should these not be optional?

nomis52: Ah good catch.

peternewman: Presumably that also means the functions in the code that generate them need to check for Has(field_name) too? @simark are you happy to fix these please?

nomis52: It's a repeated field, so missing is the same as length 0.

simark: You can't have a field that is optional AND repeated. We would have to create a new message type that contains the repeated fields, and add a new optional field of this new type. It seems like unnecessary complexity that we will have to keep in the long term.

As @nomis52 said, a missing field will appear to have the same length as a present field with 0 elements. But since we still have fields with the counts, you can still differentiate between the two situations (if input_port_count != len(inputs_ports), it means you are dealing with an old OLA that doesn't fill input_ports). So, it is really necessary to change the protobuf definition?

nomis52: Yup, sorry, it was right the first time.  No need to change.

nomis52:
Hmm, it looks like this has broken the RDM test server, because it does naughty things like:

    return {
      'universes': [u.__dict__ for u in universes],                                                                                                                                                                                            
      'status': True,
    }

And then tries to use json.dumps on the result.

simark:
When trying to launch rdm_test_server.py from the build directory, I am met with:

```
Traceback (most recent call last):
  File "tools/rdm/rdm_test_server.py", line 43, in <module>
    from ola.testing.rdm.DMXSender import DMXSender
ImportError: No module named testing.rdm.DMXSender
``` 
So I can't really test. Can you point me to my mistake (or some instructions)?

simark:
Hmm ok, it seems I need to add --enable-rdm-tests, make install, and use python packages from the install directory.

simark:
Please take a look at #703.

common/protocol/Ola.proto
peternewman: Cool, just wasn't sure if we'd cause any backwards incompatibility, but it sounds like we're fine.

PR #688, opened 2015-03-30
Clean up the RDM Discovery Agent. 

nomis52:


common/http/HTTPServer.cpp
```diff
@@ -425,7 +425,7 @@ int HTTPResponse::Send() {
  * @param options the configuration options for the server
  */
 HTTPServer::HTTPServer(const HTTPServerOptions &options)
-    : Thread(),
+    : Thread("http"),
```
peternewman: Did this work? If so we should add it back in @nomis52 !

nomis52: This is actually old code. A better version is already in HEAD.

PR #694, opened 2015-04-02
Bump protobuf java dependency

simark:
The current code does not work with 2.4.1. The latest stable version is
2.6.1, and it seems to work with that.


nomis52:
LGTM

PR #697, opened 2015-04-04
Add a watchdog to detect stalls in the Enttec Widget.

nomis52:


common/io/KQueuePoller.cpp
```diff
@@ -392,7 +392,7 @@ bool KQueuePoller::ApplyChange(int fd, int16_t filter, uint16_t flags,
     int r = kevent(m_kqueue_fd, m_change_set, m_next_change_entry, NULL, 0,
                    NULL);
     if (r < 0) {
-      OLA_WARN << "Failed to apply kqueue changes";
+      OLA_WARN << "Failed to apply kqueue changes:" << strerror(errno);
```
peternewman: SPaG, add a space after the colon.

common/utils/Watchdog.cpp
```diff
@@ -0,0 +1,62 @@
...
+  m_enabled = false;
+  m_fired = false;
+}
+
+void Watchdog::Kick() {
+  MutexLocker lock(&m_mu);
+  m_count = 0;
+}
+
+void Watchdog::Clock() {
```
peternewman: Aside from the similarity to Kick, would Tick be a better name than Clock, given it's an action?

include/ola/util/Watchdog.h
```diff
@@ -0,0 +1,113 @@
...
+   * @brief Kick the watchdog to avoid a reset.
+   *
+   * This should be called regularly while the watchdog is enabled.
+   */
+  void Kick();
+
+  /**
+   * @brief Check if the process has stalled due to a lack of Kick() calls.
+   *
+   * This should be called regularly. If it's been
```
peternewman: Missing words.

peternewman:
Just a few comments.

common/utils/Watchdog.cpp
nomis52: I took the names from  https://en.wikipedia.org/wiki/Watchdog_timer

peternewman: I guessed that may be the case, I'd still argue the input line is the clock, but it ticks.

peternewman:
I'd still prefer the function to be renamed (although I can see where you're coming from), but other than that LGTM

PR #698, opened 2015-04-04
More Ja Rule improvements

nomis52:


include/ola/rdm/QueueingRDMController.h
```diff
@@ -84,7 +84,7 @@ class QueueingRDMController: public RDMControllerInterface {
  * The DiscoverableQueueingRDMController also handles discovery, and ensures
  * that only a single discovery or RDM request sequence occurs at once.
  *
- * In this model discovery has a higher precedence that RDM messages.
+ * In this model discovery has a higher precedence than RDM messages.
```
peternewman: SPaG model,

tools/ja-rule/ja-rule-controller.cpp
```diff
@@ -102,7 +121,6 @@ class ControllerImpl : public ola::rdm::DiscoveryTargetInterface,
     RDMCommandSerializer::Pack(*request, data, &rdm_length);
     m_device->SendMessage(OpenLightingDevice::RDM_REQUEST, data, rdm_length);
 
-    // TODO(simon): do we need to check there isn't already one in-flight?
```
peternewman: Where is this being done?

tools/ja-rule/ja-rule-controller.cpp
```diff
@@ -274,12 +298,11 @@ class Controller : public ola::rdm::DiscoverableRDMControllerInterface {
...
                          bool ok, const UIDSet& uids) {
     OLA_DEBUG << "Discovery complete: " << uids;
-    if (ok) {
-      m_uids = uids;
-    }
+    m_uids = uids;
     if (callback) {
       callback->Run(m_uids);
     }
+    (void) ok;
```
peternewman: OLA_UNUSED.

tools/ja-rule/ja-rule-controller.cpp
nomis52: The next step is to wrap this in a QueuingRDMController, which enforces serialization and also handles ACK overflows.

peternewman:
How do you select a UID if you've got more than 26 nomis52? How about also allowing 0-9, or having some more paging option?

tools/ja-rule/ja-rule-controller.cpp
peternewman: Can we leave the TODO in until that's done then?

include/ola/rdm/QueueingRDMController.h
nomis52: Fixed in the next PR

tools/ja-rule/ja-rule-controller.cpp
nomis52: Added

PR #702, opened 2015-04-05
Build Java libs via Travis

peternewman:
Fix the resulting issues that appeared


configure.ac
```diff
@@ -763,6 +763,7 @@ Python API: ${enable_python_libs}
 Java API: ${enable_java_libs}
 Enable HTTP Server: ${have_microhttpd}
 RDM Responder Tests: ${enable_rdm_tests}
+Ja Rule: ${BUILD_JA_RULE}
```
peternewman: I don't think this actually works currently. If we want it to show the final state (including libusb status), I may need to have a rethink.

peternewman:
This now finishes, aside from some leftover files after distclean, but the Java tests don't actually seem to run...

nomis52:
This looks fine to me. Any reason why we shouldn't submit?

peternewman:
Just the rather comical thing that it doesn't run the Java tests:
https://travis-ci.org/OpenLightingProject/ola/jobs/87932624#L1400

peternewman:
Well I think I've answered this question. @nomis52 excluded the only test here https://github.com/OpenLightingProject/ola/commit/cd10e886fd8b7e5b255450b799ef64d73642c7ca#diff-61657881d138c2b7ac1366d88c9ee035R24 which is why it's not running any.

I think this actually makes sense as so far all the Java tests require a running olad to talk to (which we don't normally expect). Indeed Travis confirms this.

So we probably need to add a few tests of the byteString stuff and call it a day for now:
https://github.com/OpenLightingProject/ola/blob/master/java/src/main/java/ola/OlaClient.java#L408

Then this can FINALLY be merged!

peternewman:
This should allow us to run the tests except during CI.

http://stackoverflow.com/questions/8304110/skip-a-submodule-during-a-maven-build

peternewman:
This is ready for review now @nomis52 . I've tested it via a distcheck build and also a normal in path build and they both work and I've added some (fairly trivial) standalone Java tests, which hopefully others can expand on in future.

peternewman:
I spoke slightly too soon @nomis52 . Are you happy with this tweak https://github.com/OpenLightingProject/ola/pull/702/commits/a1236cc8f66d851e2c443b36792342fbcce2fe5d making those functions static given they don't reference the protocol/client at all.

PR #706, opened 2015-04-06
Clean up the Flags during exit so valgrind isn't as noisy.

nomis52:


peternewman:
LGTM given most of it just seems to be whitespace changes.

PR #707, opened 2015-04-09
Enable Dot for Doxygen where present

peternewman:


peternewman:
Retrospective @nomis52 it doesn't seem to break Travis which doesn't have dot installed.

nomis52:
Couple of things:

- We get 3 warnings: https://buildbot.openlighting.org/builders/doxygen-doc-generator/builds/325/steps/make%20doxygen-doc/logs/warnings%20%283%29
- It looks like only protected attributes appear in the diagrams. Do you think if we documented private members it would help people understand how the software is architected?

peternewman:
Warnings fixed, although I don't know how useful the resulting output is :) :
http://docs.openlighting.org/doc/latest/_callback_8h__dep__incl.png

From their example, documenting private will show them, I don't know if it will help understand the architecture, perhaps that's more a @nightrune question.

nightrune:
I really like this diagram, but I'm quite a visual learner. There isn't
much in the diagram you posted that you can't find just looking at the
headers that are included.

What do you mean documenting private?

peternewman:
If someone finds them useful, that's a good enough reason to keep them @nightrune . Regarding the private/protected members, the dark red and green on this diagram are private and protected respectively:
http://www.stack.nl/~dimitri/doxygen/manual/examples/diagrams/html/class_d.html

If we document our private and protected functions with Doxygen too (which I think is mostly adding another asterisk IIRC @nomis52) then we'd get those things showing on our diagrams too. The question is would people find it useful or confusing.

PR #712, opened 2015-04-19
Update the Ja Rule labels & fix a broadcast bug.

nomis52:


tools/ja-rule/JaRuleEndpoint.h
```diff
@@ -45,22 +45,24 @@ class JaRuleEndpoint {
...
-    RDM_REQUEST = 0x87,
-    SET_BREAK_TIME = 0x88,
-    GET_BREAK_TIME = 0x89,
-    SET_MAB_TIME = 0x90,
-    GET_MAB_TIME = 0x91,
-    SET_RDM_BROADCAST_LISTEN = 0x92,
-    GET_RDM_BROADCAST_LISTEN = 0x93,
-    SET_RDM_WAIT_TIME = 0x94,
-    GET_RDM_WAIT_TIME = 0x95
+    RESET_DEVICE = 0x00,
```
peternewman: Can we remove the fact these are duplicated in two files please.

peternewman:
LGTM

tools/ja-rule/JaRuleEndpoint.h
nomis52: Yup, that's coming in a latter change. I should be able to get rid of tools/ja-rule/JaRuleEndpoint.{h,cpp}

PR #713, opened 2015-04-19
Start writing E1.37-2 RDM tests

peternewman:


tools/rdm/TestDefinitions.py
```diff
@@ -680,6 +680,17 @@ class GetSupportedParameters(ResponderTestFixture):
       ('LOCK_STATE', 'LOCK_STATE_DESCRIPTION'),
   ]
 
+  # If any PID apart from the first in each group is supported, the first one
+  # must be.
+  PID_REVERSE_DEPENDENCIES = [
```
nomis52: I'd make this a list of tuples in the form:

[  
  ( first_pid, [other_pids]),
]

tools/rdm/TestDefinitions.py
```diff
@@ -777,6 +788,24 @@ def VerifyResult(self, response, fields):
         self.AddAdvisory('%s supported but %s is not' %
                          (pid_names[0], ','.join(unsupported_pids)))
 
+    for pid_names in self.PID_REVERSE_DEPENDENCIES:
```
nomis52: then you can write this as

for required_pid, other_pids in self.PID_REVERSE_DEPENDECIES:

tools/rdm/TestDefinitions.py
```diff
@@ -777,6 +788,24 @@ def VerifyResult(self, response, fields):
...
+      if self.LookupPid(pid_names[0]).value in supported_parameters:
+        continue
+
+      dependant_pids = []
+      for pid_name in pid_names[1:]:
+        pid = self.LookupPid(pid_name)
+        if pid is None:
+          self.SetBroken('Missing PID %s' % pid_name)
```
nomis52: Hmm, this seems wrong. Broken is supposed to indicate a code error. I think this (and the one above) should be SetFailed() 

See https://www.openlighting.org/rdm-tools/rdm-responder-tests/interpreting-the-test-output/

tools/rdm/TestDefinitions.py
```diff
@@ -6403,6 +6432,50 @@ class AllSubDevicesGetPresetMergeMode(TestMixins.AllSubDevicesGetMixin,
...
+    if not response.WasAcked():
+      self.SetProperty('list_interfaces', [])
+      return
+
+    interfaces = []
+
+    for interface in fields['interfaces']:
+      interface_id = interface['interface_identifier']
+      interfaces.append(interface_id)
+      if interface['interface_hardware_type'] != 1:
```
nomis52: Make 1 an enum ?

tools/rdm/TestDefinitions.py
peternewman: Yes, I think that's correct; it's just doing PID name (in the code) to object lookup here, if that fails it is a code error.

tools/rdm/TestDefinitions.py
peternewman: Done, and for PID_DEPENDENCIES

tools/rdm/TestDefinitions.py
peternewman: Done

tools/rdm/TestDefinitions.py
peternewman: I've clarified the error message too.

tools/rdm/TestDefinitions.py
peternewman: Done.

PR #714, opened 2015-04-20
Fix the build bots.

nomis52:


peternewman:
LGTM and seems to have fixed at least one bot so far.

PR #715, opened 2015-04-22
Fix a bug in the SetDeviceHours test.

nomis52:
I found a device that would nack set DEVICE_HOURS with a NR_WRITE_PROTECT.
In this case, we should expect NR_FORMAT_ERROR rather than
NR_UNSUPPORTED_COMMAND_CLASS when we send a malformed request.

### DIFF
--- a/tools/rdm/TestDefinitions.py
+++ b/tools/rdm/TestDefinitions.py
@@ -2959,10 +2959,10 @@ class SetDeviceHours(TestMixins.SetUInt32Mixin,
 
   def VerifyResult(self, response, fields):
     if response.command_class == PidStore.RDM_SET:
-      set_unsupported = (
-          not response.WasAcked() and
-          response.nack_reason == RDMNack.NR_UNSUPPORTED_COMMAND_CLASS)
-      self.SetProperty('set_device_hours_supported', not set_unsupported)
+      set_supported = (
+          response.WasAcked() or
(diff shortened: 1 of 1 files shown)


tools/rdm/TestDefinitions.py
```diff
@@ -2959,9 +2959,10 @@ def OldValue(self):
 
   def VerifyResult(self, response, fields):
     if response.command_class == PidStore.RDM_SET:
-      self.SetProperty('set_device_hours_supported',
-                       response.WasAcked())
-
+      set_unsupported = (
+          not response.WasAcked() and
```
peternewman: Can we have some brackets to clarify operator precedence/avoid future mistakes

tools/rdm/TestDefinitions.py
```diff
@@ -2959,9 +2959,10 @@ def OldValue(self):
 
   def VerifyResult(self, response, fields):
     if response.command_class == PidStore.RDM_SET:
-      self.SetProperty('set_device_hours_supported',
-                       response.WasAcked())
-
+      set_unsupported = (
```
peternewman: Would this not be clearer:
set_supported = response.WasAcked() or (response.nack_reason != RDMNack.NR_UNSUPPORTED_COMMAND_CLASS)

Then you can do:
self.SetProperty('set_device_hours_supported', set_supported)

tools/rdm/TestDefinitions.py
nomis52: Not here.

PR #733, opened 2015-05-19
Update PID definitions.

nomis52:


peternewman:
@nomis52 PID 34324 exists twice in this commit, for CP_TILT_HOME_ANGLE and CP_PAN_HOME_ANGLE, which must also mean its broken on the website. Did it get in before we added the validation?

peternewman:
Indeed they're both listed here, so I guess the PID isn't the primary key? http://rdm.openlighting.org/pid/manufacturer?manufacturer=17232

PR #723, opened 2015-04-30
Clean up the RDMControllerInterface so we can attach timing information

nomis52:
PR #721 has to go in first.

PR #734, opened 2015-05-19
Remove the old RDM string methods

nomis52:
#723  goes in first.


peternewman:
Resync with trunk once #723 is in. Fix Travis.

peternewman:
LGTM

PR #737, opened 2015-05-20
Add an option to display the frame data in the RDM CLI tools.

nomis52:


examples/ola-rdm.cpp
```diff
@@ -269,44 +282,59 @@ bool RDMController::Setup() {
...
+  if (!response) {
+    cerr << "Error: Missing RDM Response but response_code was "
+            "RDM_COMPLETED_OK, this is a bug, please report it!"
+         << endl;
     return;
   }
 
-  if (response_status.response_type == ola::rdm::RDM_ACK) {
-    if (response_status.pid_value == m_pending_request.pid_value ||
+  if (response->ResponseType() == ola::rdm::RDM_ACK_TIMER) {
```
peternewman: Can we not put most of the work for this in the server side somehow?

examples/ola-rdm.cpp
```diff
@@ -314,14 +342,45 @@ void RDMController::HandleResponse(
       // support queued messages.
       cout << "Empty STATUS_MESSAGES returned." << endl;
     }
-  } else if (response_status.response_type == ola::rdm::RDM_NACK_REASON) {
-    cout << "Request NACKed: " <<
-      ola::rdm::NackReasonToString(response_status.NackReason()) << endl;
+  } else if (response->ResponseType() == ola::rdm::RDM_NACK_REASON) {
```
peternewman: Likewise here.

peternewman:
Can you update the man page, and ideally the bash completion too please.

examples/ola-rdm.cpp
nomis52: Not really, it's important we return ack timers to the client so they can provide a good user experience.

nomis52:
There is no man page:)

peternewman:
Perhaps you should create one then! :)

examples/ola-rdm.cpp
peternewman: But we could add a method to fetch the backoff time in ms directly, rather than each client calculating it (and avoid having the data size validation locally in each client).

I guess the other bits vary too much between clients.

examples/ola-rdm.cpp
peternewman: A NackReasonMessage method could wrap all of this, returning NackReasonToString or Invalid NACK reason size of " << response->ParamDataSize() << " expecting " << sizeof(nack_reason);

nomis52:
I'll deal with the rest later.

peternewman:
Cheeky @nomis52 :)

PR #739, opened 2015-05-22
this is a start of WIP for OLA to compile on mingw64

sl1200mk2:
please comment / rewrite
++
Nicolas

### DIFF
--- a/common/thread/SignalThread.cpp
+++ b/common/thread/SignalThread.cpp
@@ -112,11 +112,11 @@ void* SignalThread::Run() {
   return NULL;
 }
 
+#ifndef _WIN32
 /**
  * @brief Add the signals we're interested in to the sigset.
  */
 bool SignalThread::AddSignals(sigset_t *signals) {
-#ifndef _WIN32
   SignalMap::const_iterator iter = m_signal_handlers.begin();
   for (; iter != m_signal_handlers.end(); ++iter) {
     if (sigaddset(signals, iter->first)) {
@@ -125,9 +125,9 @@ bool SignalThread::AddSignals(sigset_t *signals) {
       return false;
     }
   }
-#endif
   return true;
 }
+#endif
 
 /**
  * Block the signal
--- a/configure.ac
+++ b/configure.ac
@@ -131,8 +131,10 @@ AC_CHECK_FUNCS([bzero gettimeofday memmove memset mkdir strdup strrchr \
 LT_INIT
 AC_PROG_LIBTOOL
 
-# Decide if we're building on Windows early on.
 AM_CONDITIONAL([USING_WIN32], [test "x$host_os" = xmingw32])
+if test -z "${USING_WIN32_FALSE}"; then
+  AM_CONDITIONAL([USING_WIN32], [test "x$host_os" = xmingw64])
+fi
 
 # Epoll
 AX_HAVE_EPOLL(
@@ -307,6 +309,22 @@ ACX_PTHREAD([
              CC="$PTHREAD_CC"
              CXX="$PTHREAD_CXX"],
             [AC_MSG_ERROR([Missing pthread, please install it])])
+
+if test -z "${USING_WIN32_TRUE}"; then
+  has_winpthread=no
+  dnl this will tell us if the implementation of pthread is winpthread
+  AC_CHECK_LIB(winpthread, pthread_create, [has_winpthread="yes"],
+               [echo "found library winpthread"], )
+  PTHREAD_CFLAGS=-pthread
+  PTHREAD_LIBS=-lwinpthread
+AM_CONDITIONAL([HAVE_WINPTHREAD], [test "x$has_winpthread" = xyes])
+AS_IF([test "x$has_winpthread" = xyes],
+      [AC_DEFINE([HAVE_WINPTHREAD], [1],
+                 [define if winpthread is installed])])
+fi
+
+AC_SUBST(PTHREAD_CFLAGS)
+AC_SUBST(PTHREAD_LIBS)
 
 # pthread_setname_np can take either 1 or 2 arguments.
 PTHREAD_SET_NAME()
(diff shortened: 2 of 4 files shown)


include/ola/thread/Thread.h
```diff
@@ -33,18 +33,6 @@
 
 #include <string>
 
-#if defined(_WIN32) && defined(__GNUC__)
```
nomis52: I recall @nightrune  or @peternewman  adding this.

nightrune: Yeah @peternewman did this in this commit:

https://github.com/OpenLightingProject/ola/commit/c30041c1095246ae578c66bc53f5a8f464052e32

I don't think we can remove.

peternewman: I moved it: https://github.com/OpenLightingProject/ola/commit/c30041c1095246ae578c66bc53f5a8f464052e32 
@lukase added it initially: https://github.com/OpenLightingProject/ola/commit/5f4aa5ed067012093b058d30afa6b3f32847fe28

So it needs to stay @sl1200mk2 , by all means add some more ifdefs if its breaking your combination of libraries.

common/thread/SignalThread.cpp
```diff
@@ -112,11 +112,11 @@ void* SignalThread::Run() {
   return NULL;
 }
 
+#ifndef _WIN32
 /**
  * @brief Add the signals we're interested in to the sigset.
  */
 bool SignalThread::AddSignals(sigset_t *signals) {
-#ifndef _WIN32
```
peternewman: FWIW sigset_t must be defined on the other flavours of Windows/using pthreads. The problem with ifndefing the whole function is you have to fix up every bit of code. It might make more sense to define sigset_t if it doesn't exist.

configure.ac
```diff
@@ -131,8 +131,10 @@ AC_CHECK_FUNCS([bzero gettimeofday memmove memset mkdir strdup strrchr \
 LT_INIT
 AC_PROG_LIBTOOL
 
-# Decide if we're building on Windows early on.
```
peternewman: Can we keep the comment please.

configure.ac
```diff
@@ -131,8 +131,10 @@ AC_CHECK_FUNCS([bzero gettimeofday memmove memset mkdir strdup strrchr \
...
 AM_CONDITIONAL([USING_WIN32], [test "x$host_os" = xmingw32])
+if test -z "${USING_WIN32_FALSE}"; then
```
peternewman: Perhaps we should create a USING_WIN and also define 32 and 64 bit variants?

configure.ac
```diff
@@ -308,6 +310,22 @@ ACX_PTHREAD([
              CXX="$PTHREAD_CXX"],
             [AC_MSG_ERROR([Missing pthread, please install it])])
 
+if test -z "${USING_WIN32_TRUE}"; then
```
peternewman: It probably makes sense to extend https://github.com/OpenLightingProject/ola/blob/master/config/acx_pthread.m4

include/ola/thread/SignalThread.h
```diff
@@ -64,7 +64,9 @@ class SignalThread : public ola::thread::Thread {
 
     SignalMap m_signal_handlers;
 
+#ifndef _WIN32
```
peternewman: Again as above.

configure.ac
nightrune: Are there going to be instances where we don't need to know WIN32 vs WIN64?

peternewman: I expect so, stuff that is Windows code, as opposed to *nix, e.g. some of the networking stuff.

peternewman: E.g. https://github.com/OpenLightingProject/ola/blob/master/common/network/InterfacePicker.cpp#L142 It kind of feels wrong using _WIN32 to decide if we're on 64 bit, but perhaps that's just what everyone does for compatability

peternewman: Apparently it should be that screwy: http://stackoverflow.com/questions/6679396/should-i-define-both-win32-and-win64-in-64bit-build

include/ola/thread/Thread.h
```diff
@@ -33,18 +33,6 @@
...
-                       const ptw32_handle_t &right) {
-  return (left.p == right.p) && (left.x == right.x);
-}
-
-inline std::ostream& operator<<(std::ostream &stream,
-                                 const ptw32_handle_t &handle) {
-  stream << handle.p;
-  return stream;
-}
-#endif
```
sl1200mk2: why can't we use pthread_equal() instead of operator==
thus we can pass pthread_t and forget about ptw32_handle_t

if so, for the operator<< we can use HAVE_WINPTHREAD ifdef and write 
#ifdef HAVE_WINPTHREAD
inline std::ostream& operator<<(std::ostream &stream,
                                const pthread_t &handle) {
 stream << handle;
 return stream;
}
#endif

nomis52: Yup, we should be using pthread_equal. The == operator isn't defined for pthread_t , see http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_equal.html

nomis52: https://github.com/OpenLightingProject/ola/pull/742 fixes the one case I found of directly comparing pthread_t

sl1200mk2: thanx

nomis52: #742 is merged. Can you merge with that and confirm it's safe to remove the operator==

sl1200mk2: Hi Simon,
now it's movie time and tomorrow is a 'no work' day, so i'll test in the middle of next week.
thanx,
++

nomis52:
@sl1200mk2  can you pull and merge please, so this can be merged cleanly.

sl1200mk2:
Hi,
closing this one.
further work will be done on a fresh clone.

PR #742, opened 2015-05-23
Switch to using pthread_equal rather than operator==

nomis52:


peternewman:
If you're doing this, can you remove the == operator too please.

peternewman:
As mentioned on IRC, I suspect based on https://github.com/OpenLightingProject/ola/commit/c30041c1095246ae578c66bc53f5a8f464052e32 it may be needed elsewhere, but I guess we just need to suck it and see on Windows again and see what breaks.

peternewman:
Actually can't we just redefine the == operator to work on ThreadId and not touch/specialise the test?

nomis52:
Since ThreadId is typedefs to the native thread type, I don't want to define a operator== since it may conflict with the system definition.

I'm going to leave the operator== in place and let @nightrune remove it. I don't want to break the Windows build any more than I have to.

peternewman:
Okay @nomis52

PR #750, opened 2015-05-26
Fix the Ja-Rule Build

peternewman:


nomis52:
LGTM

nomis52:
LGTM

peternewman:
Okay I think this should be ready to merge finally if Travis goes green, as it will actually be building Ja Rule.

nomis52:
LGTM

PR #755, opened 2015-05-27
Report the config via the JSON Server Stats

peternewman:


include/olad/Preferences.h
```diff
@@ -327,6 +327,12 @@ class PreferencesFactory {
    */
   virtual Preferences *NewPreference(const std::string &name);
 
+  /**
+   * @brief The location where preferences will be stored.
+   * @return the location
+   */
+  virtual std::string Source() const = 0;
```
nomis52: Maybe call this "ConfigLocation" or something similar.

peternewman: Can do, I was just matching the Preferences class name.

peternewman: Fixed

peternewman:
This should be good to go when Travis is done if you're happy.

Issue #682, opened 2015-03-23
Add a handler to enable/disable a plugin from the Web and CLI UIs


peternewman:
Based on comments in #617.

peternewman:
Started in https://github.com/OpenLightingProject/ola/pull/756 .

peternewman:
#756 is in now, so over to @daveol

PR #756, opened 2015-05-28
Add the ability to enable and disable plugins while running - part of #682

peternewman:
Initial work to enable/disable plugins while running, the C++ client and RPC bits. Still need to actually do the work on the server side and implement in the HTTP/JSON and Python APIs


common/protocol/Ola.proto
```diff
@@ -403,6 +406,7 @@ service OlaServerService {
   rpc GetDeviceInfo (DeviceInfoRequest) returns (DeviceInfoReply);
   rpc GetCandidatePorts (OptionalUniverseRequest) returns (DeviceInfoReply);
   rpc ConfigureDevice (DeviceConfigRequest) returns (DeviceConfigReply);
+  rpc SetPluginState (PluginStateChange) returns (Ack);
```
nomis52: I'd call this PluginStateRequest

peternewman: I can't, it already exists to get the plugin info; see line 402/405.

nomis52: How about 

PluginStateChangeRequest

I know it's long, but it's much easier to keep track of the types if they have a Request / Reply suffix.

peternewman: Done.

nomis52:
It looks like this is pretty close to being done. Do you need anything else from me?

peternewman:
Not yet, I need to add the code to actually do the work in olad and expose it via the HTTP interface.

peternewman:
@nomis52 this should be ready for review and merge now.

examples/ola-client.cpp
```diff
@@ -278,6 +279,12 @@ void DisplayDevices(SelectServer *ss,
   ss->Terminate();
 }
 
+void SetPluginStateComplete(SelectServer *ss, const Result &result) {
```
nomis52: We should just merge all of these into a HandleAck(..) but that can be done later.

examples/ola-client.cpp
```diff
@@ -813,8 +826,23 @@ int FetchPluginState(OlaClientWrapper *wrapper, const options &opts) {
     DisplayPluginStateHelp(opts);
     exit(1);
   }
-  client->FetchPluginState((ola::ola_plugin_id) opts.plugin_id,
-                           NewSingleCallback(&DisplayPluginState, ss));
+  if (opts.state.length() > 0) {
```
nomis52: !opts.state.empty()

include/ola/client/OlaClient.h
```diff
@@ -130,6 +131,16 @@ class OlaClient {
                            CandidatePortsCallback *callback);
 
   /**
+   * @brief Set the state of a plugin.
+   * @param plugin_ids the ids of the plugins to set the state of.
+   * @param state the state to set the plugin to
+   * @param callback the SetCallback to invoke upon completion.
+   */
+  void SetPluginState(const std::vector<ola_plugin_id> &plugin_ids,
```
nomis52: Make this take a single ID.

The problem with passing multiple plugins in a single call is how we handle failures. Do we change all or none? That means we have to handle rollbacks.  It gets complicated.

common/protocol/Ola.proto
```diff
@@ -196,6 +196,11 @@ message PluginStateReply {
   repeated PluginInfo conflicts_with = 5;
 }
 
+message PluginStateChangeRequest {
+  repeated int32 plugin_id = 1;
```
nomis52: Make this singular - see comments below.

olad/OladHTTPServer.cpp
```diff
@@ -508,6 +513,44 @@ int OladHTTPServer::ModifyUniverse(const HTTPRequest *request,
...
+
+  string state_string = request->GetPostParameter("state");
+
+  bool state;
+  if (!StringToBoolTolerant(state_string, &state)) {
+    OLA_INFO << "Invalid state " << state_string;
+    return ServeHelpRedirect(response);
+  }
+
+  string plugin_id_string = request->GetPostParameter("plugin_ids");
```
nomis52: We can fan out here if we really need to support changing multiple plugins at once, but I'd avoid it if we can.

examples/ola-client.cpp
peternewman: Done.

common/protocol/Ola.proto
peternewman: Done.

olad/OladHTTPServer.cpp
peternewman: Removed.

include/ola/client/OlaClient.h
peternewman: Done.

examples/ola-client.cpp
peternewman: Done.

ola/OlaClientCore.h
```diff
@@ -23,6 +23,7 @@
 
 #include <memory>
 #include <string>
+#include <vector>
```
nomis52: Can drop this now

include/ola/client/OlaClient.h
```diff
@@ -34,6 +34,7 @@
 
 #include <memory>
 #include <string>
+#include <vector>
```
nomis52: And here

ola/OlaClientCore.h
peternewman: Done

include/ola/client/OlaClient.h
peternewman: Done

PR #757, opened 2015-05-29
Clean up the Daemonise method a bit.

nomis52:


common/base/Init.cpp
```diff
@@ -360,19 +359,27 @@ void Daemonise() {
...
+  int maxfd = sysconf(_SC_OPEN_MAX);
+  if (maxfd == -1) {
+    if (rl.rlim_max == RLIM_INFINITY) {
+      maxfd = 1024;
+    } else {
+      maxfd = rl.rlim_max;
+    }
+  }
+
+  for (unsigned int fd = 0; fd < rl.rlim_max; fd++) {
```
peternewman: Should this not be maxfd?

peternewman:
Generally LGTM

common/base/Init.cpp
nomis52: Good catch.

peternewman: I imagine Coverity would have thrown an unused variable warning too.

common/base/Init.cpp
```diff
@@ -360,19 +359,27 @@ void Daemonise() {
...
 
   // close all fds
-  if (rl.rlim_max == RLIM_INFINITY)
-    rl.rlim_max = 1024;
-  for (i = 0; i < rl.rlim_max; i++)
-    close(i);
+  int maxfd = sysconf(_SC_OPEN_MAX);
+  if (maxfd == -1) {
+    if (rl.rlim_max == RLIM_INFINITY) {
+      maxfd = 1024;
```
peternewman: So why 1024, or is this a 1024 ought to be enough for anyone type situation?

peternewman:
LGTM, merge if my outstanding comment doesn't throw up a new bug.

common/base/Init.cpp
nomis52: Correct. 

For our use case, we only ever have 3 fds open at this point anyway.

Issue #773, opened 2015-06-08
Building with Ja Rule requires the usbdmx plugin


simark:
I just tried building with "--enable-rdm-tests --enable-ja-rule" to try to mimic what Travis does. However, I also add "--disable-all-plugins --enable-dummy", so that my build iterations are faster. It gave me this:

```
libtool: link: g++ -I./include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -pthread -Werror -I/usr/include/libusb-1.0 -g -O2 -pthread -o tools/ja-rule/.libs/ja-rule tools/ja-rule/tools_ja_rule_ja_rule-JaRuleEndpoint.o tools/ja-rule/tools_ja_rule_ja_rule-USBDeviceManager.o tools/ja-rule/tools_ja_rule_ja_rule-ja-rule.o  -lusb-1.0 common/.libs/libolacommon.so -lresolv -lprotobuf -lpthread plugins/usbdmx/.libs/libolausbdmxwidget.a -ldns_sd -ldl -pthread -Wl,-rpath -Wl,/usr/local/lib
tools/ja-rule/tools_ja_rule_ja_rule-JaRuleEndpoint.o: In function `JaRuleEndpoint::Init()':
/home/simark/src/gridss/ola/tools/ja-rule/JaRuleEndpoint.cpp:121: undefined reference to `ola::plugin::usbdmx::LibUsbAdaptor::ErrorCodeToString(int)'
tools/ja-rule/tools_ja_rule_ja_rule-JaRuleEndpoint.o: In function `JaRuleEndpoint::SubmitInTransfer()':
/home/simark/src/gridss/ola/tools/ja-rule/JaRuleEndpoint.cpp:280: undefined reference to `ola::plugin::usbdmx::LibUsbAdaptor::ErrorCodeToString(int)'
tools/ja-rule/tools_ja_rule_ja_rule-JaRuleEndpoint.o: In function `JaRuleEndpoint::_InTransferComplete()':
/home/simark/src/gridss/ola/tools/ja-rule/JaRuleEndpoint.cpp:183: undefined reference to `ola::plugin::usbdmx::LibUsbAdaptor::ErrorCodeToString(int)'
tools/ja-rule/tools_ja_rule_ja_rule-JaRuleEndpoint.o: In function `JaRuleEndpoint::MaybeSendRequest()':
[...]

PR #775, opened 2015-06-08
configure.ac: Ja Rule dependency management, auto support for plugins, various small improvements

simark:
… met

Configure now fails if Ja Rule is enabled, but either of:

 - the usbdmx plugin is not enabled
 - libusb doesn't support hotplug

is true.

### DIFF
--- a/config/ola.m4
+++ b/config/ola.m4
@@ -104,22 +104,23 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
 #
 # Build the specified plugin if requested and its dependencies are met.
 #
-# Each plugin specified with PLUGIN_SUPPORT gets its own configure switch named
-# --enable_${plugin_key}. The three acceptable values the user can give are:
+# Each plugin specified with PLUGIN_SUPPORT gets its own configure switches
+# --enable-X/--disable-X, which translate to the enable_X variable. The three
+# acceptable states for this variable are:
 #
-#  - auto (default): Build the plugin if its dependencies are met, otherwise
-#    don't.
+#  - auto: Build the plugin if its dependencies are met, otherwise
+#    don't. This is the default state used when the user does not specify
+#    anything regarding this plugin.
 #  - yes: Build the plugin and error out if its dependencies are not met.
 #  - no: Never build the plugin.
 #
-# If the input value of enable_${plugin_key} is "auto" and EITHER of the
-# following is true:
+# If the input value of enable_X is "auto" and EITHER of the following is true:
 #
 #  - --disable-all-plugins is specified
 #  - the plugin's dependencies are not met
 #
-# then PLUGIN_SUPPORT will overwrite the value of enable_${plugin_key} to "no".
-# In all other cases, its original value is preserved.
+# then PLUGIN_SUPPORT will overwrite the value of enable_X to "no". In all
+# other cases, the original value is preserved.
 #
 # This means that after PLUGIN_SUPPORT has "ran" for a particular plugin, a value
 # of "auto" or "yes" means that the plugin will get built, while a value of "no"
--- a/config/ola.m4
+++ b/config/ola.m4
@@ -148,13 +148,13 @@ AC_DEFUN([PLUGIN_SUPPORT],
 
   # If dependencies are not met...
   if test "$3" == "no"; then
-    # ...and the user has explicitely requested this plugin to be enabled,
+    # ...and the user has explicitly requested this plugin to be enabled,
     # error out.
     if test "${enable_plugin}" = "yes"; then
       AC_MSG_ERROR([Dependencies for plugin ${plugin_key} are not met.])
     fi
 
-    # Otherwise, force the plugin disabled silently.
+    # Otherwise, silently force disable the plugin.
     enable_plugin="no";
   fi
 
--- a/config/ola.m4
+++ b/config/ola.m4
@@ -160,7 +160,7 @@ AC_DEFUN([PLUGIN_SUPPORT],
 
   # If we've disabled all plugins, disable this one unless it has been
   # explicitly enabled.
-  if test "${enable_all_plugins}" = "no" -a "${enable_plugin}" != "yes"; then
+  if test "${enable_all_plugins}" = "no" -a "${enable_plugin}" = "auto"; then
     enable_plugin="no";
   fi
 
--- a/config/ola.m4
+++ b/config/ola.m4
@@ -100,9 +100,30 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
 ])
 
 
-# PLUGIN_SUPPORT(plugin, conditional, prerequisites_found)
-# Build the specified plugin, unless it was disabled at configure time.
-# -----------------------------------------------------------------------------
+# PLUGIN_SUPPORT(plugin_key, conditional, prerequisites_found)
+#
+# Build the specified plugin if requested and its dependencies are met.
+#
+# Each plugin specified with PLUGIN_SUPPORT gets its own configure switch named
+# --enable_${plugin_key}. The three acceptable values the user can give are:
+#
+#  - auto (default): Build the plugin if its dependencies are met, otherwise
+#    don't.
+#  - yes: Build the plugin and error out if its dependencies are not met.
+#  - no: Never build the plugin.
+#
+# If the input value of enable_${plugin_key} is "auto" and EITHER of the
+# following is true:
+#
+#  - --disable-all-plugins is specified
+#  - the plugin's dependencies are not met
+#
+# then PLUGIN_SUPPORT will overwrite the value of enable_${plugin_key} to "no".
+# In all other cases, its original value is preserved.
+#
+# This means that after PLUGIN_SUPPORT has "ran" for a particular plugin, a value
+# of "auto" or "yes" means that the plugin will get built, while a value of "no"
+# means that it won't.
 AC_DEFUN([PLUGIN_SUPPORT],
 [
   plugin_key=$1
@@ -110,19 +131,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
 
   AC_ARG_ENABLE(
     [$1],
-    AS_HELP_STRING([--disable-$1], [Disable the $1 plugin]))
+    AS_HELP_STRING([--disable-$1], [Disable the $1 plugin]),
+    [],
+    [eval $enable_arg=auto])
 
   eval enable_plugin=\$$enable_arg;
+
+  # Input validation.
+  if test "${enable_plugin}" != "yes" -a \
+          "${enable_plugin}" != "no" -a \
+          "${enable_plugin}" != "auto"; then
+    AC_MSG_ERROR([Invalid value for --enable-${plugin_key}/\
+--disable-${plugin_key}. Valid values are yes, no and auto.])
+  fi
+
+  # If dependencies are not met...
   if test "$3" == "no"; then
+    # ...and the user has explicitely requested this plugin to be enabled,
+    # error out.
+    if test "${enable_plugin}" = "yes"; then
+      AC_MSG_ERROR([Dependencies for plugin ${plugin_key} are not met.])
+    fi
+
+    # Otherwise, force the plugin disabled silently.
     enable_plugin="no";
   fi
 
   # If we've disabled all plugins, disable this one unless it has been
-  # explicitly enabled
-  if test "${enable_all_plugins}" == "no"; then
-    if test "${enable_plugin}" != "yes"; then
-      enable_plugin="no";
-    fi
+  # explicitly enabled.
+  if test "${enable_all_plugins}" = "no" -a "${enable_plugin}" != "yes"; then
+    enable_plugin="no";
   fi
 
   if test "${enable_plugin}" != "no"; then
--- a/config/ola.m4
+++ b/config/ola.m4
@@ -122,7 +122,7 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
 # then PLUGIN_SUPPORT will overwrite the value of enable_X to "no". In all
 # other cases, the original value is preserved.
 #
-# This means that after PLUGIN_SUPPORT has "ran" for a particular plugin, a value
+# This means that after PLUGIN_SUPPORT has "run" for a particular plugin, a value
 # of "auto" or "yes" means that the plugin will get built, while a value of "no"
 # means that it won't.
 AC_DEFUN([PLUGIN_SUPPORT],
--- a/configure.ac
+++ b/configure.ac
@@ -663,6 +663,11 @@ AC_ARG_ENABLE(
 # knows what to link against.
 PLUGINS=""
 
+# Force enable_usbdmx to yes when Ja Rule is explicitly requested.
+AS_IF([test "x$enable_ja_rule" = xyes],
+      [AS_ECHO(["Ja Rule is enable, enabling the usbdmx plugin."])]
+      [enable_usbdmx="yes"])
+
 PLUGIN_SUPPORT(artnet, USE_ARTNET)
 PLUGIN_SUPPORT(dmx4linux, USE_DMX4LINUX, [$have_dmx4linux])
 PLUGIN_SUPPORT(dummy, USE_DUMMY)
@@ -717,13 +722,20 @@ AC_SUBST(OLA_CLIENT_LIBS)
 
 # Extra tools
 #####################################################
-AM_CONDITIONAL([BUILD_JA_RULE],
-               [test "x$enable_ja_rule" = xyes && test "x$have_libusb" = xyes \
-                && test "x$have_libusb_hotplug_api" = xyes])
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$enable_usbdmx" != xyes],
+      AC_MSG_ERROR([Ja Rule requires the usbdmx plugin, but it could not be enabled.]))
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb" != xyes],
+      AC_MSG_ERROR([Ja Rule requires libusb, but it was not found on your system.]))
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb_hotplug_api" != xyes],
+      AC_MSG_ERROR([Ja Rule requires libusb to have the hotplug capability \
+(versions 1.0.16 and above), but the version found on your system doesn't \
+appear to have it.]))
+
+AM_CONDITIONAL([BUILD_JA_RULE], [test "x$enable_ja_rule" = xyes])
 
 # Status just for configure
 BUILDING_JA_RULE='no'
-if test -z "${BUILD_JA_RULE_TRUE}"; then
+if test "x$enable_ja_rule" = xyes; then
   BUILDING_JA_RULE='yes'
 fi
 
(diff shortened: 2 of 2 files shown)


simark:
I added a second commit that does the same for libmicrohttpd.

peternewman:
See my later comments in https://github.com/OpenLightingProject/ola/issues/773.

Also you've currently forced people to add the --enable-usbdmx argument if they --enable-ja-rule, instead we should be smarter and just check the dependencies for --enable-usbdmx are met, rather than the explicit presence of the flag.

configure.ac
```diff
@@ -717,13 +721,16 @@ AC_SUBST(OLA_CLIENT_LIBS)
 
 # Extra tools
 #####################################################
-AM_CONDITIONAL([BUILD_JA_RULE],
-               [test "x$enable_ja_rule" = xyes && test "x$have_libusb" = xyes \
-                && test "x$have_libusb_hotplug_api" = xyes])
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$enable_usbdmx" != xyes],
+      AC_MSG_ERROR([Ja Rule requires the usbdmx plugin to be enable (--enable-usbdmx).]))
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb_hotplug_api" != xyes],
```
peternewman: I'd like to be belt and braces and keep this check in too "x$have_libusb" = xyes.

simark:
Now, when --enable-ja-rule is specified, we automatically try to enable the usbdmx plugin. However, there is still the possibility that usbdmx misses some dependency and gets silently disabled, so we still need to check afterwards if it has been successfully enabled.

It will make more sense if/when we apply the same yes/no/auto logic to the plugins. When Ja Rule is explicitly enabled, we will automatically explicitly enable usbdmx. If usbdmx can't be enabled for some reason, configure will stop and report an error. So the check for "x$enable_usbdmx" != "xyes" at the end will become pointless.

simark:
Arrrg, being used to Gerrit, I forgot that Github does not preserve patches versions, so I did a force push. I'll push new patches next time so you can see the diff between versions.

configure.ac
```diff
@@ -663,6 +667,11 @@ AC_ARG_ENABLE(
 # knows what to link against.
 PLUGINS=""
 
+# Force enable_usbdmx to yes when Ja Rule is explicitly requested.
+AS_IF([test "x$enable_ja_rule" = xyes],
+      [AS_ECHO(["Ja Rule is enable, enabling the usbdmx plugin."])]
```
peternewman: SPaG enabled

configure.ac
```diff
@@ -717,13 +726,20 @@ AC_SUBST(OLA_CLIENT_LIBS)
 
 # Extra tools
 #####################################################
-AM_CONDITIONAL([BUILD_JA_RULE],
-               [test "x$enable_ja_rule" = xyes && test "x$have_libusb" = xyes \
-                && test "x$have_libusb_hotplug_api" = xyes])
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$enable_usbdmx" != xyes],
```
peternewman: It probably makes sense to check for != xno for ja_rule and = xno for usbdmx to support a future auto mode.

configure.ac
```diff
@@ -717,13 +726,20 @@ AC_SUBST(OLA_CLIENT_LIBS)
...
+      AC_MSG_ERROR([Ja Rule requires the usbdmx plugin, but it could not be enabled.]))
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb" != xyes],
```
peternewman: Again for ja_rule

configure.ac
```diff
@@ -717,13 +726,20 @@ AC_SUBST(OLA_CLIENT_LIBS)
...
+      AC_MSG_ERROR([Ja Rule requires libusb, but it was not found on your system.]))
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb_hotplug_api" != xyes],
```
peternewman: And again for ja_rule

configure.ac
simark: Hmm I don't think it's that simple. Outputting the error message only makes sense if $enable_ja_rule is exactly "yes" (forced enabled by the user). Assuming Ja Rule has an auto mode (enabled if possible, otherwise disabled), if the check is "$enable_ja_rule != no && $enable_usbdmx == no" as you suggest, then this will fail:

    ./configure --enable-usbdmx=no

whereas we would just want to silently disable Ja Rule.

If we want to have an auto mode, I think we will have to add an "else if" clause to the existing AS_IF. In pseudo-code, it would look like:

    if $enable_ja_rule == yes && $enable_usbdmx != yes:
        error("Ja Rule needs libusb...")
    elif $enable_ja_rule == auto $$ $enable_usbdmx != yes:
        $enable_ja_rule = no

This would be repeated for each of Ja Rule's dependencies. Note that if $enable_usbdmx was originally set to auto, at this point it would have been forced to either "yes" or "no" by the code in PLUGIN_SUPPORT. Similarly, after checking for all Ja Rule's dependencies, we should force enable_ja_rule to either "yes" or "no", so that if some other code depends on it, it can check whether Ja Rule will actually be enabled or not.

Does this make sense?

peternewman: FWIW, it's --disable-usbdmx normally.

You're right about the ja_rule check, but I'd say the usbdmx one should still be = xno. I'm not sure we need to change the variable state in PLUGIN_SUPPORT unless its a case of changing it to no. It feels an auto should track through the whole way as an auto, so you can tell the intent later on.

simark: > FWIW, it's --disable-usbdmx normally.

`--disable-<feature>` is the same as `--enable-<feature>=no` [1], but you are right that `--disable-<feature>` is more clear.

> You're right about the ja_rule check, but I'd say the usbdmx one should still be = xno. I'm not sure we need to change the variable state in PLUGIN_SUPPORT unless its a case of changing it to no. It feels an auto should track through the whole way as an auto, so you can tell the intent later on.

I agree, we can keep more information by keeping it to auto.

[1] http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Package-Options.html

configure.ac
simark: Do you prefer
    test "x$have_libusb" != xyes
or
    test "x$have_libusb" = xno
? This one is really a boolean so functionally it's the same, but do you have a preference?

configure.ac
simark: Done in new commit.

simark:
Ok, I *think* I managed to get all cases working. Configure will now fail if a plugin is explicitely enabled, but its dependencies are not met.

config/ola.m4
```diff
@@ -100,9 +100,30 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
...
-# Build the specified plugin, unless it was disabled at configure time.
-# -----------------------------------------------------------------------------
+# PLUGIN_SUPPORT(plugin_key, conditional, prerequisites_found)
+#
+# Build the specified plugin if requested and its dependencies are met.
+#
+# Each plugin specified with PLUGIN_SUPPORT gets its own configure switch named
+# --enable_${plugin_key}. The three acceptable values the user can give are:
+#
+#  - auto (default): Build the plugin if its dependencies are met, otherwise
```
peternewman: I don't think we should ever get the user to give this option, it should just be the default.

config/ola.m4
```diff
@@ -100,9 +100,30 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
...
 
-# PLUGIN_SUPPORT(plugin, conditional, prerequisites_found)
-# Build the specified plugin, unless it was disabled at configure time.
-# -----------------------------------------------------------------------------
+# PLUGIN_SUPPORT(plugin_key, conditional, prerequisites_found)
+#
+# Build the specified plugin if requested and its dependencies are met.
+#
+# Each plugin specified with PLUGIN_SUPPORT gets its own configure switch named
+# --enable_${plugin_key}. The three acceptable values the user can give are:
```
peternewman: We should encourage users to use --disable-${plugin_key} (or --enable-${plugin_key} in conjunction with --disable-all-plugins rather than passing options into a single option. Also it's hyphens not underscores.

config/ola.m4
```diff
@@ -100,9 +100,30 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
...
+# If the input value of enable_${plugin_key} is "auto" and EITHER of the
+# following is true:
+#
+#  - --disable-all-plugins is specified
+#  - the plugin's dependencies are not met
+#
+# then PLUGIN_SUPPORT will overwrite the value of enable_${plugin_key} to "no".
+# In all other cases, its original value is preserved.
+#
+# This means that after PLUGIN_SUPPORT has "ran" for a particular plugin, a value
```
peternewman: SPaG run

config/ola.m4
```diff
@@ -100,9 +100,30 @@ AM_CONDITIONAL(BUILD_OLA_PROTOC_PLUGIN, test "${with_ola_protoc_plugin}" == "no"
...
+#
+# Each plugin specified with PLUGIN_SUPPORT gets its own configure switch named
+# --enable_${plugin_key}. The three acceptable values the user can give are:
+#
+#  - auto (default): Build the plugin if its dependencies are met, otherwise
+#    don't.
+#  - yes: Build the plugin and error out if its dependencies are not met.
+#  - no: Never build the plugin.
+#
+# If the input value of enable_${plugin_key} is "auto" and EITHER of the
```
peternewman: auto or yes, so a yes becomes no if a dependency is missing, unless we're going to have a seperate output status variable.

config/ola.m4
```diff
@@ -110,19 +131,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
...
+    [],
+    [eval $enable_arg=auto])
 
   eval enable_plugin=\$$enable_arg;
+
+  # Input validation.
+  if test "${enable_plugin}" != "yes" -a \
+          "${enable_plugin}" != "no" -a \
+          "${enable_plugin}" != "auto"; then
+    AC_MSG_ERROR([Invalid value for --enable-${plugin_key}/\
```
peternewman: Perhaps change to "invalid state for ${plugin_key}"

config/ola.m4
```diff
@@ -110,19 +131,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
...
   fi
 
   # If we've disabled all plugins, disable this one unless it has been
-  # explicitly enabled
-  if test "${enable_all_plugins}" == "no"; then
-    if test "${enable_plugin}" != "yes"; then
-      enable_plugin="no";
-    fi
+  # explicitly enabled.
+  if test "${enable_all_plugins}" = "no" -a "${enable_plugin}" != "yes"; then
```
peternewman: If we've got the third state of auto, we either need an output plugin to check, or need to set auto's to no if disabling all plugins.

configure.ac
```diff
@@ -717,13 +726,19 @@ AC_SUBST(OLA_CLIENT_LIBS)
...
-                && test "x$have_libusb_hotplug_api" = xyes])
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$enable_usbdmx" = xno],
+      [AC_MSG_ERROR([Ja Rule requires the usbdmx plugin, but it could not be enabled.])])
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb" = xno],
+      [AC_MSG_ERROR([Ja Rule requires libusb, but it was not found on your system.])])
+AS_IF([test "x$enable_ja_rule" = xyes && test "x$have_libusb_hotplug_api" = xno],
+      [AC_MSG_ERROR([Ja Rule requires libusb to have the hotplug capability (versions 1.0.16 \
+and above), but the version found on your system doesn't appear to have it.])])
+
+AM_CONDITIONAL([BUILD_JA_RULE], [test "x$enable_ja_rule" = xyes])
```
peternewman: You don't seem to be changing enable_ja_rule when the conditions fail, or is the argument the AC_MSG_ERROR's will trap it?

config/ola.m4
simark: That's what it is. If the user doesn't specify anything, the `enable_${plugin_key}` variable will hold the value "auto". See the call to AC_ARG_ENABLE. My sentence is not clear, I'll change it to:

    ... the three acceptable states are:
    
    - auto: Build the plugin if its dependencies are not met, otherwise
      don't. This is the default state used when the user did not specify
      anything regarding this plugin.
    ....

config/ola.m4
simark: > We should encourage users to use --disable-${plugin_key} (or --enable-${plugin_key} in conjunction with --disable-all-plugins rather than passing options into a single option.

I don't understand your point here, nor what I should change.

> Also it's hyphens not underscores.
Thanks, fixed.

config/ola.m4
simark: Actually, if the input state is "yes" and dependencies are missing, we throw an error and the script exits. So if the input state is "yes", the output state will always be yes.

config/ola.m4
simark: I think I should always use the same terminology. If I use state instead of value here, I'll replace it everywhere. Is this ok?

config/ola.m4
simark: > If we've got the third state of auto, we either need an output plugin to check, or need to set auto's to no if disabling all plugins.

I am not sure I understand. That's exactly what this line is doing. If the plugin's state is auto and we want to disable all plugins, we override the state to "no" so the plugin doesn't get built.

configure.ac
simark: AC_MSG_ERROR prints the error and exits the script with failure.

Currently, the default value for Ja Rule is that it doesn't get built (therefore it doesn't have an auto mode). You just enable it if you want it. If you enable it and a dependency is missing, configure stops and tells you there was a problem.

If we want Ja Rule to be auto by default, then we'll need to refactor this so that if a condition fails, we either fail the configure script or silently disable the feature, depending on the input state.

config/ola.m4
nomis52: I think value is find, the user cares about values not states.

config/ola.m4
```diff
@@ -110,19 +132,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
...
+          "${enable_plugin}" != "auto"; then
+    AC_MSG_ERROR([Invalid value for --enable-${plugin_key}/\
+--disable-${plugin_key}. Valid values are yes, no and auto.])
+  fi
+
+  # If dependencies are not met...
   if test "$3" == "no"; then
+    # ...and the user has explicitely requested this plugin to be enabled,
+    # error out.
+    if test "${enable_plugin}" = "yes"; then
```
nomis52: You can use AS_IF here if you like.

simark: What is the advantage? Is it preferred because it's more portable?

config/ola.m4
peternewman: You've fixed it now you say --enable-X/--disable-X . I was reading your previous text as suggesting --enable-foo=no, where we should be getting them to use --enable-foo/--disable-foo. All sorted now.

config/ola.m4
peternewman: Fixed.

config/ola.m4
peternewman: Okay. You've missed my main point though, it should be ${plugin_key}, not explicitly --enable-${plugin_key} which will potentially confuse them if they've only passed in --disable-${plugin_key}.

peternewman: I was trying to mean the status of the plugin had got in an invalid state/one that it shouldn't have.

config/ola.m4
```diff
@@ -110,19 +132,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
...
+      AC_MSG_ERROR([Dependencies for plugin ${plugin_key} are not met.])
```
peternewman: Will $3 have already been evaluated, or can we return the evaluation that's failing to give them a clue what's broken?

config/ola.m4
peternewman: Ah okay, so we'll only build if it's not no.

config/ola.m4
peternewman: Also they shouldn't ever be passing in a value to the arg, just setting different flags.

config/ola.m4
peternewman: Ah yes, my apologies.

configure.ac
peternewman: Is it worth us doing that now, then we can just change the default state when it launches. @nomis52 ?

config/ola.m4
simark: > Okay. You've missed my main point though, it should be ${plugin_key}, not explicitly --enable-${plugin_key} which will potentially confuse them if they've only passed in --disable-${plugin_key}.

Oh ok I understand.

> Also they shouldn't ever be passing in a value to the arg, just setting different flags.

We should encourage them to use --enable_X/--disable_X, but nothing prevents them from writing --enable_X=giraffe. If they do so, it's this message that will print.

config/ola.m4
simark: In the current situation, $3 is just "yes" or "no". We'd need good refactor to get this information.

Just thinking out loud and I have no idea how that's possible in M4/sh. We could pass a list of tuples (in pseudo-Python syntax):

    [("dep1", $have_dep1), ("dep2", $have_dep2)]

In plugin support, we can iterate on the dependencies and print precisely which one is missing.

config/ola.m4
```diff
@@ -110,19 +132,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
...
+  if test "${enable_plugin}" != "yes" -a \
+          "${enable_plugin}" != "no" -a \
+          "${enable_plugin}" != "auto"; then
+    AC_MSG_ERROR([Invalid value for --enable-${plugin_key}/\
+--disable-${plugin_key}. Valid values are yes, no and auto.])
+  fi
+
+  # If dependencies are not met...
   if test "$3" == "no"; then
+    # ...and the user has explicitely requested this plugin to be enabled,
```
peternewman: SPaG explicitly

config/ola.m4
```diff
@@ -110,19 +132,36 @@ AC_DEFUN([PLUGIN_SUPPORT],
...
+    # error out.
+    if test "${enable_plugin}" = "yes"; then
+      AC_MSG_ERROR([Dependencies for plugin ${plugin_key} are not met.])
+    fi
+
+    # Otherwise, force the plugin disabled silently.
```
peternewman: SPaG Otherwise silently force disable the plugin.

config/ola.m4
peternewman: Actually, checking "${enable_plugin}" = "auto" would be a lot clearer here; we don't care about yes, and no is already no, so no point in changing that either.

simark: Right.

config/ola.m4
simark: Fixed (I always do that mistake, because of the spelling in french).

config/ola.m4
simark: Fixed.

config/ola.m4
peternewman: Or a multi-dimensional array (which I guess is roughly the same). Probably one for a seperate PR anyway.

configure.ac
nomis52: I think it's fine as is. The next step will be to remove the conditional entirely.

nomis52:
This looks fine to me. Was there anything else or can we merge it?

config/ola.m4
peternewman: Ah, I may have missed the other half of the message before @simark

peternewman:
LGTM, I think the only thing we possibly talked about was showing the user which dependencies had failed for a plugin, but that can probably wait for another PR.

PR #776, opened 2015-06-08
Don't fail the InterfacePickerTest if we only have the loopback inter…

### LINKED ISSUE #763: Tests fail when no non-loopback interfaces are present
See the Debian bug here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=769670

It probably makes sense to fix this locally, in which case we can remove testGetLoopbackInterfaces, we just need to work out what we do on Windows given: https://github.com/OpenLightingProject/ola/blob/master/common/network/InterfacePickerTest.cpp#L99

Do we skip the test on Windows, or let it run, possibly with a log line on Windows, for the tiny userbase who will run Windows with no network interface. Or ideally fix it up so the loopback interface is reported on Windows too.

nomis52:
…face.

Closes #763


peternewman:
I don't feel we should merge this as is, but I'll continue the discussion in #763 .

nomis52:
See discussion in #763

PR #777, opened 2015-06-09
Switch just Coverity back to g++-4.9 as it doesn't work with g++-5

peternewman:


nomis52:
LGTM

peternewman:
jshint is the only thing that's failing, SEP.

PR #774, opened 2015-06-08
Fix some tests for Python 3

simark:
Python 3 complains that our types are unorderable if they don't define
at least __lt__. That's the only one needed for sorting only (as well as
__eq__ for equality).

These tests now pass:

  $ python -V
  Python 3.4.3
  $ PYTHONPATH=$PWD nosetests ola/DUBDecoderTest.py ola/MACAddressTest.py ola/UIDTest.py
  ..........
  ----------------------------------------------------------------------
  Ran 10 tests in 0.003s

  OK

Nothing seemed to test equality of MACAddresses, so I added a test for
that.

PR #778, opened 2015-06-10
Avoid using sys.version_info.major to keep Python 2.6 support

simark:
Refs #774


nomis52:
This looks fine, the js lint failure is being worked on.

PR #780, opened 2015-06-13
Fix a Coverity bug

peternewman:


olad/PluginManager.cpp
```diff
@@ -88,10 +88,15 @@ void PluginManager::LoadAll() {
...
+        AbstractPlugin *conflicting_plugin = GetPlugin(*set_iter);
         OLA_WARN << "Skipping " << plugin->Name()
-                 << " because it conflicts with "
-                 << GetPlugin(*set_iter)->Name()
-                 << " which is also enabled";
+                 << " because it conflicts with ";
+        if (conflicting_plugin) {
+          OLA_WARN << conflicting_plugin->Name();
+        } else {
+          OLA_WARN << "plugin id " << *set_iter;
```
nomis52: This is going to split the message across 3 lines.

I'd just do:

<< " because it conflicts with "
<< PluginIdentifier(*set_iter) << "which is also enabled";

then

string PluginIdentifier(plugin_id id) {
  AbstractPlugin *plugin = GetPlugin(*set_iter);
  return plugin ? plugin->Name() : "plugin_id: " + id;
}

peternewman: I've gone for an equivalent, but alternative option.

nomis52:
LGTM

peternewman:
Merge if you're happy with the final tweak @nomis52

PR #781, opened 2015-06-14
fix #682 - Plugin Enable/Disable in Web UI

### LINKED ISSUE #682: Add a handler to enable/disable a plugin from the Web and CLI UIs

daveol:
also includes a fix for the adding of the header and the tabs in the universe view, previously these got added by really bad jquery code, now it uses a angular template

 - [x] Implement in the angular code
 - [x] Implement in the plugins view


daveol:
I feel I need to change the look of how the status is displayed, because now it doesn't really say there is interaction in the "lights"

peternewman:
LGTM so far.

daveol:
I think this is done now

javascript/new-src/src/css/style.css
```diff
@@ -217,3 +217,28 @@ div#header-universe > div{
 .settings, .input-group, .save-settings {
  margin-top: 5px;
 }
+
+/* Custom checkboxes for plugins page */
```
nomis52: The cursor is a bit funny. On my chrome it's a hand icon for the entire <tr> apart from the checkbox. I think we should have the standard pointer icon unless it moves over the Info icon or Enable checkbox, then it should be the hand cursor.

javascript/new-src/src/js/app.js
```diff
@@ -713,18 +733,24 @@ angular
...
-        $ola.action.Reload().then();
+        $ola.action.Reload();
         $scope.getInfo();
       };
       $scope.go = function(id) {
         $location.path('/plugin/' + id);
       };
+      $scope.changeStatus = function(id, current) {
+        $ola.post.PluginState(id, current);
+        $scope.Reload();
```
nomis52: We should do a better job on the server side of handling this. We shouldn't be reloading all the plugins just to change the state of a single plugin.  

@peternewman  or I can make the change on the server side so you can remove this call.

nomis52:
Tested with my backend changes. Seems to work.

javascript/new-src/src/css/style.css
peternewman: I've historically found custom pointer support a bit lacking generally, I think Chrome was the only one that did it. Although from memory that was a completely custom .ico pointer.

peternewman: Actually that was custom hotspots where they did or didn't work in Chrome.

peternewman:
LGTM

PR #796, opened 2015-06-22
Open the correct interface on the EurolitePro Widget in async mode.

nomis52:


nomis52:
I checked the other calls and Eurolight is the only one where we dynamically try to find the correct interface #.

peternewman:
LGTM. I thought Coverity might have flagged that as an unused variable, although I guess its like not using a return code.

peternewman:
Do we need to update NEWS, or was this change all in 0.9.6?

nomis52:
Added

Issue #799, opened 2015-06-24
kqueue causes a infinite loop when a usb serial device is removed.


nomis52:

nomis52:
See http://stackoverflow.com/questions/30946728/kevent-usb-serial-ports for details.

I've disabled kqueue for now.

PR #800, opened 2015-06-24
Disable the use of kqueue.

nomis52:
Use of kqueue causes an infinite loop if a USB serial device is removed.
See #799 for details.


common/io/SelectServer.cpp
```diff
@@ -60,8 +60,8 @@ DEFINE_default_bool(use_epoll, true,
 
 #ifdef HAVE_KQUEUE
 #include "common/io/KQueuePoller.h"
-DEFINE_default_bool(use_kqueue, true,
-                    "Disable the use of kqueue(), revert to select()");
+DEFINE_default_bool(use_kqueue, false,
+                    "Use of kqueue() rather than select()");
```
peternewman: SPaG Use kqueue()

peternewman:
Merge when SPaG fixed.

Issue #788, opened 2015-06-16
Avoid obsolete network functions


nomis52:
We should use the newer versions of the network functions where available.

inet_aton -> inet_pton
inet_ntoa -> inet_ntop

nomis52:
The code that uses these is in common/network/IPV4Address.cpp

PR #803, opened 2015-06-29
Add support for auto mode of Ja Rule

peternewman:
Not sure which branch I'm supposed to merge into, I still need to fix input validation too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -504,6 +504,9 @@ AM_CONDITIONAL([BUILD_JAVA_LIBS], [test "x$enable_java_libs" = xyes])
 AC_ARG_ENABLE(
   [ja-rule],
   [AS_HELP_STRING([--enable-ja-rule], [Build the Ja Rule test tool])])
+# Input validation.
+AS_IF([test "x$enable_ja_rule" != xyes && test "x$enable_ja_rule" != xno && "x$enable_ja_rule" != xauto],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -124,9 +124,9 @@ AC_FUNC_STAT
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_VPRINTF
 AC_CHECK_FUNCS([bzero gettimeofday memmove memset mkdir strdup strrchr \
-                if_nametoindex inet_ntoa inet_aton select socket strerror \
-                getifaddrs getloadavg getpwnam_r getpwuid_r getgrnam_r \
-                getgrgid_r secure_getenv])
+                if_nametoindex inet_ntoa inet_ntop inet_aton inet_pton select \
```
peternewman: These are ready for https://github.com/OpenLightingProject/ola/issues/788 .

nomis52:
I think it would be much easier if we just avoid passing --enable-jarule to the pi build slaves.

I want to remove this flag as soon as I can, so there isn't much to gain by trying to make it complicated.

peternewman:
We could do, but that involves a chunk of buildbot copy/paste, and when we remove the flag, we'll still need some logic to build Ja Rule automatically if possible (although I guess that could be in ifdefs in the code).

nomis52:
I'd prefer to copy and paste in buildbot rather than here.

My goal is for the ja-rule code to be build if the libusb plugin is enabled and the appropriate version is met. That way we can get rid of the flag all together.

peternewman:
Okay, understood.

PR #804, opened 2015-06-29
Update the developer doc with information about branches.

nomis52:


README.developer
```diff
@@ -46,6 +46,29 @@ command can be run within a sub directory to only run a particular set of
...
+Branches, Versioning & Releases
+-------------------------------
+
+Version numbers take the form MAJOR.MINOR.PATCH. The release lifecycle is:
+
+- New feature work occurs on the master branch.
+- Once the new features are considered stable or enough time has passed, a new
+  minor release branch will be created, e.g. 0.9.
+- The minor release branch will be stablized with bugfixes, these bug fixes
+  will be merged back into master.
```
peternewman: SPaG will also be

README.developer
```diff
@@ -46,6 +46,29 @@ command can be run within a sub directory to only run a particular set of
...
+- Once declared stable, a new patch branch 0 will be created e.g. 0.9.0
```
peternewman: Can't the patch release numbers just be done with commits and tags to the minor release branch, as we shouldn't make further commits to them/they're going to get merged back into minor anyway?

nomis52: They could, but I suspect we'll start doing X.Y.Zrc1 and the like. Branches are cheap, so it makes sense to use them.

peternewman: If we do that, then we need to expand our version model don't we?

PR #805, opened 2015-06-30
Bump version and cleanup NEWS

nomis52:


NEWS
```diff
@@ -1,11 +1,9 @@
-x/y/2015 ola-0.9.6
+30/6/2015 ola-0.9.6
  Features:
- * Default to using epoll() and kqueue() rather than select() if they exist,
-   apart from for the HTTP thread.
+ * Default to using epoll() rather than select() if it exists,
```
peternewman: Isn't kqueue still a default? Do we not need the note about the HTTP thread?

nomis52: I turned it off until we can get the serial port problems fixed.

I don't think the HTTP stuff is relevant for the NEWS

peternewman: Okay

PR #806, opened 2015-06-30
Debian Updates

nomis52:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+ola (0.9.6-1) unstable; urgency=medium
```
peternewman: I guess in theory this should be urgency low still.

debian/control
```diff
@@ -1,43 +1,63 @@
...
 Architecture: any
 Depends: ola (= ${binary:Version}), ${misc:Depends}
-Description: Open Lighting Architecture
+Description: Open Lighting Architecture - development libraries
+ The DMX512 standard for Digital MultipleX is used for digital
+ communication networks commonly used to control stage lighting and
+ effects.
+ .
  The Open Lighting Architecture (OLA) provides a plugin framework for
  distributing DMX512 control signals on Mac and Linux.
```
peternewman: We should probably add Windows now, plus *BSD?

debian/control
```diff
@@ -1,43 +1,63 @@
...
+Description: Open Lighting Architecture - Python Classes
+ The DMX512 standard for Digital MultipleX is used for digital
+ communication networks commonly used to control stage lighting and
+ effects.
+ .
  The Open Lighting Architecture (OLA) provides a plugin framework for
  distributing DMX512 control signals on Mac and Linux.
  .
- OLA is written and maintained by Simon Newton <nomis52@gmail.com>
+ This package contains the files required for plugins or to control the OLA
```
peternewman: No plugins here.

debian/control
nomis52: Removed the platform bits, since this isn't relevant for the debian packages.

PR #808, opened 2015-07-03
Use inet_ntop and inet_pton where available

peternewman:


common/network/IPV4Address.cpp
```diff
@@ -93,7 +95,17 @@ bool IPV4Address::IsWildcard() const {
 string IPV4Address::ToString() const {
   struct in_addr addr;
   addr.s_addr = m_address;
+#ifdef HAVE_INET_NTOP
+  char str[INET_ADDRSTRLEN];
+  if (inet_ntop(AF_INET, &addr, str, INET_ADDRSTRLEN) == NULL) {
+    OLA_WARN << "Failed to convert address to string using inet_ntop, failing "
```
nomis52: Do you think we need the fallback here? From reading the man page I can't see how inet_ntoa would pass if inet_ntop failed. 

I'd just remove the warning and return false.

peternewman: We can't return false, we could return the empty string, but that's confusing. If somehow "ENOSPC The converted address string would exceed the size given by size." then inet_ntoa would still succeed.

nomis52:
LGTM

PR #809, opened 2015-07-03
Add the missing views HTML file

peternewman:
I don't know if we want this in some other branches too, as a patch?


peternewman:
Sounds like we may have more files than just this then @nomis52 ?

nomis52:
Oh, this was merged into master, not 0.9 . I was wondering why my dist verify script kept failing on this :)

peternewman:
Yes I was just looking at the other PR @nomis52 and seeming like it seemed familiar. We need to get them back in line somehow via cherry picking or other means.

PR #811, opened 2015-07-04
Bugfixes for the s390x and ppc architecture

nomis52:


configure.ac
```diff
@@ -242,6 +242,9 @@ if test -z "${USING_WIN32_FALSE}" && test "${have_msg_no_signal}" = "no" && \
  AC_MSG_ERROR([Your system needs either MSG_NOSIGNAL or SO_NOSIGPIPE])
 fi
 
+AC_CHECK_TYPE(termios2,
```
peternewman: Will you not need to include asm/termios.h here to find it?

plugins/usbdmx/LibUsbThreadTest.cpp
```diff
@@ -87,6 +87,10 @@ void LibUsbThreadTest::testNonHotplug() {
 
 #if defined(LIBUSB_API_VERSION) && (LIBUSB_API_VERSION >= 0x01000102)
 void LibUsbThreadTest::testHotplug() {
+  if (!m_context) {
+    return;
```
peternewman: Can we have a message here so we know its not running all the tests.

peternewman:
LGTM aside from comments.

configure.ac
nomis52: Yup, added :)

plugins/usbdmx/LibUsbThreadTest.cpp
nomis52: There are other reasons for this not running, like the #if checks. I don't want to get into the pattern of adding logs for all of these.

peternewman: But there's two classes of skipping here, the #if, which means the code in olad won't run either and this check, which means the olad code should run if it's got access to /sys or whatever is blocked here. I just think we should warn where we've got a coverage gap and the tests aren't fully testing the code you're able to run in olad on that machine.

nomis52: We already have the log line on 68 that lets us infer this.

peternewman: Ah yes, fair point.

PR #813, opened 2015-07-07
Fix some little-endian bugs.

nomis52:


common/messaging/MessagePrinterTest.cpp
```diff
@@ -96,7 +97,9 @@ void GenericMessagePrinterTest::testSimplePrinter() {
   // try a simple print first
   vector<const ola::messaging::MessageFieldInterface*> fields;
   fields.push_back(new BoolMessageField(&bool_descriptor, false));
-  fields.push_back(new IPV4MessageField(&ipv4_descriptor, 0x0100000a));
+  fields.push_back(
+      new IPV4MessageField(&ipv4_descriptor,
+                           ola::network::HostToNetwork(0x0a000001)));
```
peternewman: Is it worth going the whole hog and passing in an IPV4Address object, so it's more like the UID line below?

peternewman:
LGTM aside from comment.

common/messaging/MessagePrinterTest.cpp
nomis52: I don't think it gains us anything and it's more to type.

PR #817, opened 2015-07-09
Add a travis check to verify we don't miss files from the tarball.

nomis52:


.travis-ci.sh
```diff
@@ -69,5 +69,12 @@ elif [[ $TASK = 'jshint' ]]; then
   grunt test
 else
   # Otherwise compile and check as normal
-  autoreconf -i && ./configure --enable-rdm-tests --enable-ja-rule && make distcheck DISTCHECK_CONFIGURE_FLAGS='--enable-rdm-tests --enable-ja-rule'
+  autoreconf -i;
```
peternewman: You should keep the ampersands, there's no point running configure if autoreconf failed etc.

olad/www/Makefile.mk
```diff
@@ -76,10 +77,10 @@ dist_angular_DATA = \
 dist_bootjs_DATA = \
     olad/www/new/libs/bootstrap/js/bootstrap.min.js
 dist_bootfonts_DATA = \
-    olad/www/new/libs/bootstrap/fonts/glyphicons-halflings-regular.woff \
+    olad/www/new/libs/bootstrap/fonts/glyphicons-halflings-regular.eot \
     olad/www/new/libs/bootstrap/fonts/glyphicons-halflings-regular.svg \
     olad/www/new/libs/bootstrap/fonts/glyphicons-halflings-regular.ttf \
-    olad/www/new/libs/bootstrap/fonts/glyphicons-halflings-regular.eot \
+    olad/www/new/libs/bootstrap/fonts/glyphicons-halflings-regular.woff \
```
peternewman: I thought this seemed familiar; as we've both worked out I'd done it here first https://github.com/OpenLightingProject/ola/pull/809 .

scripts/verify_trees.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/python
+#  This program is free software; you can redistribute it and/or modify
```
peternewman: Our Python standard is only one space before the license.

scripts/verify_trees.py
```diff
@@ -0,0 +1,124 @@
...
+
+import os
+import fnmatch
+import textwrap
+import sys
+
+# File & directory patterns that differ between what's in the git repo and
+# what's in the tarball.
+IGNORE_PATTERNS = [
+  '*.log',
```
peternewman: Surely this shouldn't be in either, or I guess actually you should update your comment to say a checkout of the git repo that had had make and make check run. We should probably also note that while this may be similar to, it shouldn't match .gitignore.

daveol:
I'm not sure we need the files given, it depends on what you want to do with the tarball... do you want users to make adjustments with them only having a tarball? If yes we should include them because they are technically part of the build process (except we've done it already) since the gruntfile contains the minify job for the source of the JavaScript.

Off topic:
- Maybe a good idea to combine all the JavaScript in one file? so the Angular, jQuery, Bootstrap and our JavaScript sources just get delivered in one big file.
- Can we also have a check if people minified their JavaScript/css (or add a  commit hook for it)?

peternewman:
So the errant files are:
javascript/new-src/bower.json
javascript/new-src/package.json
javascript/new-src/Gruntfile.js
javascript/new-src/.jshintrc
olad/www/new/js/app.min.js.map
javascript/new-src/README.md - Include, same as other readme files.
javascript/new-src/.jscsrc

The .map file just helps debugging doesn't it @daveol ?

.travis-ci.sh
nomis52: I used set -e. The use of && was getting clumsy.

scripts/verify_trees.py
nomis52: Not sure what happened there. Fixed.

scripts/verify_trees.py
nomis52: Yeah, this includes all the build artifacts, like trs, .deps etc. I don't want to run to clean the build every time this runs.

PR #819, opened 2015-07-09
Various fixes

### LINKED ISSUE #812: RDM NACK codes aren't correctly reported in 0.9.6
As mentioned here:
https://groups.google.com/d/msg/open-lighting/ra-jK5hZ8z4/f2cFHDTNUrgJ

It seems to be this commit:
https://github.com/OpenLightingProject/ola/commit/d971464004622c777af6062e6fa1eff122818345#diff-d63f9b92a627a457f40db37fa20e6550L317

Is there any reason we've dropped NackReason or not added one to RDMResponse? If we're not going to go down that route, I guess we just need a NetworkToHost around the ParamData call, but it would be nice to put it server/API side rather than client.

nomis52:
Fix the little endian bug in the shownet code
Clean up the debian metadata files.
Fix the endian bug in the C++ RDM client - closes #812

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,7 @@ Description: Open Lighting Architecture - development libraries
  effects.
  .
  The Open Lighting Architecture (OLA) provides a plugin framework for
- distributing DMX512 control signals on Mac and Linux.
+ distributing DMX512 control signals.
  .
  This package contains the development files for libola and various related
  libraries, used to write plugins or control the OLA daemon.
@@ -34,10 +34,9 @@ Description: Open Lighting Architecture - Python Classes
  effects.
  .
  The Open Lighting Architecture (OLA) provides a plugin framework for
- distributing DMX512 control signals on Mac and Linux.
+ distributing DMX512 control signals.
  .
- This package contains the files required for plugins or to control the OLA
- daemon from python.
+ This package contains the Python client module.
 
 Package: ola-rdm-tests
 Section: libs
@@ -54,7 +53,7 @@ Description: Open Lighting Architecture - RDM Responder Tests
  without disturbing other devices on the same connection.
  .
  The Open Lighting Architecture (OLA) provides a plugin framework for
- distributing DMX512 control signals on Mac and Linux.
+ distributing DMX512 control signals.
  .
  This package provides an automated way to check protocol compliance in
  RDM devices.
@@ -70,7 +69,7 @@ Description: Open Lighting Architecture
  effects.
  .
  The Open Lighting Architecture (OLA) provides a plugin framework for
- distributing DMX512 control signals on Mac and Linux.
+ distributing DMX512 control signals.
  .
  This package contains olad, the OLA daemon used to control lighting,
  and a number of command-line tools to control and manipulate olad.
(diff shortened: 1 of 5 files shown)


examples/ola-rdm.cpp
```diff
@@ -315,7 +317,7 @@ void RDMController::HandleResponse(const ola::client::Result &result,
     } else {
       memcpy(reinterpret_cast<uint8_t*>(&backoff_time), response->ParamData(),
              sizeof(backoff_time));
-      unsigned int timeout = 100 * backoff_time;
+      unsigned int timeout = 100 * NetworkToHost(backoff_time);
```
peternewman: This is feeling more and more to me like we should be providing a method as part of RDMResponse to do most of the leg work for you; it's a lot of potential for bugs for someone to implement in their own client.

nomis52: Maybe, that doesn't belong in the 0.9 branch though.

examples/ola-rdm.cpp
```diff
@@ -349,6 +351,7 @@ void RDMController::HandleResponse(const ola::client::Result &result,
     } else {
       memcpy(reinterpret_cast<uint8_t*>(&nack_reason), response->ParamData(),
              sizeof(nack_reason));
+      nack_reason = NetworkToHost(nack_reason);
```
peternewman: Likewise here.

plugins/shownet/ShowNetNode.cpp
```diff
@@ -321,16 +321,18 @@ unsigned int ShowNetNode::BuildCompressedPacket(shownet_packet *packet,
...
+  compressed_dmx->slotSize[0] = HostToLittleEndian(
+      static_cast<uint16_t>(buffer.Size()));
 
   unsigned int enc_len = sizeof(packet->data);
   if (!m_encoder.Encode(buffer, compressed_dmx->data, &enc_len))
     OLA_WARN << "Failed to encode all data (used " << enc_len << " bytes";
 
-  compressed_dmx->indexBlock[0] = HostToLittleEndian(MAGIC_INDEX_OFFSET);
+  compressed_dmx->indexBlock[0] = HostToLittleEndian(
+      static_cast<uint16_t>(MAGIC_INDEX_OFFSET));
```
peternewman: I'm kind of surprised this needs the cast, or does it get treated as an int otherwise?

debian/control
```diff
@@ -19,7 +19,7 @@ Description: Open Lighting Architecture - development libraries
  effects.
  .
  The Open Lighting Architecture (OLA) provides a plugin framework for
- distributing DMX512 control signals on Mac and Linux.
```
peternewman: Good call.

plugins/shownet/ShowNetNode.cpp
nomis52: It doesn't, this is defensive in case anyone changes the type later.

examples/ola-rdm.cpp
peternewman: Okay.

plugins/shownet/ShowNetNode.cpp
peternewman: Ah okay, presumably they were all being treated as ints before and hence breaking?

peternewman:
LGTM

plugins/shownet/ShowNetNode.cpp
nomis52: Yup

PR #821, opened 2015-07-10
More NEWS

peternewman:


nomis52:
This should be going into the 0.9 branch, so when we cut 0.9.7 we don't have to re-do this.

peternewman:
Doh, that's where I'd meant to aim it.

PR #823, opened 2015-07-10
let's see what breaks! (using travis containers) 

daveol:
let's see what actually breaks in travis if we try to use the container based... thing


daveol:
okay I've hit my first block I think... there is no easy way to install python-protobuf without compiling it....

peternewman:
That's not python-protobuf, its protobuf proper that's sticking on.

daveol:
okay now the build fails out because libusb isn't recent enough for ja-rule support

peternewman:
So does it work cleanly if you remove the --enable-ja-rule configure options?

peternewman:
Also we're now on clang-3.7 via "add-apt-repository "deb http://llvm.org/apt/precise/ llvm-toolchain-precise main" -y" which I think is already in their list.

daveol:
it's green! it's alive!

.travis.yml
```diff
@@ -42,36 +42,52 @@ env:
...
- - sudo apt-get update -qq
- - sudo apt-get install -qq protobuf-compiler libprotobuf-lite7 libprotobuf-dev libprotoc-dev bison flex libftdi-dev libftdi1 libusb-1.0-0-dev liblo-dev libavahi-client-dev uuid-dev libcppunit-dev libcppunit-1.12-1 libncurses5-dev python-protobuf libmicrohttpd-dev
 #Coverity doesn't work with g++ 5, so only upgrade to g++ 4.9 for that
- - if [ "$TASK" != "coverity" -a "$CXX" = "g++" ]; then sudo apt-get install -qq g++-5; fi
  - if [ "$TASK" != "coverity" -a "$CXX" = "g++" ]; then export CXX="g++-5" CC="gcc-5"; fi
- - if [ "$TASK" = "coverity" -a "$CXX" = "g++" ]; then sudo apt-get install -qq g++-4.9; fi
  - if [ "$TASK" = "coverity" -a "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="gcc-4.9"; fi
-#Remove the old g++/gcc to ensure we're using the latest ones
- - if [ "$CXX" = "g++-4.9" -o "$CXX" = "g++-5" ]; then sudo rm /usr/bin/g++; sudo rm /usr/bin/gcc; fi
 #Install latest clang if we're compiling with clang
```
nomis52: Change 'install' to something else in the comment

.travis.yml
```diff
@@ -42,36 +42,52 @@ env:
...
+      - uuid-dev
+      - libcppunit-dev
+      - libncurses5-dev
+      - python-protobuf
+      - libmicrohttpd-dev
+      - doxygen
+      - graphviz
+      - g++-5
+      - g++-4.9
+      - clang-3.6
```
nomis52: Can you add a comment as to why we use clang-3.6 rather than clang-3.7

peternewman:
Have we got an issue open for the later libusb @daveol ? Can you mention it in here. It may also be worth waiting for https://github.com/travis-ci/apt-package-whitelist/issues/474 to go in, depending on how it looks time wise.

daveol:
yes I believe it was the ppa:ubuntu-cloud-archive/icehouse-staging that provided a more recent one @peternewman . I made a issue to add the ppa:ubuntu-cloud-archive/icehouse-staging to the source whitelist travis-ci/apt-source-whitelist#21

.travis-ci.sh
```diff
@@ -57,7 +57,7 @@ elif [[ $TASK = 'doxygen' ]]; then
 elif [[ $TASK = 'coverage' ]]; then
   # Compile with coverage for coveralls
   autoreconf -i;
-  ./configure --enable-ja-rule --enable-gcov;
+  ./configure --enable-rdm-tests --enable-gcov # --enable-ja-rule;
```
peternewman: Coverage is only C++ based currently, hence we skip the Python based RDM tests.

.travis-ci.sh
```diff
@@ -76,8 +76,8 @@ elif [[ $TASK = 'jshint' ]]; then
 else
   # Otherwise compile and check as normal
   autoreconf -i;
-  ./configure --enable-rdm-tests --enable-ja-rule;
-  make distcheck DISTCHECK_CONFIGURE_FLAGS='--enable-rdm-tests --enable-ja-rule';
+  ./configure --enable-rdm-tests # --enable-ja-rule;
```
peternewman: Can you resync this to use your new improved DISTCHECK_CONFIGURE_FLAGS

peternewman:
@daveol this seems to be a way to avoid installing packages like Doxygen on matrix lines that don't need them, which may speed us up slightly https://github.com/feelpp/feelpp/blob/develop/.travis.yml

nomis52:
Merging.

PR #827, opened 2015-07-12
Restore binary compatibility with 0.9.5.

### LINKED ISSUE #826: ola::io::selectserver broke binary compatibility
Hi guys, this commit
3026c750faa25f557444778e66f4c93dbaabb0cd
broke binary compatibility and this is the result:
http://www.qlcplus.org/forum/viewtopic.php?f=5&t=8713

This happened between OLA 0.9.5 and OLA 0.9.6.
Unfortunately things like that happen on OSX but don't happen in Linux.

What should I do ? Rebuild the QLC+ plugin on OLA 0.9.6 and hope it will work with versions <= 0.9.5 ?

nomis52:
Closes #826

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,9 +1,10 @@
-11/7/2015 ola-0.9.7
+12/7/2015 ola-0.9.7
  Bugs:
  * RDM NACK codes displayed with incorrect byte ordering in the C++ client #812
  * Fix endian issues on s390x and PPC architectures.
  * Fix ShowNet endian issues.
  * Add some missing files to the tarball.
+ * Restore binary compatability with 0.9.5 #826
 
 30/6/2015 ola-0.9.6
  Features:
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-12/7/2015 ola-0.9.7
+x/y/2015 ola-0.9.7
  Bugs:
  * RDM NACK codes displayed with incorrect byte ordering in the C++ client #812
  * Fix endian issues on s390x and PPC architectures.
--- a/NEWS
+++ b/NEWS
@@ -1,21 +1,9 @@
-x/y/2015 ola-0.9.7
- Features:
- * 
-
- API:
- * 
-
- RDM Tests:
- * 
-
+11/7/2015 ola-0.9.7
  Bugs:
  * RDM NACK codes displayed with incorrect byte ordering in the C++ client #812
  * Fix endian issues on s390x and PPC architectures.
  * Fix ShowNet endian issues.
  * Add some missing files to the tarball.
-
- Internal:
- *
 
 30/6/2015 ola-0.9.6
  Features:
--- a/common/io/SelectServer.cpp
+++ b/common/io/SelectServer.cpp
@@ -321,8 +321,9 @@ void SelectServer::Init(const Options &options) {
 
   // TODO(simon): this should really be in an Init() method that returns a
   // bool.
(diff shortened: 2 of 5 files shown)


common/io/SelectServer.cpp
```diff
@@ -313,6 +276,58 @@ void SelectServer::DrainCallbacks() {
...
+
+  // Default to the SelectPoller
+  if (!m_poller.get()) {
+    m_poller.reset(new SelectPoller(m_export_map, m_clock));
+  }
+#endif
+
+  // TODO(simon): this should really be in an Init() method that returns a
+  // bool.
+  if (!m_incoming_descriptor.Init())
```
peternewman: {} at least please.

NEWS
```diff
@@ -1,21 +1,10 @@
...
- Features:
- * 
-
- API:
- * 
-
- RDM Tests:
- * 
-
+12/7/2015 ola-0.9.7
```
peternewman: As soon as that. We need to get this resynced into master and gets the bots to run across it before we release.

common/io/SelectServer.cpp
```diff
@@ -313,6 +276,59 @@ void SelectServer::DrainCallbacks() {
...
+#ifdef HAVE_EPOLL
+  if (FLAGS_use_epoll && !options.force_select) {
+    m_poller.reset(new EPoller(m_export_map, m_clock));
+  }
+  if (m_export_map) {
+    m_export_map->GetBoolVar("using-epoll")->Set(FLAGS_use_epoll);
+  }
+#endif
+
+#ifdef HAVE_KQUEUE
```
peternewman: Is it worth indenting these for clarity?

common/io/SelectServer.cpp
```diff
@@ -75,62 +75,25 @@ using std::max;
...
 
-  // TODO(simon): this should really be in an Init() method.
-  if (!m_incoming_descriptor.Init())
-    OLA_FATAL << "Failed to init LoopbackDescriptor, Execute() won't work!";
-  m_incoming_descriptor.SetOnData(
-      ola::NewCallback(this, &SelectServer::DrainAndExecute));
-  AddReadDescriptor(&m_incoming_descriptor);
+SelectServer::SelectServer(const Options &options)
+    : m_export_map(options.export_map),
+      m_terminate(false),
```
peternewman: This seems rather repetitive; can't we move the initialisation into the Init function?

peternewman: This seems rather repetitive; can't we move the initialisation into the Init function?

peternewman:
LGTM, just a few comments. We need to get this resynced into master and gets the bots to run across it before we release.

NEWS
nomis52: Changed back to x/y . I plan on releasing tomorrow.

common/io/SelectServer.cpp
nomis52: Nope, this is faster since it avoids an initialization & assignment.

C++11 avoids this, but we can't use those features (yet).

common/io/SelectServer.cpp
nomis52: No, #defines are always non-indented.

peternewman: Okay.

common/io/SelectServer.cpp
peternewman: Okay.

PR #839, opened 2015-07-20
Correcting Xcode build errors

rawbengal:
These two changes correct some build errors when including OLA headers in other programs.

### DIFF
--- a/include/ola/client/ClientArgs.h
+++ b/include/ola/client/ClientArgs.h
@@ -90,9 +90,9 @@ struct SendDMXArgs {
   /**
    * @brief Create a new SendDMXArgs object
    */
-  explicit SendDMXArgs(GeneralSetCallback *callback)
+  explicit SendDMXArgs(GeneralSetCallback *_callback)
       : priority(ola::dmx::SOURCE_PRIORITY_DEFAULT),
-        callback(callback) {
+        callback(_callback) {
   }
 };
 
@@ -111,8 +111,8 @@ struct SendRDMArgs {
    */
   bool include_raw_frames;
 
-  explicit SendRDMArgs(RDMCallback *callback)
-    : callback(callback),
+  explicit SendRDMArgs(RDMCallback *_callback)
+    : callback(_callback),
       include_raw_frames(false) {
   }
 };
--- a/include/ola/messaging/StringMessageBuilder.h
+++ b/include/ola/messaging/StringMessageBuilder.h
@@ -38,7 +38,7 @@ class StringMessageBuilder: public FieldDescriptorVisitor {
     explicit StringMessageBuilder(const std::vector<std::string> &input)
         : m_input(input) {
     }
-    ~StringMessageBuilderVisitor() {}
+    ~StringMessageBuilder() {}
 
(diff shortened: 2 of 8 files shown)


peternewman:
Thanks @rawbengal , we need to cherry-pick this into 0.9 when it's in.

include/ola/client/ClientArgs.h
```diff
@@ -90,9 +90,9 @@ struct SendDMXArgs {
   /**
    * @brief Create a new SendDMXArgs object
    */
-  explicit SendDMXArgs(GeneralSetCallback *callback)
+  explicit SendDMXArgs(GeneralSetCallback *_callback)
```
peternewman: We normally name the field m_, would you mind doing that please and dropping the leading underscores from the constructor arguments.

include/ola/messaging/StringMessageBuilder.h
```diff
@@ -38,7 +38,7 @@ class StringMessageBuilder: public FieldDescriptorVisitor {
     explicit StringMessageBuilder(const std::vector<std::string> &input)
         : m_input(input) {
     }
-    ~StringMessageBuilderVisitor() {}
+    ~StringMessageBuilder() {}
```
peternewman: Good catch, was it Clang reported this (and if so which version), or Xcode itself?

include/ola/client/ClientArgs.h
nomis52: Not for option structs. We shouldn't force the clients to do:

options.m_priority = X;

but rather:

options.priority = X;

I don't mind what the constructor args are called. I would have gone with callback_arg but _callback works as well.

nomis52:
Can you please close this PR and open another one targeting the 0.9. This is a bugfix so it belongs there. It'll get merged back into the mainline later.

include/ola/client/ClientArgs.h
peternewman: Ah okay.

include/ola/messaging/StringMessageBuilder.h
rawbengal: Xcode.  I don't recall seeing any warnings when compiling the library...  just when using the header in an Xcode project.

rawbengal:
Copy that.  Closing this PR and will target 0.9.   It might take me a bit to figure out how to do move the change to 0.9.  Git is rarely obvious to me.

peternewman:
@rawbengal look at git cherry-pick, if you switch to the 0.9 branch, you can then fetch, rather than pull your master an then cherry pick the commits from it.

nomis52:
When you open the PR, you can select which branch you target. See  https://help.github.com/articles/using-pull-requests/#changing-the-branch-range-and-destination-repository  for screenshot

peternewman:
Unfortunately he's too late for that @nomis52 I tried another comparison against 0.9 and he's initially pulled from our master, so merging in like that ends up merging all our changes from master into 0.9. I'd suggest we re-open this, merge given we seem to be happy and cherry-pick ourselves.

nomis52:
https://github.com/OpenLightingProject/ola/pull/840 is correct

PR #841, opened 2015-07-22
Fix some warnings on Windows

peternewman:


peternewman:
@lukase after fixing these bits, I don't seem to have any issues compiling with -Werror enabled in Makefile.am. Which compiler and protobuf versions did you have the issue with?

PR #842, opened 2015-07-22
Add the new Model ID pids

nomis52:


data/rdm/draft_pids.proto
```diff
@@ -849,4 +849,4 @@ pid {
   }
   get_sub_device_range: ROOT_DEVICE
 }
-version: 1432052711
+version: 1437580798
```
peternewman: I normally revert and don't bother committing these.

peternewman:
LGTM

PR #845, opened 2015-07-25
Fix Travis Clang version

peternewman:


nomis52:
It looks like we're going back to 3.6 with the VM travis change.

peternewman:
This will work now, as it's running the same file as master.

PR #852, opened 2015-07-27
Add some more E1.37-2 tests

peternewman:


peternewman:
@nomis52 can I have an initial review please, before I get too far along?

tools/rdm/TestMixins.py
```diff
@@ -125,16 +161,36 @@ def VerifyResult(self, response, fields):
...
+      self.SetFailed(
+          '%s field in %s was longer than expected, was %d, expected %d' %
+          (self.EXPECTED_FIELD.capitalize(), self.PID,
+           len(fields[self.EXPECTED_FIELD]), self.MAX_LENGTH))
+
 class GetWithDataMixin(object):
-  """GET a PID with junk param data."""
+  """GET a PID with junk param data.
+
+    If ALLOWED_NACK is defined, this adds an additional custom NackGetResult to
```
nomis52: I'd change to ALLOWED_NACKS which is a list so we can allow more than one later if / when we need to.

tools/rdm/TestMixins.py
```diff
@@ -853,28 +909,46 @@ def Test(self):
...
 
     Subclasses must define EXPECTED_FIELD, which is the field to validate the
     index against and DESCRIPTION_FIELD, which is the field to check for
     unprintable characters.
+
+    If ALLOWED_NACK is defined, this adds a custom NackGetResult to the list of
+    allowed results.
   """
   CATEGORY = TestCategory.DIMMER_SETTINGS
+  ALLOWED_NACK = None
```
nomis52: Same here, make this plural and default to the empty list.

tools/rdm/TestMixins.py
```diff
@@ -853,28 +909,46 @@ def Test(self):
...
+  FIRST_INDEX_OFFSET = 1
 
-  def NumberOfSettings(self):
+  def ListOfSettings(self):
     # By default we use the first property from REQUIRES
-    return self.Property(self.REQUIRES[0])
+    if isinstance(self.Property(self.REQUIRES[0]), list):
```
nomis52: The implicit behavior here is confusing to begin with and this makes it harder.

How about we say that each test implements the ListOfSettings() method. Then many of them can be 

def ListOfSettings(self):
  return self.Property(self.REQUIRES[0])

and some can be:

def ListOfSettings(self):
  max_count = self.Property(self.REQUIRES[0])
  if max_count is None:
    return []
  return xrange(0, max_count)

I think it's better to be explicit about these, even if it is slightly more code.

tools/rdm/TestDefinitions.py
```diff
@@ -6473,14 +6433,177 @@ class GetListInterfacesWithData(TestMixins.GetWithDataMixin,
...
+  MIN_LENGTH = RDM_MIN_HOSTNAME_LENGTH
+  MAX_LENGTH = RDM_MAX_HOSTNAME_LENGTH
+
+class GetDNSHostnameWithData(TestMixins.GetWithDataMixin,
+                             OptionalParameterTestFixture):
+  """Get DNS hostname with param data."""
+  CATEGORY = TestCategory.ERROR_CONDITIONS
+  PID = 'DNS_HOSTNAME'
+  # Allow NR_HARDWARE_FAULT in case they're checking length then PDL
+  ALLOWED_NACK = RDMNack.NR_HARDWARE_FAULT
```
nomis52: I'd remove this until we find a device that does this. With the order of checks we're going to spec in E1.20 this will go away.

tools/rdm/TestDefinitions.py
```diff
@@ -6473,14 +6433,177 @@ class GetListInterfacesWithData(TestMixins.GetWithDataMixin,
...
+
+# TODO(Peter): Need to restrict these somehow so we don't saw off the branch
+#class SetDnsHostname(TestMixins.UnsupportedSetMixin,
```
nomis52: Changing DNS hostname should not result in a disconnect, even in E1.33

tools/rdm/TestMixins.py
peternewman: I was beginning to think I might need to do this anyway, so don.e

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Changed to a stub method and two subclasses that implement the two options.

tools/rdm/TestDefinitions.py
peternewman: It seems we do for starters, as we run a custom check regarding the length, then pass it onto our general string handler. Shall we therefore leave it in, given others may code it in the same way? Or is the updated E1.20 going to mandate checking for format errors first?

nomis52: Looks like you already fixed it - that's what I would have suggested . I think we can remove it now, thanks.

peternewman: I've removed it already.

peternewman:
Aside from https://github.com/OpenLightingProject/ola/issues/854 these additions/changes are clean on all the dummy responders now.

peternewman:
And the SPI responder.

tools/rdm/TestMixins.py
```diff
@@ -857,28 +913,38 @@ def Test(self):
...
-  def NumberOfSettings(self):
-    # By default we use the first property from REQUIRES
-    return self.Property(self.REQUIRES[0])
+  def ListOfSettings(self):
+    self.SetBroken('base method of GetSettingDescriptionsMixin called')
 
   def Test(self):
-    count = self.NumberOfSettings()
+    count = len(self.ListOfSettings())
     if count is None:
```
nomis52: indices = self.ListOfSettings()
if not indices:
  ....

tools/rdm/TestMixins.py
```diff
@@ -907,3 +976,49 @@ def VerifyResult(self, response, fields):
...
+  """Perform a GET for each setting in a range.
+
+    The range is a count, it will check FIRST_INDEX_OFFSET to
+    FIRST_INDEX_OFFSET + NumberOfSettings().
+
+    Subclasses must define EXPECTED_FIELD, which is the field to validate the
+    index against and DESCRIPTION_FIELD, which is the field to check for
+    unprintable characters.
+
+    If ALLOWED_NACKS is non-empty, this adds a custom NackGetResult to the list of
```
nomis52: Remove the parts of the comment that are covered in the base class

e.g. Subclasses must define EXPECTED_FIELD,...
If ALLOWED_NACKS is non-empty, ....

tools/rdm/TestMixins.py
```diff
@@ -907,3 +976,49 @@ def VerifyResult(self, response, fields):
...
+      return range(self.FIRST_INDEX_OFFSET,
+                   self.NumberOfSettings() + self.FIRST_INDEX_OFFSET)
+
+class GetSettingDescriptionsMixinList(GetSettingDescriptionsMixin):
+  """Perform a GET for each setting in a list.
+
+    The list is an array of settings, which don't need to be 
+    sequential
+
+    Subclasses must define EXPECTED_FIELD, which is the field to validate the
```
nomis52: Same here

tools/rdm/TestMixins.py
```diff
@@ -907,3 +976,49 @@ def VerifyResult(self, response, fields):
            self.DESCRIPTION_FIELD,
            self.current_item,
            fields[self.DESCRIPTION_FIELD].encode('string-escape')))
+
+
+class GetSettingDescriptionsMixinRange(GetSettingDescriptionsMixin):
```
nomis52: The convention I've tried to maintain is that mixins end with 'Mixin', so use GetSettingDescriptionsRangeMixin.

Same for the one below as well

tools/rdm/TestMixins.py
```diff
@@ -59,16 +60,33 @@ def Test(self):
 class GetMixin(object):
   """GET Mixin for an optional PID. Verify EXPECTED_FIELD is in the response.
 
-    This mixin also sets a property if PROVIDES is defined.  The target class
-    needs to defined EXPECTED_FIELD and optionally PROVIDES.
+    This mixin also sets a property/mulitiple properties if PROVIDES is
```
nomis52: sets one or more properties if ...

tools/rdm/TestDefinitions.py
```diff
@@ -4964,22 +4930,13 @@ def VerifyResult(self, response, fields):
...
-    # We can't use the GetWithDataMixin because NR_UNSUPPORTED_COMMAND_CLASS is
-    # a valid response here.
-    self.AddIfGetSupported([
-      self.NackGetResult(RDMNack.NR_UNSUPPORTED_COMMAND_CLASS),
-      self.NackGetResult(RDMNack.NR_FORMAT_ERROR),
-      self.AckGetResult(
-        warning='Get %s with data returned an ack' % self.pid.name)
-    ])
-    self.SendRawGet(PidStore.ROOT_DEVICE, self.pid, self.DATA)
+  ALLOWED_NACKS = [RDMNack.NR_UNSUPPORTED_COMMAND_CLASS]
```
nomis52: Just add a note that some responders may not let you GET the pin.

tools/rdm/TestDefinitions.py
```diff
@@ -6473,14 +6437,150 @@ class GetListInterfacesWithData(TestMixins.GetWithDataMixin,
...
+#------------------------------------------------------------------------------
+class GetIPv4DefaultRoute(TestMixins.GetMixin,
+                          OptionalParameterTestFixture):
+  """GET the IPv4 default route."""
+  # TODO(Peter): Check interface identifier is a valid interface
+  CATEGORY = TestCategory.IP_DNS_CONFIGURATION
+  PID = 'IPV4_DEFAULT_ROUTE'
+  EXPECTED_FIELD = ['ipv4_address', 'interface_identifier']
+  # TODO(Peter): Is this required?
+  #ALLOWED_NACKS = [RDMNack.NR_HARDWARE_FAULT]
```
nomis52: Not required, remove

tools/rdm/TestMixins.py
```diff
@@ -797,22 +853,22 @@ def ResetState(self):
     self._wrapper.Run()
 
 class GetZeroByteMixin(object):
-  """Get a single byte parameter with value 0, expect NR_DATA_OUT_OF_RANGE"""
+  """Get a parameter with value 0, defaults to a byte, expect NR_DATA_OUT_OF_RANGE"""
   CATEGORY = TestCategory.ERROR_CONDITIONS
+  DATA = struct.pack('!B', 0)
```
nomis52: Don't redefine this - Byte means 8 bits. If you want to create a GetZeroInt32Mixin I'm ok with that.

tools/rdm/TestDefinitions.py
```diff
@@ -6473,14 +6437,150 @@ class GetListInterfacesWithData(TestMixins.GetWithDataMixin,
...
-    self.SendRawSet(ROOT_DEVICE, self.pid)
+    self.AddIfGetSupported(self.NackGetResult(RDMNack.NR_FORMAT_ERROR))
+    self.SendRawGet(ROOT_DEVICE, self.pid, 'foobar')
+
+class GetZeroInterfaceLabel(TestMixins.GetZeroByteMixin,
+                            OptionalParameterTestFixture):
+  """GET INTERFACE_LABEL for interface 0."""
+  CATEGORY = TestCategory.ERROR_CONDITIONS
+  PID = 'INTERFACE_LABEL'
+  DATA = struct.pack('!L', 0)
```
nomis52: See other comment - don't change the width from 8 to 32, use a separate mixin instead.

tools/rdm/TestMixins.py
```diff
@@ -77,6 +95,9 @@ class GetStringMixin(GetMixin):
     This mixin also sets a property if PROVIDES is defined.  The target class
     needs to defined EXPECTED_FIELD and optionally PROVIDES.
   """
+  MIN_LENGTH = 0
+  MAX_LENGTH = RDM_MAX_STRING_LENGTH
+
   def VerifyResult(self, response, fields):
     if response.WasAcked() and self.PROVIDES:
```
nomis52: Add 

string_field = fields[self.EXPECTED_FIELD]

so we can avoid repeating fields[self.EXPECTED_FIELD] so many times.

tools/rdm/TestMixins.py
```diff
@@ -90,6 +111,18 @@ def VerifyResult(self, response, fields):
           (self.EXPECTED_FIELD.capitalize(), self.PID,
            fields[self.EXPECTED_FIELD].encode('string-escape')))
 
+    if self.MIN_LENGTH and len(fields[self.EXPECTED_FIELD]) < self.MIN_LENGTH:
```
nomis52: This can then become 

if self.MIN_LENGTH and len(string_field) < self.MIN_LENGTH:

tools/rdm/TestMixins.py
```diff
@@ -90,6 +111,18 @@ def VerifyResult(self, response, fields):
...
+      self.SetFailed(
+          '%s field in %s was shorter than expected, was %d, expected %d' %
+          (self.EXPECTED_FIELD.capitalize(), self.PID,
+           len(fields[self.EXPECTED_FIELD]), self.MIN_LENGTH))
+
+    if self.MAX_LENGTH and len(fields[self.EXPECTED_FIELD]) > self.MAX_LENGTH:
```
nomis52: and here

tools/rdm/TestMixins.py
```diff
@@ -112,6 +145,9 @@ class GetRequiredStringMixin(GetRequiredMixin):
     This mixin also sets a property if PROVIDES is defined.  The target class
     needs to defined EXPECTED_FIELD and optionally PROVIDES.
   """
+  MIN_LENGTH = 0
+  MAX_LENGTH = RDM_MAX_STRING_LENGTH
+
   def VerifyResult(self, response, fields):
     if response.WasAcked() and self.PROVIDES:
       self.SetProperty(self.PROVIDES[0], fields[self.EXPECTED_FIELD])
```
nomis52: Same comment here , use 

string_field = fields[self.EXPECTED_FIELD]

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestDefinitions.py
peternewman: Done.

tools/rdm/TestDefinitions.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestDefinitions.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
peternewman: Done.

tools/rdm/TestMixins.py
```diff
@@ -59,16 +60,32 @@ def Test(self):
...
+
+    If ALLOWED_NACKS is non-empty, this adds a custom NackGetResult to the list of
+    allowed results for each entry.
   """
+  ALLOWED_NACKS = []
+
   def Test(self):
-    self.AddIfGetSupported(self.AckGetResult(field_names=[self.EXPECTED_FIELD]))
+    if isinstance(self.EXPECTED_FIELD, list):
+      expected_fields = self.EXPECTED_FIELD
```
nomis52: Same comment as before here, we should make this EXPECTED_FIELDS and avoid the use of isinstance . Do this in another change though - this one is already pretty big

nomis52:
LGTM, please merge as is and do the other change in a separate PR.

tools/rdm/TestMixins.py
peternewman: Fixed in https://github.com/OpenLightingProject/ola/pull/859

PR #855, opened 2015-07-29
Fail with format error rather than hardware fault for some more netwo…

peternewman:
…rk PIDs

(cherry picked from commit 42f843824f4292bb44ee6c70b69dd63e254463b5)


nomis52:
LGTM

Thanks for fixing this.

PR #857, opened 2015-07-29
Fix a bug in artnet universe mapping

### LINKED ISSUE #856: Can't set Artnet universe 15
There seems to be a bug in the Artnet mapping of universes.
If I create an Artnet output with universe 15, it will show up as 0:0:0
From what I understand it should show up as 0:0:15
Then to address universe 16 you would set the subnet to 1 instead of 0...

nomis52:
Closes #856

### DIFF
--- a/plugins/artnet/ArtNetPort.cpp
+++ b/plugins/artnet/ArtNetPort.cpp
@@ -41,11 +41,12 @@ static const uint8_t ARTNET_UNIVERSE_COUNT = 16;
 
 void ArtNetInputPort::PostSetUniverse(Universe *old_universe,
                                       Universe *new_universe) {
-  if (new_universe)
+  if (new_universe) {
     m_node->SetOutputPortUniverse(
         PortId(), new_universe->UniverseId() % ARTNET_UNIVERSE_COUNT);
-  else
+  } else {
     m_node->DisableOutputPort(PortId());
+  }
 
   if (new_universe && !old_universe) {
     m_node->SetDMXHandler(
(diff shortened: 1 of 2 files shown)


plugins/artnet/ArtNetPort.cpp
```diff
@@ -35,10 +35,15 @@ using ola::rdm::RDMCommand;
 using std::string;
 using std::vector;
 
+namespace {
+static const uint8_t ARTNET_UNIVERSE_COUNT = 16;
+};  // namespace
+
 void ArtNetInputPort::PostSetUniverse(Universe *old_universe,
                                       Universe *new_universe) {
   if (new_universe)
```
peternewman: {} please

plugins/artnet/ArtNetPort.cpp
```diff
@@ -139,7 +144,7 @@ void ArtNetOutputPort::PostSetUniverse(Universe *old_universe,
                                        Universe *new_universe) {
   if (new_universe)
```
peternewman: Again

plugins/artnet/ArtNetPort.cpp
```diff
@@ -35,10 +35,15 @@ using ola::rdm::RDMCommand;
...
 using std::vector;
 
+namespace {
+static const uint8_t ARTNET_UNIVERSE_COUNT = 16;
+};  // namespace
+
 void ArtNetInputPort::PostSetUniverse(Universe *old_universe,
                                       Universe *new_universe) {
   if (new_universe)
-    m_node->SetOutputPortUniverse(PortId(), new_universe->UniverseId() % 0xf);
```
peternewman: ArtNetNode uses (port->universe_address & 0xf) for stuff, we should probably switch to the same constant throughout and ideally use the same method of processing.

plugins/artnet/ArtNetPort.cpp
nomis52: Done

plugins/artnet/ArtNetPort.cpp
nomis52: Done

peternewman:
Looks good, just a query regarding matching behaviour with ArtNetNode

nomis52:
I think that's out of scope for this change.

peternewman:
Okay.

peternewman:
LGTM

PR #863, opened 2015-08-01
Fix some bugs in the RDM tests.

nomis52:


tools/rdm/TestDefinitions.py
```diff
@@ -925,7 +925,7 @@ def _CheckForSubDevice(self):
 
     if self._current_index >= PidStore.MAX_VALID_SUB_DEVICE:
       self.SetFailed('Only found %d of %d sub devices' %
-                     (len(self._sub_devices), self._device_count))
+                     (len(self._sub_device_addresses), self._device_count))
```
peternewman: Is there some way we could do static validation to catch issues like this?

tools/rdm/TestDefinitions.py
```diff
@@ -5071,7 +5071,7 @@ def Test(self):
       self.SetNotRun('Unable to determine pin code')
       return
 
-    self.AddIfSetSupported(self.NackSetResult(RDMNack.NR_FORMAT_ERROR))
+    self.AddIfSetSupported(self.NackSetResult(RDMNack.NR_DATA_OUT_OF_RANGE))
```
peternewman: Actually this was right before. Section 3.9 in E1.37-1 says "If a device receives a PIN code outside of this range, then it shall send a NACK with a NACK Reason Code of NR_FORMAT_ERROR." because NR_DATA_OUT_OF_RANGE is reserved for reporting the current PIN is incorrect.

tools/rdm/TestDefinitions.py
nomis52: I'm not aware of a tool that does that.

tools/rdm/TestDefinitions.py
nomis52: Good point, add a comment.

tools/rdm/TestDefinitions.py
```diff
@@ -5071,6 +5071,8 @@ def Test(self):
       self.SetNotRun('Unable to determine pin code')
       return
 
+    # Section 3.9, out of range pins return NR_FORMAT_ERROR rather than
+    # NR_DATA_OUT_OF_RANGE like one may expect.
```
peternewman: Explain what NR_DATA_OUT_OF_RANGE is used for.

nomis52: Isn't the name obvious?

peternewman: Not when NR_DATA_OUT_OF_RANGE is returned for an invalid PIN. I'm just thinking a message at the end of the comment saying NR_DATA_OUT_OF_RANGE is returned when an incorrect PIN is given.

nomis52: I'm still not following, can you propose the comment you want to see?

The test is for an out-of-range PIN. Normally out-of-range fields return NR_DATA_OUT_OF_RANGE, in this case we're supposed to return NR_FORMAT_ERROR. That's what the comment is trying to say.

peternewman: How about (with appropriate line wrapping):
    # Section 3.9, out of range pins return NR_FORMAT_ERROR rather than
    # NR_DATA_OUT_OF_RANGE like one may expect
    # (NR_DATA_OUT_OF_RANGE is reserved for reporting when an
    # incorrect PIN is given with this PID)

peternewman:
The new commit looks fine, just one comment left.

tools/rdm/TestDefinitions.py
nomis52: Done in https://github.com/OpenLightingProject/ola/pull/872

tools/rdm/TestDefinitions.py
peternewman: Flake8 should catch this in future.

PR #865, opened 2015-08-02
Lint and licence check built files

peternewman:


nomis52:
Hmm, I think this may be more work than it's worth. At the very least you'll have to ignore the yacc & lex files.

protoc/CppFileGenerator.cpp
```diff
@@ -89,7 +89,8 @@ void FileGenerator::GenerateHeader(Printer *printer) {
     "// Generated by the protocol buffer compiler.  DO NOT EDIT!\n"
     "// source: $filename$\n"
     "\n"
-    "#ifndef PROTOBUF_$filename_identifier$__INCLUDED\n"
+    "#ifndef PROTOBUF_$filename_identifier$__INCLUDED  "
+    "// NOLINT(build/header_guard)\n"
```
peternewman: Hopefully I can fix this properly if I can get the filename out somehow.

peternewman:
Should be ready to go I think.

PR #866, opened 2015-08-02
Add new OLP PIDs

nomis52:
Blocked behind #863

### DIFF
--- a/data/rdm/manufacturer_pids.proto
+++ b/data/rdm/manufacturer_pids.proto
@@ -8182,5 +8182,85 @@ manufacturer {
     }
     get_sub_device_range: ROOT_DEVICE
   }
+  pid {
+    name: "PIXEL_TYPE"
+    value: 32773
+    get_request {
+    }
+    get_response {
+      field {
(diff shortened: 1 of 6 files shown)


data/rdm/manufacturer_pids.proto
```diff
@@ -8182,5 +8182,85 @@ manufacturer {
     }
     get_sub_device_range: ROOT_DEVICE
   }
+  pid {
+    name: "PIXEL_TYPE"
+    value: 32773
```
peternewman: What are PIDs 0x8003 and 4 used for? Can you add them all to https://wiki.openlighting.org/index.php/Open_Lighting_Allocations .

We should also at least open an issue to support these in the SPI plugin (and ideally stop using personalities to change pixel types).

peternewman:
Should this not be going into 0.9?

data/rdm/manufacturer_pids.proto
nomis52: Done

PR #867, opened 2015-08-03
Update with the new Ja Rule USB ID.

nomis52:


nomis52:
Blocked behind #863

peternewman:
LGTM, should this not be going into 0.9?

PR #869, opened 2015-08-03
Update Ja Rule USB IDs

nomis52:


peternewman:
LGTM

peternewman:
You missed this file @nomis52 https://github.com/OpenLightingProject/ola/blob/0.9/tools/ja-rule/USBDeviceManager.cpp

PR #890, opened 2015-08-28
Tidy the output of the Python ola_rdm_get and other minor tidying

peternewman:


peternewman:
The change to python/examples/ola_rdm_get.py fixes the "bug" rmull reported. If it doesn't know the PID name, it report PID: 0x1234, then the data returned. If it does, it used to report PID_NAME 0x1234. This meant for supported parameters you got:
SUPPORTED_PARAMETERS 0x0050

Then the list of supported PIDs, which was in an identical format and hence made it appear that it was listed twice. It now reports as:
PID: SUPPORTED_PARAMETERS 0x0050

Which is less ambiguous and a closer match to what gets returned with an unknown PID.

I don't know if we want to merge that bit into 0.9.8 too, as a bug fix?

nomis52:
Yup, please target this at the 0.9 branch.

peternewman:
I take it it's just the Python change we want in 0.9.8 and not the random bit of C++ tidying?

nomis52:
I'm fine with both.

peternewman:
Okay, done in https://github.com/OpenLightingProject/ola/pull/895, can you merge this too though.

PR #892, opened 2015-09-06
Add Nodle U1 support for usbdmx plugin

StefanKrupop:
I added support for the DMXControl Projects e.V. Nodle U1 to the usbdmx plugin. The interface supports an output and an input port (at the same time), so I also added support to read from USB devices in synchronous and asynchronous mode. Some smaller changes to the existing code where necessary.
I hope I did everything right. On my Linux Test-VM everything seems to work fine.


peternewman:
Thanks for this @StefanKrupop . I've got a few comments on the code and @nomis52 may have others.

debian/ola.udev
```diff
@@ -10,6 +10,9 @@ SUBSYSTEM=="usb|usb_device", ACTION=="add", ATTRS{idVendor}=="0962", GROUP="plug
 # udev rules for the velleman dmx device
 SUBSYSTEM=="usb|usb_device", ACTION=="add", ATTRS{idVendor}=="10cf", ATTRS{idProduct}=="8062", GROUP="plugdev"
 
+# udev rules for the DMXControl Projects e.V. Nodle U1
```
peternewman: Can you either also open a PR against https://github.com/OpenLightingProject/raspberrypi/blob/master/etc/udev/rules.d/10-local.rules or bug one of us to do it please.

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -0,0 +1,180 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * AsyncUsbReceiver.cpp
+ * An Asynchronous DMX USB receiver.
+ * Copyright (C) 2015 Simon Newton, Stefan Krupop
```
peternewman: We generally only list one author, so just put yourself and remove @nomis52 .

plugins/usbdmx/AsyncUsbReceiver.h
```diff
@@ -0,0 +1,188 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * AsyncUsbReceiver.h
+ * An Asynchronous DMX USB receiver.
+ * Copyright (C) 2015 Simon Newton, Stefan Krupop
```
peternewman: Again

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+static const unsigned char READ_ENDPOINT = 0x81;
+static const unsigned int URB_TIMEOUT_MS = 50;
+static const int CONFIGURATION = 1;
+static const int INTERFACE = 0;
+
+static const unsigned int DATABLOCK_SIZE = 33;
+
+/*
+ * @brief Send chosen mode to the DMX device
+ * @param handle the ibusb_device_handle to use.
```
peternewman: SPaG libusb

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+  memset(usb_data, 0, sizeof(usb_data));
+  usb_data[0] = 16;
+  usb_data[1] = mode;
+
+  int ret = adaptor->InterruptTransfer(
+      handle, WRITE_ENDPOINT, reinterpret_cast<unsigned char*>(usb_data),
+      DATABLOCK_SIZE, &transferred, URB_TIMEOUT_MS);
+  if (ret) {
+    OLA_WARN << "InterruptTransfer():" << adaptor->ErrorCodeToString(ret)
```
peternewman: Add a space after : within the string.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+  }
+  return ret == 0;
+}
+
+
+/*
+ * @brief Attempt to open a handle to a Nodle widget.
+ * @param adaptor the LibUsbAdaptor to use.
+ * @param usb_device the libusb_device to use.
+ * @returns A libusb_device_handle of NULL if it failed.
```
peternewman: SPaG or

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+             << adaptor->ErrorCodeToString(ret_code);
+    adaptor->Close(usb_handle);
+    return NULL;
+  }
+
+  // this device only has one configuration
+  ret_code = adaptor->SetConfiguration(usb_handle, CONFIGURATION);
+  if (ret_code) {
+    OLA_WARN << "Nodle set config failed, with libusb error code "
+             << ret_code;
```
peternewman: ErrorCodeToString.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+    adaptor->Close(usb_handle);
+    return NULL;
+  }
+
+  if (adaptor->ClaimInterface(usb_handle, INTERFACE)) {
+    OLA_WARN << "Failed to claim Nodle usb device";
```
peternewman: SPaG USB

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+// NodleU1ThreadedSender
+// -----------------------------------------------------------------------------
+
+/*
+ * Sends messages to a Nodle device in a separate thread.
+ */
+class NodleU1ThreadedSender: public ThreadedUsbSender {
+ public:
+  NodleU1ThreadedSender(LibUsbAdaptor *adaptor,
+                         libusb_device *usb_device,
```
peternewman: Align this, and the row below up with (

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+  DmxBuffer m_last_tx_buffer;
+
+  bool TransmitBuffer(libusb_device_handle *handle,
+                      const DmxBuffer &buffer);
+  bool SendDataChunk(libusb_device_handle *handle,
+                     uint8_t *usb_data);
+};
+
+bool NodleU1ThreadedSender::TransmitBuffer(libusb_device_handle *handle,
+                                            const DmxBuffer &buffer) {
```
peternewman: Again

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+}
+
+bool NodleU1ThreadedSender::SendDataChunk(libusb_device_handle *handle,
+                                           uint8_t *usb_data) {
+  int transferred;
+  int ret = m_adaptor->InterruptTransfer(
+      handle, WRITE_ENDPOINT, reinterpret_cast<unsigned char*>(usb_data),
+      DATABLOCK_SIZE, &transferred, URB_TIMEOUT_MS);
+  if (ret) {
+    OLA_WARN << "InterruptTransfer():" << m_adaptor->ErrorCodeToString(ret)
```
peternewman: Add a space after : please

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+}
+
+bool NodleU1ThreadedReceiver::ReadDataChunk(libusb_device_handle *handle,
+                                            uint8_t *usb_data) {
+  int transferred;
+  int ret = m_adaptor->InterruptTransfer(
+      handle, READ_ENDPOINT, reinterpret_cast<unsigned char*>(usb_data),
+      DATABLOCK_SIZE, &transferred, URB_TIMEOUT_MS);
+  if (ret && ret != LIBUSB_ERROR_TIMEOUT) {
+    OLA_WARN << "InterruptTransfer():" << m_adaptor->ErrorCodeToString(ret)
```
peternewman: Again

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+  return true;
+}
+
+bool SynchronousNodleU1::SendDMX(const DmxBuffer &buffer) {
+  return m_sender.get() ? m_sender->SendDMX(buffer) : false;
+}
+
+void SynchronousNodleU1::SetDmxCallback(Callback0<void> *callback) {
+  if (m_receiver.get())
```
peternewman: {} please

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+ private:
+  unsigned int m_mode;
+  uint8_t *m_packet;
+
+  DISALLOW_COPY_AND_ASSIGN(NodleU1AsyncUsbReceiver);
+};
+
+bool NodleU1AsyncUsbReceiver::PerformTransfer() {
+  FillInterruptTransfer(READ_ENDPOINT, m_packet,
+                          DATABLOCK_SIZE, URB_TIMEOUT_MS);
```
peternewman: Align with (

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+  }
+  return false;
+}
+
+// NodleU1AsyncUsbSender
+// -----------------------------------------------------------------------------
+class NodleU1AsyncUsbSender : public AsyncUsbSender {
+ public:
+  NodleU1AsyncUsbSender(LibUsbAdaptor *adaptor,
+                         libusb_device *usb_device,
```
peternewman: Align with (

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+  }
+
+  bool PerformTransfer(const DmxBuffer &buffer);
+
+  void PostTransferHook();
+
+ private:
+  unsigned int m_mode;
+  DmxBuffer m_tx_buffer;
+  // This tracks where were are in m_tx_buffer. A value of 0 means we're at the
```
peternewman: SPaG s/were/we/

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+  unsigned int m_buffer_offset;
+  uint8_t *m_packet;
+
+  bool ContinueTransfer();
+
+  bool SendInitialChunk(const DmxBuffer &buffer);
+
+  bool SendChunk() {
+    FillInterruptTransfer(WRITE_ENDPOINT, m_packet,
+                            DATABLOCK_SIZE, URB_TIMEOUT_MS);
```
peternewman: Align (

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,526 @@
...
+  }
+  return !err;
+}
+
+bool AsynchronousNodleU1::SendDMX(const DmxBuffer &buffer) {
+  return m_sender.get() ? m_sender->SendDMX(buffer) : false;
+}
+
+void AsynchronousNodleU1::SetDmxCallback(Callback0<void> *callback) {
+  if (m_receiver.get())
```
peternewman: {} please

plugins/usbdmx/NodleU1.h
```diff
@@ -0,0 +1,150 @@
...
+namespace plugin {
+namespace usbdmx {
+
+/**
+ * @brief The interface for the Nodle Widgets
+ */
+class NodleU1: public BaseWidget {
+ public:
+  explicit NodleU1(LibUsbAdaptor *adaptor,
+                    const std::string &serial,
```
peternewman: Align (

plugins/usbdmx/NodleU1.h
```diff
@@ -0,0 +1,150 @@
...
+ public:
+  /**
+   * @brief Create a new SynchronousNodleU1.
+   * @param adaptor the LibUsbAdaptor to use.
+   * @param usb_device the libusb_device to use for the widget.
+   * @param serial the serial number of this widget
+   * @param mode the send/receive mode to be used by the widget.
+   */
+  SynchronousNodleU1(LibUsbAdaptor *adaptor,
+                           libusb_device *usb_device,
```
peternewman: Align (

plugins/usbdmx/NodleU1.h
```diff
@@ -0,0 +1,150 @@
...
+ public:
+  /**
+   * @brief Create a new AsynchronousNodleU1.
+   * @param adaptor the LibUsbAdaptor to use.
+   * @param usb_device the libusb_device to use for the widget.
+   * @param serial the serial number of this widget
+   * @param mode the send/receive mode to be used by the widget.
+   */
+  AsynchronousNodleU1(LibUsbAdaptor *adaptor,
+                            libusb_device *usb_device,
```
peternewman: Align (

plugins/usbdmx/NodleU1Factory.cpp
```diff
@@ -0,0 +1,72 @@
...
+  OLA_INFO << "Found a new Nodle U1 device";
+  LibUsbAdaptor::DeviceInformation info;
+  if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
+    return false;
+  }
+
+  OLA_INFO << "Nodle U1 serial: " << info.serial;
+
+  unsigned int mode;
+  if (!StringToInt(m_preferences->GetValue(NodleU1::NODLE_MODE_KEY) ,
```
peternewman: Remove the space before ,

plugins/usbdmx/NodleU1Factory.h
```diff
@@ -0,0 +1,59 @@
...
+namespace plugin {
+namespace usbdmx {
+
+/**
+ * @brief Creates Nodle widgets.
+ */
+class NodleU1Factory : public BaseWidgetFactory<class NodleU1> {
+ public:
+  explicit NodleU1Factory(class LibUsbAdaptor *adaptor,
+                           Preferences *preferences)
```
peternewman: Align (

plugins/usbdmx/NodleU1Port.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+bool NodleU1OutputPort::WriteDMX(const DmxBuffer &buffer,
+                                 OLA_UNUSED uint8_t priority) {
+  m_widget->SendDMX(buffer);
+  return true;
+}
+
+NodleU1InputPort::NodleU1InputPort(Device *parent,
+                                     unsigned int id,
```
peternewman: Align (

plugins/usbdmx/NodleU1Port.h
```diff
@@ -0,0 +1,102 @@
...
+ public:
+  /**
+   * @brief Create a new NodleU1InputPort.
+   * @param parent The parent device for this port.
+   * @param id The port id.
+   * @param plugin_adaptor a PluginAdaptor object, used by the base class.
+   * @param widget The widget to use to receive DMX frames.
+   */
+  NodleU1InputPort(Device *parent,
+                    unsigned int id,
```
peternewman: Align (

plugins/usbdmx/ThreadedUsbReceiver.cpp
```diff
@@ -0,0 +1,94 @@
...
+void *ThreadedUsbReceiver::Run() {
+  DmxBuffer buffer;
+  buffer.Blackout();
+  if (!m_usb_handle)
+    return NULL;
+
+  while (1) {
+    {
+      ola::thread::MutexLocker locker(&m_term_mutex);
+      if (m_term)
```
peternewman: {} please

plugins/usbdmx/ThreadedUsbReceiver.cpp
```diff
@@ -0,0 +1,94 @@
...
+    libusb_close(m_usb_handle);
+    return false;
+  }
+  return true;
+}
+
+void *ThreadedUsbReceiver::Run() {
+  DmxBuffer buffer;
+  buffer.Blackout();
+  if (!m_usb_handle)
```
peternewman: {}

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -91,13 +93,25 @@ string UsbDmxPlugin::Description() const {
...
 "libusb_debug_level = {0,1,2,3,4}\n"
 "The debug level for libusb, see http://libusb.sourceforge.net/api-1.0/ .\n"
 "0 = No logging, 4 = Verbose debug.\n"
+"\n"
+"nodle_mode = {0,1,2,3,4,5,6,7}\n"
+"The mode for the Nodle U1 interface to operate in.\n"
+"0 - Standby\n"
+"1 - DMX In -> DMX Out\n"
+"2 - PC Out -> DMX Out\n"
+"3 - DMX In + PC Out -> DMX Out\n"
```
peternewman: Is this merging DMX in and PC control?

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -91,13 +93,25 @@ string UsbDmxPlugin::Description() const {
...
+"4 - DMX In -> PC IN\n"
```
peternewman: SPaG In

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -91,13 +93,25 @@ string UsbDmxPlugin::Description() const {
...
 "--- Config file : ola-usbdmx.conf ---\n"
 "\n"
 "libusb_debug_level = {0,1,2,3,4}\n"
 "The debug level for libusb, see http://libusb.sourceforge.net/api-1.0/ .\n"
 "0 = No logging, 4 = Verbose debug.\n"
+"\n"
+"nodle_mode = {0,1,2,3,4,5,6,7}\n"
+"The mode for the Nodle U1 interface to operate in.\n"
+"0 - Standby\n"
+"1 - DMX In -> DMX Out\n"
```
peternewman: My gut feeling is this should be presented within OLA's preferences as a series of bools, rather than having to know which mode number you want.

StefanKrupop:
I fixed the typos and formatting issues.
FYI, most of them actually came from the files I copied from (I used the Vellemann interface as my starting point), so maybe you want to also change these?

plugins/usbdmx/UsbDmxPlugin.cpp
StefanKrupop: Yes, the interface itself does the merging in this mode. IIRC, it uses HTP for that.

plugins/usbdmx/UsbDmxPlugin.cpp
StefanKrupop: The mode value is actually sent to the interface and used to configure it.
I am not sure whether having a bunch of booleans to configure the behaviour would really be easier...

debian/ola.udev
StefanKrupop: Done.

peternewman:
Thanks. I'll have another look through later @StefanKrupop . Feel free to open a PR to fix up the source files :) !

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -64,10 +65,11 @@ bool UsbDmxPlugin::StartHook() {
   std::auto_ptr<PluginImplInterface> impl;
   if (FLAGS_use_async_libusb) {
     impl.reset(
-        new AsyncPluginImpl(m_plugin_adaptor, this, debug_level));
+        new AsyncPluginImpl(m_plugin_adaptor, this, debug_level,
+            m_preferences));
```
nomis52: please align under the m_ on the line above.

plugins/usbdmx/UsbDmxPlugin.cpp
nomis52: Yeah I think this is fine for now. Ideally once the new preferences system is done these will be enums / labeled values and the user won't need to worry about ints.

plugins/usbdmx/AsyncPluginImpl.h
```diff
@@ -60,10 +61,12 @@ class AsyncPluginImpl: public PluginImplInterface, public WidgetObserver {
    * @param plugin The parent Plugin object which is used when creating
    * devices.
    * @param debug_level the debug level to use for libusb.
+   * @param preferences The preferences container used by the plugin
    */
   AsyncPluginImpl(PluginAdaptor *plugin_adaptor,
                   Plugin *plugin,
-                  unsigned int debug_level);
+                  unsigned int debug_level,
+                  Preferences* preferences);
```
nomis52: Please move the * to the preferences so it's consistent with the rest of the file.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,528 @@
...
+}
+
+bool SynchronousNodleU1::SendDMX(const DmxBuffer &buffer) {
+  return m_sender.get() ? m_sender->SendDMX(buffer) : false;
+}
+
+void SynchronousNodleU1::SetDmxCallback(Callback0<void> *callback) {
+  if (m_receiver.get()) {
+    m_receiver->SetReceiveCallback(callback);
+  }
```
nomis52: else {
  delete callback;
}

otherwise it'll leak

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,528 @@
...
+}
+
+const DmxBuffer &SynchronousNodleU1::GetDmxInBuffer() const {
+  return m_receiver->GetDmxInBuffer();
```
nomis52: You should wrap this in a if (m_receiver.get()) check as well

StefanKrupop: But this would require me to return another DmxBuffer...
So it left it like that because it should not get called anyway when the receiver was not initialized.

nomis52:
Sorry I got most of the way through and then had to leave for work. Will try to do the rest tonight.

plugins/usbdmx/ThreadedUsbReceiver.cpp
```diff
@@ -0,0 +1,96 @@
...
+      break;
+    }
+
+    if (buffer_updated) {
+      {
+        ola::thread::MutexLocker locker(&m_data_mutex);
+        m_buffer.Set(buffer);
+      }
+      if (m_receive_callback.get()) {
+        m_receive_callback->Run();
```
nomis52: This is going to execute the receive callback from the RX thread., but the port methods aren't thread safe.

You'll need to pass in an Executor (the PluginAdaptor) and use that to execute the callback.

StefanKrupop: I am sorry, I understand the problem, but not your supposed solution.
Could you please elaborate on how you would like to have that fixed? I do not see how I can use the PluginAdaptor for that...
Thanks!

debian/ola.udev
peternewman: Thanks.

plugins/usbdmx/UsbDmxPlugin.cpp
peternewman: Okay, it sort of felt to me like the 8 modes are actually three bits of independent options.

plugins/usbdmx/UsbDmxPlugin.cpp
peternewman: That's cool.

plugins/usbdmx/ThreadedUsbReceiver.cpp
nomis52: Take a look at 

https://docs.openlighting.org/ola/doc/latest/classola_1_1thread_1_1_executor_interface.html

The PluginAdaptor implements the ExecutorInterface . If you pass a pointer to the PluginAdaptor into the ThreadedUSBReceiver you can then call m_plug_adaptor->Execute(m_receive_callback) which will run the callback from the PluginAdaptor's thread (the main I/O thread)

plugins/usbdmx/AsyncUsbReceiver.h
```diff
@@ -0,0 +1,188 @@
...
+  void SetReceiveCallback(Callback0<void> *callback) {
+    m_receive_callback.reset(callback);
+  }
+
+  /**
+   * @brief Get DMX Buffer
+   * @returns DmxBuffer with current input values.
+   */
+  const DmxBuffer &GetDmxInBuffer() const {
+    return m_rx_buffer;
```
nomis52: m_rx_buffer is accessed without a lock here

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -0,0 +1,180 @@
...
+      DISCONNECTED : IDLE);
+
+  if (m_suppress_continuation) {
+    return;
+  }
+
+  if (transfer->status != LIBUSB_TRANSFER_TIMED_OUT) {
+    if (TransferCompleted(&m_rx_buffer)) {  // Input changed
+      if (m_receive_callback.get()) {
+        m_receive_callback->Run();
```
nomis52: This suffers from the same issue, the callback is getting run in a thread which isn't the main event thread.

However if you run the callback in the main thread, you'll then have the problem of calling GetDmxInBuffer() and accessing m_rx_buffer without a lock. 

I think what you want to do it:
 - receive the completion event in the libusb thread
 - call TransferComplete to unpack the response into a DMXBuffer
 - Schedule a callback on the Executor to run in the main thread

Now there are two ways to transfer the DMX data:
i) Have the main thread call GetDMX(DMXBuffer *dmx), the method takes a lock, copies the data from the receiver's buffer to the argument buffer and returns.
ii) Change the RX callback to take a DMXBuffer*, in which case the receiver would allocate a new DMX buffer for each frame and then pass off ownership of the DMXBuffer to whatever receives the callback. 

i) involves a mutex acquisition & a memcpy, ii) involves 2 calls to new (one for the DMXBuffer & one for the outer callback binding to the DMXBuffer) & two calls to delete.

nomis52:
Sorry some of the code in this plugin changed at the same time. Can you resolve the merge please and ping when it's ready.

StefanKrupop:
*ping* :)

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -86,10 +89,12 @@ class DeviceState {
 
 AsyncPluginImpl::AsyncPluginImpl(PluginAdaptor *plugin_adaptor,
                                  Plugin *plugin,
-                                 unsigned int debug_level)
+                                 unsigned int debug_level,
+                                 Preferences* preferences)
```
nomis52: Please put the * on the right hand side.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+void AsynchronousNodleU1::SetDmxCallback(Callback0<void> *callback) {
+  if (m_receiver.get()) {
+    m_receiver->SetReceiveCallback(callback);
+  } else {
+    delete callback;
+  }
+}
+
+const DmxBuffer &AsynchronousNodleU1::GetDmxInBuffer() {
+  m_receiver->GetDmx(&m_buffer);
```
nomis52: I think this needs to be protected with a if (m_receiver.get()) {  ... }

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+  }
+
+  if (mode & 4) {  // input port active
+    m_receiver.reset(new NodleU1AsyncUsbReceiver(m_adaptor, usb_device,
+                                                 plugin_adaptor, mode));
+  }
+}
+
+bool AsynchronousNodleU1::Init() {
+  bool err = false;
```
nomis52: It would be more readable if you wrote this as

bool ok = true;
ok &= m_sender->Init();
...
return ok;

That avoids all the negations.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+  bool TransmitBuffer(libusb_device_handle *handle,
+                      const DmxBuffer &buffer);
+  bool SendDataChunk(libusb_device_handle *handle,
+                     uint8_t *usb_data);
+};
+
+bool NodleU1ThreadedSender::TransmitBuffer(libusb_device_handle *handle,
+                                           const DmxBuffer &buffer) {
+  m_tx_buffer.SetRange(0, buffer.GetRaw(), buffer.Size());
```
nomis52: Use .Set(buffer) instead.

nomis52: And do you need the copy?  I don't think so.

plugins/usbdmx/ThreadedUsbReceiver.h
```diff
@@ -0,0 +1,122 @@
...
+ protected:
+  /**
+   * @brief Perform the DMX transfer.
+   * @param handle the libusb_device_handle to use for the transfer.
+   * @param buffer The DmxBuffer to be updated.
+   * @param buffer_updated set to true when buffer was updated (=data
+       received)
+   * @returns true if the transfer was completed, false otherwise.
+   *
+   * This is called from the sender thread.
```
nomis52: s/sender/receiver/

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+};
+
+bool NodleU1ThreadedReceiver::ReceiveBuffer(libusb_device_handle *handle,
+                                            DmxBuffer *buffer,
+                                            bool *buffer_updated) {
+  unsigned char usb_data[DATABLOCK_SIZE];
+
+  if (ReadDataChunk(handle, usb_data)) {
+    if (usb_data[0] < 16) {
+      uint16_t startOff = usb_data[0] * 32;
```
nomis52: s/startOff/start_offset/

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+      for (int i = 0; i < 32; i++) {
```
nomis52: Use .SetRange( ... )

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+    return handle;
+  }
+
+  bool PerformTransfer();
+
+  bool TransferCompleted(DmxBuffer *buffer);
+
+ private:
+  unsigned int m_mode;
+  uint8_t *m_packet;
```
nomis52: Change to m_packet[DATABLOCK_SIZE] and then you can avoid the new  / delete[]

plugins/usbdmx/AsyncUsbReceiver.h
```diff
@@ -0,0 +1,194 @@
...
+    m_receive_callback.reset(callback);
+  }
+
+  /**
+   * @brief Get DMX Buffer
+   * @param buffer DmxBuffer that will get updated with the current input.
+   */
+  void GetDmx(DmxBuffer *buffer) {
+    ola::thread::MutexLocker locker(&m_mutex);
+    *buffer = m_rx_buffer;
```
nomis52: You need to use .Set( ... ) here to force a copy, otherwise the underlying memory will be accessed by 2 threads later once the lock is unlocked.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+  FillInterruptTransfer(READ_ENDPOINT, m_packet,
+                        DATABLOCK_SIZE, URB_TIMEOUT_MS);
+  return (SubmitTransfer() == 0);
+}
+
+bool NodleU1AsyncUsbReceiver::TransferCompleted(DmxBuffer *buffer) {
+  if (m_packet[0] < 16) {
+    uint16_t startOff = m_packet[0] * 32;
+    for (int i = 0; i < 32; i++) {
+      buffer->SetChannel(startOff + i, m_packet[i + 1]);
```
nomis52: SetRange(...)

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -0,0 +1,181 @@
...
+  ola::thread::MutexLocker locker(&m_mutex);
+  m_transfer_state = (transfer->status == LIBUSB_TRANSFER_NO_DEVICE ?
+      DISCONNECTED : IDLE);
+
+  if (m_suppress_continuation) {
+    return;
+  }
+
+  if (transfer->status != LIBUSB_TRANSFER_TIMED_OUT) {
+    if (TransferCompleted(&m_rx_buffer)) {  // Input changed
```
nomis52: You should pass in the actual transfer size to TransferComplete() , in case the USB device sends less than the DATABLOCK size.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+  void PostTransferHook();
+
+ private:
+  unsigned int m_mode;
+  DmxBuffer m_tx_buffer;
+  // This tracks where we are in m_tx_buffer. A value of 0 means we're at the
+  // start of a DMX frame.
+  unsigned int m_buffer_offset;
+  uint8_t *m_packet;
```
nomis52: I'd make this m_packet[DATABLOCK_SIZE] as well.

nomis52:
Looks pretty good. Just a couple of comments. 

Please ping the thread when it's updated.

StefanKrupop:
Done!
Also gave you push rights on my fork.

nomis52:
Looks good, just the one question about small packets.

plugins/usbdmx/AsyncPluginImpl.h
```diff
@@ -65,10 +66,12 @@ class AsyncPluginImpl: public PluginImplInterface, public WidgetObserver {
    * @param plugin The parent Plugin object which is used when creating
    * devices.
    * @param debug_level the debug level to use for libusb.
+   * @param preferences The preferences container used by the plugin
```
peternewman: The second preferences should be Preferences.

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -0,0 +1,182 @@
...
+  return m_usb_handle ? true : false;
+}
+
+bool AsyncUsbReceiver::Init(libusb_device_handle* handle) {
+  m_usb_handle = handle;
+  m_inited_with_handle = true;
+  return true;
+}
+
+bool AsyncUsbReceiver::Start() {
```
peternewman: Can we avoid most of this duplication with some inheritance?

StefanKrupop:
> What if the device is only receiving DMX frames
> that are 80 bytes long, does it pad the rest with 0s? 

Just tried that: It's always 33 bytes. If the universe is shorter, the interface sends the last received values for the remaining slots (or 0 when they never where touched).
I added the check for the length "just in case" and to not have an unused parameter ;)

nomis52:
@StefanKrupop  thanks for checking.  Just the one comment from @peternewman  left I think.

plugins/usbdmx/AsyncUsbReceiver.cpp
StefanKrupop: Ok, I created a base class "AsyncUsbTransceiverBase" and moved the common parts there.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -175,6 +183,14 @@ bool AsyncPluginImpl::NewWidget(JaRuleWidget *widget) {
                                 new JaRuleDevice(m_plugin, widget, str.str()));
 }
 
+bool AsyncPluginImpl::NewWidget(NodleU1 *widget) {
+  return StartAndRegisterDevice(
+      widget,
+      new NodleU1Device(m_plugin, widget, "DMXControl Projects e.V. Nodle U1",
+                        "NodleU1-" + widget->SerialNumber(),
```
peternewman: Can you lowercase this to be consistent with the other preferences bits.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -175,6 +183,14 @@ bool AsyncPluginImpl::NewWidget(JaRuleWidget *widget) {
                                 new JaRuleDevice(m_plugin, widget, str.str()));
 }
 
+bool AsyncPluginImpl::NewWidget(NodleU1 *widget) {
+  return StartAndRegisterDevice(
+      widget,
+      new NodleU1Device(m_plugin, widget, "DMXControl Projects e.V. Nodle U1",
```
peternewman: Users may benefit from seeing the serial number in the title here, if patching multiple units in.

plugins/usbdmx/AsyncPluginImpl.h
```diff
@@ -78,6 +81,7 @@ class AsyncPluginImpl: public PluginImplInterface, public WidgetObserver {
   bool NewWidget(class AnymauDMX *widget);
   bool NewWidget(class EurolitePro *widget);
   bool NewWidget(ola::usb::JaRuleWidget *widget);
+  bool NewWidget(class NodleU1 *widget);
```
peternewman: Can you add DMXProjects to the start of the class name please.

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -0,0 +1,98 @@
...
+AsyncUsbReceiver::~AsyncUsbReceiver() {
+  if (!m_inited_with_handle) {
+    m_adaptor->Close(m_usb_handle);
+  }
+}
+
+bool AsyncUsbReceiver::Init() {
+  m_usb_handle = SetupHandle();
+  m_inited_with_handle = false;
+  return m_usb_handle ? true : false;
```
peternewman: Can you just do != NULL to avoid the ternary operator.

plugins/usbdmx/AsyncUsbReceiver.h
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * @brief Start receiving DMX
+   * @returns the value of PerformTransfer().
+   */
+  bool Start();
+
+  /**
+   * @brief Set the callback to be called when the receive buffer is updated.
+   * @param callback The callback to call.
+   * @returns NULL.
```
peternewman: Remove this please, Doxygen is clever enough to not display return on voids.

plugins/usbdmx/AsyncUsbReceiver.h
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * @brief Called from the libusb callback when the asynchronous transfer
+   *   completes.
+   * @param transfer the completed transfer.
+   */
+  void TransferComplete(struct libusb_transfer *transfer);
+
+ protected:
+  /**
+   * @brief Perform the DMX transfer.
```
peternewman: Can you be more explicit that this is receiving data from the device, rather than sending DMX.

plugins/usbdmx/AsyncUsbSender.h
```diff
@@ -73,24 +68,9 @@ class AsyncUsbSender {
...
-   */
-  libusb_device* const m_usb_device;
-
-  /**
-   * @brief Open the device handle.
-   * @returns A valid libusb_device_handle or NULL if the device could not be
-   *   opened.
-   */
-  virtual libusb_device_handle* SetupHandle() = 0;
+  libusb_device_handle *GetHandle() { return m_usb_handle; }
```
peternewman: Can we have some Doxygen for this please.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+AsyncUsbTransceiverBase::~AsyncUsbTransceiverBase() {
+  CancelTransfer();
+  m_adaptor->UnrefDevice(m_usb_device);
+  m_adaptor->FreeTransfer(m_transfer);
+}
+
+bool AsyncUsbTransceiverBase::Init() {
+  m_usb_handle = SetupHandle();
+  return m_usb_handle ? true : false;
```
peternewman: Again

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,525 @@
...
+static const unsigned int URB_TIMEOUT_MS = 50;
+static const int CONFIGURATION = 1;
+static const int INTERFACE = 0;
+
+static const unsigned int DATABLOCK_SIZE = 33;
+
+/*
+ * @brief Send chosen mode to the DMX device
+ * @param handle the libusb_device_handle to use.
+ * @returns true when mode was set
```
peternewman: SPaG true if mode was set

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,525 @@
...
+  libusb_device_handle *usb_handle = OpenNodleU1Widget(
+      m_adaptor, m_usb_device);
+
+  if (!usb_handle) {
+    return false;
+  }
+
+  SetInterfaceMode(m_adaptor, usb_handle, m_mode);
+
+  if (m_mode & 2) {  // output port active
```
peternewman: Can we make the 2 a constant please.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,525 @@
...
+    std::auto_ptr<NodleU1ThreadedSender> sender(
+        new NodleU1ThreadedSender(m_adaptor, m_usb_device, usb_handle));
+    if (!sender->Start()) {
+      return false;
+    }
+    m_sender.reset(sender.release());
+  }
+
+  if (m_mode & 4) {  // input port active
```
peternewman: Again

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,525 @@
...
+      return false;
+    }
+    m_receiver.reset(receiver.release());
+  }
+
+  return true;
+}
+
+bool SynchronousNodleU1::SendDMX(const DmxBuffer &buffer) {
+  return m_sender.get() ? m_sender->SendDMX(buffer) : false;
```
peternewman: Can you format this more like SetDmxCallback, or the same kind of style but with early returns please.

plugins/usbdmx/NodleU1.cpp
```diff
@@ -0,0 +1,525 @@
...
+void SynchronousNodleU1::SetDmxCallback(Callback0<void> *callback) {
+  if (m_receiver.get()) {
+    m_receiver->SetReceiveCallback(callback);
+  } else {
+    delete callback;
+  }
+}
+
+const DmxBuffer &SynchronousNodleU1::GetDmxInBuffer() {
+  return m_receiver->GetDmxInBuffer();
```
peternewman: Does this need a check that m_receiver has a valid pointer first?

plugins/usbdmx/NodleU1Device.h
```diff
@@ -0,0 +1,72 @@
...
+#include "plugins/usbdmx/NodleU1.h"
+
+namespace ola {
+namespace plugin {
+namespace usbdmx {
+
+/**
+ * @brief A NodleU1 device.
+ *
+ * This device creates a single output port around a Widget.
```
peternewman: Can you correct either this comment or the one above, they appear to conflict currently.

plugins/usbdmx/NodleU1Factory.cpp
```diff
@@ -0,0 +1,72 @@
...
+  OLA_INFO << "Found a new Nodle U1 device";
+  ola::usb::LibUsbAdaptor::DeviceInformation info;
+  if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
+    return false;
+  }
+
+  OLA_INFO << "Nodle U1 serial: " << info.serial;
+
+  unsigned int mode;
+  if (!StringToInt(m_preferences->GetValue(NodleU1::NODLE_MODE_KEY),
```
peternewman: You should really add the serial number into the preference you're getting, otherwise all Nodles have to be in the same mode.

plugins/usbdmx/NodleU1Port.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+NodleU1InputPort::NodleU1InputPort(Device *parent,
+                                   unsigned int id,
+                                   PluginAdaptor *plugin_adaptor,
+                                   NodleU1 *widget)
+    : BasicInputPort(parent, id, plugin_adaptor),
+      m_widget(widget) {
+  m_widget->SetDmxCallback(NewCallback(
+    static_cast<BasicInputPort*>(this),
```
peternewman: Indent by 4 from m_widget please.

plugins/usbdmx/NodleU1Port.cpp
```diff
@@ -0,0 +1,55 @@
...
+    &BasicInputPort::DmxChanged));
```
peternewman: Again

plugins/usbdmx/NodleU1Port.h
```diff
@@ -0,0 +1,102 @@
...
+   */
+  NodleU1OutputPort(Device *parent,
+                    unsigned int id,
+                    class NodleU1 *widget);
+
+  bool WriteDMX(const DmxBuffer &buffer, uint8_t priority);
+
+  std::string Description() const {
+    std::ostringstream str;
+    str << "Serial #: " << m_widget->SerialNumber();
```
peternewman: As above, I'd put the serial in the device, as the ports all have the same serial number, and just number these, or simply label them Input and Output if there is only ever one of each.

plugins/usbdmx/NodleU1Port.h
```diff
@@ -0,0 +1,102 @@
...
+                   PluginAdaptor *plugin_adaptor,
+                   class NodleU1 *widget);
+
+  const DmxBuffer &ReadDMX() const {
+    return m_widget->GetDmxInBuffer();
+  }
+
+  std::string Description() const {
+    std::ostringstream str;
+    str << "Serial #: " << m_widget->SerialNumber();
```
peternewman: As above.

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -56,13 +59,17 @@ using std::vector;
...
+                               Preferences *preferences)
     : m_plugin_adaptor(plugin_adaptor),
       m_plugin(plugin),
       m_debug_level(debug_level),
+      m_preferences(preferences),
       m_context(NULL) {
   m_widget_factories.push_back(new AnymauDMXFactory(&m_usb_adaptor));
   m_widget_factories.push_back(new EuroliteProFactory(&m_usb_adaptor));
+  m_widget_factories.push_back(
+    new NodleU1Factory(&m_usb_adaptor, m_plugin_adaptor, m_preferences));
```
peternewman: Indent by 4

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -129,6 +136,14 @@ bool SyncPluginImpl::NewWidget(OLA_UNUSED ola::usb::JaRuleWidget *widget) {
   return false;
 }
 
+bool SyncPluginImpl::NewWidget(NodleU1 *widget) {
+  return StartAndRegisterDevice(
+      widget,
+      new NodleU1Device(m_plugin, widget, "DMXControl Projects e.V. Nodle U1",
+                        "NodleU1-" + widget->SerialNumber(),
```
peternewman: Lower case this please.

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -129,6 +136,14 @@ bool SyncPluginImpl::NewWidget(OLA_UNUSED ola::usb::JaRuleWidget *widget) {
   return false;
 }
 
+bool SyncPluginImpl::NewWidget(NodleU1 *widget) {
+  return StartAndRegisterDevice(
+      widget,
+      new NodleU1Device(m_plugin, widget, "DMXControl Projects e.V. Nodle U1",
```
peternewman: Add the serial here.

plugins/usbdmx/SyncPluginImpl.h
```diff
@@ -58,10 +59,12 @@ class SyncPluginImpl: public PluginImplInterface,  public WidgetObserver {
    * @param plugin The parent Plugin object which is used when creating
    * devices.
    * @param debug_level the debug level to use for libusb.
+   * @param preferences The preferences container used by the plugin
```
peternewman: Make the second pref Preferences

plugins/usbdmx/ThreadedUsbReceiver.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+  Join();
+  libusb_unref_device(m_usb_device);
+}
+
+bool ThreadedUsbReceiver::Start() {
+  bool ret = ola::thread::Thread::Start();
+  if (!ret) {
+    OLA_WARN << "Failed to start receiver thread";
+    libusb_release_interface(m_usb_handle, 0);
```
peternewman: Ideally the interface number should be an option when instantiating the clase.

plugins/usbdmx/ThreadedUsbReceiver.cpp
```diff
@@ -0,0 +1,98 @@
...
+    return false;
+  }
+  return true;
+}
+
+void *ThreadedUsbReceiver::Run() {
+  DmxBuffer buffer;
+  buffer.Blackout();
+  if (!m_usb_handle) {
+    return NULL;
```
peternewman: Just return;

plugins/usbdmx/ThreadedUsbReceiver.cpp
```diff
@@ -0,0 +1,98 @@
...
+      {
+        ola::thread::MutexLocker locker(&m_data_mutex);
+        m_buffer.Set(buffer);
+      }
+      if (m_receive_callback.get()) {
+        m_plugin_adaptor->Execute(m_receive_callback.get());
+      }
+    }
+  }
+  libusb_release_interface(m_usb_handle, 0);
```
peternewman: Again

plugins/usbdmx/ThreadedUsbReceiver.h
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * @brief Start the new thread.
+   * @returns true if the thread is running, false otherwise.
+   */
+  bool Start();
+
+  /**
+   * @brief Entry point for the new thread.
+   * @returns NULL.
```
peternewman: Remove this

plugins/usbdmx/ThreadedUsbReceiver.h
```diff
@@ -0,0 +1,122 @@
...
+   */
+  void *Run();
+
+  /**
+   * @brief Set the callback to be called when the receive buffer is updated.
+   * @param callback The callback to call.
+   * @returns NULL.
```
peternewman: Remove this

peternewman:
Sorry, some more comments from me. I'd not got through all the code before.

plugins/usbdmx/AsyncPluginImpl.cpp
StefanKrupop: The serial number is already shown in the "Description" field like (IIRC) the Enttec plugin does. Is that not okay? I think the description is visible everywhere you see the device name in the UI...

plugins/usbdmx/NodleU1.cpp
StefanKrupop: There was a comment on that line before, but I do not find it right now.
No, the check should not be necessary because the function will not get called when the initialization failed.
If i add a check, I would also have to return some other (static?) DmxBuffer...

plugins/usbdmx/AsyncPluginImpl.cpp
nomis52: This is fine, the title should be generic 'Nodle U1', the description should be specific (serial # like you have).

plugins/usbdmx/NodleU1.cpp
nomis52: Yeah, that's fine.

plugins/usbdmx/ThreadedUsbReceiver.cpp
nomis52: The Run() method must return a void* .

plugins/usbdmx/ThreadedUsbReceiver.h
StefanKrupop: There is actually a void pointer returned, so I should rather not remove it?

plugins/usbdmx/NodleU1Port.h
StefanKrupop: Is this still valid if we keep the serial in the description just like it is now?

plugins/usbdmx/ThreadedUsbReceiver.cpp
peternewman: Sorry, I missed the asterisk.

plugins/usbdmx/ThreadedUsbReceiver.h
peternewman: Ignore me, I missed the asterisk.

plugins/usbdmx/NodleU1.cpp
peternewman: Okay

StefanKrupop:
Ok, I think I am finished with the comments so far.
Now we can start the next round of change requests ;)

nomis52:
I'm happy with this. Can we merge it now please?

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -43,6 +43,9 @@
 #include "plugins/usbdmx/GenericDevice.h"
 #include "plugins/usbdmx/JaRuleDevice.h"
 #include "plugins/usbdmx/JaRuleFactory.h"
+#include "plugins/usbdmx/DMXCProjectsNodleU1.h"
```
peternewman: Can you fix the ordering of these includes now please.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -115,6 +120,9 @@ bool AsyncPluginImpl::Start() {
   m_widget_factories.push_back(
       new JaRuleFactory(m_plugin_adaptor, m_usb_adaptor));
   m_widget_factories.push_back(
+      new DMXCProjectsNodleU1Factory(m_usb_adaptor,
+                                     m_plugin_adaptor, m_preferences));
```
peternewman: Can you move m_plugin_adaptor up to the line above.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -115,6 +120,9 @@ bool AsyncPluginImpl::Start() {
   m_widget_factories.push_back(
       new JaRuleFactory(m_plugin_adaptor, m_usb_adaptor));
   m_widget_factories.push_back(
+      new DMXCProjectsNodleU1Factory(m_usb_adaptor,
```
peternewman: Can you re-order these please.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -175,6 +183,15 @@ bool AsyncPluginImpl::NewWidget(JaRuleWidget *widget) {
                                 new JaRuleDevice(m_plugin, widget, str.str()));
 }
 
+bool AsyncPluginImpl::NewWidget(DMXCProjectsNodleU1 *widget) {
```
peternewman: Can you re-order these please.

plugins/usbdmx/AsyncPluginImpl.h
```diff
@@ -78,6 +81,7 @@ class AsyncPluginImpl: public PluginImplInterface, public WidgetObserver {
   bool NewWidget(class AnymauDMX *widget);
   bool NewWidget(class EurolitePro *widget);
   bool NewWidget(ola::usb::JaRuleWidget *widget);
+  bool NewWidget(class DMXCProjectsNodleU1 *widget);
```
peternewman: Can you re-order these please.

plugins/usbdmx/AsyncUsbReceiver.cpp
peternewman: Thanks

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -38,6 +38,9 @@
 #include "plugins/usbdmx/AnymauDMXFactory.h"
 #include "plugins/usbdmx/EurolitePro.h"
 #include "plugins/usbdmx/EuroliteProFactory.h"
+#include "plugins/usbdmx/DMXCProjectsNodleU1.h"
```
peternewman: Can you re-order these please.

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -56,13 +59,17 @@ using std::vector;
...
+                               unsigned int debug_level,
+                               Preferences *preferences)
     : m_plugin_adaptor(plugin_adaptor),
       m_plugin(plugin),
       m_debug_level(debug_level),
+      m_preferences(preferences),
       m_context(NULL) {
   m_widget_factories.push_back(new AnymauDMXFactory(&m_usb_adaptor));
   m_widget_factories.push_back(new EuroliteProFactory(&m_usb_adaptor));
+  m_widget_factories.push_back(new DMXCProjectsNodleU1Factory(&m_usb_adaptor,
```
peternewman: Can you re-order these please.

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -129,6 +136,15 @@ bool SyncPluginImpl::NewWidget(OLA_UNUSED ola::usb::JaRuleWidget *widget) {
   return false;
 }
 
+bool SyncPluginImpl::NewWidget(DMXCProjectsNodleU1 *widget) {
```
peternewman: Can you re-order these please.

plugins/usbdmx/SyncPluginImpl.h
```diff
@@ -71,6 +74,7 @@ class SyncPluginImpl: public PluginImplInterface,  public WidgetObserver {
   bool NewWidget(class AnymauDMX *widget);
   bool NewWidget(class EurolitePro *widget);
   bool NewWidget(ola::usb::JaRuleWidget *widget);
+  bool NewWidget(class DMXCProjectsNodleU1 *widget);
```
peternewman: Can you re-order these please.

plugins/usbdmx/SyncPluginImpl.h
```diff
@@ -78,6 +82,7 @@ class SyncPluginImpl: public PluginImplInterface,  public WidgetObserver {
   void WidgetRemoved(OLA_UNUSED class AnymauDMX *widget) {}
   void WidgetRemoved(OLA_UNUSED class EurolitePro *widget) {}
   void WidgetRemoved(OLA_UNUSED ola::usb::JaRuleWidget *widget) {}
+  void WidgetRemoved(OLA_UNUSED class DMXCProjectsNodleU1 *widget) {}
```
peternewman: Can you re-order these please.

plugins/usbdmx/SyncronizedWidgetObserver.h
```diff
@@ -59,6 +59,10 @@ class SyncronizedWidgetObserver : public WidgetObserver {
     return DispatchNewWidget(widget);
   }
 
+  bool NewWidget(class DMXCProjectsNodleU1 *widget) {
```
peternewman: Can you re-order these please.

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -91,13 +92,26 @@ string UsbDmxPlugin::Description() const {
 "----------------------------\n"
 "\n"
 "This plugin supports various USB DMX devices including the \n"
-"Anyma uDMX, Eurolite, Fadecandy, Sunlite USBDMX2 & Velleman K8062.\n"
+"Anyma uDMX, Eurolite, Fadecandy, Sunlite USBDMX2, Velleman K8062 "
```
peternewman: Can you re-order these please.

plugins/usbdmx/WidgetFactory.h
```diff
@@ -78,6 +78,15 @@ class WidgetObserver {
   virtual bool NewWidget(ola::usb::JaRuleWidget *widget) = 0;
 
   /**
+   * @brief Called when a new Nodle U1 widget is added.
```
peternewman: Can you re-order these please.

peternewman:
I think my main outstanding point, apart from some minor style bits, is the serial number. Perhaps this is something we need some internal consistency on too @nomis52 .

I see three bits of info presented in the UI:
Device Name
  Port Direction
  Port Description

Direction is defined by the port type, so we can ignore that. The ports are all a subset of the device.

My test install currently lists the following (IPs obscured):
E1.31 (DMX over ACN) [127.0.0.1]	Input	
E1.31 (DMX over ACN) [127.0.0.1]	Output	
OSC Device	Input	/dmx/universe/%d
OSC Device	Input	/dmx/universe/%d
OSC Device	Input	/dmx/universe/%d
OSC Device	Output	127.0.0.1:7770/4/multifader1, 

As I see it, the E1.31 device is the way to go, the IP is common across all ports, so it's listed at the device level, the ports don't actually have a description.

Consider the following:
Nodle 1 <1234> Input
Nodle 1 <1234> Output
Nodle 1 <5678> Input
Nodle 1 <5678> Output

I think that's more correct than:
Nodle 1 Input <1234>
Nodle 1 Output <1234>
Nodle 1 Input <5678>
Nodle 1 Output <5678>

Especially when we get onto the next level, say a two port Nodle, or the Enttec Pro Mk 2, this feels like the right way to go:
Nodle 1 <1234> Input Port 1
Nodle 1 <1234> Input Port 2
Nodle 1 <1234> Output Port 1
Nodle 1 <1234> Output Port 2
Nodle 1 <5678> Input Port 1
Nodle 1 <5678> Input Port 2
Nodle 1 <5678> Output Port 1
Nodle 1 <5678> Output Port 2

Plus if we rewrote the UI, or someone write another UI using a tree or a wizard or whatever, you pick the device "Nodle 1 <1234>" with serial 1234, then you can pick a port (or a specific class of port in/out) from the list to patch to it.

nomis52:
@peternewman  device / port naming is a larger issue - this PR is consistent with how we've named USB widgets, with the exception of the FadeCandy & JaRule.

We shouldn't block this PR on sorting out naming. Lets fix the last of the style comments and merge.

peternewman:
Okay, fair enough.

peternewman:
I've raised #975 to cover that.

peternewman:
LGTM I think, or at least I think you've now fixed the other bits I'd flagged thanks @StefanKrupop .

nomis52:
LGTM.

Thanks for adding this, please continue to send PRs!

Psykoman1990:
Thank you all :+1:

PR #898, opened 2015-09-15
Add flake8 support

peternewman:


peternewman:
Just a few bits to fix :):
https://travis-ci.org/OpenLightingProject/ola/jobs/80431424

peternewman:
Any thoughts on which bits we want to exclude?

nomis52:
It all looks useful.

I'd turn E121 off and then try to fix the rest in this PR.

peternewman:
Yep, I'd go with skipping E121 @nomis52 as it seems to only handle indents of four spaces.

What are your thoughts on E129, e.g. "python/ola/PidStore.py:890:7: E129 visually indented line with same indent as next logical line". I think we may have to drop that too, as otherwise it stops us nicely indenting within if etc like we'd do in C++.

nomis52:
Agreed with skipping E129

peternewman:
@nomis52 I've set allow failures for flake8 on Travis. Can you review what I've done so far so we can merge it in, as the remaining changes are not just whitespace and so probably need a more careful review, as they may actually break the code (not that whitespace can't do that to Python anyway).

nomis52:
LGTM

PR #905, opened 2015-09-23
Create philips_hue_lux.conf

gobo-ws:
OLA Trigger config to control Philips Hue Lux lights


peternewman:
Looks good @gobo-ws you'll need to add the file to https://github.com/OpenLightingProject/ola/blob/master/tools/ola_trigger/Makefile.mk#L7 to stop Travis complaining. You may also want to credit yourself here: https://github.com/OpenLightingProject/ola/blob/master/AUTHORS#L7

gobo-ws:
Thanks Peter. I am new to Github. I hope everything is ok now? Regards, Johan

peternewman:
Almost, you just need to get them all in one branch. Either make the second and third edits again here https://github.com/gobo-ws/ola/tree/patch-1, or generate and merge local PRs to merge at your end https://github.com/gobo-ws/ola/compare/patch-1...gobo-ws:patch-2?expand=1 and https://github.com/gobo-ws/ola/compare/patch-1...gobo-ws:patch-3?expand=1

tools/ola_trigger/contrib/hue_trigger.conf
```diff
@@ -0,0 +1,47 @@
...
+# Thanks to Simon Newton and Sean Sill
+# Resources used from developers.meethue.com
+
+
+
+# DMX protocol
+
+###############################################################################
+
+# Channel 1 lamp on/off. Value 0-140 lamp off. 141-255 lamp on
```
peternewman: Is there some reason the split is 140, rather than 127 in the middle of the range?

tools/ola_trigger/contrib/hue_trigger.conf
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+# Comments
+
+###############################################################################
+
+# Responsiveness of the Hue bridge will slow down if you repeatedly send the on command, that's why we are using 2 channels
+# Please notice that the Hue bridge has some frame rate limits
+# Hue use the range 1-254 to control the intensity, that's why the DMX value 0 is the same as 1 and the DMX value 255 is the same as 254
```
peternewman: SPaG uses the range

peternewman:
Excellent, we're all green on the tests now @gobo-ws . Just a couple of minor comments from me, you should be able to edit those directly from the files changed bit of the PR, then we should be good to merge.

Thanks for your work on this!

gobo-ws:
Thanks for the comments!
The split where at first 127 but Sean Sill made this comment this on IRC: "Generally people use 55% to turn on and 45% to turn off so there is a deadband for noisy faders"
SPaG?
Do not really know what is "standard"?

peternewman:
Ah, fair enough. the OLA GPIO plugin goes with 127/128 as the default switchover levels. Perhaps @nightrune should submit a PR to fix that too :). I can see the arguments for both sides; if you're going to go with the dead band, do you want to implement it on both sides then, by my reckoning, it should be off from 0-114 and on from 140-255?

SPaG is spelling, punctuation and grammar, in this case just a spelling error of s/use/uses/.

tools/ola_trigger/contrib/hue_trigger.conf
```diff
@@ -0,0 +1,47 @@
+# OLA Trigger config to control Philips Hue Lux lights
```
peternewman: It may also be worth renaming the file to something like philips_hue_lux.conf, and dropping the trigger bit, given it's already in the ola_trigger folder now.

peternewman:
Can you add the file to line 76 of the the Makefile.mk ( https://github.com/gobo-ws/ola/blob/patch-1/tools/ola_trigger/Makefile.mk#L76 ) too please, so it gets tested to ensure ola_trigger can parse it.

gobo-ws:
Done. But still some Travis CI build error.

gobo-ws:
"coverage/coveralls  — Coverage decreased (-0.01%) to 54.424%" @peternewman is there anything I can do to solve this?

peternewman:
Don't worry about the coverage @gobo-ws . You could write some more tests, but the fact its only gone down by 0.01% it's probably just down to some of the generated code.

peternewman:
LGTM, @nomis52 any final comments?

nomis52:
LGTM

Issue #854, opened 2015-07-28
Ack Timer Responder fails Set Identify Device and Set Start Address tests


peternewman:
From running ola_rdm_set I get:
ola_rdm_set -u 0 --uid 7a70:ffffff03 identify_device 1
Empty STATUS_MESSAGE returned.

And the RDM tests:
./tools/rdm/rdm_responder_test.py -u 0 -d -t SetIdentifyDevice -s 7a70:ffffff03
OLA Responder Tests Version 0.9.7
Fetching UID list from server
Found UID 7a70:ffffff03
The following devices were detected and will be reconfigured
 7a70:ffffff00
 7a70:ffffff01
 7a70:ffffff02
 7a70:ffffff03
 7a70:ffffff04
 7a70:ffffff05
 7a70:ffffff06
Restricting tests to SetIdentifyDevice
Starting tests, universe 0, UID 7a70:ffffff03, broadcast write delay 0ms, inter-test delay 0ms
Test order is [GetIdentifyDevice, SetIdentifyDevice]
GetIdentifyDevice: Get the identify state.
 GET: uid: 7a70:ffffff03, pid: IDENTIFY_DEVICE (0x1000), sub device: 0, args: []
 Response: RDMResponse(type=ACK, command_class=GET), PID: 0x1000, TN: 0, PDL: 1, data: {'identify_state': True}
GetIdentifyDevice: Passed
SetIdentifyDevice: Set the identify state.
 SET: uid: 7a70:ffffff03, pid: IDENTIFY_DEVICE (0x1000), sub device: 0, args: [False]
 Received ACK TIMER set to 500 ms
 GET: pid: QUEUED_MESSAGE (0x0020), data: ['error']
 Response: RDMResponse(type=ACK, command_class=GET), PID: 0x0020, TN: 0, PDL: 0, data: {}
 GET: pid: QUEUED_MESSAGE (0x0020), data: ['error']
 Response: RDMResponse(type=ACK, command_class=GET), PID: 0x0030, TN: 0, PDL: 0, data: {'messages': []}
 Failed: ACK_TIMER issued but the response was never queued
 SET: uid: 7a70:ffffff03, pid: IDENTIFY_DEVICE (0x1000), sub device: 0, args: [True]
 Received ACK TIMER set to 500 ms
 GET: pid: QUEUED_MESSAGE (0x0020), data: ['error']
[...]

Issue #917, opened 2015-09-26
OLA fails to build with microhttpd 0.9.43


nomis52:

peternewman:
Closed in #919

PR #921, opened 2015-09-26
Fix the ACK Timer responder so that it passes the RDM tests

nomis52:


nomis52:
blocked behind #917, closes #854

peternewman:
Can you resync please @nomis52

common/rdm/AckTimerResponder.cpp
```diff
@@ -208,13 +208,29 @@ RDMResponse *AckTimerResponder::ResponseFromQueuedMessage(
...
+        request->SourceUID(),
+        request->TransactionNumber(),
+        RDM_ACK,
+        QueuedMessageCount(),
+        ROOT_RDM_DEVICE,
+        queued_response->Pid(),
+        queued_response->ParamData(),
+        queued_response->ParamDataSize());
     case RDMCommand::SET_COMMAND_RESPONSE:
-      return GetResponseFromData(request,
```
peternewman: Ah, so it was just that it had the wrong PID in the response? It looks like GetResponseWithPid will be a neater fit, or does it have to come from the root device? Of course in that case you could extend GetResponseWithPid to have a version that passes in the subdevice too.

nomis52: GetResponseWithPid doesn't work because you need to override the command class as well.

peternewman: Ah yes, GET_COMMAND_RESPONSE becomes GET_RESPONSE.

peternewman:
LGTM

common/rdm/AckTimerResponder.cpp
nomis52: Not just that, a GET_REQUEST can reply with a SET_RESPONSE. Queued messages are funky.

peternewman: Ah, I'd missed the fact you were switching on the queued message!

PR #925, opened 2015-09-27
Fixes to the Ja Rule tools so they run on the BBB.

nomis52:


peternewman:
LGTM

What compiler is BBB using so we can try and catch this on Travis?

nomis52:
gcc 4.4 but I think it's more of an architecture thing rather than a compiler thing.

peternewman:
Ah okay

PR #926, opened 2015-09-27
Add the Hotplug agent

nomis52:

PR #929, opened 2015-09-30
More USB code cleanup

nomis52:


libs/usb/HotplugAgent.cpp
```diff
@@ -77,25 +77,23 @@ AsyncronousLibUsbAdaptor *HotplugAgent::GetUSBAdaptor() const {
...
 
   m_use_hotplug = HotplugSupported();
   OLA_DEBUG << "HotplugSupported(): " << m_use_hotplug;
-  if (m_use_hotplug) {
 #ifdef HAVE_LIBUSB_HOTPLUG_API
+  if (m_use_hotplug) {
     m_usb_thread.reset(new ola::usb::LibUsbHotplugThread(
           m_context, hotplug_callback, this));
-#else
-    OLA_FATAL << "Mismatch between m_use_hotplug and HAVE_LIBUSB_HOTPLUG_API";
```
peternewman: Don't we want the error message still?

libs/usb/Types.cpp
```diff
@@ -0,0 +1,50 @@
...
+bool USBDeviceID::operator<(const USBDeviceID &id) const {
+  if (bus_number < id.bus_number) {
+    return true;
+  } else if (bus_number == id.bus_number) {
+    return device_address < id.device_address;
+  }
+  return false;
+}
+
+ostream& operator<<(ostream& os, const USBDeviceID &id) {
```
peternewman: Is this not similar duplication to the other << operator you added just recently?

peternewman:
LGTM

libs/usb/HotplugAgent.cpp
nomis52: There shouldn't be any way for this message to occur so I removed it to make the code simpler.

libs/usb/Types.cpp
nomis52: Which one? I can't find one.

libs/usb/HotplugAgent.cpp
peternewman: Okay, I thought it may be worth it for the edge case.

libs/usb/Types.cpp
peternewman: I found it eventually: https://github.com/OpenLightingProject/ola/pull/926/files#diff-034547a73202f2f4f37ca79aabba4010R39

peternewman: Ah, it's the same one, it's just no longer in the header...

PR #932, opened 2015-09-30
drop AC_PROG_LIBTOOL in configure.ac

### LINKED ISSUE #931: Obsolete autotools macro: AC_PROG_LIBTOOL
we already have LT_INIT on the previous line.

daveol:
since AC_PROG_LIBTOOL is obsolete in newer versions op autotools (according to redhat epel6+) they now use LT_INIT instead.

fixes #931


peternewman:
Should we cherry-pick this into 0.9.9 @nomis52 ?

nomis52:
I don't think so.

peternewman:
Okay

PR #930, opened 2015-09-30
Simplify the Async USB plugin

nomis52:

PR #933, opened 2015-10-01
Fix the shutdown bug

nomis52:


nomis52:
Depends on #930

nomis52:
Ok, this one is ready now.

nomis52:
Ah sorry I merged without you looking. @peternewman  can you take a look and I'll fix anything else you find.

common/rdm/DiscoveryAgent.cpp
```diff
@@ -217,7 +217,9 @@ void DiscoveryAgent::BranchComplete(const uint8_t *data, unsigned int length) {
   OLA_INFO << "BranchComplete, got " << length;
   if (length == 0) {
     // timeout
-    FreeCurrentRange();
+    if (!m_uid_ranges.empty()) {
+      FreeCurrentRange();
```
peternewman: Are we better off fixing this within FreeCurrentRange, in case it gets called from elsewhere?

libs/usb/JaRuleWidgetPort.cpp
```diff
@@ -102,6 +102,8 @@ JaRuleWidgetPort::~JaRuleWidgetPort() {
                   "CancelTransfer()?";
     }
 
+    // Cancelling may take up to a second if the endpoint has stalled. I can't
```
peternewman: SPaG Canceling is the American English, which the rest of the code seems to use.

plugins/usbdmx/JaRuleOutputPort.h
```diff
@@ -63,7 +75,9 @@ class JaRuleOutputPort: public BasicOutputPort {
   void PostSetUniverse(Universe *old_universe, Universe *new_universe);
 
  private:
-  ola::usb::JaRulePortHandle* const m_port_handle;  // not owned
+  const unsigned int m_port_index;
+  ola::usb::JaRuleWidget *m_widget;  // not ownedk
```
peternewman: SPaG owned

peternewman:
Generally LGTM, just a few comments.

libs/usb/JaRuleWidgetPort.cpp
peternewman: Although we seem to have cancelling universally elsewhere.

common/rdm/DiscoveryAgent.cpp
nomis52: The other 2 calls to FreeCurrentRange already have this or some variant of.

plugins/usbdmx/JaRuleOutputPort.h
nomis52: Fixed

common/rdm/DiscoveryAgent.cpp
peternewman: Hence is it worth removing the duplication and putting it in FreeCurrentRange?

nomis52: No, because they do different things one is a loop, the other is singular.

PR #938, opened 2015-10-06
Fix the build on Ubuntu

peternewman:


peternewman:
I don't know if we want class versus no class standardised or is there some reason for that which I've missed?

nomis52:
You need the class if you haven't included the .h file:

https://en.wikipedia.org/wiki/Forward_declaration#Classes

peternewman:
Do we want to standardise one way or the other then?

nomis52:
It depends on the use case.

If the class only appears as a pointer or a template arg like this, there is an advantage to not including the *.h since it'll speed up the build.

Issue #940, opened 2015-10-08
Consider switching to help2man


peternewman:
https://www.gnu.org/software/help2man/

Possibly for the Python tools at least, or extend our flag code to use some similar options.

peternewman:
It seems to work okay, with only minor initial tweaking needed:
help2man -N -n "Gather RDM model info from responders" --version-string=" " -o man/rdm_model_collector.py.1 tools/rdm/rdm_model_collector.py

nomis52:
Can we close this?

peternewman:
help2man -N -n "Get the value of a PID for an RDM device" --version-string=" " -o man/ola_rdm_get.1 ./examples/ola_rdm_get

peternewman:
help2man -N -n "Sets the DMX values for a universe." --version-string=" " -o man/ola_set_dmx.1 ./examples/ola_set_dmx

PR #941, opened 2015-10-10
Add some more man pages

peternewman:
These are just generated using help2man, so may not match our house style, but are better than nothing.


peternewman:
See https://github.com/OpenLightingProject/ola/issues/940 for some background.

peternewman:
We should probably cherry-pick this into 0.9 too.

nomis52:
Yeah, can you target the 0.9 branch please.

PR #953, opened 2015-10-15
Detect & handle JaRule timeouts.

nomis52:
Also add the Return Code enums

### DIFF
--- a/tools/ja-rule/ja-rule.cpp
+++ b/tools/ja-rule/ja-rule.cpp
@@ -63,6 +63,7 @@ using ola::rdm::RDMStatusCode;
 using ola::rdm::UID;
 using ola::strings::ToHex;
 using ola::usb::CommandClass;
+using ola::usb::JaRuleReturnCode;
 using ola::usb::JaRuleWidget;
 using ola::utils::JoinUInt8;
 using ola::utils::SplitUInt16;
@@ -158,6 +159,9 @@ class Controller {
     m_actions['.'] = Action(
         "Get the hardware info",
         NewCallback(this, &Controller::GetHardwareInfo));
+    m_actions[','] = Action(
+        "Run the self test",
+        NewCallback(this, &Controller::RunSelfTest));
 
     // Timing Options
     // For each of the options below, we allow a bigger range than the device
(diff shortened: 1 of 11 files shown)


tools/ja-rule/ja-rule.cpp
```diff
@@ -292,80 +296,69 @@ class Controller {
...
-      OLA_INFO << "RC: " << static_cast<int>(return_code);
-      return;
-    }
-
     string response;
     if (!payload.empty()) {
        response.append(reinterpret_cast<const char*>(payload.data()),
                        payload.size());
     }
-    OLA_INFO << "Echo Reply (" << static_cast<int>(return_code) << "): "
```
peternewman: Is it not worth leaving the Echo Reply bit, for what type of command it was?

tools/ja-rule/ja-rule.cpp
```diff
@@ -292,80 +296,69 @@ class Controller {
...
   void AckCommandComplete(
       ola::usb::USBCommandResult result,
-      uint8_t return_code, uint8_t status_flags,
+      JaRuleReturnCode return_code, uint8_t status_flags,
       const ByteString &payload) {
     if (!CheckResult(result, status_flags)) {
       return;
     }
 
-    OLA_INFO << "ACK (" << static_cast<int>(return_code)
```
peternewman: Again

tools/ja-rule/ja-rule.cpp
```diff
@@ -292,80 +296,69 @@ class Controller {
...
-             << "): payload_size: " << payload.size();
-    OLA_INFO << "Flags (" << static_cast<int>(return_code) << "):";
```
peternewman: Again

peternewman:
Generally LGTM

tools/ja-rule/ja-rule.cpp
nomis52: If the command isn't an echo the result will be COMMAND_RESULT_CLASS_MISMATCH

tools/ja-rule/ja-rule.cpp
nomis52: Same

tools/ja-rule/ja-rule.cpp
nomis52: Same

tools/ja-rule/ja-rule.cpp
peternewman: I'm meaning more so that the output is actual useful/friendly for a human, but you'll see more in my upcoming PR.

PR #954, opened 2015-10-16
Print the hardware information to stdout

nomis52:


peternewman:
LGTM

Although for the @mrpackethead I assumed it would want to just list the info immediately when plugged in?

nomis52:
That's coming later

PR #956, opened 2015-10-23
Use the build matrix subset for packages

peternewman:
https://github.com/travis-ci/travis-ci/issues/3505

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: false
 script:
  - "bash -ex .travis-ci.sh"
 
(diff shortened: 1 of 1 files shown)


nomis52:
Travis seems unhappy.

peternewman:
Yes it still needs some work @nomis52, but I think I'm getting there now.

nomis52:
LGTM

Does this help with caching at all?

peternewman:
It won't really affect caching, but reduces what needs installing, so we'll get better results when llvm is being flaky, as well as contributing to the flakiness less by reducing how much we hammer them.

PR #961, opened 2015-10-28
Fix invalid shebangs

susnux:
There should not be any shebangs in python-modules.
Encountered this issue when packaging for openSUSE.
Will fix a part of issue #904

PR #965, opened 2015-10-29
Fixed wrong shebangs in tools/rdm.

susnux:
Shebangs only in executable scripts not in modules.
Use "/usr/bin/env python", so the system will figure out the correct position of python itself.
See @peternewman s request in #961


peternewman:
Thanks @susnux . LGTM, will merge when Travis is green.

PR #968, opened 2015-11-07
add files to gitignore

daveol:
Git wanted to add these, they should probably be excluded


peternewman:
Can we add the .exe variants too please and remove the plugins/e131/e131 versions.

peternewman:
I've fixed this here: https://github.com/OpenLightingProject/ola/commit/b1ccbee03283e0d75f2eb5f262321b202de95d62

PR #969, opened 2015-11-07
Web cleanup

daveol:
Split the source in multiple files so we can unit test them at a later date


nomis52:
LGTM

javascript/new-src/Gruntfile.js
```diff
@@ -18,7 +18,24 @@ module.exports = function(grunt) {
       build: {
         files: [{
           dest: '../../olad/www/new/js/app.min.js',
-          src: 'src/js/app.js'
+          src: [
```
peternewman: Can you not define this array outside of the function to avoid duplication if/when you need to add more files to it?

PR #973, opened 2015-11-12
Autotools macros update

daveol:
A lot of macro's where obsolete look [here](http://www.gnu.org/software/autoconf/manual/autoconf-2.62/html_node/Obsolete-Macros.html) for more info

### DIFF
--- a/config/stl_hash.m4
+++ b/config/stl_hash.m4
@@ -19,10 +19,10 @@ AC_DEFUN([AC_CXX_STL_HASH],
    # First try unordered_map, but not on gcc's before 4.2 -- I've
    # seen unexplainable unordered_map bugs with -O2 on older gcc's.
    AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
-                      [#if defined(__GNUC__) && (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 2))
+                      [[#if defined(__GNUC__) && (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 2))
                        # error GCC too old for unordered_map
-                       #endif],
-                      [/* no program body necessary */])],
+                       #endif]],
+                      [[/* no program body necessary */]])],
                      [stl_hash_old_gcc=no],
                      [stl_hash_old_gcc=yes])
    for location in unordered_map tr1/unordered_map; do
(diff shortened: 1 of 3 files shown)


config/stl_hash.m4
```diff
@@ -1,4 +1,4 @@
-# Copied from the protobuf package, http://code.google.com/p/protobuf/
+	# Copied from the protobuf package, http://code.google.com/p/protobuf/
```
peternewman: Remove these extra spaces, we should probably update their URL too.

nomis52:
We need to be careful with this, not everyone has the latest autoconf. Unfortunately the page linked to doesn't tell us when the new macros became available.

Looking at releases, precise is the oldest I know of with 2.68. Luckily this is what travis runs!

nomis52:
That said, LGTM.

config/stl_hash.m4
daveol: I don't even know where those spaces came from... I will update the url but I updated the macro's myself

daveol:
CentOS6 (a.k.a. el6) is also on autoconf 2.68, and debian squeeze (oldoldstabe a.k.a. 2 versions behind) has 2.67 so I think we are safe

nomis52:
LGTM

peternewman:
FWIW, our limitation was actually 'AM_PROG_AR' which was introduced in Automake 1.11.2. Ubuntu 12.04 LTS has it, but 10.04 (EOL April this year) didn't have it. We weren't actually specifying a minimum version of automake, so I've added that in https://github.com/OpenLightingProject/ola/pull/981 .

peternewman:
Further info, @daveol and @nomis52 the Automake 1.11.2/AM_PROG_AR thing breaks mingw32. It looks like we should actually be moving to mingw-w64, which despite the name supports 32 and 64 bit. However, in the short term I wonder if we should revert that particular bit of this change, until we've got mingw-w64 working.

nomis52:
Can you paste links or log output please. 

Do you know if mingw-w64 works for us??

peternewman:
Finally as promised:

autoreconf -i
configure.ac:30: warning: macro `AM_PROG_AR' not found in library
configure.ac:30: warning: macro `AM_PROG_AR' not found in library
configure.ac:30: error: possibly undefined macro: AM_PROG_AR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf-2.68: /mingw/bin/autoconf-2.68 failed with exit status: 1

We've discussed on IRC @daveol and @nomis52 agreed we'd revert this change for 0.10, then move to mingw-w64 for 0.11. Do you know if it's just that one line AM_PROG_AR we need to revert @daveol it seems to build fine with the other changes still present, but it would be good to know if the LT_INIT win32-dll stuff is related, or just a parallel tidy?

daveol:
Probably only the AM_PROG_AR since the LT_INIT is a separate thing the only thing related is that libtool (a.k.a. LT_INIT ,a.k.a (deprecated) AC_PROG_LIBTOOL) uses ar (a.k.a. AM_PROG_AR)

peternewman:
Okay thanks @daveol . Out of interest, did you have Doxygen on the box? When I run with -Wall -Werror I get lots of Doxygen errors too.

daveol:
I had doxygen, but I couldn't figure out how to fix those errors

peternewman:
Ah okay @daveol. Done in #998 anyway.

PR #974, opened 2015-11-12
Get the build working on Windows again

peternewman:


tools/ja-rule/USBDeviceManager.cpp
```diff
@@ -180,7 +183,7 @@ void USBDeviceManager::SignalEvent(EventType event, JaRuleWidget* widget) {
     return;
   }
 
-  if (pthread_equal(m_start_thread_id, Thread::Self())) {
+  if (m_in_start && pthread_equal(m_start_thread_id, Thread::Self())) {
```
nomis52: Why do we need this bit?

peternewman: Because on Windows pthread_t is a struct ( https://github.com/OpenLightingProject/ola/blob/master/include/ola/thread/Thread.h#L39 ), we can't assign the int zero to it. If you're happy that just initialising it again with pthread_t() will be safe enough, we can drop it, the Internet seemed to suggest a separate bool would be the safest option.

nomis52: Links?

peternewman: http://stackoverflow.com/questions/6276939/is-there-an-invalid-pthread-t-id and http://www.linuxforums.org/forum/programming-scripting/123602-pthread_t-question.html

If you're happy with just "m_start_thread_id = pthread_t()" we can stick with that.

tools/ja-rule/USBDeviceManager.cpp
```diff
@@ -77,7 +79,8 @@ bool USBDeviceManager::Start() {
   }
 
   m_cleanup_thread.Start();
-  m_start_thread_id = 0;
+  m_start_thread_id = pthread_t();
```
nomis52: Can you remove this line then, it's not needed

tools/ja-rule/USBDeviceManager.cpp
nomis52: Ok, please add a comment explaining why this is required.

tools/ja-rule/USBDeviceManager.cpp
peternewman: Done

tools/ja-rule/USBDeviceManager.cpp
peternewman: Done

nomis52:
LGTM

PR #978, opened 2015-11-20
Fix more flake8 issues and make flake8 passing mandatory

peternewman:


nomis52:
LGTM

peternewman:
@nomis52 this should now all be done apart from:
"python/examples/rdm_snapshot.py:423:81: E501 line too long (81 > 80 characters)"

I'm not really sure how best to solve that last one apart from splitting the print statements, splitting the line tracks through to output, which kind of spoils the usage output.

peternewman:
Do we also want this check merged into the main lint one on Travis, rather than run separately?

nomis52:
I'd keep the lint separate. I don't have any good ideas on how to fix rdm_snapshot.py

peternewman:
Okay, lint kept separate. This is finally green @nomis52 including fixing up the other Python files I'd missed.

include/ola/gen_callbacks.py
```diff
@@ -41,14 +42,15 @@ def Header():
    *
    * You should have received a copy of the GNU Lesser General Public
    * License along with this library; if not, write to the Free Software
-   * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+   * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 \
```
peternewman: The other option here is to take this one line out of the triple quotes and print it separately, if we'd rather avoid the backslash.

python/examples/rdm_snapshot.py
```diff
@@ -419,9 +419,9 @@ def _RDMRequestComplete(self, response, unpacked_data, unpack_exception):
 
 
 def Usage():
+  print("Usage: rdm_snapshot.py --universe <universe> [--input <file>] "
+        "[--output <file>]\n")
   print(textwrap.dedent("""\
-  Usage: rdm_snapshot.py --universe <universe> [--input <file>] [--output <file>]
```
peternewman: We could use the backslash here instead to avoid a separate print line if you'd rather?

peternewman:
Just doing a final check for python files, I've found the tests folder with recvdmx.py and regression.py which both still refer to LLA. Is it worth fixing them up, they seem to want an Observer, is that now ClientWrapper? Or should we just delete them?

nomis52:
tests/ can be deleted.

python/examples/rdm_snapshot.py
nomis52: Up to you.

nomis52:
Still LGTM

python/examples/rdm_snapshot.py
peternewman: I'll stick with the current mix then, I think this is cleaner than backslash, but probably not in the middle of a block where I'd need to split it into three print statements.

peternewman:
Removed tests/.

peternewman:
Final review time @nomis52 , merge if you're happy.

nomis52:
LGTM

PR #981, opened 2015-11-23
Specify the minimum automake version we require

peternewman:


peternewman:
Rather unhelpfully it throws an error because AM_PROG_AR doesn't exist before it gets to the version check, but at least we'll have it noted down for the future.

nomis52:
LGTM

PR #989, opened 2015-12-13
Update NEWS

peternewman:


peternewman:
Targetting 0.10 in #990 instead.

PR #1014, opened 2016-01-14
Fix new flake8 issues due to latest pep8

peternewman:

PR #1015, opened 2016-01-22
OLA Trigger tidying and improvements

peternewman:


tools/ola_trigger/config.lex
```diff
@@ -40,7 +40,7 @@ whitespace   [ \t]
 ^#[^\n]*             {}
 {whitespace}+#[^\n]* {}
 {whitespace}+     { count(); return WHITESPACE; }
-\n                { count(); yylineno++; return NEW_LINE; }
+\n                { count(); return NEW_LINE; }
```
nomis52: Don't we need this to tell you which line the error occurs on?

tools/ola_trigger/config.ypp
```diff
@@ -65,6 +67,12 @@ assignment :
         delete $1;
         delete $5;
       }
+    | WORD_LITERAL optional_whitespace '=' optional_whitespace INTEGER_LITERAL {
```
nomis52: Rather than repeating this change 

value:
        INTEGER_LITERAL { $$ = ola::strings::IntToString($3); } 
     | WORD_LITERAL
     | QUOTED_VALUE

Same for `argument` below.

peternewman:
Thanks for the comments. This is blocking on https://github.com/OpenLightingProject/ola/pull/1014#issuecomment-172871600 in #1014 though.

peternewman:
Thanks for the comments @nomis52 . This is blocking on https://github.com/OpenLightingProject/ola/pull/1014#issuecomment-172871600 in #1014 though.

tools/ola_trigger/config.lex
peternewman: I found with this in place, the line numbers were wrong; with it removed, they were correct. It's probably worth testing on some other OSes though.

I think we could remove our yywrap function too, and just use the noyywrap option if it's present in old enough releases.

tools/ola_trigger/config.ypp
peternewman: Thanks. That's good to know. I take it you actually mean "INTEGER_LITERAL { $$ = ola::strings::IntToString($1); }" though.

Done but not tested yet.

peternewman:
@nomis52 do you want to merge this if you're happy?

nomis52:
LGTM

PR #1018, opened 2016-01-26
Resync with 0.10

peternewman:


peternewman:
I think https://github.com/OpenLightingProject/ola/pull/1019 should fix the merge conflicts, and I believe this will result in no files changed when the other one is merged.

PR #1029, opened 2016-02-07
Python RpcChannel incorrectly decodes message header on big-endian architectures

aleden:
On the C++ server side, RpcChannel::SendMsg in common/rpc/RpcChannel.cpp does

```
   RpcHeader::EncodeHeader(&header, PROTOCOL_VERSION,
                                length - sizeof(header));
```

to produce the header, and then sends it by doing

```
  output.replace(
      0, sizeof(header),
      reinterpret_cast<const char*>(&header), sizeof(header));

  ssize_t ret = m_descriptor->Send(
      reinterpret_cast<const uint8_t*>(output.data()), length);
```

On the OLA client program's Python API side, `StreamRpcChannel::_ProcessIncomingData` in `ola/rpc/StreamRpcChannel.py` does

        header = struct.unpack('<L', raw_header)[0]
        version, size = self._DecodeHeader(header)

to receive the header. The problem is that the format string passed to `struct.unpack` instructs to use little-endian decoding instead of the native byte order.

Consequently, on big-endian architectures

> Protocol mismatch: 0 != 1

is printed and the value of `expected_size` becomes `100663312`, which results in the OLA client program never receiving a message, since `_ProcessIncomingData` will always return at `# not enough data yet`). The program will use more and more CPU as `self._buffer` grows larger since `data_size = sum([len(s) for s in self._buffer])` will take increasingly longer to execute.

Tested on powerpc.

### DIFF
--- a/python/ola/rpc/StreamRpcChannel.py
+++ b/python/ola/rpc/StreamRpcChannel.py
@@ -240,7 +240,7 @@ class StreamRpcChannel(service.RpcChannel):
         if not raw_header:
           # not enough data yet
           return
-        header = struct.unpack('<L', raw_header)[0]
+        header = struct.unpack('=L', raw_header)[0]
         version, size = self._DecodeHeader(header)
 
         if version != self.PROTOCOL_VERSION:


peternewman:
Excellent spot thanks @aleden . I saw the email notification and assumed it was going to be an issue, so it being a PR is a nice treat.

Would you mind targeting 0.10 branch with this please, so we can get it in the latest release (or we can do it afterwards if you'd rather not)?

@nomis52 I believe I'm right in thinking this shouldn't be an issue in future, as we don't support cross machine RPC, i.e. we don't need to switch to network/away from native encoding?

peternewman:
I assume you've tested comms in both directions @aleden , but it looks like the pack is native already:
https://github.com/OpenLightingProject/ola/blob/master/python/ola/rpc/StreamRpcChannel.py#L148

aleden:
@peternewman Should I submit another pull request to 0.10?

Yeah, the pack looks native. I ran the Python "Receiving DMX" example on powerpc (Debian Jessie) with no issues to speak of. No "Protocol mismatch" message, CPU usage looks stable :)

peternewman:
We normally merge into 0.10, if it's going into there, then sync that into master on a semi-regular basis. If I've gone into master first, I cherry-pick back into 0.10.

peternewman:
@aleden do you want to create the new 0.10 PR, if not we can merge this and cherry-pick?

aleden:
Hi, @peternewman created a new 0.10 PR.

peternewman:
Closing this as it will appear via 0.10's branch and the PR associated here https://github.com/OpenLightingProject/ola/pull/1036

PR #1031, opened 2016-02-09
Fix a configure warning on OpenBSD

peternewman:
Also handle another type of pthread_set_name


peternewman:
I think this is good to merge now @nomis52 .

nomis52:
LGTM

PR #1035, opened 2016-02-10
Grunt cleanup

daveol:
This cleans up some of the mess that was going on in the gruntfile


peternewman:
You've added app.js, rather than updating app.min.js. If you intend to keep it, it needs adding to the makefiles.

javascript/new-src/src/constants.js
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
peternewman: Is this a personal choice to change them? It's upset the licence enforcer. So could we change it back, or make all the old SRC the same too and change this (although that may upset the other linter):
https://github.com/OpenLightingProject/ola/blob/master/scripts/enforce_licence.py#L142

javascript/new-src/src/app.js
```diff
@@ -1,4 +1,4 @@
-/**
+/*!
```
peternewman: Is the ! a typo?

javascript/new-src/Gruntfile.js
```diff
@@ -14,28 +49,20 @@ module.exports = function(grunt) {
...
-            'src/controllers/add_universe.js',
-            'src/controllers/plugin_info.js',
-            'src/controllers/setting_universe.js',
-            'src/controllers/header.js',
-            'src/controllers/overview.js',
-            'src/constants.js',
-            'src/factories/ola.js',
-            'src/filters/start_form.js'
-          ]
+          src: ['../../olad/www/new/js/app.js']
```
peternewman: Are you creating the interim file, or is this just testing and the above will become another array eventually?

javascript/new-src/Gruntfile.js
```diff
@@ -1,4 +1,39 @@
 /*jshint node: true*/
+// array of source files to concat
+var targets = {
+  js: [
```
peternewman: Could this be js_src or something perhaps, so it looks more obvious below in the concats?

javascript/new-src/src/constants.js
daveol: No, it's because the header gets stripped for app.js so we don't end up with a load of license headers.

javascript/new-src/src/app.js
daveol: No, it's to prevent the header from being dropped by the concat job that creates app.js

javascript/new-src/Gruntfile.js
daveol: sure

javascript/new-src/Gruntfile.js
daveol: I intend to create (and keep) the app.js file for debugging purpouses, I will add makefile/server handlers for it

peternewman: Okay cool. I take it the minified one will continue to be the default?

javascript/new-src/src/constants.js
peternewman: I see more specifically it's because it only strips /* and not /**. It looks like the Closure Linter is happy with /* too, so if you can change all the JS comments to /* and then fix this line:
https://github.com/OpenLightingProject/ola/blob/master/scripts/enforce_licence.py#L142

javascript/new-src/Gruntfile.js
daveol: yes

javascript/new-src/src/constants.js
daveol: It works now, so it is not really needed anymore

peternewman:
We don't have an updated olad/www/new/js/app.min.js, or has that not actually changed despite all the newline fiddling?

daveol:
olad/www/new/js/app.min.js has not changed since the minifier strips white space whenever possible

peternewman:
Ah of course.

peternewman:
Can you resync please @daveol .

javascript/new-src/Gruntfile.js
```diff
@@ -124,7 +119,7 @@ module.exports = function(grunt) {
     cssmin: {
       build: {
         files: [{
-          src: 'css/style.css',
+          src: targets.css,
```
peternewman: From the documentation, this looks like it might not work properly if you add another file to the array:
https://github.com/gruntjs/grunt-contrib-cssmin#combine-two-files-into-one-output-file

peternewman:
Just one comment, otherwise LGTM.

javascript/new-src/Gruntfile.js
daveol: It still works, the way they document is just a simpler way of writing it. The dest/src properties get handled by grunt if I recall correctly

peternewman: Okay cool.

peternewman:
LGTM, any final comments @nomis52 ?

nomis52:
LGTM

peternewman:
Sorry @daveol , bad timing, another resync then it's good to merge.

PR #1036, opened 2016-02-11
message header is native byte order, decode it as such

aleden:


peternewman:
LGTM.

Retrospective review if you fancy @nomis52

nomis52:
LGTM

PR #1044, opened 2016-02-25
(Fixes #1043): Use pkg-config for cppunit

### LINKED ISSUE #1043: cppunit-config removed in fedora
In the latest version of the cppunit package for fedora, cppunit-config is removed in favor for pkg-config based solutions ([see here](https://apps.fedoraproject.org/packages/cppunit/changelog/))

This is the patch I made for the package in the mean time, it's not pretty and probably not a proper solution:
```patch
diff --git a/config/cppunit.m4 b/config/cppunit.m4
index 41f067b..7a7834c 100644
--- a/config/cppunit.m4
+++ b/config/cppunit.m4
@@ -25,12 +25,11 @@ AC_ARG_WITH(cppunit-exec-prefix,[  --with-cppunit-exec-prefix=PFX  Exec prefix w
   AC_PATH_PROG(CPPUNIT_CONFIG, cppunit-config, no)
   cppunit_version_min=$1
 
-  AC_MSG_CHECKING(for Cppunit - version >= $cppunit_version_min)
   no_cppunit=""
   if test "$CPPUNIT_CONFIG" = "no" ; then
-    AC_MSG_RESULT(no)
-    no_cppunit=yes
+    PKG_CHECK_MODULES(CPPUNIT,[cppunit >= $cppunit_version_min],,[no_cppunit="yes"])
   else
+    AC_MSG_CHECKING(for Cppunit - version >= $cppunit_version_min)
     CPPUNIT_CFLAGS=`$CPPUNIT_CONFIG --cflags`
     CPPUNIT_LIBS=`$CPPUNIT_CONFIG --libs`
     cppunit_version=`$CPPUNIT_CONFIG --version`
``` 
This will do in the mean time, but either a better version of this fix has to be made or some other solution.

daveol:
The old cppunit macro AM_PATH_CPPUNIT still used cppunit-config which
has been deprecated since pkgconfig came along. The cppunit macro has
been removed and the check now uses PKG_CHECK_MODULES instead.


nomis52:
Seems legit.

peternewman:
Should we backport this into 0.10.1, if it's going to break the builds on the latest OSes?

nomis52:
Yeah that's probably a good idea.

Issue #1046, opened 2016-02-27
Provide warning in web UI when JavaScript is turned off


brunchboy:
I just had a panic (helpfully resolved by kind folks on the IRC channel) thinking that my `olad` installation had gotten broken somehow, when it was because I left JavaScript turned off in my browser after trying to read news on a site which is plagued by aggressive ads. The pressure of preparing for a show prevented me from figuring this out on my own. It would be nice if the UI could help people in similar brain-impaired situations before panic sets in. I may be able to submit a PR myself to do this once I am past the pressure of preparing for my current show, but wanted to add this issue in case I forget, or in case someone else wants to get to it first.

daveol:
Basically we want [<noscript>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/noscript) tags with the appropriate text, or maybe just an element that gets removed by a javascript call

brunchboy:
Yup. I just investigated taking a crack at this because it seemed straightforward enough, but your Javascript is built using a complicated procedure and the [README](https://github.com/OpenLightingProject/ola/tree/master/javascript) is out of date and refers to resources which are no longer at the locations specified, and likely no longer available using `svn`. I am spending a few minutes to see if I can get past that, but need to get back to designing my show soon.

brunchboy:
Yeah, I can’t even figure out how to compile it to test using a simple `<noscript>` approach. I found `mac.build.sh` but when I try it, I get:

    ➜ ola git:(master) ✗ mkdir /tmp/ola
    ➜ ola git:(master) ✗ ./mac_build.sh /tmp/ola/
    ./mac_build.sh: line 18: ./configure: No such file or directory
    make: *** No targets specified and no makefile found.  Stop.
    make: *** No rule to make target `check'.  Stop.
    Password:

I will leave this to someone who has practice making changes to OLA. :smile:

Oh, but wait! I found where the HTML lives in my installation from Homebrew, so I can test my changes there. I have come up with a simple solution that works and looks fine, so a PR is coming shortly.

PR #1047, opened 2016-02-28
Add warning if JavaScript is off, #1046

brunchboy:
Now a simple `noscript` tag will inform the user why the `olad` web
interface is not working, and how to remedy the situation.

### DIFF
--- a/olad/www/mobile.html
+++ b/olad/www/mobile.html
@@ -447,6 +447,15 @@
       <div id="plugin_frame"><div id="plugin_list"></div></div>
       <div id="plugin_info_frame"></div>
     </div>
+    <noscript>
+      <br>
+      <p align="center">
+        <strong>This interface requires JavaScript to function. If you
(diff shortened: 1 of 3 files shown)


peternewman:
Can you fix https://github.com/brunchboy/ola/blob/add-javascript-help/olad/www/mobile.html the old mobile UI too please @brunchboy . It would make sense to target your patch at 0.10 branch so we can get it in 0.10.1.

brunchboy:
Sure, sorry I missed that, I never tried the mobile UI, and I should!

brunchboy:
I have recreated this PR agains the 0.10 branch, so perhaps this one can be discarded?

brunchboy:
Closing because I opened against the wrong branch, per Peter Newman.

peternewman:
See https://github.com/OpenLightingProject/ola/pull/1048 instead.

PR #1048, opened 2016-02-28
Add warning if JavaScript is off, #1046

brunchboy:
Now a simple `noscript` tag will inform the user why the `olad` web
interface is not working, and how to remedy the situation.

Includes new, old, and old mobile interfaces. Recreated to submit to 0.10 branch.


olad/www/new/index.html
```diff
@@ -54,6 +54,15 @@
  </div>
  <!-- /.container-fluid -->
 </nav>
+<noscript>
+  <br>
+  <p align="center">
+    <strong>This interface requires JavaScript to function. If you
```
peternewman: If I'm being really pedantic, it would be nice if the indentation was consistent in each code block, you've got some on the would line here, but not on the others.

olad/www/mobile.html
```diff
@@ -448,6 +448,15 @@
       <div id="plugin_info_frame"></div>
     </div>
 
+    <noscript>
+      <br>
+      <p align="center">
+        <strong>This interface requires JavaScript to function. If you
+        would like to use it, enable JavaScript and reload the
```
peternewman: We could be polite and say "please enable"

peternewman:
LGTM, only minor nits @brunchboy .

olad/www/new/index.html
brunchboy: That is how Emacs indents HTML. To change that I could put the `strong` tags on their own lines, would you prefer that?

brunchboy: Or wait, let me look at the others, perhaps I misunderstood.

brunchboy: Right you are, in copying and pasting, the indentation got messed up. Fixing now.

olad/www/mobile.html
brunchboy: I thought about that, and almost wrote it that way, but decided that it was the user’s choice. If you like I’d be happy to add it while I am in there fixing the indentations.

brunchboy:
That’s fine! Since we both thought about saying “please” in the message, that seems like reason enough to do so. All those changes are now pushed.

peternewman:
I'd agree it's their choice, although given it doesn't work at all if they disable Javascript, its not really their choice.

Can you hit the update button please, as I've just merged another change in.

brunchboy:
Done! :smile:

peternewman:
LGTM

nomis52:
LGTM

PR #1050, opened 2016-02-28
Correct the reported min and max for DMX start address via the JSON API

### LINKED ISSUE #1042: Old UI doesn't let you set DMX address of a one slot device to 512
See the original report here:
https://groups.google.com/forum/#!topic/open-lighting/Vf2kRA1PMII

Interestingly it does work via the RDM Patcher.

It would appear we've got an off by one error here:
https://github.com/OpenLightingProject/ola/blob/master/olad/RDMHTTPModule.cpp#L2107

The returned JSON is:
`{
  "error": "",
  "items": [
    {
      "description": "DMX Start Address",
      "id": "address",
      "max": 511,
      "min": 0,
      "type": "uint",
      "value": 512
    }
  ],
  "refresh": true
}`

It also allows you to set the address -1, and then gets NAcked by the responder.

@nomis52 is there some reason behind this, or is it just a straight bug?

peternewman:
Closes #1042

### DIFF
--- a/include/ola/Constants.h
+++ b/include/ola/Constants.h
@@ -37,6 +37,16 @@ enum {
 };
 
 /**
+ * @brief The minimum number for a DMX512 slot.
+ */
+static const uint16_t DMX_MIN_SLOT_NUMBER = 1;
+
+/**
+ * @brief The maximum number for a DMX512 slot.
+ */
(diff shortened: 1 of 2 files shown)


nomis52:
LGTM

PR #1052, opened 2016-02-28
Fix a Python include.

nomis52:


peternewman:
LGTM but I want to check the Python linters when Travis is done.

peternewman:
Yes, it's not got any worse at least.

PR #1056, opened 2016-03-01
New placeholder for NEWS

peternewman:


peternewman:
Just use #1057, which has this in too.

PR #1062, opened 2016-03-12
Make DNS-SD functionality optional at build time.

bk138:
When building for embedded systems, for instance, we
want to limit dependencies to a minimum and it seems
that OLA is able to perform its core tasks without
DNS-SD perfectly fine.

Signed-off-by: Christian Beier <dontmind@freeshell.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -411,22 +411,25 @@ AS_IF([test "x$use_ossp_uuid" = xyes],
 
 # DNS-SD support
 # We use either avahi or the Apple DNS-SD library.
+AC_ARG_WITH(dns-sd,
+[AS_HELP_STRING([--without-dns-sd], [disable DNS-SD support])],,)
+if test "x$with_dns_sd" != "xno"; then
+	# dns_sd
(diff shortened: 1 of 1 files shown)


peternewman:
LGTM @nomis52 ? Also @bk138 you're correct, DNS-SD is just a nice to have for us, not a requirement.

configure.ac
```diff
@@ -411,23 +411,26 @@ AS_IF([test "x$use_ossp_uuid" = xyes],
 
 # DNS-SD support
 # We use either avahi or the Apple DNS-SD library.
-
-# dns_sd
-AC_CHECK_HEADER(
-  [dns_sd.h],
-  [AC_SEARCH_LIBS(DNSServiceRegister, [dns_sd], [have_dnssd="yes"])])
+AC_ARG_WITH(dns-sd,
+[  --without-dns-sd        disable DNS-SD support],,)
```
nomis52: Use AS_HELP_STRING() please

bk138:
Now using AS_HELP_STRING @nomis52

PR #1076, opened 2016-05-19
Fix a segfault with the client if auto start is off or olad fails to start

peternewman:
As spotted by @s-light on IRC.


peternewman:
With auto start on, and olad not able to start for some reason, e.g. sudo, the same thing happened too (which has also been fixed):
sudo ./examples/ola_rdm_discover -u 99

peternewman:
This seemed worth getting into 0.10.2 @nomis52 .

PR #1077, opened 2016-05-20
Final NEWS update and release 0.10.2

peternewman:
I suspect this may fail Travis due to the following in our ola_protoc which the latest clang 3.9.0 has brought up. I don't know your thoughts but I'd be tempted to force it through regardless, then we can work on fixing this issue for the next release.

```
common/rpc/TestServiceService.pb.cpp:137:14: error: binding dereferenced null
      pointer to reference has undefined behavior [-Werror,-Wnull-dereference]
      return *reinterpret_cast< ::google::protobuf::Message*>(NULL);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
common/rpc/TestServiceService.pb.cpp:153:14: error: binding dereferenced null
      pointer to reference has undefined behavior [-Werror,-Wnull-dereference]
      return *reinterpret_cast< ::google::protobuf::Message*>(NULL);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-x/5/2016 ola-0.10.2
+21/5/2016 ola-0.10.2
  Features:
  * Add support for Enttec USB Pro Mk 2 B
  * Make DNS-SD functionality optional at build time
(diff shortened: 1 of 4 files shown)


nomis52:
LGTM

PR #1078, opened 2016-05-21
Resync 0.10 into master

peternewman:


peternewman:
Need to do it in one of my branches.

PR #1097, opened 2016-06-29
Update gpio_pins to uint16_t closes #1095

### LINKED ISSUE #1095: C.H.I.P GPIO nonfunctional
Attempting to use the GPIO plugin on a CHIP, which starts at either 408 or 1023 depending on kernel version (http://docs.getchip.com/chip.html#gpio) fails. I think this might be because of this line:

https://github.com/OpenLightingProject/ola/blob/master/plugins/gpio/GPIOPlugin.cpp#L81

I'm going to try and recompile with 'pin' initialized as a uint16_t tonight and see if that clears up the problem, but there may be things I've overlooked.

loansindi:
CHIP GPIO starts at either 408 or 1023, so the pins vars should be uint16_t, rather than uint8_t


plugins/spi/SPIDevice.h
```diff
@@ -86,7 +86,7 @@ class SPIDevice: public ola::Device {
   static const char SPI_DEVICE_NAME[];
   static const char HARDWARE_BACKEND[];
   static const char SOFTWARE_BACKEND[];
-  static const uint8_t MAX_GPIO_PIN = 25;
+  static const uint16_t MAX_GPIO_PIN = 25;
```
peternewman: You'll need to increase the value here too, or you won't be able to use those GPIO pins.

loansindi: i didn't change that, and i'm using GPIO pin 415 successfully - is there maybe something funny happening with whatever check looks at MAX_GPIO_PIN?

loansindi: So is the point of this to set which pins can be used as additional chip selects? If the goal is to let any GPIO pin be defined for SPI use, we should set it to (i believe) 1024 (as in kernel 4.4, CHIP GPIO starts at 1016).

peternewman: That's what the code seems to do.

@nomis52 what are your thoughts, I'm inclined to say if the range is so large it becomes a bit of a pointless check. We could either remove it, or test against a series of offsets (0, 415, 1016), with each having a range up to 24. Or we could just have the warning and not ignore the value, so it would show in the log as a warning about a possible typo.

plugins/gpio/GPIOPort.cpp
```diff
@@ -44,11 +44,11 @@ bool GPIOOutputPort::Init() {
 }
 
 string GPIOOutputPort::Description() const {
-  vector<uint8_t> pins = m_driver->PinList();
+  vector<uint16_t> pins = m_driver->PinList();
   std::ostringstream str;
   str << "Pins ";
 
-  vector<uint8_t>::const_iterator iter = pins.begin();
+  vector<uint16_t>::const_iterator iter = pins.begin();
```
peternewman: I know this wasn't your code, but can you replace this whole block with our StringJoin function, rather than duplicating code please.

peternewman:
Just a few minor comments, generally LGTM.

nomis52:
LGTM.

You can increase the SPi pin # if you like. It's intended as a sanity check.

plugins/gpio/GPIOPort.cpp
loansindi: I've had a busy couple weeks so I haven't checked in on this - I'll try to look into this this week and see if I can do that.

peternewman:
@loansindi Can you resync please.

peternewman:
Thanks @loansindi . Have you got time to address those outstanding issues?

loansindi:
I've looked at the string join thing a little bit and have to admit I'm not
100% confident in my ability to do that (I don't do a lot of c++) but I can
try taking a crack at it. I can also increase the spi count if we want to
do that.

peternewman:
Okay, well if you could at least do the SPI count, I can always pick up the string join bit if you get stuck.

It should just be a case of deleting the lines involving the iterator and switching to:
http://docs.openlighting.org/ola/doc/latest/namespaceola.html#a3a63159a0ab0cb63dca869fa38f47047

str << ola::StringJoin(", ", pins);

Or something similar, and possibly adding an include. There are plenty of examples elsewhere in the code. But if you get stuck, no worries, we can sort it out.

loansindi:
So something like this?

string GPIOOutputPort::Description() const {
  vector<uint8_t> pins = m_driver->PinList();
  std::ostringstream str;
  str << "Pins ";
  str << ola::StringJoin(", ", pins);
  }
  return str.str();
}

I just kicked off a build to see if that makes anything obviously explode.

peternewman:
That looks about right @loansindi .

peternewman:
If you resync, the clang error should go away @loansindi .

loansindi:
I pushed my changes - I'm still having trouble building locally but it looks like travis isn't choking where I am, I'm not really sure what's going on with that.

plugins/gpio/GPIOPort.cpp
```diff
@@ -47,12 +48,7 @@ string GPIOOutputPort::Description() const {
...
   std::ostringstream str;
   str << "Pins ";
-
-  vector<uint16_t>::const_iterator iter = pins.begin();
-  while (iter != pins.end()) {
-    str << static_cast<unsigned int>(*iter++);
-    if (iter != pins.end())
-      str << ", ";
+  str << ola::StringJoin(", ", pins);
   }
```
peternewman: I think this } is errant/extra, I'm not sure why Travis isn't complaining about it.

plugins/spi/SPIDevice.h
```diff
@@ -86,7 +86,7 @@ class SPIDevice: public ola::Device {
   static const char SPI_DEVICE_NAME[];
   static const char HARDWARE_BACKEND[];
   static const char SOFTWARE_BACKEND[];
-  static const uint16_t MAX_GPIO_PIN = 25;
+  static const uint8_t MAX_GPIO_PIN = 1023;
```
peternewman: This needs to be uint16_t. Again I'm not clear why Travis isn't catching it.

loansindi: I had some git troubles, sorry. Will rectify in the morning.

peternewman:
You seem to have managed to pull all the changes from master into your 0.10 branch at some point @loansindi . I'm not aware of an easy way to unpick that, apart from starting again on a clean 0.10 based branch and either re-applying your changes or cherry-picking from your existing PR.

Travis is indeed choking, if your code is failing in a different place, do you want to post the error and we can try and diagnose it.

plugins/spi/SPIDevice.h
peternewman: No worries, thanks @loansindi . I think it's almost there when you fix the git issues.

peternewman:
Due to issues with master getting merged in, this PR has been closed and work continues in #1123 .

PR #1116, opened 2016-09-01
Work around <ernno.h> include failure on mips.

yoe:
On mips architectures, `<asm/termios.h>` sets some cpp macros which cause
`<cerrno>` (included by `<ostream>`, used by `<ola/Logging.h>`) to not define
ERANGE, EDOM, or EILSEQ, causing a spectacular [compile failure](https://buildd.debian.org/status/fetch.php?pkg=ola&arch=mips&ver=0.10.2-2&stamp=1472208742) there.

Explicitly include `<cerrno>` at the right location to avoid the issue.

### DIFF
--- a/common/io/ExtendedSerial.cpp
+++ b/common/io/ExtendedSerial.cpp
@@ -42,7 +42,7 @@
 // ERANGE, EDOM, or EILSEQ, causing a spectacular compile failure there.
 //
 // Explicitly include <cerrno> now to avoid the issue.
-#include <cerrno>
+#include <errno.h>
 #include <asm/termios.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


common/io/ExtendedSerial.cpp
```diff
@@ -36,6 +36,15 @@
 
 #ifdef HAVE_ASM_TERMIOS_H
 // use this not standard termios for custom baud rates
+//
+// On mips architectures, <asm/termios.h> sets some cpp macros which cause
+// <cerrno> (included by <ostream>, used by <ola/Logging.h>) to not define
+// ERANGE, EDOM, or EILSEQ, causing a spectacular compile failure there.
+//
+// Explicitly include <cerrno> now to avoid the issue.
+#if HAVE_ERRNO_H
```
peternewman: As discussed on IRC, I wonder if we should die during configure if errno.h isn't present and then take these guards out here.

peternewman:
Just one comment from me.

peternewman:
Thanks, LGTM.

peternewman:
@nomis52 this just needs a quick review too.

PR #1123, opened 2016-09-13
Update GPIO and SPI pin limits for CHIP

loansindi:
New PR because I broke the old one pretty bad. 

This updates maximum pin values to use uint_16 (from uint_8) because the CHIP's GPIO starts above 255 in all cases, making GPIO and SPI assignments fail due to overflow.

### DIFF
--- a/plugins/gpio/GPIOPort.cpp
+++ b/plugins/gpio/GPIOPort.cpp
@@ -44,11 +44,11 @@ bool GPIOOutputPort::Init() {
 }
 
 string GPIOOutputPort::Description() const {
-  vector<uint8_t> pins = m_driver->PinList();
+  vector<uint16_t> pins = m_driver->PinList();
   std::ostringstream str;
   str << "Pins ";
 
-  vector<uint8_t>::const_iterator iter = pins.begin();
+  vector<uint16t>::const_iterator iter = pins.begin();
   while (iter != pins.end()) {
     str << static_cast<unsigned int>(*iter++);
     if (iter != pins.end())
--- a/plugins/gpio/GPIOPort.cpp
+++ b/plugins/gpio/GPIOPort.cpp
@@ -19,12 +19,12 @@
  */
 
 #include "plugins/gpio/GPIOPort.h"
-#include "ola/StringUtils.h"
 
 #include <sstream>
 #include <string>
 #include <vector>
 
+#include "ola/StringUtils.h"
 #include "plugins/gpio/GPIODriver.h"
 
 namespace ola {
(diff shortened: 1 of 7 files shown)


plugins/gpio/GPIOPort.cpp
```diff
@@ -44,15 +45,10 @@ bool GPIOOutputPort::Init() {
...
   std::ostringstream str;
   str << "Pins ";
-
-  vector<uint8_t>::const_iterator iter = pins.begin();
-  while (iter != pins.end()) {
-    str << static_cast<unsigned int>(*iter++);
-    if (iter != pins.end())
-      str << ", ";
+  str << ola::StringJoin(", ", pins);
   }
```
peternewman: I think this } is errant/extra.

plugins/spi/SPIDevice.h
```diff
@@ -86,7 +86,7 @@ class SPIDevice: public ola::Device {
   static const char SPI_DEVICE_NAME[];
   static const char HARDWARE_BACKEND[];
   static const char SOFTWARE_BACKEND[];
-  static const uint8_t MAX_GPIO_PIN = 25;
+  static const uint8_t MAX_GPIO_PIN = 1023;
```
peternewman: This needs to be uint16_t.

plugins/gpio/GPIOPort.cpp
```diff
@@ -19,6 +19,7 @@
  */
 
 #include "plugins/gpio/GPIOPort.h"
+#include "ola/StringUtils.h"
```
peternewman: This needs to be just above #include "plugins/gpio/GPIODriver.h" rather than up here please.

plugins/gpio/GPIOPort.cpp
```diff
@@ -44,16 +45,10 @@ bool GPIOOutputPort::Init() {
...
   str << "Pins ";
-
-  vector<uint8_t>::const_iterator iter = pins.begin();
-  while (iter != pins.end()) {
-    str << static_cast<unsigned int>(*iter++);
-    if (iter != pins.end())
-      str << ", ";
-  }
+  str << ola::StringJoin(", ", pins);
   return str.str();
```
peternewman: This whole bit could now just be rewritten as:
`return "Pins " << ola::StringJoin(", ", pins);`

Then you can drop the std::ostringstream usage too.

peternewman:
One minor comment, otherwise LGTM.

nomis52:
LGTM

peternewman:
The build failures are an outstanding issue with the Travis Mac build so I'm merging anyway.

PR #1132, opened 2016-09-27
Remove replacing -I with -isystem

joerg-krause:
Building OLA with a GCC 6 cross-toolchain fails:

```
/usr/bin/arm-linux-g++ -DHAVE_CONFIG_H -I.   -D_LARGEFILE_SOURCE
-D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -I./include -I./include
-Wall -Wformat -W -isystem
/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include -pthread
-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -Os
-pthread -c -o libs/acn/e131_transmit_test.o
libs/acn/e131_transmit_test.cpp
/usr/bin/arm-linux-g++ -DHAVE_CONFIG_H -I.   -D_LARGEFILE_SOURCE
-D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -I./include -I./include
-Wall -Wformat -W -isystem
/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include -pthread
-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -Os
-pthread -c -o libs/acn/E131TestFramework.o
libs/acn/E131TestFramework.cpp
In file included from
/opt/ext-toolchain/arm-buildroot-linux-gnueabihf/include/c++/6.1.0/ext/string_conversions.h:41:0,
                 from
/opt/ext-toolchain/arm-buildroot-linux-gnueabihf/include/c++/6.1.0/bits/basic_string.h:5402,
                 from
/opt/ext-toolchain/arm-buildroot-linux-gnueabihf/include/c++/6.1.0/string:52,
                 from ./tools/ola_trigger/config.ypp:2:
/opt/ext-toolchain/arm-buildroot-linux-gnueabihf/include/c++/6.1.0/cstdlib:75:25:
fatal error: stdlib.h: No such file or directory
 #include_next <stdlib.h>
                         ^
compilation terminated.
```

The C++ library in GCC 6 now provides its own `<stdlib.h>` header that
wraps the C library header of the same name, so in `<cstdlib>` the
header include

```
#include <stdlib.h>
```

has become

```
#include_next <stdlib.h>
```

`#include_next` is sensitive to the order of directories in the
preprocessor's search path, so if that order is changed with `-isystem`
then the compiler can't find the right header:

```
[1] /usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include
[2] /opt/ext-toolchain/arm-buildroot-linux-gnueabihf/include/c++/6.1.0
[..]
End of search list.
```

`<cstdlib>` is located in [2] whereas `<stdlib.h>` (C library header) is
in [1]. In this case, the `#include_next <stdlib.h>` statement in
`<cstdlib>`, located in [2], is evaluated **after** the search path [1],
so the compiler does not find the right system header.

The problem is that the OLA build system replaces the `-I` in the CFLAGS
from libprotobuf with `-isystem` to fix some warnings treated as errors
in the libprotobuf header files.

`-isystem` should be used to suppress warnings in system headers only
and the libprotobuf header files are not system files.

The correct fix is to compile with less restrictions and remove
`-Werror` for the build.

As using `-isystem` is reordering GCCs search path and using `-isystem`
is really not necessary, remove the faulty replacement of `-I`.

Closes: #1125

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -38,6 +38,10 @@ AM_DISTCHECK_CONFIGURE_FLAGS = --enable-python-libs
 # COMMON_TESTING_FLAGS_ONLY_WARNINGS
 # The common C++ flags for testing with (non-fatal) warnings enabled. Use this
 # for 3rd party code that does not build correctly with COMMON_TESTING_FLAGS.
+#
+# COMMON_TESTING_PROTOBUF_FLAGS
+# The common C++ flags for testing with some warnings protobuf code throws
+# disabled.
 COMMON_CXXFLAGS_ONLY_WARNINGS = \
     -I$(top_srcdir)/include \
     -I$(top_builddir)/include \
@@ -53,6 +57,7 @@ COMMON_TESTING_FLAGS_ONLY_WARNINGS = $(COMMON_CXXFLAGS_ONLY_WARNINGS) \
                                      -DTEST_BUILD_DIR=\"$(builddir)\"
 
 COMMON_TESTING_FLAGS = $(COMMON_TESTING_FLAGS_ONLY_WARNINGS)
+COMMON_TESTING_PROTOBUF_FLAGS = $(COMMON_TESTING_FLAGS)
 
 # The genererated protobuf files don't compile with -Werror on win32 so we
 # disable fatal warnings on WIN32.
@@ -61,6 +66,7 @@ if FATAL_WARNINGS
   COMMON_CXXFLAGS += -Werror
   COMMON_PROTOBUF_CXXFLAGS += -Werror -Wno-error=unused-parameter
   COMMON_TESTING_FLAGS += -Werror
+  COMMON_TESTING_PROTOBUF_FLAGS += -Werror -Wno-error=unused-parameter
 endif
 endif
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -64,9 +64,11 @@ COMMON_TESTING_PROTOBUF_FLAGS = $(COMMON_TESTING_FLAGS)
 if ! USING_WIN32
 if FATAL_WARNINGS
   COMMON_CXXFLAGS += -Werror
-  COMMON_PROTOBUF_CXXFLAGS += -Werror -Wno-error=unused-parameter
+  COMMON_PROTOBUF_CXXFLAGS += -Werror -Wno-error=unused-parameter \
+                              -Wno-error=deprecated-declarations
   COMMON_TESTING_FLAGS += -Werror
-  COMMON_TESTING_PROTOBUF_FLAGS += -Werror -Wno-error=unused-parameter
+  COMMON_TESTING_PROTOBUF_FLAGS += -Werror -Wno-error=unused-parameter \
+                                   -Wno-error=deprecated-declarations
 endif
 endif
 
--- a/common/rdm/Makefile.mk
+++ b/common/rdm/Makefile.mk
@@ -95,7 +95,8 @@ common_rdm_DiscoveryAgentTester_LDADD = $(COMMON_TESTING_LIBS)
 common_rdm_PidStoreTester_SOURCES = \
     common/rdm/DescriptorConsistencyCheckerTest.cpp \
     common/rdm/PidStoreTest.cpp
-common_rdm_PidStoreTester_CXXFLAGS = $(COMMON_TESTING_FLAGS)
+common_rdm_PidStoreTester_CXXFLAGS = $(COMMON_TESTING_FLAGS) \
+                                     $(COMMON_PROTOBUF_CXXFLAGS)
 common_rdm_PidStoreTester_LDADD = $(COMMON_TESTING_LIBS)
 
 common_rdm_RDMHelperTester_SOURCES = common/rdm/RDMHelperTest.cpp
@@ -109,7 +110,8 @@ common_rdm_RDMMessageTester_SOURCES = \
     common/rdm/RDMMessageInterationTest.cpp \
     common/rdm/StringMessageBuilderTest.cpp \
     common/rdm/VariableFieldSizeCalculatorTest.cpp
-common_rdm_RDMMessageTester_CXXFLAGS = $(COMMON_TESTING_FLAGS)
+common_rdm_RDMMessageTester_CXXFLAGS = $(COMMON_TESTING_FLAGS)\
+                                       $(COMMON_PROTOBUF_CXXFLAGS)
 common_rdm_RDMMessageTester_LDADD = $(COMMON_TESTING_LIBS)
 
 common_rdm_RDMAPITester_SOURCES = \
--- a/common/rdm/Makefile.mk
+++ b/common/rdm/Makefile.mk
@@ -95,8 +95,7 @@ common_rdm_DiscoveryAgentTester_LDADD = $(COMMON_TESTING_LIBS)
 common_rdm_PidStoreTester_SOURCES = \
     common/rdm/DescriptorConsistencyCheckerTest.cpp \
     common/rdm/PidStoreTest.cpp
-common_rdm_PidStoreTester_CXXFLAGS = $(COMMON_TESTING_FLAGS) \
-                                     $(COMMON_PROTOBUF_CXXFLAGS)
+common_rdm_PidStoreTester_CXXFLAGS = $(COMMON_TESTING_PROTOBUF_FLAGS)
 common_rdm_PidStoreTester_LDADD = $(COMMON_TESTING_LIBS)
 
 common_rdm_RDMHelperTester_SOURCES = common/rdm/RDMHelperTest.cpp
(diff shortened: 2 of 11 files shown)


joerg-krause:
Why building with `-Werror` at all? Warnings come and go with various compiler versions, so using `-Werror` is prone to cause build failures with various compiler versions, especially newer versions that introduce new warnings.

joerg-krause:
Like building ola 0.10.2 with protobuf 2.6.1 additionally throws an error because of `-Werror=deprecated-declarations`.

peternewman:
The warnings are generally there for a reason though, I don't believe the compiler writers add them in for fun. It's also one of a number of techniques we use to ensure and improve our code quality, alongside coverage, linting and static analysis tools.

The warnings also give us an early warning of features that may be being deprecated, so we can act on it before end users get hit by finding it's missing from the language.

I'm surprised to hear you had issues with Protobuf 2.6, our Mac builds are currently on 3.0 I believe and compiling with no issues.

There's a --disable-fatal-warnings configure option, to bypass the -Werror bit for the time before we can get a fix out.

joerg-krause:
I added a commit that passes `-Wno-error=unused-parameter` to the protobuf build.

Makefile.am
```diff
@@ -60,6 +60,7 @@ if ! USING_WIN32
 if FATAL_WARNINGS
   COMMON_CXXFLAGS += -Werror
   COMMON_PROTOBUF_CXXFLAGS += -Werror -Wno-error=unused-parameter
+  libprotobuf_CFLAGS += -Wno-error=unused-parameter
```
peternewman: Shouldn't this be CXXFLAGS given we're compiling C++?

peternewman:
While this does seem to work @joerg-krause it seems to be rather scattergun, we've gone from 332 instances of -Wno-error=unused-parameter to 1928.

In comparison my attempt here (which isn't complete) has only taken it up to 346 instances https://github.com/OpenLightingProject/ola/compare/0.10...peternewman:0.10-fix-travis . Which should mean mroe code is getting more in depth -Werror coverage.

Makefile.am
joerg-krause: No, pkg-config sets <VARIABLE-PREFIX>_CFLAGS no matter whether C or C++.

peternewman: Ah okay.

peternewman:
Can you change to my proposed fix style instead ( https://github.com/OpenLightingProject/ola/compare/0.10...peternewman:0.10-fix-travis ) please @joerg-krause , otherwise the -Werror overriding of that test is so scattergun it's barely going to hit any code.

joerg-krause:
@peternewman Sorry for late reply. Due to illness I am not able to do this right now, but properly next week.

peternewman:
No worries @joerg-krause thanks for letting us know and I hope you're feeling better soon.

joerg-krause:
@peternewman Sorry for the delay, some more important stuff had to be done first :smile:

I used your patches, I hope everything is already now!

peternewman:
Sorry @joerg-krause I'd not finished the job. You'll need to fix the Travis clang and gcc failures here with more of the same as what I'd done (is it currently building cleanly for you?):
https://travis-ci.org/OpenLightingProject/ola/jobs/175842798

Hope your now well again and thanks for coming back.

joerg-krause:
Yes, I'm well again! Thanks!

It builds fine for me on a Linux host. The travis failures are on an Apple machine, right? I'll check the error messages and update the commits accordingly.

peternewman:
Yes it's just the Mac builds you need to fix.

The Flake8 tests are failing as they've improved their testing. I've got a fix for that here: #1152 so you're blocked behind that going in.

peternewman:
The Flake8 test issue is now merged in @joerg-krause so it's just the Mac builds that still failing, which really are down to your changes.

joerg-krause:
@peternewman Finally, all checks have passed :smile:

common/rdm/Makefile.mk
```diff
@@ -109,7 +110,8 @@ common_rdm_RDMMessageTester_SOURCES = \
     common/rdm/RDMMessageInterationTest.cpp \
     common/rdm/StringMessageBuilderTest.cpp \
     common/rdm/VariableFieldSizeCalculatorTest.cpp
-common_rdm_RDMMessageTester_CXXFLAGS = $(COMMON_TESTING_FLAGS)
+common_rdm_RDMMessageTester_CXXFLAGS = $(COMMON_TESTING_FLAGS)\
```
peternewman: Can we have a space before the slash please.

olad/Makefile.mk
```diff
@@ -49,6 +49,7 @@ olad_libolaserver_la_SOURCES = $(ola_server_sources) \
                                olad/OlaServer.cpp \
                                olad/OlaDaemon.cpp
 olad_libolaserver_la_CXXFLAGS = $(COMMON_CXXFLAGS) \
+                                $(COMMON_PROTOBUF_CXXFLAGS) \
```
peternewman: You can just swap to COMMON_PROTOBUF_CXXFLAGS, as COMMON_CXXFLAGS is just a subset of the flags, with less skipping of errors.

olad/plugin_api/Makefile.mk
```diff
@@ -28,7 +28,8 @@ olad_plugin_api_libolaserverplugininterface_la_SOURCES = \
     olad/plugin_api/Universe.cpp \
     olad/plugin_api/UniverseStore.cpp \
     olad/plugin_api/UniverseStore.h
-olad_plugin_api_libolaserverplugininterface_la_CXXFLAGS = $(COMMON_CXXFLAGS)
+olad_plugin_api_libolaserverplugininterface_la_CXXFLAGS = $(COMMON_CXXFLAGS) \
```
peternewman: Again

joerg-krause:
> Can you add a COMMON_TESTING_PROTOBUF_FLAGS rather than using multiple sets of flags please, as I think that will more closely match what we've got already. The others are defined in Makefile.am

Sorry, I am not sure what you mean.

peternewman:
In https://github.com/OpenLightingProject/ola/blob/0.10/Makefile.am#L20 we define a group of different interlinked flag variables. Rather than setting $(COMMON_TESTING_CXXFLAGS) $(COMMON_PROTOBUF_CXXFLAGS) in the Makefile.mks as you've done. If you instead create a new flag COMMON_TESTING_PROTOBUF_FLAGS in that Makefile.am, which just needs to set  -Wno-error=unused-parameter when -Werror is set (the same as COMMON_PROTOBUF_CXXFLAGS compared to COMMON_CXXFLAGS. Then you can just use the new flag, rather than a pair, in the places where its required.

joerg-krause:
@peternewman 

> All checks have passed

Makefile.am
```diff
@@ -61,6 +65,8 @@ if FATAL_WARNINGS
   COMMON_CXXFLAGS += -Werror
   COMMON_PROTOBUF_CXXFLAGS += -Werror -Wno-error=unused-parameter
   COMMON_TESTING_FLAGS += -Werror
+  COMMON_TESTING_PROTOBUF_FLAGS = $(COMMON_TESTING_FLAGS) \
```
peternewman: Please can you move the "COMMON_TESTING_PROTOBUF_FLAGS = $(COMMON_TESTING_FLAGS)" bit outside the if, like COMMON_TESTING_FLAGS is done. Then just add the -Werror -Wno-error=unused-parameter inside the if; just like COMMON_PROTOBUF_CXXFLAGS. I think you'll find this currently fails to compile if you do --disable-fatal-warnings.

joerg-krause:
@peternewman The macOS build fails now because of some deprecation warnings.

Makefile.am
```diff
@@ -61,6 +66,7 @@ if FATAL_WARNINGS
   COMMON_CXXFLAGS += -Werror
   COMMON_PROTOBUF_CXXFLAGS += -Werror -Wno-error=unused-parameter
```
peternewman: I think you need to add -Wno-error=deprecated-declarations here to fix the latest issue.

Makefile.am
peternewman: Done

peternewman:
Other than that, LGTM.

joerg-krause:
@peternewman I guess some of the test scripts are failing now. Can you have a look, please?

peternewman:
See initial high level review from @nomis52 here https://github.com/OpenLightingProject/ola/pull/1126#issuecomment-249462380 .

joerg-krause:
Thanks for merging!

PR #1136, opened 2016-10-14
Added DMXCreator 512 Basic USB interface

FloEdelmann:
see http://www.dmx512.ch/512.html

I reverse-engineered the USB protocol using WireShark and USBPcap. This is my first try, so there may be some functions that I didn't implement or that could be better, please just let me know.

DMXCreator sends two URB packets for each change:
1. A constant byte string to endpoint 1 that indicates new data.
2. The actual DMX data to endpoint 2.

~~It currently only supports 256 channels, I may work on that later.~~ (Fixed in df44d5a)
Also, there is a bug that lets it consume way too much memory in synchronous libusb mode until it eventually gets killed.


FloEdelmann:
There is one more bug: It doesn't handle unplugging correctly so the next time it gets plugged in, it refuses to connect because of the missing serial number (there can always only be one DMXCreator device at a time).

debian/ola.udev
```diff
@@ -13,6 +13,9 @@ SUBSYSTEM=="usb|usb_device", ACTION=="add", ATTRS{idVendor}=="10cf", ATTRS{idPro
 # udev rules for the DMXControl Projects e.V. Nodle U1
 SUBSYSTEM=="usb|usb_device", ACTION=="add", ATTRS{idVendor}=="16d0", ATTRS{idProduct}=="0830", GROUP="plugdev"
 
+# udev rules for the DMXCreator 512 Basic device
```
peternewman: Can you also open a PR here to patch the pi file: https://github.com/OpenLightingProject/raspberrypi/blob/master/etc/udev/rules.d/10-local.rules

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -274,6 +284,7 @@ void AsyncPluginImpl::SetupUSBDevice(libusb_device *usb_device) {
             << strings::ToHex(descriptor.idProduct);
 
   WidgetFactories::iterator factory_iter = m_widget_factories.begin();
+
```
peternewman: Can you remove this extra space you've added please.

plugins/usbdmx/DMXCreatorFactory.cpp
```diff
@@ -0,0 +1,89 @@
...
+    WidgetObserver *observer,
+    libusb_device *usb_device,
+    const struct libusb_device_descriptor &descriptor) {
+  if (descriptor.idVendor != VENDOR_ID || descriptor.idProduct != PRODUCT_ID) {
+    return false;
+  }
+
+  OLA_INFO << "Found a new DMXCreator device";
+  LibUsbAdaptor::DeviceInformation info;
+  /*if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
```
peternewman: Can you put these checks back in please.

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -56,7 +56,7 @@ bool UsbDmxPlugin::StartHook() {
   }
 
   unsigned int debug_level;
-  if (!StringToInt(m_preferences->GetValue(LIBUSB_DEBUG_LEVEL_KEY) ,
+  if (!StringToInt(m_preferences->GetValue(LIBUSB_DEBUG_LEVEL_KEY),
```
peternewman: Good spot!

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -92,8 +92,8 @@ string UsbDmxPlugin::Description() const {
 "----------------------------\n"
 "\n"
 "This plugin supports various USB DMX devices including the \n"
-"Anyma uDMX, DMXControl Projects e.V. Nodle U1, Eurolite, Fadecandy, "
-"Sunlite USBDMX2 and Velleman K8062.\n"
+"Anyma uDMX, DMXControl Projects e.V. Nodle U1, DMXCreator 512 Basic,\n"
```
peternewman: Again good spot with the \n.

peternewman:
I think the unplugging bug will be because m_missing_serial_number never gets reset when the serial number less widget gets unplugged. I suspect this bug affects all the widgets with the same code.

peternewman:
I'd imagine you send a different constant to endpoint 1 to send the other half of the universe, or perhaps just a second packet to endpoint 2 before you send the endpoint 1 data (i.e. sending to endpoint 1 may reset the slot counter back to 0).

peternewman:
Hopefully the synchronous issue can be fixed, if not, it probably needs to be disabled for the moment.

peternewman:
The other PR is in, so you can resync this.

plugins/usbdmx/DMXCreatorFactory.cpp
FloEdelmann: This is what WireShark captures when the device descriptor is received:

```
Frame 2: 82 bytes on wire (656 bits), 82 bytes captured (656 bits) on interface 0
USB URB
DEVICE DESCRIPTOR
    bLength: 18
    bDescriptorType: 0x01 (DEVICE)
    bcdUSB: 0x0100
    bDeviceClass: Vendor Specific (0xff)
    bDeviceSubClass: 0
    bDeviceProtocol: 0
    bMaxPacketSize0: 16
    idVendor: Unknown (0x0a30)
    idProduct: Unknown (0x0002)
    bcdDevice: 0x0001
    iManufacturer: 0
    iProduct: 0
    iSerialNumber: 0
    bNumConfigurations: 25
```

FloEdelmann:
It was exactly as you said: Just send two packets after another to endpoint 2. But I also had to change the constant that is sent to endpoint 1. It took me quite a while to figure that out :see_no_evil: 

The unplugging bug and the synchronous mode issue unfortunately do both still exist. I hope to figure at least the latter out soon.

peternewman:
Please can you add the device to the list in plugins/usbdmx/UsbDmxPlugin.h too.

plugins/usbdmx/AsyncUsbSender.cpp
```diff
@@ -64,7 +64,8 @@ void AsyncUsbSender::TransferComplete(struct libusb_transfer *transfer) {
   }
 
   if (transfer->status != LIBUSB_TRANSFER_COMPLETED) {
-    OLA_WARN << "Transfer returned " << transfer->status;
+    OLA_WARN << "Transfer returned "
+      << m_adaptor->ErrorCodeToString(transfer->status);
```
peternewman: Nice spot. Can you align the << with the row above please.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -48,10 +49,14 @@ namespace {
 static const unsigned char ENDPOINT_1 = 0x01;
 static const unsigned char ENDPOINT_2 = 0x02;
 static const unsigned int URB_TIMEOUT_MS_SYNC = 1000;
-static const unsigned int URB_TIMEOUT_MS_ASYNC = 500;
+static const unsigned int URB_TIMEOUT_MS_ASYNC = 50;
 static const unsigned int CHANNELS_PER_PACKET = 256;
 
-void AsyncCallback(struct libusb_transfer *transfer);
+enum {
```
peternewman: Can you typedef the enum to give it a name please.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -178,31 +185,49 @@ class DMXCreatorAsyncUsbSender : public AsyncUsbSender {
   }
 
   bool PerformTransfer(const DmxBuffer &buffer) {
+    // if we only wanted to send the first half, the last byte would be 0x01
     unsigned char status_buffer[6] = {
-      0x80, 0x01, 0x00, 0x00, 0x00, 0x01
+      0x80, 0x01, 0x00, 0x00, 0x00, 0x02
```
peternewman: I take it you've not worked out what the other bytes in the packet do?

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -178,31 +185,49 @@ class DMXCreatorAsyncUsbSender : public AsyncUsbSender {
...
 
-    m_adaptor->FillBulkTransfer(m_transfer, m_usb_handle, ENDPOINT_1,
-      status_buffer, sizeof(status_buffer), &AsyncCallback, this,
-      URB_TIMEOUT_MS_ASYNC);
+    length = CHANNELS_PER_PACKET;
+    memset(m_dmx_buffer_2, 0, length);
+    buffer.GetRange(CHANNELS_PER_PACKET, m_dmx_buffer_2, &length);
 
+    m_state = STATE_SEND_FIRST_HALF;
+    FillBulkTransfer(ENDPOINT_1, status_buffer, sizeof(status_buffer),
```
peternewman: Any reason you've switched away from m_adaptor->?

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -178,31 +185,49 @@ class DMXCreatorAsyncUsbSender : public AsyncUsbSender {
   }
 
   bool PerformTransfer(const DmxBuffer &buffer) {
+    // if we only wanted to send the first half, the last byte would be 0x01
     unsigned char status_buffer[6] = {
```
peternewman: Can you not initialise the status_buffer once given it never changes?

debian/ola.udev
peternewman: Thanks, merged.

plugins/usbdmx/DMXCreator.h
```diff
@@ -0,0 +1,113 @@
...
+      : SimpleWidget(adaptor, usb_device) {}
+
+  virtual ~DMXCreator() {}
+
+  /**
+   * @brief Get a (fake) serial number of this widget.
+   * @returns The serial number of the widget.
+   */
+  std::string SerialNumber() const {
+    return "?";
```
peternewman: You can just return "", that's what other devices do, but more on that later.

plugins/usbdmx/DMXCreatorFactory.cpp
```diff
@@ -0,0 +1,73 @@
...
+    libusb_device *usb_device,
+    const struct libusb_device_descriptor &descriptor) {
+  if (descriptor.idVendor != VENDOR_ID || descriptor.idProduct != PRODUCT_ID) {
+    return false;
+  }
+
+  OLA_INFO << "Found a new DMXCreator device";
+  // Unfortunately, DMXCreator devices don't provide any additional information
+  // that identify them. So we have to stick with just testing vendor and
+  // product ids. Also, since DMXCreator devices don't have serial numbers and
```
peternewman: Do you know for certain that none have serial numbers?

The m_adaptor->GetDeviceInfo(usb_device, descriptor, &info) call ought to succeed if the device is present, even if it returns no other info, so you could do the following still, and just remove the vendor/device name checks:

```
  LibUsbAdaptor::DeviceInformation info;
  if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
    return false;
  }
```

plugins/usbdmx/DMXCreatorFactory.cpp
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  DMXCreator *widget = NULL;
+  if (FLAGS_use_async_libusb) {
+    widget = new AsynchronousDMXCreator(m_adaptor, usb_device);
+  } else {
+    // Synchronous mode consumes way too much memory and eventually gets killed
+    // Until this is fixed, disable synchronous mode.
+    // widget = new SynchronousDMXCreator(m_adaptor, usb_device);
+    return false;
```
peternewman: Please log an OLA_WARN while this is disabled.

plugins/usbdmx/DMXCreatorFactory.h
```diff
@@ -0,0 +1,62 @@
...
+  bool DeviceAdded(
+      WidgetObserver *observer,
+      libusb_device *usb_device,
+      const struct libusb_device_descriptor &descriptor);
+
+ private:
+  bool m_missing_serial_number;
+  ola::usb::LibUsbAdaptor *m_adaptor;
+
+  static const char EXPECTED_MANUFACTURER[];
```
peternewman: This and the const below can be removed if it doesn't return names to check.

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -92,8 +92,9 @@ string UsbDmxPlugin::Description() const {
 "----------------------------\n"
 "\n"
 "This plugin supports various USB DMX devices including the \n"
-"Anyma uDMX, AVLdiy D512, DMXControl Projects e.V. Nodle U1, Eurolite, "
-"Fadecandy, Sunlite USBDMX2 and Velleman K8062.\n"
+"Anyma uDMXX, AVLdiy D512, DMXControl Projects e.V. Nodle U1,\n"
+"DMXCreator 512 Basic, Eurolite, Fadecandy, Sunlite USBDMX2 and \n"
```
peternewman: Would you mind removing the space before the \n please.

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -92,8 +92,9 @@ string UsbDmxPlugin::Description() const {
 "----------------------------\n"
 "\n"
 "This plugin supports various USB DMX devices including the \n"
```
peternewman: Would you mind removing the space before the \n please.

peternewman:
Can you also add the 512 Basic bit to the filenames too please, given they already have other interfaces we might want to support in future.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+  libusb_device_handle* SetupHandle() {
+    libusb_device_handle *usb_handle;
+    bool ok = m_adaptor->OpenDeviceAndClaimInterface(
+        m_usb_device, 0, &usb_handle);
+    return ok ? usb_handle : NULL;
+  }
+
+  bool PerformTransfer(const DmxBuffer &buffer) {
+    // if we only wanted to send the first half, the last byte would be 0x01
```
FloEdelmann: Unfortunately not. I just have this one device for testing, I guess the other DMXCreator dongles use other bytes.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -0,0 +1,257 @@
...
+    unsigned char status_buffer[6] = {
```
FloEdelmann: Yeah right, that would make more sense.

plugins/usbdmx/DMXCreatorFactory.cpp
FloEdelmann: Ok, I'll do that. I'm not completely sure that none of those devices have serial numbers but I think they don't since they not even return valid manufaturer / product names.

plugins/usbdmx/UsbDmxPlugin.cpp
peternewman: I've done this (partly to try out the new collaborative editing options).

plugins/usbdmx/UsbDmxPlugin.cpp
peternewman: I've done this too.

plugins/usbdmx/DMXCreatorFactory.cpp
peternewman: Thanks. As you say, I suspect they're unlikely to have serial numbers, but it would be a shame to be missing the functionality if they did (given it's already been written).

AUTHORS
```diff
@@ -8,6 +8,7 @@ Contributors:
   Bastien Andrès, contributed to the dmx4linux plugin
   Carsten Presser, Karate USB plugin
   Dave Olsthoorn, working on a new web-ui
+  Florian Edelmann, added DMXCreator512Basic USB support
```
peternewman: Sorry, I probably didn't explain very well, feel free to stick to normal English when it's not a class name or whatever, so you can add some spaces back in here. It was more about being specific which model from DMX Creator it was, if someone wants to add the 1024 version.

debian/ola.udev
```diff
@@ -16,6 +16,9 @@ SUBSYSTEM=="usb|usb_device", ACTION=="add", ATTRS{idVendor}=="10cf", ATTRS{idPro
 # udev rules for the DMXControl Projects e.V. Nodle U1
 SUBSYSTEM=="usb|usb_device", ACTION=="add", ATTRS{idVendor}=="16d0", ATTRS{idProduct}=="0830", GROUP="plugdev"
 
+# udev rules for the DMXCreator512Basic device
```
peternewman: Same again.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -189,6 +192,13 @@ bool AsyncPluginImpl::NewWidget(DMXCProjectsNodleU1 *widget) {
           m_plugin_adaptor));
 }
 
+bool AsyncPluginImpl::NewWidget(DMXCreator512Basic *widget) {
+  return StartAndRegisterDevice(
+      widget,
+      new GenericDevice(m_plugin, widget, "DMXCreator512Basic USB Device",
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,270 @@
...
+static const unsigned int URB_TIMEOUT_MS_ASYNC = 50;
+static const unsigned int CHANNELS_PER_PACKET = 256;
+
+// if we only wanted to send the first half of the universe, the last byte would
+// be 0x01
+static const unsigned char status_buffer[6] = {
+  0x80, 0x01, 0x00, 0x00, 0x00, 0x02
+};
+
+typedef enum SendState {
```
peternewman: Can you remove SendState from this line, given you've got it at the end.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,270 @@
...
+  STATE_SEND_STATUS,
+  STATE_SEND_FIRST_HALF,
+  STATE_SEND_SECOND_HALF
+} SendState;
+
+}  // namespace
+
+// DMXCreatorThreadedSender
```
peternewman: Change this to the full Basic name.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,270 @@
...
+// -----------------------------------------------------------------------------
+
+/**
+ * Sends messages to a DMXCreator512Basic device in a separate thread.
+ */
+class DMXCreatorThreadedSender: public ThreadedUsbSender {
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,270 @@
...
+  }
+  m_sender.reset(sender.release());
+  return true;
+}
+
+bool SynchronousDMXCreator512Basic::SendDMX(const DmxBuffer &buffer) {
+  return m_sender.get() ? m_sender->SendDMX(buffer) : false;
+}
+
+// DMXCreatorAsyncUsbSender
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,270 @@
...
+// -----------------------------------------------------------------------------
+class DMXCreatorAsyncUsbSender : public AsyncUsbSender {
```
peternewman: Again

plugins/usbdmx/DMXCreator512BasicFactory.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (descriptor.idVendor != VENDOR_ID || descriptor.idProduct != PRODUCT_ID) {
+    return false;
+  }
+
+  LibUsbAdaptor::DeviceInformation info;
+  if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
+    return false;
+  }
+
+  OLA_INFO << "Found a new DMXCreator512Basic device";
```
peternewman: You can put spaces back here.

plugins/usbdmx/DMXCreator512BasicFactory.cpp
```diff
@@ -0,0 +1,84 @@
...
+  // Unfortunately, DMXCreator512Basic devices don't provide any additional
```
peternewman: And here.

plugins/usbdmx/DMXCreator512BasicFactory.cpp
```diff
@@ -0,0 +1,84 @@
...
+  // information that identify them. So we have to stick with just testing
+  // vendor and product ids. Also, since DMXCreator512Basic devices don't have
```
peternewman: Again

plugins/usbdmx/DMXCreator512BasicFactory.cpp
```diff
@@ -0,0 +1,84 @@
...
+  DMXCreator512Basic *widget = NULL;
+  if (FLAGS_use_async_libusb) {
+    widget = new AsynchronousDMXCreator512Basic(m_adaptor, usb_device,
+                                                info.serial);
+  } else {
+    // Synchronous mode consumes way too much memory and eventually gets killed
+    // Until this is fixed, disable synchronous mode.
+    // widget = new SynchronousDMXCreator512Basic(m_adaptor, usb_device,
+    //                                            info.serial);
+    OLA_WARN << "Synchronous mode is disabled for DMXCreator512Basic at the "
```
peternewman: Again

plugins/usbdmx/DMXCreator512BasicFactory.h
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * @brief Creates DMXCreator512Basic widgets.
+ */
+class DMXCreator512BasicFactory
+    : public BaseWidgetFactory<class DMXCreator512Basic> {
+ public:
+  explicit DMXCreator512BasicFactory(ola::usb::LibUsbAdaptor *adaptor)
+      : BaseWidgetFactory<class DMXCreator512Basic>
+            ("DMXCreator512BasicFactory"),
```
peternewman: Can you leave the ( on the previous line.

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -143,6 +146,13 @@ bool SyncPluginImpl::NewWidget(DMXCProjectsNodleU1 *widget) {
           m_plugin_adaptor));
 }
 
+bool SyncPluginImpl::NewWidget(DMXCreator512Basic *widget) {
+  return StartAndRegisterDevice(
+      widget,
+      new GenericDevice(m_plugin, widget, "DMXCreator512Basic USB Device",
```
peternewman: Again

plugins/usbdmx/UsbDmxPlugin.cpp
```diff
@@ -91,9 +91,10 @@ string UsbDmxPlugin::Description() const {
 "USB DMX Plugin\n"
 "----------------------------\n"
 "\n"
-"This plugin supports various USB DMX devices including the \n"
-"Anyma uDMX, AVLdiy D512, DMXControl Projects e.V. Nodle U1, Eurolite, "
-"Fadecandy, Sunlite USBDMX2 and Velleman K8062.\n"
+"This plugin supports various USB DMX devices including the\n"
+"Anyma uDMXX, AVLdiy D512, DMXControl Projects e.V. Nodle U1,\n"
+"DMXCreator512Basic, Eurolite, Fadecandy, Sunlite USBDMX2 and\n"
```
peternewman: Again

FloEdelmann:
Everything should be fine now :wink:

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -59,23 +59,23 @@ static const unsigned char status_buffer[6] = {
...
 /**
- * Sends messages to a DMXCreator512Basic device in a separate thread.
+ * Sends messages to a DMXCreator 512 Basic device in a separate thread.
  */
-class DMXCreatorThreadedSender: public ThreadedUsbSender {
+class DMXCreator512BasicThreadedSender: public ThreadedUsbSender {
  public:
-  DMXCreatorThreadedSender(LibUsbAdaptor *adaptor,
+  DMXCreator512BasicThreadedSender(LibUsbAdaptor *adaptor,
                            libusb_device *usb_device,
```
peternewman: Can you align these with the (

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -93,7 +93,7 @@ class DMXCreatorThreadedSender: public ThreadedUsbSender {
                       const DmxBuffer &buffer);
 };
 
-bool DMXCreatorThreadedSender::TransmitBuffer(libusb_device_handle *handle,
+bool DMXCreator512BasicThreadedSender::TransmitBuffer(libusb_device_handle *handle,
                                               const DmxBuffer &buffer) {
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,272 @@
...
+// DMXCreator512BasicThreadedSender
+// -----------------------------------------------------------------------------
+
+/**
+ * Sends messages to a DMXCreator 512 Basic device in a separate thread.
+ */
+class DMXCreator512BasicThreadedSender: public ThreadedUsbSender {
+ public:
+  DMXCreator512BasicThreadedSender(LibUsbAdaptor *adaptor,
+                           libusb_device *usb_device,
```
peternewman: Can you align these with the (

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,272 @@
...
+  uint8_t *m_dmx_buffer_1;
+  uint8_t *m_dmx_buffer_2;
+  LibUsbAdaptor* const m_adaptor;
+
+  bool TransmitBuffer(libusb_device_handle *handle,
+                      const DmxBuffer &buffer);
+};
+
+bool DMXCreator512BasicThreadedSender::TransmitBuffer(
+      libusb_device_handle *handle, const DmxBuffer &buffer) {
```
peternewman: Just indent by 4.

plugins/usbdmx/AsyncPluginImpl.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreatorFactory.cpp
peternewman: Sorted.

plugins/usbdmx/AsyncUsbSender.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+namespace {
+
+static const unsigned char ENDPOINT_1 = 0x01;
+static const unsigned char ENDPOINT_2 = 0x02;
+static const unsigned int URB_TIMEOUT_MS_SYNC = 1000;
+static const unsigned int URB_TIMEOUT_MS_ASYNC = 50;
+static const unsigned int CHANNELS_PER_PACKET = 256;
+
+enum {
```
peternewman: Sorted.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -0,0 +1,257 @@
...
+      : ThreadedUsbSender(usb_device, handle),
+        m_adaptor(adaptor) {
+    m_control_setup_buffer = new uint8_t[CHANNELS_PER_PACKET];
+  }
+
+ private:
+  uint8_t *m_control_setup_buffer;
+  LibUsbAdaptor* const m_adaptor;
+
+
```
peternewman: Sorted.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -0,0 +1,257 @@
...
+  bool TransmitBuffer(libusb_device_handle *handle,
+                      const DmxBuffer &buffer);
+};
+
+bool DMXCreatorThreadedSender::TransmitBuffer(libusb_device_handle *handle,
```
peternewman: Sorted.

plugins/usbdmx/DMXCreator.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator.cpp
```diff
@@ -0,0 +1,257 @@
...
+    memset(m_dmx_buffer_1, 0, length);
+    buffer.Get(m_dmx_buffer_1, &length);
+
+    length = CHANNELS_PER_PACKET;
+    memset(m_dmx_buffer_2, 0, length);
+    buffer.GetRange(CHANNELS_PER_PACKET, m_dmx_buffer_2, &length);
+
+    m_state = STATE_SEND_FIRST_HALF;
+    FillBulkTransfer(ENDPOINT_1, status_buffer, sizeof(status_buffer),
+      URB_TIMEOUT_MS_ASYNC);
```
peternewman: Sorted.

plugins/usbdmx/DMXCreator.h
peternewman: Resolved.

plugins/usbdmx/DMXCreatorFactory.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/AsyncPluginImpl.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512BasicFactory.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512BasicFactory.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512BasicFactory.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512BasicFactory.h
peternewman: Sorted.

plugins/usbdmx/SyncPluginImpl.cpp
peternewman: Sorted.

plugins/usbdmx/UsbDmxPlugin.cpp
peternewman: Sorted.

peternewman:
Any comments @nomis52 ?

What do you want to do about the issue with replugging with no serial number, shall we just raise an issue for now, as it probably affects all the existing devices too?

Any thoughts on what's causing the memory leak? @FloEdelmann can you just confirm the leak is still present with all the changes that have happened.

FloEdelmann:
As far as I can tell, the memory leak is not present anymore, but I can only transmit one time before the thread shuts down itself. I'll look into that later.

FloEdelmann:
Wohoo, got it working! :tada:
Now the only remaining issue affects all devices without a serial number. I think we should open another issue for that.

peternewman:
I'm inclined to agree w.r.t the replugging serial number issue and leaving it for future fixing. Do you want to open an issue (ideally with some log lines in -l 4, even if it's just from your plugin).

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+bool DMXCreator512BasicThreadedSender::TransmitBuffer(
+    libusb_device_handle *handle, const DmxBuffer &buffer) {
+
+  if (*m_dmx_buffer == buffer) {
+    // no need to update -> sleep 1ms to avoid timeout errors
+    usleep(1000);
+    return true;
+  }
+
+  m_dmx_buffer = new DmxBuffer(buffer); // force copy
```
peternewman: Add another space before the //.

FloEdelmann:
Just saw that there was a whitespace issue again. I fixed that.

I'll open a new issue :+1:

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+  bool TransmitBuffer(libusb_device_handle *handle,
+                      const DmxBuffer &buffer);
+};
+
+bool DMXCreator512BasicThreadedSender::TransmitBuffer(
+    libusb_device_handle *handle, const DmxBuffer &buffer) {
+
+  if (*m_dmx_buffer == buffer) {
+    // no need to update -> sleep 1ms to avoid timeout errors
```
peternewman: This is different to the behaviour on all other dongles. Are you saying if you send the same packet again it fails?

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+    usleep(1000);
+    return true;
+  }
+
+  m_dmx_buffer = new DmxBuffer(buffer);  // force copy
```
peternewman: You're copying the buffer just to see if it's changed?

FloEdelmann: Is there a better way?

peternewman: It was partly just a comment that you could use it for the buffer.Get/GetRange calls.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: If you only make one sendDmx call to OLA every second for example, does the device keep up a valid refresh rate?

Are you saying the dongle crashes/stops responding if you send more frequently? Is this just in sync mode?

peternewman: Hmm, I wonder if both sets of code need some rate limiting then. It sounds like it may fall over if someone sends data to OLA faster than it can cope with?

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Sorted.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Not that I can immediately think of.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: It certainly sounds like it's got a microprocessor inside it. I guess for the true test, send data every ten seconds or more say. In a gap, depower the LED PAR and plug it back into the mains again; if it comes back in the same state, there must be a microprocessor in the dongle.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Actually, there is an improvement, you can switch to m_dmx_buffer = buffer; Due to some magic we do in the background, it will copy the data if it needs to, but otherwise just leave it as a pointer.

plugins/usbdmx/DMXCreator512Basic.cpp
FloEdelmann: Confirmed. So it certainly does keep up a decent refresh rate itself.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+      : ThreadedUsbSender(usb_device, handle),
+        m_adaptor(adaptor) {
+    m_dmx_buffer = new DmxBuffer();
+    m_dmx_buffer_1 = new uint8_t[CHANNELS_PER_PACKET];
+    m_dmx_buffer_2 = new uint8_t[CHANNELS_PER_PACKET];
+  }
+
+ private:
+  DmxBuffer *m_dmx_buffer;
+  uint8_t *m_dmx_buffer_1;
```
nomis52: these should be uint8_t m_dmx_buffer[CHANNELS_PER_PACKET] to avoid having to manually manage memory - as it stands we're leaking the arrays & the dmx buffer.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+                                   libusb_device_handle *handle)
+      : ThreadedUsbSender(usb_device, handle),
+        m_adaptor(adaptor) {
+    m_dmx_buffer = new DmxBuffer();
+    m_dmx_buffer_1 = new uint8_t[CHANNELS_PER_PACKET];
+    m_dmx_buffer_2 = new uint8_t[CHANNELS_PER_PACKET];
+  }
+
+ private:
+  DmxBuffer *m_dmx_buffer;
```
nomis52: Don't use a pointer here - DMXBuffers use ref counting & copy-on-write underneath so you don't have to use pointers.

plugins/usbdmx/DMXCreator512Basic.cpp
nomis52: can you call this 

universe_data_lower & universe_data_upper please. That's makes it clear it's not a DMXBuffer.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+  if (*m_dmx_buffer == buffer) {
+    // no need to update -> sleep 1ms to avoid timeout errors
+    usleep(1000);
+    return true;
+  }
+
+  m_dmx_buffer = buffer;
+
+  unsigned int length = CHANNELS_PER_PACKET;
+  memset(m_dmx_buffer_1, 0, length);
```
nomis52: you can avoid clearing the bytes that we then set in the next line.

Do the get first, then memset(0) any remaining bytes in the buffer (CHANNELS_PER_PACKET - length)

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+                              URB_TIMEOUT_MS);
+
+  if (r < 0 && r != LIBUSB_ERROR_PIPE) {
+    OLA_WARN << "Sending data bytes (1) failed: "
+             << m_adaptor->ErrorCodeToString(r);
+    return false;
+  }
+
+  bytes_sent = 0;
+  r = m_adaptor->BulkTransfer(handle, ENDPOINT_2,
```
nomis52: I'd move this code into  a function since it's repeated 3 times.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+                         const_cast<unsigned char*>(m_dmx_buffer_2),
+                         CHANNELS_PER_PACKET, URB_TIMEOUT_MS);
+        SubmitTransfer();
+        break;
+    }
+  }
+
+
+ private:
+  uint8_t *m_dmx_buffer_1;
```
nomis52: same comment here, make these arrays instead of pointers - at least in this case we're not leaking them.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,287 @@
...
+  libusb_device_handle* SetupHandle() {
+    libusb_device_handle *usb_handle;
+    bool ok = m_adaptor->OpenDeviceAndClaimInterface(
+        m_usb_device, 0, &usb_handle);
+    return ok ? usb_handle : NULL;
+  }
+
+  bool PerformTransfer(const DmxBuffer &buffer) {
+    unsigned int length = CHANNELS_PER_PACKET;
+    memset(m_dmx_buffer_1, 0, length);
```
nomis52: same comment here about avoiding the double-set.

FloEdelmann:
Sorry, I pushed by accident. Just ignore this commit until I comment again.

FloEdelmann:
Now the BulkTransfer works as before. The size was wrong.
Also, I reduced the wait time to 50ns, that works well, too.

FloEdelmann:
Travis failed again due to network outage. Don't know if it was Sourceforge or Travis itself. At least it was not me ;)

FloEdelmann:
Any updates on this? I believe I have addressed all of your requested changes @peternewman @nomis52

peternewman:
Hi @FloEdelmann , sorry I've been rather busy, I'll try and look in the next day or two.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  r = BulkTransferPart(handle, ENDPOINT_2, m_universe_data_upper,
+                       CHANNELS_PER_PACKET, "upper data");
+  return r;
+}
+
+bool DMXCreator512BasicThreadedSender::BulkTransferPart(
+    libusb_device_handle *handle, unsigned char endpoint,
+    const uint8_t *buffer, size_t size, const char name[]) {
```
peternewman: Can you switch name to a std::string please.

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+  bool BulkTransferPart(libusb_device_handle *handle, unsigned char endpoint,
+                        const uint8_t *buffer, size_t size, const char name[]);
+};
+
+bool DMXCreator512BasicThreadedSender::TransmitBuffer(
+    libusb_device_handle *handle, const DmxBuffer &buffer) {
+
+  if (m_dmx_buffer == buffer) {
+    // no need to update -> sleep 100ns to avoid timeout errors
```
peternewman: Can you correct the comment please, or the usleep.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,275 @@
...
+    usleep(50);
+    return true;
+  }
+
+  m_dmx_buffer = buffer;
+
+  unsigned int length = CHANNELS_PER_PACKET;
+  m_dmx_buffer.Get(m_universe_data_lower, &length);
+  memset(m_universe_data_lower+length, 0, CHANNELS_PER_PACKET - length);
```
peternewman: Can we have spaces around the + please.

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+  length = CHANNELS_PER_PACKET;
+  m_dmx_buffer.GetRange(CHANNELS_PER_PACKET, m_universe_data_upper, &length);
+  memset(m_universe_data_upper+length, 0, CHANNELS_PER_PACKET - length);
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,275 @@
...
+    libusb_device_handle *usb_handle;
+    bool ok = m_adaptor->OpenDeviceAndClaimInterface(
+        m_usb_device, 0, &usb_handle);
+    return ok ? usb_handle : NULL;
+  }
+
+  bool PerformTransfer(const DmxBuffer &buffer) {
+    unsigned int length = CHANNELS_PER_PACKET;
+    buffer.Get(m_universe_data_lower, &length);
+    memset(m_universe_data_lower+length, 0, CHANNELS_PER_PACKET - length);
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    length = CHANNELS_PER_PACKET;
+    buffer.GetRange(CHANNELS_PER_PACKET, m_universe_data_upper, &length);
+    memset(m_universe_data_upper+length, 0, CHANNELS_PER_PACKET - length);
```
peternewman: Again

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

plugins/usbdmx/DMXCreator512Basic.cpp
peternewman: Done

peternewman:
@nomis52 any final comments from you?

FloEdelmann:
Yay, it's done! :tada:

peternewman:
Yep thanks for all your hard work @FloEdelmann , now onto the next one...

PR #1138, opened 2016-10-21
Allow keyboard input for keypad

FloEdelmann:
Until now, one had to click the buttons in the web interface to control the keypad. I've enabled keyboard input for this.


olad/www/new/js/app.js
```diff
@@ -310,9 +310,56 @@ ola.controller('keypadUniverseCtrl',
...
+          case '7':
+          case '8':
+          case '9':
+            $scope.input(key);
+            break;
+          case '@':
+          case 'a':
+            $scope.input(' @ ');
+            break;
+          case '-':
```
peternewman: I'd be inclined to leave out -, so we can do  1 THRU 10 - 5 + 15 in future, without confusing things.

peternewman:
Thanks for this!

olad/www/new/js/app.js
peternewman: It was popular on Strand, and probably other consoles. The above would select channels 1, 2, 3, 4, 6, 7, 8, 9, 10 and 15.

FloEdelmann:
Should the directive also go into the file you mentioned?

olad/www/new/js/app.js
peternewman: What else would you expect it to do (aside from maths)?

peternewman:
Possibly a question for @daveol but from my understanding of directives and given what he did with filters, I'd say make a new folder and put it in there, similar to this:
https://github.com/OpenLightingProject/ola/tree/master/javascript/new-src/src/filters

I'm assuming that directive can be used to grab focus anywhere in the UI we might need to?

peternewman:
Yeah * did all channels before (min channel to max channel).

peternewman:
Travis is now working again, there are a chunk of JS lint issues you'll need to fix please:
https://travis-ci.org/OpenLightingProject/ola/jobs/169594371#L417

FloEdelmann:
jshint and jscs are don't agree on the case indentation. When I comply with one, the other fails.

peternewman:
I'm inclined to say we should stick with jscs as you've currently done, as that's our own custom JS style, rather than a default one. So then it's just a case of stopping jshint complaining.

It looks like this got resolved in jshint here, which should be in jshint 2.2.0 upwards:
https://github.com/jshint/jshint/commit/c64fb5c680f67538b095e496a56f48e79ee670ec

Which we're already running, so needs a little more digging.

This is the first switch/case in our JS code, hence the issue.

peternewman:
Actually I think it may be the {} block within the case @FloEdelmann . Have you tried removing that?

FloEdelmann:
I actually never used `{}` blocks for case statements but added them after jscs complained.

peternewman:
I've taken the {} back out, as we need to fix our JCSC config instead.

peternewman:
Travis should be fixed by https://github.com/OpenLightingProject/javascript-style/pull/9 .

peternewman:
As promised, now all green.

javascript/new-src/src/directives/autofocus.js
```diff
@@ -0,0 +1,41 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Allows an element to obtail focus whenever a variable is set to true.
```
peternewman: Fixed.

javascript/new-src/src/controllers/keypad_universe.js
```diff
@@ -79,9 +79,61 @@ ola.controller('keypadUniverseCtrl',
...
           $scope.field = fields[0];
         }
+        $scope.focusInput = true;
+      };
+
+      $scope.keypress = function($event) {
+        var key = $event.key;
+
+        if ($event.altKey || $event.ctrlKey || $event.metaKey
+            || ($event.which == 0 && key != 'Enter' && key != 'Backspace')) {
```
peternewman: Sorted

javascript/new-src/src/directives/autofocus.js
peternewman: Fixed

FloEdelmann:
Nice, thank you :)

FloEdelmann:
Exactly. Ok I'll do this then.


Am 21.10.2016 um 18:36 schrieb Peter Newman:

peternewman:
@daveol if you want to retrospectively review this, let us know if there are any issues.

@FloEdelmann I realised I forgot to regenerate the JS after fixing lint issues and it looks like you did the same. Would you mind updating them and pushing a new PR please.

FloEdelmann:
Unfortunately, I won't find the time until maybe Monday. If it's still relevant then, just let me know.

PR #1142, opened 2016-10-23
Markdown for plugin descriptions

### LINKED ISSUE #678: Switch to markdown in all the Plugin info strings
Right now the plugin info text is free form.

If we switched to markdown, then we could use a Javascript markdown renderer to make the description in web UI look prettier.

Maybe we can use https://code.google.com/p/pagedown/ ?

FloEdelmann:
I started progress to close #678 and [project 1](https://github.com/OpenLightingProject/ola/projects/1).

The plugin descriptions are converted to Markdown (`README.md` in each directory). I've also written a shell script to convert those Markdown files into C++ header files which can be called automatically during build. I don't know how to integrate that into the automake toolchain, so I leave this open for you ;)

You can tell me how to start and I'll try to do it myself.


peternewman:
Here's an example of integrating a generated script into our Makefiles:
https://github.com/OpenLightingProject/ola/blob/master/include/ola/Makefile.mk#L46

peternewman:
Thanks, BTW, this is a pleasant surprise. :)

plugins/usbdmx/README.md
```diff
@@ -1,5 +1,37 @@
-USBDMX Plugin
```
peternewman: Can you split the dev info out into a readme developer file of some sort please, as we don't need it to appear in the UI for the users.

It probably makes sense to standardise plugins/osc/README in the same way at the same time.

plugins/convert_README_to_header.sh
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+plugin=`basename "$path"`
+
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/\\\\\\\\n"\n"/g' "$path/README.md"`;
+
+identifier="PLUGINS_${plugin}_DESCRIPTION_H_";
+identifier=`echo "$identifier" | tr '[:lower:]' '[:upper:]'`
+
+echo "#ifndef $identifier" > $outfile;
```
peternewman: You'll need to add an appropriate header with licenses too. See https://github.com/OpenLightingProject/ola/blob/master/include/ola/make_plugin_id.sh for an example.

plugins/convert_README_to_header.sh
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+if [ ! -e "$path/README.md" ]; then
+  echo "README.md file in '$path' does not exist";
+  exit;
+fi
+
+plugin=`basename "$path"`
+
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/\\\\\\\\n"\n"/g' "$path/README.md"`;
```
peternewman: That's an impressive number of backspace escapes!

FloEdelmann: Yeah, but it works :grinning:

peternewman:
Given there are a number of plugins here, it probably makes sense to go through the review cycle with just one initially, then when we're happy you can apply the reviewed change style to all the other plugins, rather than repeatedly changing everything each time until it's sorted.

FloEdelmann:
See Karate plugin for how I did it. If that suits your needs, it's easy to add it for the other plugins, too.

FloEdelmann:
Unfortunately, it doesn't work on Travis :(
I included the rule that matches all plugin description header files in `plugins/Makefile.mk` and that worked for me at home. Now Travis says it can't find a rule for target `include/ola/plugins/karate_plugin_desc.h`.

plugins/karate/Makefile.mk
```diff
@@ -17,4 +19,5 @@ plugins_karate_libolakarate_la_LIBADD = \
     olad/plugin_api/libolaserverplugininterface.la
 endif
 
-EXTRA_DIST += plugins/karate/README.protocol
+EXTRA_DIST += plugins/karate/README.md \
```
peternewman: We generally just go for += \ on the first line, then list everything indented beneath as you've done below.

plugins/karate/Makefile.mk
```diff
@@ -1,8 +1,10 @@
 # LIBRARIES
 ##################################################
 if USE_KARATE
+BUILT_SOURCES += include/ola/plugins/karate_plugin_desc.h
 lib_LTLIBRARIES += plugins/karate/libolakarate.la
 plugins_karate_libolakarate_la_SOURCES = \
+	include/ola/plugins/karate_plugin_desc.h \
```
peternewman: Can you fix the indenting please.

plugins/Makefile.mk
```diff
@@ -23,3 +23,7 @@ include plugins/dmx4linux/Makefile.mk
 include plugins/e131/Makefile.mk
 include plugins/uartdmx/Makefile.mk
 endif
+
+include/ola/plugins/%_plugin_desc.h: plugins/%/README.md plugins/convert_README_to_header.sh plugins/Makefile.mk
```
FloEdelmann: double `%`s don't work unfortunately. Should I rather go for `plugins/%/PluginDescription.h`?

FloEdelmann:
Travis is still failing, but only on Mac OS / XCode. The compiler complains ˋExpected ';' after top level declaratorˋ. It could be this issue: http://stackoverflow.com/questions/12160733/xcode-doesnt-recognize-the-word-class

So I'll try making it a ˋ.cppˋ file instead of a ˋ.hˋ header.

peternewman:
My suspicion is your sed command isn't working correctly on the Mac, see this line you've got '0.1\n"n"==='; where it should be multiple lines or something:
2190const char plugin_description[] = "KarateLight - Version 0.1\n"n"=========================\n"n"\n"n"The plugin creates devices with a single output port. Info on the\n"n"KarateLight Hardware can be found at http://karatelight.de\n"n"\n"n"Unfortunately the site is in German only, but the maintainer will respond to\n"n"emails in English.\n"n"\n"n"\n"n"## Config file: `ola-karate.conf`\n"n"\n"n"`device = /dev/kldmx0`  \n"n"The path to the KarateLight device. Multiple entries are supported.";

You could try changing travis-ci.sh to cat out the file you're having issues with in the after_failure section.

FloEdelmann:
I've accidentally included some other changes in commit 2a51b74 that shouldn't go there. In commit 775ba01, there is only the one change I actually wanted to make.

peternewman:
I see https://github.com/OpenLightingProject/ola/commit/2a51b743446f2209f0cff13675c48c3c92aad706 gives us a tantalising glimpse of things to come @FloEdelmann !

plugins/convert_README_to_header.sh
```diff
@@ -0,0 +1,72 @@
...
+
+if [ ! -e "$path/README.md" ]; then
+  echo "README.md file in '$path' does not exist";
+  exit 1;
+fi
+
+plugin=`basename "$path"`;
+outfilename=`basename $outfile`;
+
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/\\\\n"\\
```
peternewman: What a daft choice on the Mac front. Can we have a link to the Stackoverflow article in a comment here, to explain the craziness please.

peternewman: What a daft choice on the Mac front. Can we have a link to the Stackoverflow article in a comment here, to explain the craziness please.

FloEdelmann:
Actually, I've started on that one before all the others but it turned out to be the most difficult one 😅

I have problems detecting the Break signal to distinguish a new Universe from the one before. If I make some progress here, I'll let you know in another pull request 😉

plugins/convert_README_to_header.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/sh
+
+# A simple script to build a C++ header file containing the plugin description
+# from the plugin's README.md
+# The output file then contains one variable 'plugin_description'.
+
+if [ $# != 2 ]; then
+  echo "Usage: convert_README_to_header.sh <path> <dest file>";
+  echo "<path>: path to plugin dir, e.g. plugins/artnet";
+  echo "<dest file>: e.g. include/ola/plugins/artnet_description.h";
```
peternewman: Can you update the example dest file.

plugins/convert_README_to_header.sh
```diff
@@ -0,0 +1,75 @@
...
+plugin=`basename "$path"`;
+outfilename=`basename $outfile`;
+
+# See http://stackoverflow.com/a/16576291
+# On Mac OS's sed, \n is not recognized as a newline character, but
+# \[actual newline] works
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/\\\\n"\\
+"/g' "$path/README.md"`;
+
+identifier="PLUGINS_${plugin}_PLUGIN_DESCRIPTION_H_";
```
peternewman: Why don't you just run the echo, and tr inline, rather than processing the whole line afterwards?

plugins/karate/Makefile.mk
```diff
@@ -1,8 +1,10 @@
 # LIBRARIES
 ##################################################
 if USE_KARATE
+BUILT_SOURCES += plugins/karate/PluginDescription.h
```
peternewman: Could you not add them all the BUILT_SOURCES in one go with the % expansion?

plugins/artnet/Makefile.mk
```diff
@@ -41,3 +43,6 @@ plugins_artnet_ArtNetTester_CXXFLAGS = $(COMMON_TESTING_FLAGS)
 plugins_artnet_ArtNetTester_LDADD = $(COMMON_TESTING_LIBS) \
                                     plugins/artnet/libolaartnetnode.la
 endif
+
+EXTRA_DIST += \
```
peternewman: Sorry, I didn't explain very well, you can leave this on one line if it's just one file.

plugins/convert_README_to_header.sh
FloEdelmann: Actually, because I don't know if I should stick to the line length limit in shell scripts, too. So I just bypassed the issue ;)

plugins/karate/Makefile.mk
FloEdelmann: I think that this is not possible. Especially if we want to respect the `USE_KARATE` etc. variables.

FloEdelmann:
What shall we do about the licenses? Both JS libraries are MIT licensed (see [here](https://github.com/Hypercubed/angular-marked/blob/master/lib/angular-marked.js#L1) and [here](https://github.com/chjj/marked/blob/master/lib/marked.js#L1)) but don't use the regular comment.

peternewman:
You could try and submit PRs to add it back in at their end. Realistically though, just add them to the ignored file list, like the others already are:
https://github.com/OpenLightingProject/ola/blob/master/scripts/enforce_licence.py#L52

plugins/artnet/Makefile.mk
peternewman: I think only for the few that have a README and a dev README.

plugins/Makefile.mk
```diff
@@ -23,3 +23,8 @@ include plugins/dmx4linux/Makefile.mk
 include plugins/e131/Makefile.mk
 include plugins/uartdmx/Makefile.mk
 endif
+
+dist_noinst_SCRIPTS += plugins/convert_README_to_header.sh
+
+plugins/%/PluginDescription.h: plugins/%/README.md plugins/convert_README_to_header.sh plugins/Makefile.mk
```
peternewman: Travis is currently logging the following:
```
plugins/Makefile.mk:29: warning: '%'-style pattern rules are a GNU make extension
Makefile.am:155:   'plugins/Makefile.mk' included from here
plugins/Makefile.mk:30: warning: shell dirname $<: non-POSIX variable name
plugins/Makefile.mk:30: (probably a GNU make extension)
```

So I'd be tempted to switch to a more long-hand method and put the config in each plugin's Makefile.mk.

FloEdelmann: Yeah, I saw that too, but since I'm not very familiar with automake (just a bit with pure make) I couldn't find a good solution. I don't want to put these lines in every plugin because then e.g. a filename change of the shell script or dependency changes would require all the plugin Makefiles to be updated.

peternewman: I think that might be easier; it also stops it being built if you're not using that plugin.

Your first issue could be sorted via sed and prob the second issue too.

FloEdelmann: Currently, the `PluginDescription.h`s are also not built if the flag is false, since it only gets added to `BUILT_SOURCES` if the flag is set.

Your solution would definitely be easier to setup, I do just wonder if it's also easy to maintain.

peternewman: Ah I see.

I don't think it will have too much of a hit on maintenance. We've barely changed the other similar script based config in the Makefiles in the past. There's also the balance of it being obvious and in the expected place, rather than hidden in another file.

FloEdelmann: Ok, that's a good point I had not thought about yet. I guess we should then just go with the easiest way to prevent that error.

FloEdelmann:
That was the trivial part :wink:

plugins/karate/Makefile.mk
```diff
@@ -1,8 +1,14 @@
 # LIBRARIES
 ##################################################
 if USE_KARATE
+# Plugin description is generated from README.md
+BUILT_SOURCES += plugins/karate/PluginDescription.h
+plugins/karate/PluginDescription.h: plugins/karate/README.md plugins/karate/Makefile.mk plugins/convert_README_to_header.sh
+	sh $(top_srcdir)/plugins/convert_README_to_header.sh plugins/karate PluginDescription.h
```
peternewman: Can you change the paths to similar to this:
https://github.com/OpenLightingProject/ola/blob/master/include/ola/Makefile.mk#L48

So you'll need to revert part of your most recent change, to pass in the paths separately.

I'm rather surprised it's not breaking Travis TBH.

plugins/karate/Makefile.mk
```diff
@@ -1,8 +1,14 @@
 # LIBRARIES
 ##################################################
 if USE_KARATE
+# Plugin description is generated from README.md
+BUILT_SOURCES += plugins/karate/PluginDescription.h
```
peternewman: Can you switch the name to KaratePluginDescription please (which I can ask for now it's in each file)

peternewman:
Can you resync too please.

plugins/karate/Makefile.mk
peternewman: Ah it looks like it's currently being built so early it doesn't break the distcheck, but I'm not sure how. Anyway I think it needs changing please.

plugins/usbdmx/README.md
peternewman: Done

plugins/convert_README_to_header.sh
peternewman: Done

plugins/artnet/Makefile.mk
```diff
@@ -11,9 +11,15 @@ plugins_artnet_libolaartnetnode_la_SOURCES = \
     plugins/artnet/ArtNetNode.h
 plugins_artnet_libolaartnetnode_la_LIBADD = common/libolacommon.la
 
+# Plugin description is generated from README.md
+BUILT_SOURCES += plugins/artnet/ArtNetPluginDescription.h
```
peternewman: Can you switch to built_sources like we've used elsewhere.

plugins/Makefile.mk
peternewman: Sorted

plugins/karate/Makefile.mk
peternewman: Done

plugins/karate/Makefile.mk
peternewman: Done

plugins/convert_README_to_header.sh
peternewman: Done

peternewman: Done

plugins/convert_README_to_header.sh
peternewman: Done

plugins/karate/Makefile.mk
peternewman: Understood

plugins/artnet/Makefile.mk
peternewman: Sorted

plugins/Makefile.mk
peternewman: Done

plugins/karate/Makefile.mk
peternewman: Done

plugins/karate/Makefile.mk
peternewman: Done

peternewman:
@nomis52 can you review the changes so far, before @FloEdelmann replicates them to the other plugins. @daveol can you take a look at the JS changes please.

plugins/artnet/Makefile.mk
FloEdelmann: I tried that at the beginning, but it didn't work because it was being built too late, I guess. After switching to `BUILT_SOURCES` in bc8667a, that part worked at least.

FloEdelmann:
Okay, let's see. I currently can not make locally, as I'm on my tablet right now, so I'll have to rely on Travis.

peternewman:
The Flake8 tests are failing as they've improved their testing. I've got a fix for that here: #1152 so you're blocked behind that going in.

plugins/openpixelcontrol/Makefile.mk
```diff
@@ -48,3 +48,5 @@ plugins_openpixelcontrol_OPCServerTester_LDADD = \
     plugins/openpixelcontrol/libolaopc.la
 endif
 
+EXTRA_DIST += plugins/openpixelcontrol/README.md
+
```
peternewman: Can you remove the blank line please.

plugins/osc/README.developer
```diff
@@ -1,3 +1,6 @@
+OSC Plugin Developer Information
```
peternewman: Is the file Markdown, or just something that looks similar? If the former, can we give it a .md extension.

plugins/pathport/README.md
```diff
@@ -0,0 +1,24 @@
...
+universe 1 is xDMX channels 512 - 1023.
+
+
+## Config file: `ola-pathport.conf`
+
+`dscp = <int>`  
+Set the DSCP value for the packets. Range is 0-63.
+
+`ip = [a.b.c.d|<interface_name>]`  
+The ip address or interface name to bind to. If not specified it will use
```
peternewman: IP

plugins/sandnet/README.md
```diff
@@ -0,0 +1,17 @@
...
+This plugin creates a single device with 2 output and 8 input ports.
+
+The universe bindings are offset by one from those displayed in sandnet. For
+example, SandNet universe 1 is OLA universe 0.
+
+
+## Config file: `ola-sandnet.conf`
+
+`ip = [a.b.c.d|<interface_name>]`  
+The ip address or interface name to bind to. If not specified it will use
```
peternewman: IP

plugins/shownet/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The ports correspond to the DMX channels used in the shownet protocol. For
+example the first input and output port 0 is channels 1 - 512 and the second
+input and output ports are channels 513 - 1024.
+
+
+## Config file: `ola-shownet.conf`
+
+`ip = [a.b.c.d|<interface_name>]`  
+The ip address or interface name to bind to. If not specified it will use
```
peternewman: IP

plugins/stageprofi/Makefile.mk
```diff
@@ -18,3 +18,5 @@ plugins_stageprofi_libolastageprofi_la_LIBADD = \
     olad/plugin_api/libolaserverplugininterface.la
 endif
 
+EXTRA_DIST += plugins/stageprofi/README.md
+
```
peternewman: Can we remove the blank line please.

peternewman:
@nomis52 can you review the changes. @daveol can you take a look at the JS changes please.

I'm tempted to say we merge this as is, when the outstanding nits are fixed. Then get @FloEdelmann to do the other plugins promptly afterwards and merge them in a second PR to keep the diff sizes down. As long as he checks that no-one has updated plugin descriptions in the meantime (which seems unlikely aside from #1147 , then we should be fine).

plugins/osc/README.developer
FloEdelmann: It was just something similar but I transformed it.

plugins/osc/README.developer.md
```diff
@@ -0,0 +1,675 @@
...
+  script.
+* Make sure you can complete a full OLA build & test cycle (`make`,
+  `make check`).
+
+The initial commit of this plugin was Rev
+`b9d6a47e9a2586d51dac521d7532fb11d76e0ae5`. It's worth having that diff
+open as you step through the explanations.
+
+Let's get to it. If you make it all the way through this doc and I ever meet
+you in person I'll buy you a beer!
```
FloEdelmann: Ok I already made it through the doc :laughing:

peternewman: Heh, you need to talk to @nomis52 for that, and fly to the states! Or I'll buy you one if I see you in the UK.

plugins/stageprofi/Makefile.mk
peternewman: Done

plugins/osc/README.developer.md
FloEdelmann: Those are both a bit far away for me unfortunately...

FloEdelmann:
The output for the old web UI / `ola_plugin_info` also improved for my taste.

plugins/osc/README.developer.md
peternewman: Well if you ever end up over either way...

peternewman:
Excellent. As long as they look readable don't have formatting quirks and the line breaks etc still work, that's fine.

plugins/shownet/README.md
peternewman: Done

plugins/pathport/README.md
peternewman: Done

plugins/openpixelcontrol/Makefile.mk
peternewman: Done

plugins/karate/Makefile.mk
```diff
@@ -1,8 +1,14 @@
...
 ##################################################
 if USE_KARATE
+# Plugin description is generated from README.md
+built_sources += plugins/karate/KaratePluginDescription.h
+plugins/karate/KaratePluginDescription.h: plugins/karate/README.md plugins/karate/Makefile.mk plugins/convert_README_to_header.sh
+	sh $(top_srcdir)/plugins/convert_README_to_header.sh $(top_srcdir)/plugins/karate $(top_builddir)/plugins/karate/KaratePluginDescription.h
+
 lib_LTLIBRARIES += plugins/karate/libolakarate.la
 plugins_karate_libolakarate_la_SOURCES = \
+    plugins/karate/KaratePluginDescription.h \
```
peternewman: Done

plugins/artnet/README.md
```diff
@@ -0,0 +1,72 @@
...
+| 0          | 15            | 0            | 240                 |
+| 0          | 15            | 15           | 255                 |
+| 1          | 0             | 0            | 256                 |
+| 1          | 0             | 1            | 257                 |
+| 1          | 0             | 15           | 271                 |
+| 1          | 1             | 0            | 272                 |
+| 1          | 15            | 0            | 496                 |
+| 1          | 15            | 15           | 511                 |
+
+That is `Port Address = (Net << 8) + (Subnet << 4) + (Universe % 4)`
```
peternewman: Done

plugins/dummy/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+* Dummy Device (original)
+* Dimmer Rack, with a configurable number of sub-devices
+* Moving Light
+* Advanced Dimmer Rack, with E1.37-1 PIDs
+* A device that responds with ack timers
+* Sensor Device, with a number of sensors implemented
+* Network Device, with E1.37-2 PIDs
+
+The number of each device is configurable.
```
peternewman: Done

plugins/e131/README.md
```diff
@@ -0,0 +1,39 @@
...
+Enable the draft (2014) E1.31 discovery protocol.
+
+`ignore_preview = [true|false]`  
+Ignore preview data.
+
+`input_ports = [int]`  
+The number of input ports to create up to a max of 32.
+
+`ip = [a.b.c.d|<interface_name>]`  
+The ip address or interface name to bind to. If not specified it will use
```
peternewman: Done

plugins/espnet/README.md
```diff
@@ -0,0 +1,20 @@
...
+a fixed number of ports which can be patched to any universe. When sending
+data from a port, the data is addressed to the universe the port is patched
+to. For example if port 0 is patched to universe 10, the data will be sent
+to ESP universe 10.
+
+
+## Config file: `ola-espnet.conf`
+
+`ip = [a.b.c.d|<interface_name>]`  
+The ip address or interface name to bind to. If not specified it will use
```
peternewman: Done

plugins/artnet/Makefile.mk
```diff
@@ -11,9 +11,15 @@ plugins_artnet_libolaartnetnode_la_SOURCES = \
...
 
+# Plugin description is generated from README.md
+built_sources += plugins/artnet/ArtNetPluginDescription.h
+plugins/artnet/ArtNetPluginDescription.h: plugins/artnet/README.md plugins/artnet/Makefile.mk plugins/convert_README_to_header.sh
+	sh $(top_srcdir)/plugins/convert_README_to_header.sh $(top_srcdir)/plugins/artnet $(top_builddir)/plugins/artnet/ArtNetPluginDescription.h
+
 # The OLA artnet plugin
 lib_LTLIBRARIES += plugins/artnet/libolaartnet.la
 plugins_artnet_libolaartnet_la_SOURCES = \
+    plugins/artnet/ArtNetPluginDescription.h \
```
peternewman: Done

FloEdelmann:
Still no update?

As soon as this is merged, I'll update the other plugins including fixes for #1147 and fd8e5860454ef5714b997a9b71d79f3165461f99.

peternewman:
Thanks @nomis52 . I think the benefit of them being README.md is that it partially self-documents on GitHub, as you get the introduction to each plugin in it's plugin folder, like so:
https://github.com/FloEdelmann/ola/tree/82a753fec37362d61f6c340e87f67cdd1e248f9d/plugins/artnet

I'm going to force merge this, so you can get working on the other plugins @FloEdelmann , as the test failure is related to some Travis changes I made the other day rather than your code.

PR #1084, opened 2016-05-30
Try and fix some more pychecker issues

peternewman:

PR #1143, opened 2016-10-23
Pychecker fixes part 8

peternewman:
The next subset of #1084 .


peternewman:
The build should go green when this completes @nomis52 .

PR #1146, opened 2016-11-04
Several small Web UI improvements

FloEdelmann:
For the Universe's fader view:

* Page navigation wraps at 0 / max
* Fix page navigation glitch when resizing window
* Make channel values number inputs
* Make page navigation bar narrower on big screens


peternewman:
LGTM. @nomis52 and/or @daveol do you want to take a quick look too?

PR #1160, opened 2016-12-03
update ola_streaming_client help

FloEdelmann:
Is there a help text in the code, too?


peternewman:
Yes there is https://github.com/OpenLightingProject/ola/blob/0.10/examples/ola-streaming-client.cpp

peternewman:
Thanks

PR #1161, opened 2016-12-04
Cleaned up error handling of close system calls. They are now mostly …

rewolff:
…uniform.

The return value of "close" is documented to be zero on success. So if (close (..)) works. 
On error paths (except for ONE existing case) the close return value is not checked. 

I've been naughty. In this commit I've also downgraded the "Device /dev/ttyUSB0 doesn't exist, so there's no point trying to acquire a lock" message in Serial.CPP from INFO to DEBUG. It clutters the output when running with -l 3, and nothing is happening. 

What I would like to do too, as a cleanup is for where WIN32 is different create a 
CLOSEHandle(h) define that evaluates to CloseHandle(toHandle (h)) on WIN32 and to close(h) on others. This removes 5 lines-of-code at a time, all over that source file..... Is my suggestion for the name of the define sensible (acceptable with C++ programming practises?)

### DIFF
--- a/common/base/Init.cpp
+++ b/common/base/Init.cpp
@@ -372,7 +372,7 @@ void Daemonise() {
   }
 
   for (int fd = 0; fd < maxfd; fd++) {
-    close(fd);
+    close(fd); // ignore errors. Hope for the best. :-)
   }
 
   // send stdout, in and err to /dev/null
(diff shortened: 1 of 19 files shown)


peternewman:
Hi @rewolff thanks for this. Can you target our 0.10 branch, so we can get this into 0.10.3 potentially. Also you'll need to fix the lint issues, and the C++ build errors on Linux (probably due to not including errno.h from the looks of things).

rewolff:
Peter, I changed a lot of code, (albeit in a trivial way), that I dont' have a chance to test for myself. I'm not sure that pushing for an immediate release is a good idea. I am not familiar enough with git to know immediately what I need to do. So again this will have to wait for a "rainy sunday afternoon".

PR #1167, opened 2016-12-10
Disable warnings for readdir_r deprecation

daveol:
This disables deprecation warnings when building with newer glibc versions since readdir_r has been deprecated. This is a very temporary fix, a more structural one has to be made soon™


peternewman:
Thanks @daveol seems like a neat solution. Can you target this at 0.10 though please, as I think we ought to get it out for 0.10.4.

daveol:
that went wrong somehow... let me reopen this

PR #1168, opened 2016-12-11
Disable warnings for readdir_r deprecation

daveol:
This disables deprecation warnings when building with newer glibc
versions. This is a very temporary fix, a more structural one has to be
made soon™


peternewman:
This patch does work, as per the discussion here:
https://groups.google.com/forum/#!topic/open-lighting/GYt4Ne3qQzM

daveol:
I'll try and fix this today

daveol:
Travis mac builds are currently failing because of #1192

peternewman:
Fixed in #1194 @daveol just waiting for @nomis52 to have a chance to review it so we can merge it in.

configure.ac
```diff
@@ -168,6 +168,29 @@ AC_CHECK_FUNCS([bzero gettimeofday memmove memset mkdir strdup strrchr \
                 socket strerror getifaddrs getloadavg getpwnam_r getpwuid_r \
                 getgrnam_r getgrgid_r secure_getenv])
 
+AC_MSG_CHECKING(for readdir_r deprecation)
+old_cxxflags=$CXXFLAGS
+CXXFLAGS="-Werror=deprecated-declarations"
```
peternewman: I think you want to leave the ${CXXFLAGS} variable in the new CXXFLAGS, in case the user has put anything weird in there to make it work on their machine.

From looking at your changeset, it may just be that fixing arg would have solved the problem. Although I suspect this is perhaps one case where we don't necessarily want to fail if the test fails for any reason, hence just testing for deprecated-declarations causing an error.

peternewman:
The latest fix looks more promising at least, none of the Travis builds are detecting it as deprecated. I assume your machine is correctly showing it as deprecated though?

configure.ac
peternewman: @daveol can you fix this comment and retest, then I think we can merge it.

daveol:
I can verify that this test fails (as intended) on fedora

PR #1177, opened 2017-01-02
Switch to my version of Pychecker

peternewman:


peternewman:
I need to fix the latest error, or understand why it's failing.

PR #1181, opened 2017-01-02
Add some more RDM tests

peternewman:


peternewman:
This is currently failing on common/network/HealthCheckedConnectionTester which seems to be flaky since I upgraded the Xcode version on the Travis Macs. I think we'd previously seen failures elsewhere on buildbots.

PR #1184, opened 2017-01-06
Master resync

peternewman:


peternewman:
Forced merge as Protobuf fixes will come in when #1194 gets into Master.

Issue #1193, opened 2017-01-31
build failures with gcc7


yoe:
Hi,

I received Debian bug [853583](https://bugs.debian.org/853583) today, which claims that ola fails to build with GCC7.

It's not urgent (gcc7 won't be made the default until after the stretch release), but you might want to look into it. The bug report also contains instructions on how to install gcc7 from experimental on a Debian unstable system, so you can try building things.

peternewman:
Looks like the failure may actually be this:
```
common/utils/StringUtils.cpp: In function 'void ola::CapitalizeLabel(std::__cxx11::string*)':
common/utils/StringUtils.cpp:420:15: error: this statement may fall through [-Werror=implicit-fallthrough=]
         *iter = ' ';
         ~~~~~~^~~~~
common/utils/StringUtils.cpp:422:7: note: here
       case ' ':
       ^~~~
```

peternewman:
From https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

> Since there are occasions where a switch case fall through is desirable, GCC provides an attribute, __attribute__ ((fallthrough)), that is to be used along with a null statement to suppress this warning that would normally occur:
> 
>           switch (cond)
>             {
>             case 1:
>               bar (0);
>               __attribute__ ((fallthrough));
>             default:
>               ...
>             }
> C++17 provides a standard way to suppress the -Wimplicit-fallthrough warning using [[fallthrough]]; instead of the GNU attribute. In C++11 or C++14 users can use [[gnu::fallthrough]];, which is a GNU extension.

You can also do comments, but these aren't sufficient in the strictest checking.

peternewman:
I think we probably want to add something to include/ola/base/Macro.h, maybe call it OLA_FALLTHROUGH and we can add detection as necessary to populate the appropriate attribute or whatever depending on the detected compiler version.

peternewman:
See support from some other compilers here too:
https://groups.google.com/forum/#!topic/mozilla.dev.platform/YT7SXFhyr_I

As well as __builtin_fallthrough() in Clang.

yoe:
after

    apt -t experimental install gcc-7 g++-7

and running `./configure CC=gcc-7 CXX=g++-7` in @daveol's `fix_fallthrough` branch, the following compile issues still occur (pardon my Dutch):

    wouter@gangtai:~/debian/ola$ make -k
    make  all-recursive
[...]

PR #1204, opened 2017-02-20
GCC7 Implicit-fallthrough

daveol:
This is the initial macro for solving switch case fallthrough errors. It
is non-empty in all major GCC versions higher than 7 and defines
```OLA_FALLTHROUGH``` as ```__attribute__ ((fallthrough));```

```OLA_FALLTHROUGH``` has to be used at the end of the case to work.

This fixes github bug #1193 and debian bug [853583](https://bugs.debian.org/853583)

### DIFF
--- a/include/ola/base/Macro.h
+++ b/include/ola/base/Macro.h
@@ -65,7 +65,8 @@
 /**
  * @def OLA_FALLTHROUGH
  * @brief Mark switch cases as fallthrough when required
- * @note we are not using [[fallthrough]]; because C++98 compatibility
+ * @note We are not currently using [[fallthrough]]; because we need C++98
+ * compatibility
  *
  * @examplepara
  *   @code
--- a/include/ola/base/Macro.h
+++ b/include/ola/base/Macro.h
@@ -63,6 +63,30 @@
 #endif  // __GNUC__
 
 /**
+ * @def OLA_FALLTHROUGH
+ * @brief Mark switch cases as fallthrough when required
+ *
+ * @examplepara
+ *   @code
+ *   switch (cond) {
+ *     case 'foo':
+ *       doFoo();
+ *       //after foo fallthrough to bar because $REASON
+ *       OLA_FALLTHROUGH
+ *     case 'bar':
+ *   @endcode
+ */
+#ifdef __GNUC__
+#if __GNUC__ >= 7
+#define OLA_FALLTHROUGH __attribute__ ((fallthrough));
+#else
+#define OLA_FALLTHROUGH
+#endif // __GNUC__ >= 7
+#else
+#define OLA_FALLTHROUGH
+#endif // __GNUC__
+
+/**
  * @def STATIC_ASSERT
  * @brief Compile time assert().
  *
--- a/include/ola/base/Macro.h
+++ b/include/ola/base/Macro.h
@@ -65,6 +65,7 @@
 /**
  * @def OLA_FALLTHROUGH
  * @brief Mark switch cases as fallthrough when required
+ * @note we are not using [[fallthrough]]; because C++98 compatibility
  *
  * @examplepara
  *   @code
--- a/include/ola/base/Macro.h
+++ b/include/ola/base/Macro.h
@@ -71,7 +71,7 @@
  *   switch (cond) {
  *     case 'foo':
  *       doFoo();
- *       //after foo fallthrough to bar because $REASON
+ *       // after foo fallthrough to bar because $REASON
  *       OLA_FALLTHROUGH
  *     case 'bar':
  *   @endcode
@@ -81,10 +81,10 @@
 #define OLA_FALLTHROUGH __attribute__ ((fallthrough));
 #else
 #define OLA_FALLTHROUGH
-#endif // __GNUC__ >= 7
+#endif  // __GNUC__ >= 7
 #else
 #define OLA_FALLTHROUGH
-#endif // __GNUC__
+#endif  // __GNUC__
 
 /**
  * @def STATIC_ASSERT
--- a/plugins/usbpro/BaseRobeWidget.cpp
+++ b/plugins/usbpro/BaseRobeWidget.cpp
@@ -120,7 +120,7 @@ void BaseRobeWidget::ReceiveMessage() {
       } while (m_header.som != SOM);
       m_state = RECV_PACKET_TYPE;
 
-      // if we don't return fallthrough
+      // if we don't return, fallthrough
       OLA_FALLTHROUGH
     case RECV_PACKET_TYPE:
       m_descriptor->Receive(&m_header.packet_type, 1, count);
@@ -128,7 +128,7 @@ void BaseRobeWidget::ReceiveMessage() {
         return;
       m_state = RECV_SIZE_LO;
 
-      // if we don't return fallthrough
+      // if we don't return, fallthrough
       OLA_FALLTHROUGH
     case RECV_SIZE_LO:
       m_descriptor->Receive(&m_header.len, 1, count);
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: include/ola/base/Macro.h [hunk 1]
```diff
@@ -63,6 +63,30 @@
...
+ *     case 'foo':
+ *       doFoo();
+ *       //after foo falltrough to bar because $REASON
+ *       OLA_FALLTROUGH
+ *     case 'bar':
+ *   @endcode
+ */
+#ifdef __GNUC__
+#if __GNUC__ >= 7
+#define OLA_FALLTROUGH __attribute__ ((fallthrough));
```
peternewman: From the issue, we want to use [[fallthrough]]; when the compiler has C++17 support, regardless of which compiler it is, but that may be a future change I imagine. Perhaps add it as a todo if it doesn't work now (which it shouldn't as we're forcing something older to bypass other issues.

peternewman:
This probably also wants to target 0.10 branch given it's fixing an existing bug.

include/ola/base/Macro.h
daveol: we compile with c++98 still.... so ```[[falltrough]];``` is not gonna work

daveol:
I don't really like targeting branches that the bug is not specific for, #1168 demonstrates this perfectly because I now still need to apply the patch to every master every time I want to test.

Besides, you can still cherry-pick this into the 0.10 branch

include/ola/base/Macro.h
peternewman: True. Perhaps just add it as a future todo/comment for why we aren't using the [[]] variant.

peternewman:
The first bit is easy enough to fix. As soon as this gets approved. https://github.com/OpenLightingProject/ola/pull/1205

I'd rather they go into 0.10 and flow nicely into master, which will be 0.10+more, than try and cherry-pick a change related to master that might relate to code that doesn't even exist in 0.10. Also the cherry-pick becomes a pain if there are multiple commits as there are with this. It's true we could squash, but personally I like to keep the history so things like the PR comments make sense.

plugins/usbpro/BaseRobeWidget.cpp
```diff
@@ -118,16 +119,25 @@ void BaseRobeWidget::ReceiveMessage() {
           return;
       } while (m_header.som != SOM);
       m_state = RECV_PACKET_TYPE;
+
+      // if we don't return fallthrough
```
peternewman: Pedant: "if we don't return, fallthrough"

include/ola/base/Macro.h
```diff
@@ -63,6 +63,30 @@
...
 /**
+ * @def OLA_FALLTHROUGH
+ * @brief Mark switch cases as fallthrough when required
+ *
+ * @examplepara
+ *   @code
+ *   switch (cond) {
+ *     case 'foo':
+ *       doFoo();
+ *       //after foo fallthrough to bar because $REASON
```
peternewman: Can you fix the lint of the example by adding a space between // and after please.

include/ola/base/Macro.h
```diff
@@ -63,6 +63,30 @@
...
+ *       OLA_FALLTHROUGH
+ *     case 'bar':
+ *   @endcode
+ */
+#ifdef __GNUC__
+#if __GNUC__ >= 7
+#define OLA_FALLTHROUGH __attribute__ ((fallthrough));
+#else
+#define OLA_FALLTHROUGH
+#endif // __GNUC__ >= 7
```
peternewman: This needs lint fixing, two spaces before the //

include/ola/base/Macro.h
```diff
@@ -63,6 +63,30 @@
...
+#else
+#define OLA_FALLTHROUGH
+#endif // __GNUC__
```
peternewman: Again

yoe:
in several other places, too. I'll follow up in #1193.

include/ola/base/Macro.h
```diff
@@ -63,6 +63,30 @@
...
+ *     case 'foo':
+ *       doFoo();
+ *       // after foo fallthrough to bar because $REASON
+ *       OLA_FALLTHROUGH
+ *     case 'bar':
+ *   @endcode
+ */
+#ifdef __GNUC__
+#if __GNUC__ >= 7
+#define OLA_FALLTHROUGH __attribute__ ((fallthrough));
```
peternewman: Can you add a future todo/comment for why we aren't using the [[fallthrough]] variant please (as we're still on C++98).

plugins/usbpro/BaseRobeWidget.cpp
peternewman: Done

include/ola/base/Macro.h
peternewman: Done

include/ola/base/Macro.h
peternewman: Done

include/ola/base/Macro.h
peternewman: Done

include/ola/base/Macro.h
peternewman: Done

daveol:
@nomis52 hey, can you review this so it can be merged?

hepiyellow:
I am getting this error running `make`:
```
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I/opt/local/include -I./include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -D_THREAD_SAFE -I/opt/local/include -Werror -Werror -Wno-error=unused-parameter -Wno-error=deprecated-declarations -Wno-error=sign-compare -DPID_DATA_DIR=\"/usr/local/share/ola/pids\" -g -O2 -std=gnu++98 -D_THREAD_SAFE -MT common/utils/common_libolacommon_la-StringUtils.lo -MD -MP -MF common/utils/.deps/common_libolacommon_la-StringUtils.Tpo -c common/utils/StringUtils.cpp  -fno-common -DPIC -o common/utils/.libs/common_libolacommon_la-StringUtils.o
common/utils/StringUtils.cpp:420:9: error: use of undeclared identifier 'OLA_FALLTHROUGH'
        OLA_FALLTHROUGH
        ^
common/utils/StringUtils.cpp:424:9: error: use of undeclared identifier 'OLA_FALLTHROUGH'
        OLA_FALLTHROUGH
        ^
2 errors generated.
make[2]: *** [common/utils/common_libolacommon_la-StringUtils.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

My gcc:
```
$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 7.3.0 (clang-703.0.31)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Running on OS X 10.11.6 (El Capitan)
and Xcode 7.3.1

Using OLA master from Git,
`commit ce879956fb035194fd1aa486e77357dce7dc61e8 `

Any ideas?

peternewman:
Hi @hepiyellow , firstly you're better off starting a new issue, rather than commenting on a pull request that's been merged and closed for a month.

In terms of your issue, I think you'll find there's an older version of Macro.h somewhere on your machine, from a previous build/install that it's loading instead (which won't have OLA_FALLTHROUGH defined), hence why it's not working.

If this is still a problem, can you open an issue where we can continue the discussion.

PR #1213, opened 2017-03-17
Add my RDM test listing/generation script

peternewman:


peternewman:
This is currently failing pychecker with "tools/rdm/list_rdm_tests.py:86: Function (main) has too many lines (265)" but I'm not sure how sensible it is to split it out into more functions to avoid the issue.

peternewman:
Forced merge as one of the OS X tests was failing due to https://github.com/OpenLightingProject/ola/issues/1244 .

PR #1223, opened 2017-04-18
Add cast to fix build error

simark:
The error is the following:

ola/AutoStart.cpp: In function 'ola::network::TCPSocket* ola::client::ConnectToServer(short unsigned int)':
ola/AutoStart.cpp:108:56: error: missing sentinel in function call [-Werror=format=]
     execlp("olad", "olad", "--daemon", "--syslog", NULL);

From execlp(3):

  The list of arguments must be terminated by a null pointer, and, since these
  are variadic functions, this pointer must be cast (char *) NULL.

This is with "gcc (Alpine 6.2.1) 6.2.1 20160822" on Alpine Linux x86_64.

Found it from

  http://stackoverflow.com/questions/12095865/spurious-missing-sentinel-in-function-call

### DIFF
--- a/ola/AutoStart.cpp
+++ b/ola/AutoStart.cpp
@@ -105,7 +105,8 @@ TCPSocket *ConnectToServer(unsigned short port) {
 
     // Try to start the server, we pass --daemon (fork into background) and
     // --syslog (log to syslog).
-    execlp("olad", "olad", "--daemon", "--syslog", NULL);
+    execlp("olad", "olad", "--daemon", "--syslog",
+           reinterpret_cast<char*>(NULL));
     OLA_WARN << "Failed to exec: " << strerror(errno);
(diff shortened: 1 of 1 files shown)


peternewman:
Thanks. Can you fix the lint error please (or tell cpplint to ignore it if it has to be like that).

simark:
New version pushed.

peternewman:
Sorry @simark can you target 0.10 too please (you'll probably need to start a new branch at your end). Then we can get this fix into the next release.

simark:
Ok, I wrapped the line, I hope the indentation is fine :).  I'll open a new PR for 0.10.

peternewman:
Now being done targetting 0.10 in https://github.com/OpenLightingProject/ola/pull/1226

PR #1237, opened 2017-04-23
Display manufacturer name from the PID files to make life easier during RDM testing

peternewman:
- Display manufacturer name from the PID files to make life easier during testing

Still need to:
- [x] ~~Decide if we store all the manufacturer names in the existing manufacturer_pids.proto or in a separate manufacturer_names.proto (same structure, no PIDs)~~ Gone for separate file for potential different update frequencies and to avoid loading when not required
- [x] Possibly tidy up the code a bit
- [x] Update RDM-app to do so
- [x] Fix up PID store loading
- [ ] Ensure it behaves the same when you do a discovery from the RDM test server
- [ ] Allow access to the names from C++
- [ ] Add to web UI?
- [ ] Add to CLI tools?


peternewman:
This probably wants a more traditional review please @nomis52 . Any thoughts on the questions at the top welcome too, i.e. manufacturer_pids.proto or a separate manufacturer_names.proto?

The current JS changes are designed to be backwards compatible, but we could just replace the old drop down data, at the risk of breaking if someone ends up running a new UI pointing at an old RDM test server, or vice versa.

data/rdm/download.sh
```diff
@@ -1,6 +1,14 @@
...
-wget -O draft_pids.proto http://rdm.openlighting.org/download?pids=esta-draft
-wget -O manufacturer_pids.proto http://rdm.openlighting.org/download?pids=manufacturers
+datahost="rdm.openlighting.org"
+if [ ! -z $1 ]; then
+	datahost=$1
+fi
+
+echo "Fetching PID data from $datahost"
+
+curl -o pids.proto -f http://$datahost/download?pids=esta
```
peternewman: This change is because curl doesn't overwrite the file if the get fails for some reason.

tools/rdm/rdm_test_server.py
```diff
@@ -591,6 +595,10 @@ def GetJson(self, request, response):
     response.SetStatus(HTTPResponse.OK)
     return {
       'uids': [str(u) for u in uids],
+      'nameduids': dict(
```
peternewman: This is to keep compatibility with mismatched JS and Python ends.

tools/rdm/static/rdm_tests.js
```diff
@@ -480,10 +480,21 @@ RDMTests.prototype.update_device_list = function() {
     'u': universe_options.val() }, function(data) {
     if (data['status'] == true) {
       devices_list.empty();
-      var uids = data.uids;
-      $.each(uids, function(item) {
-        devices_list.append($('<option />').val(uids[item]).text(uids[item]));
-      });
+      if (data['nameduids'] != undefined) {
```
peternewman: This is to keep compatibility with mismatched JS and Python ends.

peternewman:
This is probably a good point for a review and merge @nomis52 . It should load (and ignore) the manufacturer names in C++ currently, but they work as intended in Python and the RDM test stuff. The other stuff can follow in a future PR.

peternewman:
Thanks @nomis52 . I'll get this in and address the outstanding tasks in a future PR.

PR #1240, opened 2017-04-25
converted ModelCollector.py to Python3, using 2to3 tool

eddydebaere:
As requested in 
https://groups.google.com/forum/#!topic/open-lighting/j7x7X-4U6xk

wkr
Eddy

### DIFF
--- a/tools/rdm/ModelCollector.py
+++ b/tools/rdm/ModelCollector.py
@@ -51,7 +51,7 @@ class ModelCollector(object):
    LANGUAGES,
    SLOT_INFO,
    SLOT_DESCRIPTION,
-   SLOT_DEFAULT_VALUE) = xrange(13)
+   SLOT_DEFAULT_VALUE) = range(13)
 
   def __init__(self, wrapper, pid_store):
     self.wrapper = wrapper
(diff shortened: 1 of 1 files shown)


eddydebaere:
converted ModelCollector.py to Python3, using 2to3 tool

eddydebaere:
converted ModelCollector.py to Python3, using 2to3 tool

peternewman:
Is there some reason this is closed @eddydebaere ?

peternewman:
There are a few suggestions in here for how to make the code Python 2 and 3 compatible, without the speed hits of dropping to range in Python 2:
http://stackoverflow.com/questions/15014310/why-is-there-no-xrange-function-in-python3

eddydebaere:
The speed penalties are only relevant for large  (or frequently accessed) ranges/lists. In this case, 
the range is only 13 elements, and only 1 access. Therefore, I would keep it easy, and just use 'range'.

eddydebaere:
The speed penalties are only relevant for large  (or frequently accessed) ranges/lists. In this case, 
the range is only 13 elements, and only 1 access. Therefore, I would keep it easy, and just use 'range'.

Issue #1201, opened 2017-02-12
Configure doesn't check for tcmalloc's presence when trying to add it to libs


peternewman:
Instead it errors on the first compilation after that test:
```
configure:22278: g++ -o conftest -g -O2 -std=gnu++98 -pthread   -ltcmalloc conftest.cpp  -ldl  -lprotoc >&5
/usr/bin/ld: cannot find -ltcmalloc
collect2: error: ld returned 1 exit status
configure:22278: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "OLA"
| #define PACKAGE_TARNAME "ola"
| #define PACKAGE_VERSION "0.10.3"
| #define PACKAGE_STRING "OLA 0.10.3"
| #define PACKAGE_BUGREPORT "open-lighting@googlegroups.com"
| #define PACKAGE_URL ""
| #define PACKAGE "ola"
| #define VERSION "0.10.3"
| #define HAVE_DIRENT_H 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_NETINET_IN_H 1
| #define HAVE_ARPA_NAMESER_H 1
| #define HAVE_NETDB_H 1
| #define HAVE_RESOLV_H 1
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_ERRNO_H 1
| #define HAVE_ARPA_INET_H 1
| #define HAVE_BITS_SOCKADDR_H 1
| #define HAVE_FCNTL_H 1
| #define HAVE_FLOAT_H 1
| #define HAVE_LIMITS_H 1
| #define HAVE_MALLOC_H 1
| #define HAVE_NETINET_IN_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_SYS_FILE_H 1
| #define HAVE_SYS_IOCTL_H 1
| #define HAVE_SYS_SOCKET_H 1
| #define HAVE_SYS_TIME_H 1
[...]

PR #1242, opened 2017-04-30
Fix configure to mangle AM_LDFLAGS instead of LDFLAGS, which it has no business to mangle.

yoe:
Fix #1201 

While at it, fix configure to mangle AM_LDFLAGS instead of LDFLAGS, which it has no business to mangle.

PR #1243, opened 2017-04-30
Search for tcmalloc with AC_CHECK_LIB

yoe:
Currently we just unconditionally add it to LDFLAGS, without checking
whether it even exists. That creates confusing error messages down the
line.

Instead, use AC_CHECK_LIB to see whether the library is there, and
provide a proper error message if not. This should fix
OpenLightingProject/ola#1201.

(cherry picked from commit 16ee3bc890cedc1d26a84db458ba788ecb28533d)


peternewman:
Are you happy @nomis52

peternewman:
This fixes #1201 .

peternewman:
Merging as @nomis52 was happy with #1242 and this is just a subset of it.

Issue #1244, opened 2017-05-01
cppunit 1.14.0 and ac_cv_gnu_plus_plus_98=no


ilovezfs:
I've just upgraded `cppunit` to version 1.14.0 in Homebrew, and it seems that this breaks the `ola` build. I've worked around the issue with the following diff:

```
diff --git a/Formula/ola.rb b/Formula/ola.rb
index d288ec8..78947cd 100644
--- a/Formula/ola.rb
+++ b/Formula/ola.rb
@@ -3,7 +3,7 @@ class Ola < Formula
   homepage "https://www.openlighting.org/ola/"
   url "https://github.com/OpenLightingProject/ola/archive/0.10.3.tar.gz"
   sha256 "474db6752940cea6cd9493dcbeeb13429b5d29f4777973d08738cb5ef04c9dcd"
-  revision 1
+  revision 2
   head "https://github.com/OpenLightingProject/ola.git"
 
   bottle do
@@ -36,7 +36,12 @@ class Ola < Formula
     sha256 "846eb4846f19598affdc349d817a8c4c0c68fd940303e6934725c889f16f00bd"
   end
 
+  needs :cxx11
+
   def install
+    ENV.cxx11
+    ENV["ac_cv_gnu_plus_plus_98"] = "no"
+
     resource("protobuf-c").stage do
       system "./configure", "--disable-dependency-tracking",
                             "--prefix=#{buildpath}/vendor/protobuf-c"
```

Not sure if this is going to impact your CI or not.

Also, it would be good not to need to manually specify ac_cv_gnu_plus_plus_98=no since `ENV.cxx11` should be sufficient, but configure currently overrides that unless 98 is forced to no.

(See https://github.com/Homebrew/homebrew-core/pull/13103)

peternewman:
Thanks for the heads up, yes it has broken our CI.

For info our reasoning for not having moved on before is here:
https://github.com/OpenLightingProject/ola/blob/master/README.developer#L83

Unfortunately some of the newer compilers were failing if left in auto (e.g. https://github.com/OpenLightingProject/ola/issues/1028) hence why we had to pin the build own to C++98.

I may try and roll the Mac builds back to cppunit 1.13.0 for the moment, until we decide how to proceed with that.

ilovezfs:
@peternewman you might want to cherry-pick https://github.com/Homebrew/brew/pull/2406 and give that a try.

ilovezfs:
[...]

Issue #1028, opened 2016-02-05
"template<class> class std::auto_ptr" is deprecated in C++11/C++14 (GCC6 default)


daveol:
When the rebuild for Fedora 24 triggered it failed. This was due to a deprecation warning that was treated as a error. 

The deprecation warning was about `template<class> class std::auto_ptr` since this is used in [protoc/CppGenerator.cpp](https://github.com/OpenLightingProject/ola/blob/master/protoc/CppGenerator.cpp) on [line 52](https://github.com/OpenLightingProject/ola/blob/master/protoc/CppGenerator.cpp#L52) and [line 57](https://github.com/OpenLightingProject/ola/blob/master/protoc/CppGenerator.cpp#L57)

The koji build job is [here](http://koji.fedoraproject.org/koji/taskinfo?taskID=12854227) and the build log is [here](https://kojipkgs.fedoraproject.org//work/tasks/4227/12854227/build.log)

peternewman:
-Wno-deprecated-declarations appears to be the immediate workaround.

See also:
http://stackoverflow.com/questions/2404115/is-auto-ptr-deprecated

This seems to imply it can be bypassed/fixed with some compiler flags:
https://github.com/fuzzylite/fuzzylite/issues/65

peternewman:
It would appear the main cause is:
"The default mode for C++ is now -std=gnu++14 instead of -std=gnu++98."

daveol:
I can confirm this, since a build with ```-std=gnu++98``` didn't fail

daveol:
For what it's worth, this is the patch I made for the fedora package:
```patch
diff --git a/configure.ac b/configure.ac
index ef31e09..3e98692 100644
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,9 @@ AC_PROG_LN_S
 AC_PROG_MAKE_SET
 AC_PROG_MKDIR_P
 
+# Ensure we use gnu++98 instead of c++14
+CPPFLAGS="$CPPFLAGS -std=gnu++98"
+
 # Checks for header files.
 AC_HEADER_DIRENT
 AC_HEADER_RESOLV
```

peternewman:
Thanks for the confirmation @daveol . @nomis52 any feeling on how we should progress?

daveol:
[...]

PR #1250, opened 2017-05-14
Require cppunit less than 1.14.0

peternewman:
Until we've sorted our build to be happy with C++11


peternewman:
See also the discussion here https://github.com/OpenLightingProject/ola/issues/1244

nomis52:
See comments on #1028  .

peternewman:
Merged, to address the other issues after 0.10.4.

PR #1263, opened 2017-06-17
Resync with 0.10

peternewman:


peternewman:
N.B. @nomis52 this also bumps the Java Protobuf version from 2.6 to 3.1, so we can use the same up to date main protobuf as we do elsewhere:
https://github.com/OpenLightingProject/ola/pull/1263/files#diff-61657881d138c2b7ac1366d88c9ee035R91

I can revert this if you'd rather, but it then means the rest of the build has to use Protobuf 2.6 too.

PR #1196, opened 2017-02-04
Add more missing RDM tests

peternewman:

PR #1277, opened 2017-07-02
More RDM test changes

peternewman:
Part of #1196 .


peternewman:
Sorry, this one is rather large but I only did it as two commits before.

peternewman:
Hopefully this is small enough to review now as a single standalone PR @nomis52 ? I don't think I can really break it down further (while still passing the tests), unless I go and make some fresh commits to do the same code changes, as I did it in a few big chunks as it was mostly copying and pasting from the RDM test generation script output.

peternewman:
Is this small enough on it's own now @nomis52 ? There's a tiny bit more I could pull out if it helps, but most of this is now just moving some of the tests from one bit of the file to another so they're grouped together.

Issue #1285, opened 2017-07-18
SPI input sampling


FloEdelmann:
I managed to get DMX input working with Raspberry Pi's on-board hardware (and a transciever like needed for the UART plugin, but that doesn't count 😉)!

I'm sampling SPI's MISO port (the other SPI pins are not connected) at 2MHz (= 8x DMX frequency) and parse the resulting bits to get the real data back. SPI uses a dedicated clock on the Raspberry Pi so its timing is stable – especially compared to GPIO bit-banging (see #1188). I tested it with DMX data from OLA's UART plugin, the DMXCreator 512 and e:cue and it worked well.

However, one issue persists: streaming SPI data is not possible without writing a kernel module, if I have researched correctly. This means that at the moment, blocks of 8192 bytes are read as a batch operation and parsed afterwards. We must hope that a whole DMX packet is contained in these 8192 bytes. Since often a packet will be cut off by the block end, the higher channels are updated less frequently than the lower ones. However, one could make the 8192 bytes configurable for users that want a higher refresh rate for the high channels (which results in more stuttering in all channels, as the time between two SPI read operation increases). For the lower channels, it works fine though.

I will now work on wrapping my code in an OLA plugin. If I run into problems, I'll add a comment here to keep it organized. I'd name the plugin *SPI DMX* (directory name `spidmx`, C++ prefix `SpiDmx`) to distinguish it from the existing *SPI* plugin. Is there a better name?

peternewman:
I wonder about spiin/SPIIn to match the style we've used for the other one, and to be less ambiguous, given both directions have SPI and DMX involved. All the other *DMX ones are primarily output too.

Of course you could try adding it to the SPI out plugin, but that may be rather too hard.

FloEdelmann:
Generally speaking, output via SPI should also be possible. One "just" has to synchronize input read and output write, since those operations are a single "shift register"-like operation.

peternewman:
Yeah I hadn't really considered DMX out too (via the SPI pins).

Do you think standalone DMX out via SPI should work, or you've actually done a basic proof of concept that it does?

You've probably sold me on spidmx/SPIDmx as a name then. We're not very consistent, every other TLA is just camel cased (giving SpiDmx), but given our existing SPI plugin is SPI, my feeling is we should stick to that. Thoughts @nomis52 .

FloEdelmann:
[...]

PR #1286, opened 2017-07-18
Make all plugins CamelCase

FloEdelmann:
As discussed in #1285, I refactored the SPI plugin to change the C++ name prefix to `Spi`. To standardize all plugins, I did the same for the GPIO, OPC and OSC plugins.

Unfortunately, I get the following errors when `make`ing, although I cannot find references to the old names in the code anymore:

```
olad/.libs/libolaserver.so: undefined reference to `vtable for ola::plugin::osc::OscPlugin'
olad/.libs/libolaserver.so: undefined reference to `vtable for ola::plugin::spi::SpiPlugin'
olad/.libs/libolaserver.so: undefined reference to `vtable for ola::plugin::gpio::GpioPlugin'
olad/.libs/libolaserver.so: undefined reference to `vtable for ola::plugin::openpixelcontrol::OpcPlugin'
```

Am I missing something or are these old build artefacts and I just have to rebuild completely (I have already run `make clean`)?

### DIFF
--- a/olad/DynamicPluginLoader.cpp
+++ b/olad/DynamicPluginLoader.cpp
@@ -68,7 +68,7 @@
 #endif  // USE_OPENPIXELCONTROL
 
 #ifdef USE_OSC
-#include "plugins/osc/OSCPlugin.h"
+#include "plugins/osc/OscPlugin.h"
 #endif  // USE_OSC
 
 #ifdef USE_PATHPORT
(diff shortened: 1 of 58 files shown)


FloEdelmann:
d245842 actually makes it standard Markdown compliant which enabled me to preview it right in my editor which doesn't understand GFM. I'm rather confused why the line breaks worked before, since GFM does also require 2 spaces at the end of a line. It shouldn't break things though.

FloEdelmann:
Well, then UARTDMX, USBDMX, USBPro, FTDIDMX and OpenDMX can all be done in a separate pull request and this one can be closed. I'll raise an issue for that.

peternewman:
Thanks again @FloEdelmann and sorry for the change of mind.

PR #1289, opened 2017-07-21
SPI DMX receiver

### LINKED ISSUE #1285: SPI input sampling

FloEdelmann:
Closes #1285, see there for a description.

This is still a work in progress, I just want Travis to run its tests. On my laptop I compile and run it fine. On the Raspberry, even with a fresh `autoreconf`, after compiling and installing, running `olad` gives

```
olad: error while loading shared libraries: libolaspidmx.so.0: cannot open shared object file: No such file or directory
```

I'm confused since it works on my laptop, so I can't have missed something, right?

**EDIT:** Apparently, Travis does also only fail due to a mistake in protobuf. `spidev` is actually not available on Mac so it is skipped completely.


FloEdelmann:
I have to manually set the `LD_LIBRARY_PATH` environment variable to `/usr/local/lib`; it was empty before. Any idea how that could have happened?

FloEdelmann:
Now it's finished. I'd be happy to see you review it :)

plugins/spidmx/SpiDmxDevice.cpp
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Set the default preferences for this one Device
+ */
+void SpiDmxDevice::SetDefaults() {
+  if (!m_preferences) {
+    return;
+  }
+
+  bool save = m_preferences->SetDefaultValue(DeviceBlocklength(),
+                                             UIntValidator(0, 65535),
```
peternewman: Surely 0 as a minimum value is nonsense, you want at least 1 slot don't you?

plugins/spidmx/SpiDmxParser.cpp
```diff
@@ -0,0 +1,538 @@
...
+  chunk = buffer;
+
+  chunk_bitcount = 0;
+
+  ChangeState(WAIT_FOR_BREAK);
+
+  while (chunk_bitcount < buffersize) {
+    switch (state) {
+      case WAIT_FOR_BREAK:
+        // printf("%6ld  0x%02x  wait for break (%d)\n", chunk_bitcount,
```
peternewman: Can we remove this debug please, or convert it to OLA_DEBUG.

plugins/spidmx/SpiDmxParser.cpp
```diff
@@ -0,0 +1,538 @@
...
+  state_bitcount = 0;
+
+  if (state == WAIT_FOR_MAB) {
+    channel_count = -1;
+  }
+}
+
+/**
+ * Helper function that returns the number of zeros if a falling edge is
+ * detected in the given byte or -1 if the byte is no falling edge.
```
peternewman: SPaG "is not a falling edge" is I think what you're trying to say?

plugins/spidmx/SpiDmxParser.cpp
```diff
@@ -0,0 +1,538 @@
...
+      return 8;
+
+    default:
+      return -1;
+  }
+}
+
+/**
+ * Helper function that returns the number of ones if a rising edge is
+ * detected in the given byte or -1 if the byte is no rising edge.
```
peternewman: Again

plugins/spidmx/SpiDmxPlugin.cpp
```diff
@@ -0,0 +1,130 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * SpiDmxPlugin.cpp
+ * This looks for possible SPI devices to instanciate and is managed by OLAD.
```
peternewman: SPaG instantiate

plugins/spidmx/SpiDmxPlugin.h
```diff
@@ -0,0 +1,75 @@
...
+namespace spidmx {
+
+class SpiDmxPlugin : public Plugin {
+ public:
+  explicit SpiDmxPlugin(PluginAdaptor *plugin_adaptor)
+      : Plugin(plugin_adaptor),
+        m_plugin_adaptor(plugin_adaptor) {
+  }
+
+  ola_plugin_id Id() const { return OLA_PLUGIN_EXPERIMENTAL; }
```
peternewman: Can you define a new OLA_PLUGIN_SPIDMX in https://github.com/OpenLightingProject/ola/blob/master/common/protocol/Ola.proto and set the value to 10000 for now, then we'll confirm an actual value when it gets merged. But doing it that way will increase the chances we'll remember to do so!

plugins/spidmx/SpiDmxThread.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/**
+ * This thread does only have to run if ports using it are patched to a
+ * universe. Thus, they must register and unregister to notify this thread.
+ */
+void SpiDmxThread::RegisterPort() {
+  m_registered_ports++;
+
+  if (m_registered_ports == 1) {
```
peternewman: >= 1 perhaps, although it shouldn't happen.

plugins/spidmx/SpiDmxThread.cpp
```diff
@@ -0,0 +1,174 @@
...
+    Start();
+  }
+}
+void SpiDmxThread::UnregisterPort() {
+  m_registered_ports--;
+
+  if (m_registered_ports == 0) {
```
peternewman: <= 0 again just in case.

plugins/spidmx/SpiDmxParser.cpp
FloEdelmann: If I put it in OLA_DEBUG, it outputs so much that receiving gets too laggy to be usable. I'll remove those comments shortly before merging but I keep them until then for easy debugging.

plugins/spidmx/SpiDmxPlugin.h
FloEdelmann: When `make`ing, I get the following error:

```
"ola.proto.OLA_PLUGIN_EXPERIMENTAL" uses the same enum value as "ola.proto.OLA_PLUGIN_SPIDMX". If this is intended, set 'option allow_alias = true;' to the enum definition.
```

Should I add the suggested option somewhere (where?) or just set the constant to 9999?

plugins/spidmx/SpiDmxDevice.cpp
peternewman: Fixed

plugins/spidmx/SpiDmxParser.cpp
peternewman: Done

plugins/spidmx/SpiDmxParser.cpp
peternewman: Done

plugins/spidmx/SpiDmxPlugin.cpp
peternewman: Done

plugins/spidmx/SpiDmxThread.cpp
peternewman: Well done for understanding what I meant, even when the GitHub markdown confused it.

plugins/spidmx/SpiDmxThread.cpp
peternewman: Done

plugins/spidmx/SpiDmxPlugin.h
peternewman: Interestingly this only happens on the OS X builds, which I think use a later protobuf.

As you can probably tell, we've only been through this particular workflow a few times with PRs. Essentially what we're trying to do initially is stopping you grabbing 23 for example and someone else having it too, as well as ensuring your code doesn't just use the OLA_PLUGIN_EXPERIMENTAL constant. I guess we could either comment out OLA_PLUGIN_EXPERIMENTAL (which will kill your same enum value error, or perhaps just change the comment there to say an ID greater than OLA_PLUGIN_EXPERIMENTAL, and allow a free for all up there, knowing they'll get updated before merging.

plugins/spidmx/README.md
```diff
@@ -0,0 +1,46 @@
...
+will instantiate a device.
+
+### Per Device Settings
+
+**Note:** Substitute `<device>` with the full device name, e.g.
+`/dev/spidev0.0`.
+
+`<device>-blocklength = 4096`  
+How many SPI bytes (= DMX bits) should be received (optional). The default
+is 4096, but 8192 is recommended.
```
peternewman: Out of interest, how come we don't default to the value we recommend?

FloEdelmann:
On the Raspberry Pi, without further configuration, a 4kB read / trabsmit buffer is available. If one wants to read more than that in a single call, the buffer size needs to be increased via a command line call.

plugins/spidmx/README.md
FloEdelmann: On the Raspberry Pi, without further configuration, a 4kB read / trabsmit buffer is available. If one wants to read more than that in a single call, the buffer size needs to be increased via a command line call.

plugins/spidmx/SpiDmxPlugin.h
FloEdelmann: Do we need the OLA_PLUGIN_EXPERIMENTAL then at all? The comment could just say that all IDs above 10000 are experimental and used during development until a fixed IDs is assigned.

plugins/spidmx/README.md
peternewman: Ah yes, a few lines up the same file. I wonder if it's worth an N.B. here when recommending 8192 that they'll need to do some local config to enable this.

plugins/spidmx/SpiDmxPlugin.h
peternewman: Yeah you're right, we probably don't. Hopefully that will make it less confusing for people too. Do you fancy trying to update the words for this, or would you rather I did it?

plugins/spidmx/README.md
FloEdelmann: The call used to read / write ([ioctl](http://man7.org/linux/man-pages/man2/ioctl.2.html), in [SPIDMXWidget.cpp](https://github.com/FloEdelmann/ola/blob/b8ecf02095b3bff6a9c5e0a3ebdda374000394b6/plugins/spidmx/SPIDMXWidget.cpp#L112)) does not return the number of received / transmitted bytes but only an unspecified non-negative integer or -1 in an error case.

Since the call fails if the buffer is not big enough, the check should already be enough, I guess.

plugins/spidmx/SpiDmxPlugin.h
FloEdelmann: I've made an attempt in b8ecf02. Is this okay?

plugins/spidmx/README.md
peternewman: That's a shame. Perhaps the error on the line below needs a hint about buffer sizes though, if changing the value to 4097 will cause it to break?

FloEdelmann: Like this?

common/protocol/Ola.proto
```diff
@@ -66,15 +66,15 @@ enum PluginIds {
...
   OLA_PLUGIN_GPIO = 22;
+  OLA_PLUGIN_SPIDMX = 10001;
 
   /*
    * To obtain a new plugin ID, open a ticket at
    * https://github.com/OpenLightingProject/ola/issues/new
    * Plugin IDs are usually assigned just prior to merging the code into the
-   * mainline. For development of plugins please use the value of
-   * OLA_PLUGIN_EXPERIMENTAL in a plugin ID you define above.
+   * mainline. During development of new plugins, please use a value of 10000
```
peternewman: We probably want this to be values of 10000 or more (so e.g. "a value of 10000 or greater")

plugins/spidmx/SPIDMXWidget.cpp
```diff
@@ -0,0 +1,184 @@
...
+  tr.len           = blocklength;
+  tr.speed_hz      = SPI_SPEED;
+  tr.delay_usecs   = SPI_DELAY;
+  tr.bits_per_word = SPI_BITS_PER_WORD;
+  tr.cs_change     = SPI_CS_CHANGE;
+  tr.pad           = SPI_PAD;
+
+  int ret = ioctl(m_fd, SPI_IOC_MESSAGE(1), &tr);
+  if (ret < 1) {
+    OLA_WARN << Name() << " ioctl read/write error. Maybe this is due to a"
```
peternewman: Perhaps say "This may be due to a buffer size mismatch" or something similar.

peternewman: Done.

common/protocol/Ola.proto
peternewman: Done.

plugins/spidmx/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Config file: `ola-spidmx.conf`
+
+`enabled = true`  
+Enable this plugin (DISABLED by default).
+
+`device_prefix = <string>`  
+The prefix of files to match in `/dev`. Usually set to `spidev`. Each match
+will instantiate a device.
```
peternewman: You should add a note here that multiple device_prefix lines are supported.

plugins/spidmx/SPIDMXThread.cpp
```diff
@@ -0,0 +1,175 @@
...
+    m_spi_rx_buffer(blocklength),
+    m_spi_tx_buffer(blocklength) {
+}
+
+SPIDMXThread::~SPIDMXThread() {
+  Stop();
+}
+
+/**
+ * This thread does only have to run if ports using it are patched to a
```
peternewman: SPaG "This thread only has to run if..."

plugins/spidmx/SPIDMXThread.cpp
```diff
@@ -0,0 +1,175 @@
...
+  return m_dmx_rx_buffer;
+}
+
+
+/**
+  * @brief Set the callback to be called when the receive buffer is updated.
+  * @param callback The callback to call.
+  */
+bool SPIDMXThread::SetReceiveCallback(Callback0<void> *callback) {
+  if (!callback) {
```
peternewman: Shouldn't this either be m_receive_callback, or the line below do a reset to NULL?

plugins/spidmx/SPIDMXWidget.cpp
```diff
@@ -0,0 +1,185 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * SPIDMXWidget.cpp
+ * This is a wrapper around the needed SPIDEV calls.
```
peternewman: SPaG "the required SPIDEV"

plugins/spidmx/SPIDMXWidget.h
```diff
@@ -0,0 +1,110 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * SPIDMXWidget.h
+ * This is a wrapper around the needed SPIDEV calls.
```
peternewman: Again

plugins/spidmx/SPIDMXThread.cpp
FloEdelmann: Oh right, thanks!

plugins/spidmx/README.md
peternewman: Done.

plugins/spidmx/SPIDMXThread.cpp
peternewman: Done.

plugins/spidmx/SPIDMXThread.cpp
peternewman: No worries, done.

plugins/spidmx/SPIDMXWidget.cpp
peternewman: Done.

plugins/spidmx/SPIDMXWidget.h
peternewman: Done.

peternewman:
LGTM, over to @nomis52 .

FloEdelmann:
Any news on this?

peternewman:
@nomis52 have you got any time to review this please?

FloEdelmann:
Since this PR will be referenced in my Bachelor's Thesis due next month, it would be nice if it could already be merged then :)

common/protocol/Ola.proto
```diff
@@ -66,15 +66,15 @@ enum PluginIds {
   OLA_PLUGIN_UARTDMX = 20;
   OLA_PLUGIN_OPENPIXELCONTROL = 21;
   OLA_PLUGIN_GPIO = 22;
+  OLA_PLUGIN_SPIDMX = 10001;
```
nomis52: If we're merging to mainline this should have a non-dev code.

peternewman: Well spotted!

peternewman: Fixed.

FloEdelmann:
Nice, thank you! :)

peternewman:
No worries, thanks for the code @FloEdelmann .

Coverity is complaining about:
```
** CID 185518:  Uninitialized members  (UNINIT_CTOR)
/plugins/spidmx/SPIDMXParser.h: 38 in ola::plugin::spidmx::SPIDMXParser::SPIDMXParser(ola::DmxBuffer *, ola::Callback0<void> *)()


________________________________________________________________________________________________________
*** CID 185518:  Uninitialized members  (UNINIT_CTOR)
/plugins/spidmx/SPIDMXParser.h: 38 in ola::plugin::spidmx::SPIDMXParser::SPIDMXParser(ola::DmxBuffer *, ola::Callback0<void> *)()
32
33     class SPIDMXParser {
34      public:
35       SPIDMXParser(DmxBuffer *buffer, Callback0<void> *callback)
36         : m_dmx_buffer(buffer),
37           m_callback(callback) {
>>>     CID 185518:  Uninitialized members  (UNINIT_CTOR)
>>>     Non-static class member "sampling_position" is not initialized in this constructor nor in any functions that it calls.
38       }
39       void ParseDmx(uint8_t *buffer, uint64_t chunksize);
40       void SetCallback(Callback0<void> *callback) {
41         m_callback = callback;
42       }
43
```

I've not looked carefully enough to realise why it's only moaning about sampling_position and not the other members, but would you mind resolving this in a new PR please @FloEdelmann ?

PR #1268, opened 2017-06-23
Plugfest RDM changes

peternewman:

PR #1300, opened 2017-08-13
Still even more RDM tests

peternewman:
A subset of #1268 .


peternewman:
Sorry @nomis52 probably wants a quick re-review, I'd managed to put some code in twice, so I've dropped that and corrected a pair of other bugs instead.

PR #1301, opened 2017-08-13
Fix USB PRO change-of-state packet handling to allow valid packets, closes #1290

### LINKED ISSUE #1290: USB PRO: Warnings for "normal" changed packages?
I'm implementing an USB_PRO compatible device, reading the code that I'll have to work with... 

https://github.com/OpenLightingProject/ola/blob/master/plugins/usbpro/GenericUsbProWidget.cpp#L273

Here a "changed" packet is handled. The format is designed to be able to compress changes by only sending the changed bytes not all bytes. 

Now this implementation in OLA, that issues a warning if the packet is not the full maximum of 46 bytes, hints at that the original implementation was faulty. 

Suppose DMX channel 30 and 160 have changed. I'd expect the device to send: 
```
<start-of-packet: 7e><packet label: changed-data>  <length: bond, james: 0x007>
  <3: first change is at 3*8=24>  <0x40: the 6th byte has changes> <0x00: no changes in 32-39> <0> <0> <0> <newvalue for ch 30> <39 slots not sent> <end-of-packet: e7>
```
And then another 
```
<length: 7>
  <20: first change is at 20*8=160>  <0x01: the frist byte has changes> <0x00: no more changes> <0> <0> <0> <newvalue for ch 160> <39 slots not sent>
```

I don't HAVE any original devices, so I can't test them. But I'd expect the olad console to be flooded with these warnings for devices doing "the right thing". 

So... IMHO the warning and return should go. Wait! Actually...  The check should be against the minimum packet length of 6 or 7. (minimum valid packet length is 6. But it'd be odd to send one. And "return" is a valid action. So I can live with a warning in that case. The realistic minimum is 7. 

Agreed? Anybody?

Actually... I just realized, you should check agianst BIGGER than the maximum changed-data-packet as well. 
```
if (length > sizeof(widget_data_changed)) {
    OLA_WARN << "Change of state packet was too big: " << length;
    return;
  }
```

rewolff:
seems to have worked... :-)

### DIFF
--- a/plugins/usbpro/EnttecUsbProWidget.cpp
+++ b/plugins/usbpro/EnttecUsbProWidget.cpp
@@ -499,8 +499,12 @@ void EnttecPortImpl::HandleDMXDiff(const uint8_t *data, unsigned int length) {
     uint8_t data[40];
   } widget_data_changed;
 
+  // The minimum packet length is 1 byte start_pos, 5 bytes bitmap
+  // and one changed databyte. A length-6 packet would be valid (if
+  // no bits in the changed bitmap was set), but nonsensical to
+  // transmit. In any case, ignoring it would at least work.
   if ((length < 7) || (length > sizeof(widget_data_changed))) {
-    OLA_WARN << "Change of state packet was invalid: " << length;
+    OLA_WARN << "Change of state packet has an invalid length: " << length;
     return;
   }
 
--- a/plugins/usbpro/EnttecUsbProWidget.cpp
+++ b/plugins/usbpro/EnttecUsbProWidget.cpp
@@ -499,8 +499,8 @@ void EnttecPortImpl::HandleDMXDiff(const uint8_t *data, unsigned int length) {
     uint8_t data[40];
   } widget_data_changed;
 
-  if (length < sizeof(widget_data_changed)) {
-    OLA_WARN << "Change of state packet was too small: " << length;
+  if ((length < 7) || (length > sizeof(widget_data_changed))) {
+    OLA_WARN << "Change of state packet was invalid: " << length;
     return;
   }
 
--- a/plugins/usbpro/GenericUsbProWidget.cpp
+++ b/plugins/usbpro/GenericUsbProWidget.cpp
@@ -270,8 +270,8 @@ void GenericUsbProWidget::HandleDMXDiff(const uint8_t *data,
     uint8_t data[40];
   } widget_data_changed;
 
-  if (length < sizeof(widget_data_changed)) {
-    OLA_WARN << "Change of state packet was too small: " << length;
+  if ((length < 7) || (length > sizeof(widget_data_changed))) {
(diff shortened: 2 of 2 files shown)


plugins/usbpro/EnttecUsbProWidget.cpp
```diff
@@ -499,8 +499,8 @@ void EnttecPortImpl::HandleDMXDiff(const uint8_t *data, unsigned int length) {
     uint8_t data[40];
   } widget_data_changed;
 
-  if (length < sizeof(widget_data_changed)) {
-    OLA_WARN << "Change of state packet was too small: " << length;
+  if ((length < 7) || (length > sizeof(widget_data_changed))) {
```
peternewman: Can we have a comment in the code where the 7 came from please.

plugins/usbpro/EnttecUsbProWidget.cpp
```diff
@@ -499,8 +499,8 @@ void EnttecPortImpl::HandleDMXDiff(const uint8_t *data, unsigned int length) {
...
+    OLA_WARN << "Change of state packet was invalid: " << length;
```
peternewman: Can we change this to "was an invalid size" or similar, we've not actually looked at if the data is valid yet.

plugins/usbpro/GenericUsbProWidget.cpp
```diff
@@ -270,8 +270,8 @@ void GenericUsbProWidget::HandleDMXDiff(const uint8_t *data,
     uint8_t data[40];
   } widget_data_changed;
 
-  if (length < sizeof(widget_data_changed)) {
-    OLA_WARN << "Change of state packet was too small: " << length;
+  if ((length < 7) || (length > sizeof(widget_data_changed))) {
```
peternewman: Likewise for the second implementation.

peternewman:
@nomis52 it seems rather odd we need to duplicate this code for the Enttec Pro and the Generic USBPro, although I can't see any easy way to resolve it as the inheritance diagrams don't seem to overlap at all. Probably one for another day I guess.

rewolff:
I'm not too famliar with how this works, but pushing to my branch on github caused immediate changes here in this window, so I think the pull is now including the newly added changes as well....

My own compile-test now worked. Thanks for the help.

plugins/usbpro/GenericUsbProWidget.cpp
peternewman: Done

plugins/usbpro/EnttecUsbProWidget.cpp
peternewman: Done

peternewman:
Yes, because it's a PR, pushing to your branch does the trick.

Some of the unit tests are a tad flakey so I'll rerun them when the others have finished.

PR #1303, opened 2017-08-14
clarfied/expanded build instructions for debian and similar distribut…

rewolff:
I look in the README.debian whenever I've forgotten how to build things. I think my "notes" can be useful for others as well. 

I don't know when "autoreconf" (without -i) is supposed to work. So I left the original as is and added a comment about adding -i when you get complaints....


peternewman:
Thanks @rewolff . This should probably go into 0.10.6. Are you happy targeting that instead or would you rather I cherry pick once it's merged?

peternewman:
@rewolff I've made some minor style tweaks and also added more packages to the list you may want to install if you want to take a look.

@nomis52 this is ready for review and merge.

rewolff:
looks good to me.

PR #1305, opened 2017-08-15
Fix some more RDM tests

peternewman:
(cherry picked from commit 027e6933d2f972f7ef052e7cbbf388f721bc8341)

### DIFF
--- a/tools/rdm/TestMixins.py
+++ b/tools/rdm/TestMixins.py
@@ -317,7 +317,9 @@ class SetWithDataMixin(ResponderTestFixture):
   def Test(self):
     results = [
       self.NackSetResult(RDMNack.NR_FORMAT_ERROR),
-      self.NackSetResult(RDMNack.NR_UNSUPPORTED_COMMAND_CLASS), # TODO(Peter): Fix this, ideally we change behaviour based on past support of the PID
+      # TODO(Peter): Fix this, ideally we change behaviour based on past
+      # support of the PID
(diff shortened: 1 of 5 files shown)


tools/rdm/TestMixins.py
```diff
@@ -317,6 +317,9 @@ class SetWithDataMixin(ResponderTestFixture):
   def Test(self):
     results = [
       self.NackSetResult(RDMNack.NR_FORMAT_ERROR),
+      # TODO(Peter): Fix this, ideally we change behaviour based on past
+      # support of the PID
+      self.NackSetResult(RDMNack.NR_UNSUPPORTED_COMMAND_CLASS),
```
nomis52: why is this changing?

peternewman: Think of e.g device hours, a responder may not support setting it, in which case this response is valid, depending on the order packet format and command class are checked.

The future extension is to remember if we succeeded on a normal SetDeviceHours, pass that in and advisory/warn here that the normal one succeeded and this didn't fail with NR_FORMAT_ERROR.

I can add a warning/advisory to these in the interim saying this should really be an NR_FORMAT_ERROR if you'd like, but I don't think that's correct if SET isn't supported generally for the PID.

peternewman: Are you happy with that as an explanation @nomis52, or do we need a different plan going forward?

PR #1316, opened 2017-09-03
Move to the edge group on Travis so we potentially break early

peternewman:


peternewman:
See discussion here, may not actually do the job, but merging anyway:
https://github.com/travis-ci/travis-ci/issues/8315

PR #1326, opened 2017-10-08
 Add a new specific OLA_ASSERT_DMX_EQUALS test

peternewman:


common/dmx/RunLengthEncoderTest.cpp
```diff
@@ -149,9 +149,11 @@ void RunLengthEncoderTest::checkEncodeDecode(const uint8_t *data,
   OLA_ASSERT_TRUE(m_encoder.Encode(src, m_dst, &dst_size));
 
   OLA_ASSERT_TRUE(m_encoder.Decode(0, m_dst, dst_size, &dst));
-  OLA_ASSERT_TRUE(src == dst);
-  OLA_ASSERT_EQ(dst.Size(), data_size);
+  // Because we've zero padded to a full frame, we check it's a subset
```
peternewman: These few lines want a really careful check, as this set of tests wasn't being run before and I've had to make some changes to the tests so they pass now.

PR #1329, opened 2017-11-04
Initialize members of SPIDMXParser class

FloEdelmann:
Addition to #1289, as requested in a [comment](https://github.com/OpenLightingProject/ola/pull/1289#issuecomment-341681607) after merging.

I don't know how to run Coverity myself, so I initialized all members. It builds without problems and all variables are reset before first proper use anyway, so it should not break anything.

### DIFF
--- a/plugins/spidmx/SPIDMXParser.h
+++ b/plugins/spidmx/SPIDMXParser.h
@@ -35,6 +35,13 @@ class SPIDMXParser {
   SPIDMXParser(DmxBuffer *buffer, Callback0<void> *callback)
     : m_dmx_buffer(buffer),
       m_callback(callback) {
+    state = WAIT_FOR_BREAK;  // reset in ChangeState()
+    chunk = NULL;  // reset in ParseDmx()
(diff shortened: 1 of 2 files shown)


plugins/spidmx/SPIDMXParser.h
```diff
@@ -35,6 +35,13 @@ class SPIDMXParser {
   SPIDMXParser(DmxBuffer *buffer, Callback0<void> *callback)
     : m_dmx_buffer(buffer),
       m_callback(callback) {
+    state = WAIT_FOR_BREAK;  // reset in ChangeState()
```
nomis52: These should all be prefixes with m_ and set using m_state(WAIT_FOR_BREAK)

peternewman:
If you're interested @FloEdelmann , there's details on Coverity Scan here:
https://scan.coverity.com/

You can sign up, run it on your own machine, and upload the results to the website for analysis. Or you could configure your own Travis instance to run it with your key.

Issue #1192, opened 2017-01-28
protobuf 3.2.0 issue


ilovezfs:
```
libtool: compile:  clang++ -DHAVE_CONFIG_H -I. -I./include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -D_THREAD_SAFE -isystem /usr/local/Cellar/protobuf/3.2.0/include -DPID_DATA_DIR=\"/usr/local/Cellar/ola/HEAD-5f490bd_1/share/ola/pids\" -g -O2 -std=gnu++98 -D_THREAD_SAFE -c common/rpc/Rpc.pb.cc  -fno-common -DPIC -o common/rpc/.libs/common_libolacommon_la-Rpc.pb.o
libtool: compile:  clang++ -DHAVE_CONFIG_H -I. -I./include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -D_THREAD_SAFE -isystem /usr/local/Cellar/protobuf/3.2.0/include -DPID_DATA_DIR=\"/usr/local/Cellar/ola/HEAD-5f490bd_1/share/ola/pids\" -g -O2 -std=gnu++98 -D_THREAD_SAFE -c common/rpc/Rpc.pb.cc -o common/rpc/common_libolacommon_la-Rpc.pb.o >/dev/null 2>&1
/bin/sh ./libtool  --tag=CXX   --mode=compile clang++ -DHAVE_CONFIG_H -I.    -I./include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -D_THREAD_SAFE -isystem /usr/local/Cellar/protobuf/3.2.0/include -g -O2 -std=gnu++98 -D_THREAD_SAFE  -c -o common/protocol/common_protocol_libolaproto_la-Ola.pb.lo `test -f 'common/protocol/Ola.pb.cc' || echo './'`common/protocol/Ola.pb.cc
libtool: compile:  clang++ -DHAVE_CONFIG_H -I. -I./include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -D_THREAD_SAFE -isystem /usr/local/Cellar/protobuf/3.2.0/include -g -O2 -std=gnu++98 -D_THREAD_SAFE -c common/protocol/Ola.pb.cc  -fno-common -DPIC -o common/protocol/.libs/common_protocol_libolaproto_la-Ola.pb.o
[...]

PR #1336, opened 2017-12-01
Eliminate protobuf AddDescriptors() call.

cmuellner:
Eliminating protobuf's AddDescriptors() calls enabled
forward-compatibility with libprotobuf. That allows to
run OLA on recent Linux distrubutions.

Tested on x86_64 running Fedora 27 with protobuf 3.3.1
in combination with QLC+ and a uDMX controller.

See #1192.

Signed-off-by: Christoph Müllner <christophm30@gmail.com>


cmuellner:
Comment on this:
The generated code is only needed when being run during static initialization.
During a quick search I haven't found a reason to eliminate the call.
However I'm confronted with this source base only for a couple of hours.

In fact this fixes the build issue with recent libprotobuf versions.

I've created also a Fedora 27 package using OBS:

https://build.opensuse.org/package/show/home:cmuellner_vienna/ola

peternewman:
This is potentially great news thanks @cmuellner .

Can you just give a bit more detail on what level of testing you've done? E.g. have you clicked around the web UI's plugin descriptions? Did you have the dummy plugin enabled, and if you patch that to a universe, do the RDM aspects of that show data when you expand them?

I'm just trying to make a bit of sense of what the generated code does in Protobuf land and how come we don't seem to need it, seemingly on Protobuf 3.3. We probably need to double check an old version still builds, but the Linux Travis checks look promising at least.

peternewman:
I've just reread this comment from @ilovezfs in #1192 and I note he actually said the same thing before and I just got so tied up in trying to migrate that whole codebase that I missed it:
https://github.com/OpenLightingProject/ola/issues/1192#issuecomment-276083201

nomis52:
This all seems very reasonable. Thanks.

cmuellner:
Regarding my testing: I just have limited amount of devices here. I just received a uDMX (clone) yesterday, which I used to control a LED Mini Par (which has 8 channels). I did a very trivial test by enabling the output of the uDMX plugin and changing a few channels of the par. I just used the QLC+ application (new web UI).

loansindi:
Using this patch I was able to build on Arch linux. Web interface seems to be working. RDM responses from dummy device work correctly, receiving Artnet from software on another machine works correctly.

peternewman:
Okay this all sounds very promising from the various testing @cmuellner .

My plan is we merge #1338 into OLA master, then update the Travis config so it's not an allowed failure, and it should pass, and then we can merge this knowing that Travis likes it too!

peternewman:
Okay, my proposal above regarding Travis is in place, so I think as long as this Travis build passes, we're good to merge. Are you happy @nomis52 ?

nomis52:
Sgtm

peternewman:
@cmuellner this is currently failing in Travis due to mismatch of Java Protobuf and Protobuf versions, see https://travis-ci.org/OpenLightingProject/ola/jobs/314603586#L2314

We pass in a protobufGeneratedSource variable currently, so doing something similar for Protobuf version ought to do the trick, but ideally dynamically populated based on the detected C++ version.

peternewman:
From a quick look, we can't find a module's version using existing pkg-config macros, just by calling pkg-config directly. I suspect just doing something like https://github.com/OpenLightingProject/ola/blob/master/configure.ac#L79 and passing in 3.2 if libprotobuf > 3.2 ought to work initially.

mcallegari:
Hey guys, any chance to see this merged into a release anytime soon ?
I can't build QLC+ on OBS on Fedora 26/27 and OpenSUSE Tumbleweed as OLA fails to build cause of protobuf.
Thanks

peternewman:
Yep @mcallegari , I just want to fix the Java build so Travis passes before we merge this in and release.

mcallegari:
@peternewman ok cool, thanks :+1:

peternewman:
Okay I've got the Java bit working now, with some rather hacky direct pkg-config calls so I think I'm happy with this.

Do you want to review please @nomis52 then we can merge, get a few more people to test and then release 0.11.0!

mcallegari:
@peternewman sorry to reiterate on an issue closed 10 months ago, but QLC+ builds on OpenSUSE Tumbleweed are still failing cause of protobuf. Is there a OLA release planned with this fix included ?
At this point, a 0.10.x or the long waited 0.11 will do. Thanks.

peternewman:
Hi @mcallegari 

My comments from here still stand, it will be in 0.11:
https://github.com/OpenLightingProject/ola/issues/1192#issuecomment-384379273

In the meantime, there's a patch file here you can apply to fix the package build (I assume OpenSUSE supports patches):
https://github.com/peternewman/macports-ports/blob/59d235f0ee80f36b6fa5017ce5932bb90306a22f/net/ola/files/patch-protobuf3.diff

I'd consider doing it in 0.10.x, but I don't feel it's had enough testing, and I've had close to zero feedback when asking people to test the upcoming release:
https://groups.google.com/forum/#!topic/open-lighting/GpgyUwM5X_g

mcallegari:
@peternewman thanks, the patch worked! After some autoreconf magic all my Qt5 builds on OBS can now be built.

PR #1343, opened 2017-12-17
 Increment the transaction number for each RDM packet

### LINKED ISSUE #189: ArtNet RDM does not increment transaction number for each command
_From [si...@nomis52.net](https://code.google.com/u/100927661077854655312/) on June 09, 2012 22:21:16_

Like the title says, the TN isn't incremented and is always set to 0.

_Original issue: http://code.google.com/p/open-lighting/issues/detail?id=188_

peternewman:
Closes #189


peternewman:
The Robe implementation should probably be moved to the common code too for consistency.

PR #1353, opened 2018-01-04
Fix more spelling errors

peternewman:
Use the latest codespell and improve the regex to help find them

### DIFF
--- a/common/network/TCPConnectorTest.cpp
+++ b/common/network/TCPConnectorTest.cpp
@@ -91,7 +91,7 @@ class TCPConnectorTest: public CppUnit::TestFixture {
     SelectServer *m_ss;
     IPV4Address m_localhost;
     ola::SingleUseCallback0<void> *m_timeout_closure;
-    unsigned int m_sucessfull_calls;
+    unsigned int m_sucessful_calls;
     unsigned int m_failure_calls;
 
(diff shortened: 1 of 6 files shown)


common/network/TCPConnectorTest.cpp
```diff
@@ -91,7 +91,7 @@ class TCPConnectorTest: public CppUnit::TestFixture {
     SelectServer *m_ss;
     IPV4Address m_localhost;
     ola::SingleUseCallback0<void> *m_timeout_closure;
-    unsigned int m_sucessfull_calls;
+    unsigned int m_sucessful_calls;
```
FloEdelmann: shouldn't it be `m_successful_calls`, i.e. 2 `c`s?

FloEdelmann: Ah, sorry, just seen #1355.

peternewman: Well spotted @FloEdelmann . I think when https://github.com/OpenLightingProject/ola/pull/1360 is in, I'll have found all variants of suc+es+ful+ , some fixed in master, and some in 0.10. It now finds words within underscores, so should catch that sort of thing in future. CamelCase and camelCase still need resolving if anyone fancies a python regex challenge!

PR #1357, opened 2018-01-04
Fix various stuff flagged up by lintian

peternewman:
@yoe would you mind taking a look at this lot in particular, I've just based it on the documentation mostly.

### DIFF
--- a/common/rdm/DiscoveryAgent.cpp
+++ b/common/rdm/DiscoveryAgent.cpp
@@ -371,7 +371,7 @@ void DiscoveryAgent::HandleCollision() {
   UID mid_uid(mid >> 32, mid);
   mid++;
   UID mid_plus_one_uid(mid >> 32, mid);
-  OLA_INFO << "Collision, spliting into: " << lower_uid << " - " << mid_uid
+  OLA_INFO << "Collision, splitting into: " << lower_uid << " - " << mid_uid
            << " , " << mid_plus_one_uid << " - " << upper_uid;
 
   range->uids_discovered = 0;
--- a/debian/copyright
+++ b/debian/copyright
@@ -33,7 +33,7 @@ following license:
  */
 
 On Debian GNU/Linux systems, the text of the LGPL can be found under
-/usr/share/common-licenses/LGPL-2.1
+'/usr/share/common-licenses/LGPL-2.1'
 
 The files under the java directory are licensed under the following
 license:
@@ -55,7 +55,7 @@ license:
  *************************************************************************/
 
 On Debian GNU/Linux, the text of the Apache 2 license can be found in
-/usr/share/common-licenses/Apache-2.0
+'/usr/share/common-licenses/Apache-2.0'
 
 Everything else is licensed under the following license:
 
@@ -76,4 +76,4 @@ The Debian packaging is (C) 2010, Simon Newton <nomis52@gmail.com> and
 is also licensed under the GPL.
 
 On Debian GNU/Linux, the full text of the GPL can be found in
-'/usr/share/common-licenses/GPL'.
+'/usr/share/common-licenses/GPL-2'.
--- a/debian/tests/unit-tests
+++ b/debian/tests/unit-tests
@@ -1,3 +1,0 @@
-#!/bin/sh
-
-make check || (cat test-suite.log && false)
--- a/debian/tests/unit-tests
+++ b/debian/tests/unit-tests
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+make check || (cat test-suite.log && false)
--- a/debian/Makefile.mk
+++ b/debian/Makefile.mk
@@ -31,5 +31,4 @@ EXTRA_DIST += \
     debian/source/lintian-overrides \
     debian/source/local-options \
     debian/tests/control \
-    debian/tests/unit-tests \
     debian/watch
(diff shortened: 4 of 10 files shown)


debian/tests/unit-tests
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+make check || (cat test-suite.log && false)
```
yoe: we run this at package build time already.

The autopkgtest tests are meant to test the *installed* binaries, not the built ones. They could do stuff like "make sure the init script does what it's supposed to do" etc, but shouldn't run the functional test suite.

I don't think this kind of autopkgtest is of any help, so I'd prefer against it.

debian/copyright
```diff
@@ -33,7 +33,7 @@ following license:
  */
 
 On Debian GNU/Linux systems, the text of the LGPL can be found under
-/usr/share/common-licenses/LGPL-2.1
+'/usr/share/common-licenses/LGPL-2.1'
```
yoe: not sure this helps much (this file is not using the machine-parseable format of pain), but I suppose it doesn't do harm.

common/rdm/DiscoveryAgent.cpp
```diff
@@ -371,7 +371,7 @@ void DiscoveryAgent::HandleCollision() {
   UID mid_uid(mid >> 32, mid);
   mid++;
   UID mid_plus_one_uid(mid >> 32, mid);
-  OLA_INFO << "Collision, spliting into: " << lower_uid << " - " << mid_uid
+  OLA_INFO << "Collision, splitting into: " << lower_uid << " - " << mid_uid
```
yoe: This isn't Debian-specific, is it? ;-)

might want to add that separate to these series.

peternewman: Well it was found via https://lintian.debian.org/tags/spelling-error-in-binary.html and confusingly missed via spellintian (I guess because it's in a newer dictionary than I'm using) and codespell.

debian/copyright
peternewman: Well it's generating https://lintian.debian.org/tags/copyright-refers-to-symlink-license.html as we're supposedly pointing to usr/share/common-licenses/GPL. That's the only reference I could see to GPL$. Actually I guess you're referring to the random addition of extra ' by me. I did wonder, but given we've got them on the last line...

Is it worth us trying to move to the machine-readable format at some point?

debian/tests/unit-tests
peternewman: Ah that's not very well explained here https://lintian.debian.org/tags/testsuite-autopkgtest-missing.html .

Do you mean like checking command line arguments are spelt correctly/to match the code in the init script? :tongue: 

We could do something like this @yoe https://anonscm.debian.org/viewvc/python-apps/packages/vdirsyncer/trunk/debian/tests/control?revision=13541&view=markup or is even that not really worth it? Is it worth putting it in as a starting point.

debian/copyright
yoe: the GPL$ is indeed something else, which you (correctly) fixed. The addition of ' seems somewhat superfluous, but it doesn't do any harm.

I'm not convinced of the benefit of the machine-parseable copyright files. They require a *lot* of work for little gain. To me, copyright files are a necessary evil for uploading anything into Debian, I'd rather not spend too much time on them ;-)

debian/tests/unit-tests
yoe: That could work as a first attempt, yes.

What would make somewhat more sense is to have an autopkgtest that starts the daemon and then tries to run some things through command line utilities or the browser (e.g., using selenium or some such) to make sure that once installed, the packages work as designed.

That'd be quite some work to set up though.

debian/copyright
peternewman: Ah okay, I think I just added them because the last one has them and I like consistency. I'm happy to remove the ' from all of the mentions of copyright files if you think that's more appropriate?

Okay, I'm happy to skip the machine option until Lintian starts moaning then!

debian/tests/unit-tests
peternewman: Yeah, well I guess the ultimate, and something I've always wanted to do on our normal CI would be to run up olad, then run the RDM tests against the dummy responders (or even run the RDM test server, and run the suite via the webpage).

But I've no idea how you describe that sort of inter-package dependency and test requirements, and where you'd script what...

yoe: That sounds like something awesome to look into, and shouldn't need as much time (since much of the code has already been written, as opposed to the selenium stuff).

Re: package interdependencies: that should be simple, the autopkgtest spec allows you to declare dependencies needed for the test suite. The test system will ensure that the packages are installed before running the tests. Your script may then check that the right daemons are running (or fiddle with them until they are) and that's it.

peternewman: Okay, I've done the boring simple thing for now, based on the one I linked to. We can look at extending the RDM test stuff when we get it working in Travis first I guess.

peternewman:
@yoe can you check you're happy with my changes based on your comments before I merge this?

peternewman:
Agreed, it'll be nice to see it actually running as intended for starters.

peternewman:
To confirm, it is at least working as intended, and passing on Debian CI:
https://ci.debian.net/packages/o/ola/

Issue #871, opened 2015-08-03
Switch to using PIXEL_TYPE & PIXEL_COUNT for the SPI plugin


nomis52:

peternewman:
http://rdm.openlighting.org/pid/display?manufacturer=31344&pid=32773
http://rdm.openlighting.org/pid/display?manufacturer=31344&pid=32774

peternewman:
We need to check the RDM standard, and the forum to see what is said about changing the list of personalities on a fixture while keeping the same device model ID.

PR #1358, opened 2018-01-05
added APA102 with Pixel Brightness mode

### LINKED ISSUE #1356: SPI Plugin: APA102 missing 'pixel brightness'
that is a long standing feature request.
some background information:
https://groups.google.com/d/msg/open-lighting/m-op0ym8rcg/I7Q3H6TrCWYJ

the APA102 Pixel LEDs have a 'pixel-brightness' (in the datasheet called global brightness) setting.
currently this is hardcoded as 0xFF.

feature request is to add a additional personality that enables the pixel-brightness to be controlled.

its a 5bit value.
In the groups discussion it was clear that this should be handled with a nice mapping from 0..255 to 0..31
so with this every pixel gets 4 dmx slots:

1. pixel-brightness
1. first color (red)
1. second color (green)
1. third color (blue)

s-light:
this fixes #1356 
make check passed all tests. 
hardware test i to be done. i will report back if i have it tested in real live :-)


plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +585,85 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
 
+
+void SPIOutput::IndividualAPA102ControlPixelBrightness(
+  const DmxBuffer &buffer
+) {
+  // some detailed information on the protocol:
+  // https://cpldcpu.wordpress.com/2014/11/30/understanding-the-apa102-superled/
+  // Data-Struct
+  // StartFrame: 4 bytes = 32 bits zeros (APA102_START_FRAME_BYTES)
+  // LEDFrame: 1 byte FF ; 3 bytes color info (Blue, Green, Red)
```
peternewman: This comment ought to have details on the brightness, rather than FF

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +585,85 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+    // set pixel data
+    // only write pixel data if buffer has complete data for this pixel:
+    if ((buffer.Size() - offset) >= APA102PB_SLOTS_PER_PIXEL) {
+      // first Byte contains:
+      // 3 bits start mark (111) + 5 bits pixel brightness
+      // (datasheet name: global brightness)
+      uint8_t pixel_brightness = CalculateAPA102PixelBrightness(
+        buffer.Get(offset + 0));
+      // merge 3 bits start mark (111): can be written as 0xE0
+      uint8_t first_byte = 0xE0 && pixel_brightness;
```
peternewman: Can we make 0xE0 a constant please

peternewman: You could avoid storing pixel_brightness and just && it straight in.

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +585,85 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+      // set first_byte to output buffer
+      output[spi_offset + 0] = first_byte;
```
peternewman: Likewise for first_byte.

plugins/spi/SPIOutput.cpp
```diff
@@ -653,6 +738,19 @@ uint8_t SPIOutput::CalculateAPA102LatchBytes(uint16_t pixel_count) {
...
+ * Calculate Pixel Brightness for APA102:
+ * Map Input to Output range:
+ * Input is 8bit value (0..255)
+ * Output is 5bit value (0..31)
+ */
+uint8_t SPIOutput::CalculateAPA102PixelBrightness(uint8_t brightness) {
+  // brightness == 255
+  // x == 31
+  // x = brightness * 31 / 255
+  return brightness * 31 / 255;
```
peternewman: This can just be done as (brightness >> 3) and discard the lower order bits.

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +739,233 @@ void SPIOutputTest::testCombinedAPA102Control() {
   // check if the output writes are 1
   OLA_ASSERT_EQ(1u, backend.Writes(1));
 }
+
+
+/**
+ * Test DMX writes in the individual APA102 Pixel Brightness mode.
+ */
+void SPIOutputTest::testIndividualAPA102ControlPixelBrightness() {
+  // personality 7= Individual APA102
```
peternewman: This should be personality 9, 7 is individual no PB.

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +739,233 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  const uint16_t this_test_personality = 7;
```
peternewman: Again

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +739,233 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  // generate dmx data
+  buffer.SetFromString(
+        std::string("  0,0,0,0,   9,0,0,0,  16,0,0,0,  42,0,0,0,  84,0,0,0,") +
+                    "127,0,0,0, 172,0,0,0, 206,0,0,0, 247,0,0,0, 254,0,0,0," +
+                    "255,0,0,0, 255,0,0,0, 255,0,0,0, 255,0,0,0, 255,0,0,0," +
+                    "255,0,0,0, 255,0,0,0");
+  output3.WriteDMX(buffer);
+  data = backend.GetData(0, &length);
+  const uint8_t EXPECTED10[] = { 0, 0, 0, 0,
+                                0xE0 && 0, 0, 0, 0,  // Pixel 1
```
peternewman: This test can't be working with the selected personality, or there's a bigger issue.

peternewman: Add that issue is you've not added it via the CPPUNIT_TEST macro at the top of the file, so it's not actually being run, hence the tests are still passing.

plugins/spi/SPIOutputTest.cpp
peternewman: Thinking about trying to fix this, I don't know what @nomis52 thinks, but to me it would be good if we had enums for the personalties, as then this problem, and the switch case to decide what mode we're in can be neater.

If we list the enums, then we can do
personalities.insert(personalities.begin() + PERS_WS2801_INDIVIDUAL, Personality("WS2801 Individual"));

Then switch/run tests based on PERS_WS2801_INDIVIDUAL.

Or we could build an array and convert it to a vector.

peternewman:
You should also increment the RDM software version when you add a personality (the 4 in GetDeviceInfo) in SPIOutput.cpp.

plugins/spi/SPIOutput.cpp
peternewman: Also as Codacy put it "Condition '224' is always true", it should be a single ampersand for bitwise AND. I'm kind of surprised nothing else picked this up!

s-light:
thanks @peternewman for your comments - i found today that i did forget to include the test.... ;-)
the && is i think coming from a mix of languages..
and you are right the test is not passing at all.....
i will work this over - but don't know if i get to do it the next days - iam now on vacation for a week - have to see..

plugins/spi/SPIOutputTest.cpp
peternewman: Also && here.

peternewman:
Which one uses && for bitwise AND then, as opposed to logical, or does it use the same for both?

The test isn't passing, because as per one of my other comments, you're using the wrong personality, so you're not generating data via your code, but trying to test to match it.

s-light:
now `make check` is successful :-)
tomorrow i will have a look at the `APA102_LEDFRAME_START_MARK`declaration.
and if this is fixed i will have a try at your [recommendation regarding the  enum for the personalities](https://github.com/OpenLightingProject/ola/pull/1358#discussion_r159878140)

s-light:
switched to use of enum for Personalities :memo: 
`make check` succeeded for me.
@peternewman @nomis52 please check the enum part if this is correctly done.

TODO: real-live hardware test

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +184,41 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+      "P9813 Combined Control"));
+  personalities.insert(personalities.begin() + PERS_APA102_INDIVIDUAL - 1,
+    Personality(m_pixel_count * APA102_SLOTS_PER_PIXEL,
+      "APA102 Individual Control"));
+  personalities.insert(personalities.begin() + PERS_APA102_COMBINED - 1,
+    Personality(APA102_SLOTS_PER_PIXEL,
+      "APA102 Combined Control"));
+  personalities.insert(personalities.begin() + PERS_APA102PB_INDIVIDUAL - 1,
+    Personality(m_pixel_count * APA102PB_SLOTS_PER_PIXEL,
+      "APA102 with pixel brightness Individual Control"));
```
peternewman: This is too long, the personality description is only 32 characters.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +184,41 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  personalities.insert(personalities.begin() + PERS_APA102PB_COMBINED - 1,
+    Personality(m_pixel_count * APA102PB_SLOTS_PER_PIXEL,
+      "APA102 with pixel brightness Combined Control"));
```
peternewman: Again

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +605,83 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+  if (m_output_number == 0) {
+    // set APA102_START_FRAME_BYTES to zero
+    memset(output, 0, APA102_START_FRAME_BYTES);
+  }
+
+  for (uint16_t i = 0; i < m_pixel_count; i++) {
+    // Convert RGB to APA102 Pixel
+    uint16_t offset = first_slot + (i * APA102PB_SLOTS_PER_PIXEL);
+
+
```
peternewman: Remove one of these blank lines

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +605,83 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+      // start of frame delimiter
+      spi_offset += APA102_START_FRAME_BYTES;
+    }
+    // set pixel data
+    // only write pixel data if buffer has complete data for this pixel:
+    if ((buffer.Size() - offset) >= APA102PB_SLOTS_PER_PIXEL) {
+      // first Byte:
+      // 3 bits start mark (111) (APA102_LEDFRAME_START_MARK) +
+      // 5 bits pixel brightness (datasheet name: global brightness)
+      output[spi_offset + 0] = SPIOutput::APA102_LEDFRAME_START_MARK +
```
peternewman: This ought to be | for bitwise maths.

plugins/spi/SPIOutput.h
```diff
@@ -38,6 +38,22 @@ namespace spi {
 
 class SPIOutput: public ola::rdm::DiscoverableRDMControllerInterface {
  public:
+  // definitions for all SPI Personalities
+  // this hardcoded mapping is needed for backwards compatibility
+  // enum class SPI_PERSONALITY : unsigned int {
+  enum SPI_PERSONALITY {
+    PERS_WS2801_INDIVIDUAL = 1,
+    PERS_WS2801_COMBINED = 2,
```
peternewman: I believe if you specify the value for the first item, the rest will increment afterwards, so not hardcoding the other values and just adding a note to keep this order may be clearer?

plugins/spi/SPIOutput.h
```diff
@@ -38,6 +38,22 @@ namespace spi {
...
+    PERS_LDP8806_INDIVIDUAL = 3,
+    PERS_LDP8806_COMBINED = 4,
+    PERS_P9813_INDIVIDUAL = 5,
+    PERS_P9813_COMBINED = 6,
+    PERS_APA102_INDIVIDUAL = 7,
+    PERS_APA102_COMBINED = 8,
+    PERS_APA102PB_INDIVIDUAL = 9,
```
peternewman: These two don't need it anyway, due to no backwards compatibility.

Also I wonder if PERS_APA102_PB_INDIVIDUAL would be clearer, as the LED type is still the same.

plugins/spi/SPIOutput.h
```diff
@@ -38,6 +38,22 @@ namespace spi {
...
+    PERS_APA102PB_COMBINED = 10
```
peternewman: Likewise

plugins/spi/SPIOutputTest.cpp
```diff
@@ -83,30 +87,39 @@ void SPIOutputTest::setUp() {
...
   options.pixel_count = 32;
   SPIOutput output2(m_uid, &backend, options);
 
+  // check default constructor values
   OLA_ASSERT_EQ(
       string("Output 0, WS2801 Individual Control, 75 slots @ 1."
              " (707a:00000000)"),
       output1.Description());
   OLA_ASSERT_EQ(static_cast<uint16_t>(1), output1.GetStartAddress());
-  OLA_ASSERT_EQ(static_cast<uint8_t>(1), output1.GetPersonality());
```
peternewman: Perhaps have both checks here? At some point we should do a spot check that at least one of the personalities maps to it's old ID, and also that they're valid personality IDs.

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +751,388 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  const uint8_t EXPECTED3[] = { 0, 0, 0, 0,
+                                0xFF, 0x4E, 0x38, 0x22,
+                                0xFF, 0x1E, 0x14, 0x0A,
+                                0};
+  OLA_ASSERT_DATA_EQUALS(EXPECTED3, arraysize(EXPECTED3), data, length);
+  OLA_ASSERT_EQ(3u, backend.Writes(0));
+
+  // test4
+  // tests what happens if fewer then needed color information are received
+  buffer.SetFromString("7, 9");
```
peternewman: You ought to do this test with three channels, given you need four, and responding to three like other code would be a likely error.

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +751,388 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  // test8
+  // create new output with pixel_count=16 and check data length
+  // setup pixel_count to 16
+  options.pixel_count = 16;
+  // setup SPIOutput
+  SPIOutput output2(m_uid, &backend, options);
+  // set personality
+  output2.SetPersonality(this_test_personality);
+  buffer.SetFromString(
+        std::string("255,0,0,0, 255,0,0,0, 255,0,0,0, 255,0,0,0, 255,0,0,0,") +
```
peternewman: You should be able to drop the bracket here, and all the + I believe.

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +751,388 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  // generate dmx data
+  buffer.SetFromString(
+        std::string("  0,0,0,0,   8,0,0,0,  16,0,0,0,  42,0,0,0,  84,0,0,0,") +
+                    "127,0,0,0, 167,0,0,0, 206,0,0,0, 240,0,0,0, 247,0,0,0," +
+                    "248,0,0,0, 255,0,0,0, 255,0,0,0, 255,0,0,0, 255,0,0,0," +
+                    "255,0,0,0, 255,0,0,0");
+  output3.WriteDMX(buffer);
+  data = backend.GetData(0, &length);
+  const uint8_t EXPECTED10[] = { 0, 0, 0, 0,
+                                0xE0 +  0, 0, 0, 0,  // Pixel 1
```
peternewman: | for bitwise again

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +751,388 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  const uint8_t EXPECTED3[] = { 0, 0, 0, 0,
+                                0xFF, 0x4E, 0x38, 0x22,
+                                0xFF, 0x4E, 0x38, 0x22,
+                                0};
+  OLA_ASSERT_DATA_EQUALS(EXPECTED3, arraysize(EXPECTED3), data, length);
+  OLA_ASSERT_EQ(3u, backend.Writes(0));
+
+  // test4
+  // tests what happens if fewer then needed color information are received
+  buffer.SetFromString("7, 9");
```
peternewman: Again

plugins/spi/SPIOutputTest.cpp
```diff
@@ -738,3 +751,388 @@ void SPIOutputTest::testCombinedAPA102Control() {
...
+  options.pixel_count = 20;
+  // setup SPIOutput
+  SPIOutput output2(m_uid, &backend, options);
+  // set personality
+  output2.SetPersonality(this_test_personality);
+  buffer.SetFromString("127,0,0,0");
+  output2.WriteDMX(buffer);
+  data = backend.GetData(0, &length);
+  const uint8_t EXPECTED8[] = { 0, 0, 0, 0,
+                                0xE0 + 15, 0, 0, 0,  // Pixel 1
```
peternewman: Again |

plugins/spi/SPIOutput.cpp
```diff
@@ -74,6 +74,8 @@ using std::min;
 using std::string;
 using std::vector;
 
+
```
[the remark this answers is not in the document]
peternewman: Heh, I meant the blank line, but that's good, let's ditch the using's too if we're not using them!

plugins/spi/SPIOutput.cpp
```diff
@@ -208,12 +206,12 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
       "APA102 Combined Control"));
-  personalities.insert(personalities.begin() + PERS_APA102PB_INDIVIDUAL - 1,
-    Personality(m_pixel_count * APA102PB_SLOTS_PER_PIXEL,
-      "APA102 with pixel brightness Individual Control"));
-  personalities.insert(personalities.begin() + PERS_APA102PB_COMBINED - 1,
-    Personality(m_pixel_count * APA102PB_SLOTS_PER_PIXEL,
-      "APA102 with pixel brightness Combined Control"));
+  personalities.insert(personalities.begin() + PERS_APA102_PB_INDIVIDUAL - 1,
+    Personality(m_pixel_count * APA102_PB_SLOTS_PER_PIXEL,
+      "APA102 PB Individual Control"));
```
peternewman: I wonder if something less cryptic might help, or is PB a regularly used term for the APA102? Perhaps "APA102 Pixel Bright Individ Ctrl".

As an aside, if you fancy another little task, we ought to add a static assert or something when the personality string is too long, so we catch it at compile time.

plugins/spi/SPIOutput.cpp
```diff
@@ -208,12 +206,12 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  personalities.insert(personalities.begin() + PERS_APA102_PB_COMBINED - 1,
+    Personality(m_pixel_count * APA102_PB_SLOTS_PER_PIXEL,
+      "APA102 PB Combined Control"));
```
peternewman: Perhaps "APA102 Pixel Bright Combine Ctrl", but I'm open to other suggestions.

plugins/spi/SPIOutput.cpp
```diff
@@ -237,11 +235,11 @@ SPIOutput::~SPIOutput() {
 }
 
 
-string SPIOutput::GetDeviceLabel() const {
+std::string SPIOutput::GetDeviceLabel() const {
```
peternewman: Yeah sorry, let's put back the using, apologies for the confusion.

plugins/spi/SPIOutput.cpp
```diff
@@ -843,7 +840,7 @@ RDMResponse *SPIOutput::GetDeviceInfo(const RDMRequest *request) {
 RDMResponse *SPIOutput::GetProductDetailList(const RDMRequest *request) {
   // Shortcut for only one item in the vector
   return ResponderHelper::GetProductDetailList(request,
-    vector<ola::rdm::rdm_product_detail>
+    std::vector<ola::rdm::rdm_product_detail>
```
peternewman: Likewise

plugins/spi/SPIOutput.h
```diff
@@ -39,7 +39,9 @@ namespace spi {
 class SPIOutput: public ola::rdm::DiscoverableRDMControllerInterface {
  public:
   // definitions for all SPI Personalities
-  // this hardcoded mapping is needed for backwards compatibility
+  // keep this order - it is important for backwards compatibility!
+  // (the integer representation is used to store the configuration in files.)
```
peternewman: Perhaps also add a note that it's friendly for users if the RDM personality IDs don't change too (well aside from adding more).

peternewman:
Again if you're feeling really keen, using the SlotDataCollection variant of the Personality constructor might be helpful, at least for your combined control, to indicate what the channel order is (IRGB, rather than RGBI say). I suspect populating it for all slots in individual may not make sense, we could do the first pixel or two, but that might confuse desks that pay attention to that data.

plugins/spi/SPIOutput.cpp
s-light: i don't think the Pixel Brightness (in the datasheet 'Global Brightness') is regularly used at all....
and i support you in that PB is cryptic ;-)
it would break the naming convention used currently - but i don't know if the 'Control' part is helpfull at all?!

iam open to do the 'static assert thing' but i don't know how to do something like this - can you give me some keywords i can search for?

s-light: what about 'APA102 Pixel Brightness Individ.' ?

plugins/spi/SPIOutput.cpp
s-light: and 'APA102 Pixel Brightness Combined'

plugins/spi/SPIOutput.cpp
s-light: no problem ;-) i change it back.

peternewman: Thanks

plugins/spi/SPIOutput.cpp
peternewman: Here's an example of a static assert https://github.com/OpenLightingProject/ola/blob/7a96311d03b656fc53b36283d4460774706d3bc3/common/rdm/ResponderHelper.cpp#L112 although @nomis52 may say it's a bad idea or have a better idea.

peternewman:
You'll also need to add the GetSlotInfo bits from here: https://github.com/OpenLightingProject/ola/blob/7a96311d03b656fc53b36283d4460774706d3bc3/common/rdm/DummyResponder.cpp#L288 and in the ParamHandler earlier.

So it actually responds to those queries.

s-light:
i think i have added all basic Slot things...
but i have no idea how to write a test case for this. can you give me a hint?
i thought that i can do something similar like the `output.Description()` test case.
but i then noticed that i need some rdm-request as argument for the `GetSlotInfo()` function.. [(SPIOutputTest.cpp L663)](https://github.com/OpenLightingProject/ola/pull/1358/commits/b18b207f21f6cbfc9071cf98c5c5a8ae80c8f0ae#diff-c6f31f4e9fc73de40a7ff2614396d1cdR663)

would it be possible to map the 'pixelcount' configuration to an rdm parameter?
(as fare as i know currently its only settable in the config file)

other - more general question regarding the SPIOutput files:
currently my feeling is that the three files are getting to long..
do you think there is an nice / easy to read / maintain / understand solution to split this up?
i would love to help at this - but think currently my coding knowledge is not good enough to design something like this in c++..
i thought of something like one file per led type - containing all the needed const and special helper functions for this type... (as static class or namespace) ?

PS: my hw setup is on the way back to working ;-) so i think i can test this in real live this week.

peternewman:
I wouldn't particularly bother about writing a test for the SlotType stuff. Although you could use the PersonalityManager to just check you've got the right number of SlotData entries for each fixture. The rest should be covered in the ResponderHelper test code. Also we've got the RDM test suite so there's no point rewriting that in C++ (but it would help to run your current code through it).

In terms of pixel count, and pixel type, see here (but it's probably worth either waiting until this PR is in, or starting a new branch based on this one, so the PR doesn't get too huge):
https://github.com/OpenLightingProject/ola/issues/871

A tidy up probably wants to wait for another PR too, I'm not really sure of the best way to do it, given the rather dynamic nature of the code with personalities, and the fact you need access to a lot of the core data. Perhaps the trick would be functions which get passed a DmxBuffer, a pointer to an array and a start address, and write that buffer into that array. Then leave the more RDM/DMX based stuff in the existing files. See e.g. include/ola/io/Serial.h for a possible idea.

s-light:
regarding the redesigns iam on your site thats nothing for this pr...

i added a experimental static_assert test - but it did not work -
i committed it  as a discussion base.

common/rdm/ResponderPersonality.cpp
```diff
@@ -40,6 +40,10 @@ using std::string;
 Personality::Personality(uint16_t footprint, const string &description)
     : m_footprint(footprint),
       m_description(description) {
+        STATIC_ASSERT(sizeof(description) <= 32);
+        // static_assert(
+        //   sizeof(description) <= 32,
+        //   "Personality Description is to long (must be <= 32 chars)");
 }
 
 Personality::Personality(uint16_t footprint, const string &description,
```
s-light: ok i missed the second constructor.
but also with the same line adde there it does not throw an error for the 'to long' description..
so it seems iam missing something else...

peternewman: You want description.length() to find out how many characters are in the string, sizeof tells you the size of the object (which I think will be a pointer in this case).

peternewman: Does "STATIC_ASSERT(description.length() <= 32);" not work then? Are you importing Macro.h for our STATIC_ASSERT macro?

peternewman: Does "STATIC_ASSERT(description.length() <= 32);" not work then? Are you importing Macro.h for our STATIC_ASSERT macro?

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
-                                      "P9813 Combined Control"));
-  personalities.push_back(Personality(m_pixel_count * APA102_SLOTS_PER_PIXEL,
-                                      "APA102 Individual Control"));
-  personalities.push_back(Personality(APA102_SLOTS_PER_PIXEL,
-                                      "APA102 Combined Control"));
+  // personality description is max 32 characters
+
+  personalities.insert(personalities.begin() + PERS_WS2801_INDIVIDUAL - 1,
+    Personality(m_pixel_count * WS2801_SLOTS_PER_PIXEL,
+      "WS2801 Individual Control"));
```
peternewman: Can we align the descriptions with the ( on the line above please.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
-  personalities.push_back(Personality(P9813_SLOTS_PER_PIXEL,
-                                      "P9813 Combined Control"));
-  personalities.push_back(Personality(m_pixel_count * APA102_SLOTS_PER_PIXEL,
-                                      "APA102 Individual Control"));
-  personalities.push_back(Personality(APA102_SLOTS_PER_PIXEL,
-                                      "APA102 Combined Control"));
+  // personality description is max 32 characters
+
+  personalities.insert(personalities.begin() + PERS_WS2801_INDIVIDUAL - 1,
+    Personality(m_pixel_count * WS2801_SLOTS_PER_PIXEL,
```
peternewman: Can we indent this to align with the ( above ideally, or otherwise four in from personalities depending on line length.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+    Personality(P9813_SLOTS_PER_PIXEL,
+      "P9813 Combined Control"));
+
+  personalities.insert(personalities.begin() + PERS_APA102_INDIVIDUAL - 1,
+    Personality(m_pixel_count * APA102_SLOTS_PER_PIXEL,
+      "APA102 Individual Control"));
+
+  ola::rdm::SlotDataCollection::SlotDataList sd_APA102_COMBINED;
+  sd_APA102_COMBINED.push_back(
+    ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
```
peternewman: Indent to four please.

plugins/spi/SPIOutput.cpp
```diff
@@ -375,7 +428,7 @@ void SPIOutput::IndividualLPD8806Control(const DmxBuffer &buffer) {
   if (!output)
     return;
 
-  const unsigned int length = std::min(m_pixel_count * LPD8806_SLOTS_PER_PIXEL,
+  const unsigned int length = min(m_pixel_count * LPD8806_SLOTS_PER_PIXEL,
```
peternewman: I think this wants to be back to std:: unless you've done a using.

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +632,82 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
   m_backend->Commit(m_output_number);
 }
 
+
+void SPIOutput::IndividualAPA102ControlPixelBrightness(
+  const DmxBuffer &buffer
```
peternewman: Indent by four please.

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +632,82 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+) {
```
peternewman: Put this on the line above please.

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +632,82 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+  // LEDFrame:
+  //    1 byte START_MARK + pixel brightness
+  //    3 bytes color info (Blue, Green, Red)
+  // EndFrame: (n/2)bits; n = pixel_count
+
+  // calculate DMX-start-address
+  const unsigned int first_slot = m_start_address - 1;  // 0 offset
+
+  // only do something if at least 1 pixel can be updated..
+  if (buffer.Size() - first_slot < APA102_PB_SLOTS_PER_PIXEL) {
```
peternewman: Can we have some brackets around the "buffer.Size() - first_slot" bit please.

plugins/spi/SPIOutput.cpp
```diff
@@ -579,6 +632,82 @@ void SPIOutput::IndividualAPA102Control(const DmxBuffer &buffer) {
...
+      spi_offset += APA102_START_FRAME_BYTES;
+    }
+    // set pixel data
+    // only write pixel data if buffer has complete data for this pixel:
+    if ((buffer.Size() - offset) >= APA102_PB_SLOTS_PER_PIXEL) {
+      // first Byte:
+      // 3 bits start mark (111) (APA102_LEDFRAME_START_MARK) +
+      // 5 bits pixel brightness (datasheet name: global brightness)
+      output[spi_offset + 0] = SPIOutput::APA102_LEDFRAME_START_MARK |
+        CalculateAPA102PixelBrightness(buffer.Get(offset + 0));
```
peternewman: Indent to four. Can we have brackets round the bitwise or operation please.

plugins/spi/SPIOutput.cpp
```diff
@@ -636,6 +765,68 @@ void SPIOutput::CombinedAPA102Control(const DmxBuffer &buffer) {
...
 }
 
+
+void SPIOutput::CombinedAPA102ControlPixelBrightness(const DmxBuffer &buffer) {
+  // for Protocol details see IndividualAPA102Control
+  // calculate DMX-start-address
+  const uint16_t first_slot = m_start_address - 1;  // 0 offset
+
+  // check if enough data is there.
+  if (buffer.Size() - first_slot < APA102_PB_SLOTS_PER_PIXEL) {
```
peternewman: Again

plugins/spi/SPIOutput.cpp
```diff
@@ -636,6 +765,68 @@ void SPIOutput::CombinedAPA102Control(const DmxBuffer &buffer) {
...
+    memset(output, 0, APA102_START_FRAME_BYTES);
+  }
+
+  // create Pixel Data
+  uint8_t pixel_data[APA102_SPI_BYTES_PER_PIXEL];
+  // first Byte:
+  // 3 bits start mark (111) (APA102_LEDFRAME_START_MARK) +
+  // 5 bits pixel brightness (datasheet name: global brightness)
+  pixel_data[0] = SPIOutput::APA102_LEDFRAME_START_MARK |
+    CalculateAPA102PixelBrightness(buffer.Get(first_slot + 0));
```
peternewman: Again

plugins/spi/SPIOutput.cpp
```diff
@@ -689,7 +892,9 @@ RDMResponse *SPIOutput::SetDeviceLabel(const RDMRequest *request) {
 }
 
 RDMResponse *SPIOutput::GetSoftwareVersionLabel(const RDMRequest *request) {
-  return ResponderHelper::GetString(request, string("OLA Version ") + VERSION);
```
peternewman: Did this need changing? Or is this an editor change, which might explain a lot of my other comments.

plugins/spi/SPIOutput.h
```diff
@@ -38,6 +38,28 @@ namespace spi {
...
  public:
+  // definitions for all SPI Personalities
+  // keep personality order
+  // - it is important for backwards compatibility
+  // the integer representation is used to store the configuration in files
+  // and RDM Personality IDs should also be stable.
+  // new ones can be added at end.
+  // remember to increment RDM-Version in
+  // SPIOutput.cpp SPIOutput::GetDeviceInfo()
+  // enum class SPI_PERSONALITY : unsigned int {
```
peternewman: Can we drop this comment line please.

plugins/spi/SPIOutput.cpp
s-light: i deleted the using and added the namespace things like `std::` first because i misinterpreted your comment... 26c5181d683ad78c1dec2e36adab56e41bdc2db1
then re-add it. ceea93d7825470ab69ead4c47d49068f18063e21
i don't know what you are preferring (and i don't know what is 'good style' here..)

common/rdm/ResponderPersonality.cpp
s-light: i did not `#include "ola/base/Macro.h"` but it seems it is included (this was not the problem except it is overwritting some other defaults) the error message if i uncomment this is:
```
common/rdm/ResponderPersonality.cpp:44:41: error: call to non-constexpr function ‘std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::length() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type = long unsigned int]’
         STATIC_ASSERT(description.length() <= 32);
                       ~~~~~~~~~~~~~~~~~~^~
./include/ola/base/Macro.h:125:66: note: in definition of macro ‘STATIC_ASSERT’
     sizeof(::internal::STATIC_ASSERT_FAILURE< static_cast<bool>( x ) >)>\
```
so if i interpret this correctly it tells me that `.length()`does return a `non-constexpr` and that STATIC_ASSERT needs a `constexpr` --that makes sense as the code is not run because the STATIC_ASSERT is evaluated at compile time.. - as far [as i found](https://stackoverflow.com/questions/27123306/is-it-possible-to-use-stdstring-in-a-constexpr#27123850) there is no way to get a `std:string` to be constexpr compatible

plugins/spi/SPIOutput.cpp
s-light: do you mean like this:
```
  personalities.insert(personalities.begin() + PERS_WS2801_INDIVIDUAL - 1,
                       Personality(m_pixel_count * WS2801_SLOTS_PER_PIXEL,
                                   "WS2801 Individual Control"));
that would fit the line length.
```

plugins/spi/SPIOutput.cpp
s-light: you mean so:
```
  sd_APA102_COMBINED.push_back(
      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
```
i thought that the indentation rule is 2 spaces.
or is it 2spaces - and on continuing lines 4spaces?
sorry to mess up this things!

plugins/spi/SPIOutput.cpp
s-light: both no - i did it (as i thought its easier to read this way - i will revoke it..)

common/rdm/ResponderPersonality.cpp
peternewman: Yes apologies, I did some testing afterwards and found the same thing, but failed to comment.

I suspect, but haven't double checked, that OLA will already truncate the returned string to within the RDM spec (i.e. 32 characters). So perhaps we should just compromise on adding an OLA_WARN or OLA_CRIT to the constructors, so we at least see a message in the log if someone does something they shouldn't. I did find some stuff about string literals, which may possibly mean we could do it with a macro, which eventually passes the literal into the std::string, but I'm not certain about that.

plugins/spi/SPIOutput.cpp
peternewman: The top one please, it's the two arguments to insert, then the two to Personality.

plugins/spi/SPIOutput.cpp
peternewman: Yeah, we indent by two (apart from public:/private:, which are one where we've bothered/in some cases), but four for a continuing line.

plugins/spi/SPIOutput.cpp
peternewman: I don't really mind, although compactness often helps given we don't actually care about that line. If it's going to be on multiple lines, we generally live the first argument with the function/method, then indent the rest to match, unless one of the lines is too long. But we normally squeeze a lot onto the line until we hit the line length.

plugins/spi/SPIOutput.h
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
peternewman: Yeah so both lines still need to go in by one please.

plugins/spi/SPIOutput.cpp
peternewman: I think the vague theory is use using if we've got a few hits, but probably not if we've only got one, as it pollutes the namespace a little, as well as making things marginally less clear. i.e. the eternal balance of being concise versus clear.

This file seems to have a "using std::auto_ptr;" and no other mention of auto_ptr, so that's a good start. But then we have vector once, but using. So we have no hard and fast rules, on the .cpp files anyway; it's more critical in the header files, although I can't remember exactly why off the top of my head.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
-  personalities.push_back(Personality(P9813_SLOTS_PER_PIXEL,
-                                      "P9813 Combined Control"));
-  personalities.push_back(Personality(m_pixel_count * APA102_SLOTS_PER_PIXEL,
-                                      "APA102 Individual Control"));
-  personalities.push_back(Personality(APA102_SLOTS_PER_PIXEL,
-                                      "APA102 Combined Control"));
+  // personality description is max 32 characters
+
+  personalities.insert(personalities.begin() + PERS_WS2801_INDIVIDUAL - 1,
+                      Personality(m_pixel_count * WS2801_SLOTS_PER_PIXEL,
```
peternewman: These still want indenting please, like so (i.e. add one space to the second and third lines):
```
personalities.insert(personalities.begin() + PERS_WS2801_INDIVIDUAL - 1,
                     Personality(m_pixel_count * WS2801_SLOTS_PER_PIXEL,
                                 "WS2801 Individual Control"));
```

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_GREEN, 0));
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
+  personalities.insert(personalities.begin() + PERS_APA102_COMBINED - 1,
+                      Personality(APA102_SLOTS_PER_PIXEL,
+                                  "APA102 Combined Control",
+      ola::rdm::SlotDataCollection(sd_APA102_COMBINED)));
```
peternewman: The line ola::rdm::SlotDataCollection(sd_APA102_COMBINED) can become sd_APA102_COMBINED.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_GREEN, 0));
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
+  personalities.insert(personalities.begin() + PERS_APA102_PB_COMBINED - 1,
+                      Personality(APA102_PB_SLOTS_PER_PIXEL,
+                                  "APA102 Pixel Brightness Combined",
+      ola::rdm::SlotDataCollection(sd_APA102_PB_COMBINED)));
```
peternewman: Again

plugins/spi/SPIOutputTest.cpp
```diff
@@ -642,6 +658,13 @@ void SPIOutputTest::testCombinedAPA102Control() {
   unsigned int length = 0;
   const uint8_t *data = NULL;
 
+  // test0
```
peternewman: Can we remove this please.

peternewman:
Do you think you've got time to do these last few minor style changes soon @s-light so we can get this merged in, as I suspect 0.11.0 may finally appear soon...

s-light:
i try to do it as soon as possible. (hopefully i get to it until Thursday)

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutputTest.cpp
peternewman: Done

s-light:
i try to do a real-live hardware test in the next days.
(i have done some minimal testing a few commits ago on real hardware - back than it worked..)

plugins/spi/SPIOutput.cpp
s-light: hm - i was to fast with the commit - 
for me this results in an error:
```
plugins/spi/SPIOutput.cpp: In constructor ‘ola::plugin::spi::SPIOutput::SPIOutput(const ola::rdm::UID&, ola::plugin::spi::SPIBackendInterface*, const ola::plugin::spi::SPIOutput::Options&)’:
plugins/spi/SPIOutput.cpp:225:53: error: no matching function for call to ‘ola::rdm::Personality::Personality(const uint16_t&, const char [24], ola::rdm::SlotDataCollection::SlotDataList&)’
                                   sd_APA102_COMBINED));
                                                     ^
In file included from ./include/ola/rdm/ResponderHelper.h:35:0,
                 from plugins/spi/SPIOutput.cpp:42:
./include/ola/rdm/ResponderPersonality.h:41:5: note: candidate: ola::rdm::Personality::Personality(uint16_t, const string&, const ola::rdm::SlotDataCollection&)
     Personality(uint16_t footprint, const std::string &description,
     ^~~~~~~~~~~
./include/ola/rdm/ResponderPersonality.h:41:5: note:   no known conversion for argument 3 from ‘ola::rdm::SlotDataCollection::SlotDataList {aka std::vector<ola::rdm::SlotData>}’ to ‘const ola::rdm::SlotDataCollection&’
./include/ola/rdm/ResponderPersonality.h:40:5: note: candidate: ola::rdm::Personality::Personality(uint16_t, const string&)
     Personality(uint16_t footprint, const std::string &description);
     ^~~~~~~~~~~
./include/ola/rdm/ResponderPersonality.h:40:5: note:   candidate expects 2 arguments, 3 provided
./include/ola/rdm/ResponderPersonality.h:38:7: note: candidate: ola::rdm::Personality::Personality(const ola::rdm::Personality&)
 class Personality {
       ^~~~~~~~~~~
./include/ola/rdm/ResponderPersonality.h:38:7: note:   candidate expects 1 argument, 3 provided
./include/ola/rdm/ResponderPersonality.h:38:7: note: candidate: ola::rdm::Personality::Personality(ola::rdm::Personality&&)
./include/ola/rdm/ResponderPersonality.h:38:7: note:   candidate expects 1 argument, 3 provided
plugins/spi/SPIOutput.cpp:243:57: error: no matching function for call to ‘ola::rdm::Personality::Personality(const uint16_t&, const char [33], ola::rdm::SlotDataCollection::SlotDataList&)’
                                    sd_APA102_PB_COMBINED));
                                                         ^
In file included from ./include/ola/rdm/ResponderHelper.h:35:0,
                 from plugins/spi/SPIOutput.cpp:42:
./include/ola/rdm/ResponderPersonality.h:41:5: note: candidate: ola::rdm::Personality::Personality(uint16_t, const string&, const ola::rdm::SlotDataCollection&)
     Personality(uint16_t footprint, const std::string &description,
     ^~~~~~~~~~~
./include/ola/rdm/ResponderPersonality.h:41:5: note:   no known conversion for argument 3 from ‘ola::rdm::SlotDataCollection::SlotDataList {aka std::vector<ola::rdm::SlotData>}’ to ‘const ola::rdm::SlotDataCollection&’
./include/ola/rdm/ResponderPersonality.h:40:5: note: candidate: ola::rdm::Personality::Personality(uint16_t, const string&)
     Personality(uint16_t footprint, const std::string &description);
     ^~~~~~~~~~~
./include/ola/rdm/ResponderPersonality.h:40:5: note:   candidate expects 2 arguments, 3 provided
./include/ola/rdm/ResponderPersonality.h:38:7: note: candidate: ola::rdm::Personality::Personality(const ola::rdm::Personality&)
 class Personality {
       ^~~~~~~~~~~
./include/ola/rdm/ResponderPersonality.h:38:7: note:   candidate expects 1 argument, 3 provided
./include/ola/rdm/ResponderPersonality.h:38:7: note: candidate: ola::rdm::Personality::Personality(ola::rdm::Personality&&)
./include/ola/rdm/ResponderPersonality.h:38:7: note:   candidate expects 1 argument, 3 provided
```

peternewman: Oops, my apologies, sorry about that. Perhaps we should have a Personality constructor that takes a vector directly?

s-light: seems like a good idea ;-)
i can try to look at this next week.
would it be enough to add a additional overloaded constructor and from this one use the cast and call the current one? (i have not looked at the code yet..)

peternewman: Yes I think so, but don't trust what I said given the earlier comment! :)

s-light:
real-world-test done :-)
http://blog.s-light.eu/ola-apa102-pixelbrightness-support/
(made a short video of a 144pixel strip controlled by two universes -
input both E1.31 (sACN) output first 100 leds with APA102 PB individ. rest 44 leds with APA102 combined. so i also tested the old and the new in combination.. (tested other combinations without video...)
for me it seems that it is working fine.

common/rdm/ResponderPersonality.cpp
```diff
@@ -40,13 +40,21 @@ using std::string;
 Personality::Personality(uint16_t footprint, const string &description)
     : m_footprint(footprint),
       m_description(description) {
+        // STATIC_ASSERT(description.length() <= 32);
```
peternewman: I'm assuming this line doesn't work at all, in which case can we remove it please?

common/rdm/ResponderPersonality.cpp
```diff
@@ -40,13 +40,21 @@ using std::string;
...
+        // c++11:
+        // static_assert(
```
peternewman: Does this work in C++11? If so I guess we can leave it as a comment, but perhaps make the C++11 line a TODO(C++11) or something.

common/rdm/ResponderPersonality.cpp
```diff
@@ -40,13 +40,21 @@ using std::string;
...
+        //   sizeof(description) <= 32,
+        //   "Personality Description is to long (must be <= 32 chars)");
 }
 
 Personality::Personality(uint16_t footprint, const string &description,
                          const SlotDataCollection &slot_data)
     : m_footprint(footprint),
       m_description(description),
       m_slot_data(slot_data) {
+        // constexpr int descriptionlen = description.length();
```
peternewman: Can we remove this too if it doesn't work.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+                                   "P9813 Individual Control"));
+  personalities.insert(personalities.begin() + PERS_P9813_COMBINED - 1,
+                      Personality(P9813_SLOTS_PER_PIXEL,
+                                  "P9813 Combined Control"));
+
+  personalities.insert(personalities.begin() + PERS_APA102_INDIVIDUAL - 1,
+                       Personality(m_pixel_count * APA102_SLOTS_PER_PIXEL,
+                                   "APA102 Individual Control"));
+
+  ola::rdm::SlotDataCollection::SlotDataList sd_APA102_COMBINED;
```
peternewman: I was going to say lower case some/all of this variable name, but given we've made it, why don't we stick it at the top and call it sd_rgb_combined, then we can apply it to all the RGB combined SPI pixels.

Which also means my suggestion about a Personality constructor taking a SlotDataList becomes redundant, as we want to convert it to a SlotDataCollection once, then use that throughout.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  personalities.insert(personalities.begin() + PERS_APA102_COMBINED - 1,
+                      Personality(APA102_SLOTS_PER_PIXEL,
+                                  "APA102 Combined Control",
+       ola::rdm::SlotDataCollection(sd_APA102_COMBINED)));
+
+  personalities.insert(personalities.begin() + PERS_APA102_PB_INDIVIDUAL - 1,
+                       Personality(m_pixel_count * APA102_PB_SLOTS_PER_PIXEL,
+                                   "APA102 Pixel Brightness Individ."));
+
+  ola::rdm::SlotDataCollection::SlotDataList sd_APA102_PB_COMBINED;
```
peternewman: Likewise for the irgb one.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_GREEN, 0));
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
+  personalities.insert(personalities.begin() + PERS_APA102_COMBINED - 1,
+                      Personality(APA102_SLOTS_PER_PIXEL,
+                                  "APA102 Combined Control",
+       ola::rdm::SlotDataCollection(sd_APA102_COMBINED)));
```
peternewman: This wants to align with the Personality( bracket, or be offset, so you can do:
```
  personalities.insert(personalities.begin() + PERS_APA102_COMBINED - 1,
                       Personality(
                           APA102_SLOTS_PER_PIXEL,
                           "APA102 Combined Control",
                           ola::rdm::SlotDataCollection(sd_APA102_COMBINED)));
```

Or:
```
  personalities.insert(
      personalities.begin() + PERS_APA102_COMBINED - 1,
      Personality(APA102_SLOTS_PER_PIXEL,
                  "APA102 Combined Control",
                  ola::rdm::SlotDataCollection(sd_APA102_COMBINED)));
```
Whichever you fancy, although the second looks nicer to me I think.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+                                   "APA102 Individual Control"));
+
+  ola::rdm::SlotDataCollection::SlotDataList sd_APA102_COMBINED;
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_GREEN, 0));
+  sd_APA102_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
+  personalities.insert(personalities.begin() + PERS_APA102_COMBINED - 1,
```
peternewman: As below, sorry I got the wrong order.

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +185,67 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  ola::rdm::SlotDataCollection::SlotDataList sd_APA102_PB_COMBINED;
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_INTENSITY, 0));
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_RED, 0));
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_GREEN, 0));
+  sd_APA102_PB_COMBINED.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
+  personalities.insert(personalities.begin() + PERS_APA102_PB_COMBINED - 1,
```
peternewman: Again.

plugins/spi/SPIOutput.cpp
peternewman: Sorry @s-light bump, can we remove the unused "using std::auto_ptr;" please (it's too far up the file for me to comment). I realise it's not your problem either!

plugins/spi/SPIOutput.cpp
```diff
@@ -187,60 +186,79 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
       ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
-  personalities.insert(personalities.begin() + PERS_APA102_PB_COMBINED - 1,
-                       Personality(APA102_PB_SLOTS_PER_PIXEL,
-                                   "APA102 Pixel Brightness Combined",
-      ola::rdm::SlotDataCollection(sd_APA102_PB_COMBINED)));
+  ola::rdm::SlotDataCollection sdc_irgb_combined;
+  sdc_irgb_combined = ola::rdm::SlotDataCollection(sd_irgb_combined);
+
+  personalities.insert(
+      personalities.begin() + PERS_WS2801_INDIVIDUAL - 1,
```
peternewman: You don't need to move this one down, but I understand you may want to make them all look the same.

common/rdm/ResponderPersonality.cpp
peternewman: Done.

common/rdm/ResponderPersonality.cpp
peternewman: Done.

common/rdm/ResponderPersonality.cpp
peternewman: Done.

plugins/spi/SPIOutput.cpp
s-light: you understand my intention ;-)
i have made the style-change in an extra commit - so i could easily revoke this..
i like the 'list-style' this way most - but i know you mentioned the rule is normally 'all that fits in the same line'...

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: Done

plugins/spi/SPIOutput.cpp
peternewman: That's a good enough reason for me! Rule is probably too strong, general style, but with enough exceptions you can do what you want. :smile:

plugins/spi/SPIOutput.cpp
```diff
@@ -180,26 +184,86 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
...
+  sd_rgb_combined.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_GREEN, 0));
+  sd_rgb_combined.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_COLOR_ADD_BLUE, 0));
+  ola::rdm::SlotDataCollection sdc_rgb_combined;
+  sdc_rgb_combined = ola::rdm::SlotDataCollection(sd_rgb_combined);
+
+  ola::rdm::SlotDataCollection::SlotDataList sd_irgb_combined;
+  sd_irgb_combined.push_back(
+      ola::rdm::SlotData::PrimarySlot(ola::rdm::SD_INTENSITY, 0));
```
peternewman: This probably wants to default to 100% aka 255 aka DMX_MAX_SLOT_VALUE, so by default it still just works like RGB mode, but with the option of using the overall intensity.

PR #1361, opened 2018-01-06
Release 0.10.6

peternewman:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+ola (0.10.6-1) unstable; urgency=low
+
+  * New upstream release
+
+ -- Simon Newton <nomis52@gmail.com>  Sat, 01 Jan 2018 23:41:00 +0000
```
nomis52: You should put your own name here.

peternewman: Then I'd have to think about scary things like which email to publish quite so widely. :smile: 

Also we both missed the fact I'd put the wrong date!

PR #1365, opened 2018-01-18
Fix for Python 3 error

pietrodn:
Just a small patch for an error in a place where Python 3 expected a byte-like object and instead got a `str`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,7 +11,8 @@ x/y/2018 ola-0.10.7
  Bugs:
  * Fix the build on Windows
  * Entire codebase now passes codespell testing
- * 
+ * Fix a Python 3 bug in the API
+ *
 
  Internal:
  *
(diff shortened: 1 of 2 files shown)


peternewman:
Thanks!

This doesn't seem to break Python 2.7, I assume you've tested and it improves Python 3 compatibility?

peternewman:
Actually @pietrodn would you mind targetting our 0.10 branch please (or I can cherry-pick after merge if getting it into 0.10 is a bit too much of a faff for you)?

pietrodn:
I tested on Python 3, it avoids a crash.
I will rebase the PR on top of 0.10, but should I open a new PR for that?

peternewman:
We don't need 0.10 and master PRs, so if your git fu is good enough to do it in this PR it's fine, or close this and open a fresh one.

peternewman:
Great thanks. I've made the updates on the PR, and forced a Travis build with a news update, if that all passes we should be good to go!

PR #1369, opened 2018-01-31
Add link to my Bachelor's Thesis in SPI DMX Plugin's README

FloEdelmann:
The online version of my [Bachelor's Thesis](http://mnm-team.org/pub/Fopras/edel17/PDF-Version/edel17.pdf) is finally available!

I've documented the whole development process of a PC-DMX interface using OLA, including my implementation of the SPI DMX Plugin and a corresponding electrical schematic. Thus, I think it will be a useful resource to add to the plugin's README.

Maybe you want to look into it @peternewman and @nomis52 :)

### DIFF
--- a/plugins/spidmx/README.md
+++ b/plugins/spidmx/README.md
@@ -13,6 +13,10 @@ set a larger block size if you want a higher refresh rate for the high
 channels (which results in more stuttering in all channels, as the time
 between two SPI read operation increases).
 
+A detailed description of the development process and the used techniques
+can be found in Flo Edelmann's Bachelors Thesis:
+http://mnm-team.org/pub/Fopras/edel17/PDF-Version/edel17.pdf
+
 
 ## Raspberry Pi configuration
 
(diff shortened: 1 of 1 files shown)


plugins/spidmx/README.md
```diff
@@ -13,6 +13,10 @@ set a larger block size if you want a higher refresh rate for the high
 channels (which results in more stuttering in all channels, as the time
 between two SPI read operation increases).
 
+A detailed description of the development process and the used techniques
```
peternewman: Nit, grammar wise, it should probably be the techniques used.

peternewman:
An interesting read too @FloEdelmann , thanks for that.

Another possible extension, which probably ought to be fairly simple given all the leg work you've done already would be to create another RDM/DMX sniffer like the logic one:
https://github.com/OpenLightingProject/ola/tree/master/tools/logic

Which essentially just tries to decode and print whatever data you've received. I guess missing the top of some frames may make it somewhat less helpful for RDM though...

FloEdelmann:
Good that the thesis itself was proof-read several times 😃 

Yeah, I already thought receiving RDM with this plugin. However, I came to the same conclusion as you did: For DMX, missing a packet is not too bad because it will be sent again later. RDM packets have to be received correctly at the first try though, which is quite unlikely (given that the whole packet needs to fit into one chunk).

So I think it won't be too useful.

PR #1399, opened 2018-04-11
SPI Plugin - TI TLC5971 support

s-light:
this pull request adds support for the [TI TLC5971 12Ch 16Bit LED-Driver Chips](http://www.ti.com/product/TLC5971) in the SPI-Plugin.

its basically working - 
as fare as i have seen if a configuration works it will on every start.
but sometimes / some configurations leads to random `malloc(): memory corruption` and `Received Segmentation fault` crashes.  
That is not good :beetle: 
currently i think it has to do with amount of pixels/ports in use - but have no evident for this yet.

iam currently out of ideas how to start tracking down the course of this - 
and hope one of you can give me a idea how to start debugging this. -
it definite has to do somehow with my code... and only shows if i use the plugin with the new pixel type.


plugins/spi/SPIOutput.cpp
```diff
@@ -261,6 +297,10 @@ SPIOutput::SPIOutput(const UID &uid, SPIBackendInterface *backend,
                   "APA102 Pixel Brightness Combined",
                   sdc_irgb_combined));
 
+  personalities.insert(personalities.begin() + PERS_TLC5971_INDIVIDUAL - 1,
+    Personality(m_pixel_count * TLC5971_SLOTS_PER_DEVICE,
+      "TLC5971 Individual Control (16bit per channel)"));
```
peternewman: We probably ideally want both 8 and 16bit individual and combined options eventually, but feel free to fix the underlying issues first.

plugins/spi/SPIOutput.cpp
```diff
@@ -874,6 +917,239 @@ uint8_t SPIOutput::CalculateAPA102PixelBrightness(uint8_t brightness) {
...
+  //    GSG1 16bit;
+  //    GSR1 16bit;
+  //    GSB0 16bit;
+  //    GSG0 16bit;
+  //    GSR0 16bit;
+  // Device Nth-1 (244Bit = 28Byte)
+  // Device ..
+  // Device 2
+  // Device 1
+  // short brake of 8x period of clock (666ns .. 2.74ms) to generate latchpulse
```
peternewman: SPaG break

plugins/spi/SPIOutput.h
```diff
@@ -214,6 +217,91 @@ class SPIOutput: public ola::rdm::DiscoverableRDMControllerInterface {
...
+  // // byte border ------------------------------------------
+  // static const uint8_t TLC5971_PACKET_CONFIG_LSHIFT_BCB_LS = 6;
+  // static const uint8_t TLC5971_PACKET_CONFIG_LSHIFT_BCG_RS = 1;
+  // // byte border ------------------------------------------
+  // static const uint8_t TLC5971_PACKET_CONFIG_LSHIFT_BCG_LS = 7;
+  // static const uint8_t TLC5971_PACKET_CONFIG_LSHIFT_BCR = 0;
+
+  union TLC5971_packet_gsdata_t {
+    uint8_t bytes[24];
+    // the uint16_t will not work everywhere because of endianess problems..
```
peternewman: SPaG: endianness

peternewman:
Can you give us some working and broken configurations to compare please?

s-light:
i test some configurations for comparison this  evening and write up what i find. 
last time i just had to enable 4 ports 
(all @ 16 'pixels' --> every pixel in this type has 12 LEDs @ 16bit - so 24Channels as that is the smallest amount of data i need to generate a package for one driver chip..) 
and it crashed - - 1, 2, 3 ports worked fine before i tested 4, - all with the same 16 pixels... - but if i after a crash went back to 3 or 2 it also crashed.. only 1 port worked then... 
so i want to make sure i try and test this a bit more structured  (makes a restart of the system a difference for example...)

plugins/spi/SPIOutput.h
s-light: i will have a look at this..

plugins/spi/SPIOutput.cpp
s-light: yes - that is the plan when the rest works :-)
for the combined options there are more then one way to solve it...
1. repeat/copy one set of driver channels (24ch) to all others
1. repeat/copy the first 3 LED values (3 or 6 ch for 8 or 16bit modes) to all other positions

what do you think does make more sense?

plugins/spi/SPIOutput.h
peternewman: http://docs.openlighting.org/ola/doc/latest/namespaceola_1_1network.html#ae9263a09db6563136f1e46af389dbf0f

peternewman:
Possibly relevant:
https://stackoverflow.com/questions/19534051/glibc-detect-smallbin-linked-list-corrupted?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

I'd look at what you checkout from the SPI buffer, and how you access that afterwards.

On a different note, if you update your branch compared to master, the Travis build should start working again.

plugins/spi/SPIOutput.cpp
```diff
@@ -874,6 +917,239 @@ uint8_t SPIOutput::CalculateAPA102PixelBrightness(uint8_t brightness) {
...
+    // }
+
+    // OLA_WARN << "  TLC5971_packet_t size:"
+    //          << sizeof(TLC5971_packet_t);
+    // OLA_WARN << "  device_data size:"
+    //          << sizeof(device_data);
+    // should return 28byte = 224bit
+
+    // copy data to output buffer
+    // memcpy(output + spi_offset, device_data.bytes, sizeof(TLC5971_packet_t));
```
peternewman: Does the memcpy not work? Why not, that ought to be quicker and safer.

plugins/spi/SPIOutput.cpp
```diff
@@ -874,6 +917,239 @@ uint8_t SPIOutput::CalculateAPA102PixelBrightness(uint8_t brightness) {
...
+      uint8_t data_index = 0;
+      data_index < sizeof(TLC5971_packet_t);
+      data_index++
+    ) {
+      output[spi_offset + data_index] = device_data.bytes[data_index];
+    }
+  }  // for devices_in_buffer end
+
+  // write output back
+  m_backend->Commit(m_output_number);
```
peternewman: Low tech debug, add a log line before this, is it this causing the issue, or the code above.

PR #1400, opened 2018-04-12
Use top_builddir in more places and switch to libusb_set_option if available

peternewman:


peternewman:
Sorry @nomis52 too soon. Can you re-review now I've fixed the Mac build options by switching to libusb_set_option if available.

PR #1406, opened 2018-04-21
Ignore another manufacturer name in Codespell

peternewman:


nils-van-zuijlen:
There are some other manufacturers that have the same issue. See #1408

peternewman:
Thanks @nils-van-zuijlen . We actually need both for now, as our 0.10 branch is our current release with patches, whereas master is our next major release, so we need to keep both green on Travis.

PR #1409, opened 2018-04-25
Yocto recipe for building ola on any embedded device supported. 

desert:
Tested with rocko.

### DIFF
--- a/yocto/recipes-conectivity/ola/ola_git.bb
+++ b/yocto/recipes-conectivity/ola/ola_git.bb
@@ -1,0 +1,73 @@
+SUMMARY = "Open Lighting Architecture - OLA"
+DESCRIPTION = "The Open Lighting Architecture (OLA) is part of the Open Lighting Project and provides applications with a mechanism to send and receive DMX512 & RDM commands using hardware devices and DMX over IP protocols. This enables software lighting controllers to communicate with hardware either via Ethernet or traditional DMX512 networks. \
+OLA can also convert DMX512 data sent using DMX over IP protocols from one format to another, allowing devices from different manufacturers to interact with one another. For example a Strand Lighting Console using ShowNet can send DMX512 to an Enttec EtherGate. When combined with a physical DMX interface such as the DMX USB Pro, OLA can send and receive data from wired DMX512 networks."
+
+LICENSE = "GPLv2 & LGPLv2.1"
+LIC_FILES_CHKSUM = "file://LICENCE;md5=7aa5f01584d845ad733abfa9f5cad2a1"
+
+#DEPENDS_class-nativesdk = "bison libusb1 protobuf protobuf-native pkgconfig-native ossp-uuid ola-native "
+DEPENDS_class-target    = "bison libmicrohttpd avahi libusb1 libftdi cppunit protobuf protobuf-native ola-native"
+DEPENDS_class-native    = "protobuf bison-native flex-native pkgconfig-native ossp-uuid-native"
+
+PV = "0.10.6"
+
+# Esta revision de git tiene un parche para protobuf, la tag de 0.10.6 no compila.
+SRCREV = "00dc86a48ec4c528cec90166435b440f283a9c86" 
+SRC_URI = "git://github.com/OpenLightingProject/ola.git;protocol=https"
+
+S = "${WORKDIR}/git"
+B = "${WORKDIR}/git"
+
+inherit autotools-brokensep relative_symlinks
+# autotools-brokensep 
+# forces in-source bulding (ola has bugs for out-of-tree builds). 
+# See: https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#migration-1.7-autotools-class-changes
+# 
+# relative_symlinks
+# Absolute symbolic links (symlinks) within staged files are no longer permitted and now trigger an error.
+# inherit relative_symlinks within the recipe to turn those absolute symlinks into relative
+# https://www.yoctoproject.org/docs/2.4/ref-manual/ref-manual.html#migration-2.3-absolute-symlinks
+
+#EXTRA_OECONF = " --disable-unittests --disable-rdm-tests --disable-python-libs --disable-java-libs --disable-all-plugins"
+EXTRA_OECONF  = " --disable-unittests --disable-rdm-tests --disable-java-libs "
+
+# nativesdk is for the toolchain, only libola and ola_common are required, all plugins are dynamic
+EXTRA_OECONF_append_class-nativesdk = " --with-ola-protoc-plugin=${STAGING_BINDIR_NATIVE}/ola_protoc_plugin \
+                                        --disable-all-plugins --disable-e133 --disable-python-libs --disable-examples --disable-http"
+      
+
+# native    is for the strange ola_protoc_plugin required at build time
+EXTRA_OECONF_append_class-native    = " --disable-all-plugins --disable-e133 --disable-python-libs --disable-examples --disable-http"
+
+# target    are the real config options
+EXTRA_OECONF_append_class-target = " --with-ola-protoc-plugin=${STAGING_BINDIR_NATIVE}/ola_protoc_plugin --disable-python-libs\
+  --enable-http     --enable-dummy            --enable-gpio   --enable-opendmx --enable-e131 \
+  --enable-opendmx  --enable-usbdmx           --enable-usbpro --enable-spi     --enable-spidmx \
+  --enable-artnet   --enable-openpixelcontrol --enable-renard --enable-kinet \
+  --enable-nanoleaf --enable-karate           --enable-espnet --enable-milinst --enable-pathport \
+  --enable-sandnet  --enable-shownet          --enable-stageprofi \
+  --enable-e133 \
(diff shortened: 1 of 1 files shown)


desert:
So strange it did not build, my code does not even require to be compiled. It's basically a yocto recipe just for reference.

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
...
+LICENSE = "GPLv2 & LGPLv2.1"
+LIC_FILES_CHKSUM = "file://LICENCE;md5=7aa5f01584d845ad733abfa9f5cad2a1"
+
+#DEPENDS_class-nativesdk = "bison libusb1 protobuf protobuf-native pkgconfig-native ossp-uuid ola-native "
+DEPENDS_class-target    = "bison libmicrohttpd avahi libusb1 libftdi cppunit protobuf protobuf-native ola-native"
+DEPENDS_class-native    = "protobuf bison-native flex-native pkgconfig-native ossp-uuid-native"
+
+PV = "0.10.6"
+
+# Esta revision de git tiene un parche para protobuf, la tag de 0.10.6 no compila.
```
peternewman: Could we have this comment in English please :)

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
+SUMMARY = "Open Lighting Architecture - OLA"
+DESCRIPTION = "The Open Lighting Architecture (OLA) is part of the Open Lighting Project and provides applications with a mechanism to send and receive DMX512 & RDM commands using hardware devices and DMX over IP protocols. This enables software lighting controllers to communicate with hardware either via Ethernet or traditional DMX512 networks. \
+OLA can also convert DMX512 data sent using DMX over IP protocols from one format to another, allowing devices from different manufacturers to interact with one another. For example a Strand Lighting Console using ShowNet can send DMX512 to an Enttec EtherGate. When combined with a physical DMX interface such as the DMX USB Pro, OLA can send and receive data from wired DMX512 networks."
+
+LICENSE = "GPLv2 & LGPLv2.1"
+LIC_FILES_CHKSUM = "file://LICENCE;md5=7aa5f01584d845ad733abfa9f5cad2a1"
+
+#DEPENDS_class-nativesdk = "bison libusb1 protobuf protobuf-native pkgconfig-native ossp-uuid ola-native "
```
peternewman: Is this relevant still? Would this be for a native build?

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
...
+EXTRA_OECONF_append_class-nativesdk = " --with-ola-protoc-plugin=${STAGING_BINDIR_NATIVE}/ola_protoc_plugin \
+                                        --disable-all-plugins --disable-e133 --disable-python-libs --disable-examples --disable-http"
+      
+
+# native    is for the strange ola_protoc_plugin required at build time
+EXTRA_OECONF_append_class-native    = " --disable-all-plugins --disable-e133 --disable-python-libs --disable-examples --disable-http"
+
+# target    are the real config options
+EXTRA_OECONF_append_class-target = " --with-ola-protoc-plugin=${STAGING_BINDIR_NATIVE}/ola_protoc_plugin --disable-python-libs\
+  --enable-http     --enable-dummy            --enable-gpio   --enable-opendmx --enable-e131 \
```
peternewman: You don't explicitly need to enable all the options. Although it will ensure configure fails if some of them aren't available for some reason.

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
...
+  --enable-nanoleaf --enable-karate           --enable-espnet --enable-milinst --enable-pathport \
+  --enable-sandnet  --enable-shownet          --enable-stageprofi \
+  --enable-e133 \
+  --enable-examples "
+
+# -fvisibility-inlines-hidden breaks stuff
+CXXFLAGS = "${CFLAGS}"
+
+# The code is not Werror safe
+do_configure_prepend() {
```
peternewman: `--disable-fatal-warnings` will do this, without needing to be quite so hacky. Although we try our best to fix the issues, so ideally can you tell us what errors you're getting and we'll try and fix or bypass them before they become critical in future.

yocto/recipes-conectivity/ola/ola_git.bb
desert: You need to build native (host, usually x86), in order to build target (usually ARM). So yes, this is the painful one, this is the real thing that makes this build.

yocto/recipes-conectivity/ola/ola_git.bb
desert: In yocto anyone would be able to disable them, but disabling them will make ola build transparently but be useless as will have no plugins.

yocto/recipes-conectivity/ola/ola_git.bb
desert: You mean as an argument for ./configure ?

desert:
> So you need to probably add a Makefile for that directory and add it to that.

Oh, I know nothing about autotools.

yocto/recipes-conectivity/ola/ola_git.bb
peternewman: Correct @desert .

But please either comment here, or pastebin the errors and we can try adn fix them properly.

peternewman:
> >    So you need to probably add a Makefile for that directory and add it to that.
> Oh, I know nothing about autotools.

That's easy, see the Debian one here:
https://github.com/OpenLightingProject/ola/blob/master/debian/Makefile.mk

And then the new Makefile.mk needs including in the master Makefile.am:
https://github.com/OpenLightingProject/ola/blob/master/Makefile.am#L153

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
...
+
+#DEPENDS_class-nativesdk = "bison libusb1 protobuf protobuf-native pkgconfig-native ossp-uuid ola-native "
+DEPENDS_class-target    = "bison libmicrohttpd avahi libusb1 libftdi cppunit protobuf protobuf-native ola-native"
+DEPENDS_class-native    = "protobuf bison-native flex-native pkgconfig-native ossp-uuid-native"
+
+PV = "0.10.6"
+
+# 0.10.6 tag won't build, this rev has protobuf patches
+SRCREV = "00dc86a48ec4c528cec90166435b440f283a9c86" 
+SRC_URI = "git://github.com/OpenLightingProject/ola.git;protocol=https"
```
peternewman, line 16: We should be able to build 0.10.9 cleanly now, is this the right sort of syntax?
```suggestion
PV = "0.10.9"

SRC_URI = "git://github.com/OpenLightingProject/ola.git;protocol=https"
```

yocto/recipes-conectivity/ola/ola_git.bb
peternewman: Should it be uncommented though then, there seems to be a mix of both referring to nativesdk throughout the file.

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
...
+
+S = "${WORKDIR}/git"
+B = "${WORKDIR}/git"
+
+inherit autotools-brokensep relative_symlinks
+# autotools-brokensep 
+# forces in-source building (ola has bugs for out-of-tree builds).
```
peternewman, line 23: I think/hope we've fixed this too now...

yocto/recipes-conectivity/ola/ola_git.bb
```diff
@@ -0,0 +1,73 @@
...
+# autotools-brokensep 
+# forces in-source building (ola has bugs for out-of-tree builds). 
+# See: https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#migration-1.7-autotools-class-changes
+# 
+# relative_symlinks
+# Absolute symbolic links (symlinks) within staged files are no longer permitted and now trigger an error.
+# inherit relative_symlinks within the recipe to turn those absolute symlinks into relative
+# https://www.yoctoproject.org/docs/2.4/ref-manual/ref-manual.html#migration-2.3-absolute-symlinks
+
+#EXTRA_OECONF = " --disable-unittests --disable-rdm-tests --disable-python-libs --disable-java-libs --disable-all-plugins"
```
peternewman, line 31: Is this line redundant/part of testing?

yocto/recipes-conectivity/ola/ola_git.bb
peternewman: Okay, if we need to explicitly enable all plugins in here we ought to add a note to ensure we add new ones here when they're added to https://github.com/OpenLightingProject/ola/blob/d34d4f13d954894e76bf024743e6519696813c29/configure.ac#L892-915

peternewman (commented):
Some updated comments/re-review.

PR #1411, opened 2018-04-29
Fix/udev rules

bitte-ein-bit:
udev rules are invalid format


bitte-ein-bit:
@peternewman here you go.

peternewman:
How did you get the error exactly @bitte-ein-bit ? I'm wondering if we can add validating the udev rules to our Travis tasks.

bitte-ein-bit:
Well, I wondered why the rules would not apply. Rebooting did not load the rules. I googled a reload command, then it would show up in syslog. But before I had already spotted the missing commas, but was unsure on the syntax.

This error would be catched by https://github.com/pavlinux/udev/blob/master/test/rule-syntax-check.py. In this folder is also a simple test suite.

peternewman:
Could I also trouble you to update https://github.com/OpenLightingProject/raspberrypi/blob/master/etc/udev/rules.d/10-local.rules with the comma too? Or feel free to tell me to get lost and I'll do it myself...

PR #1415, opened 2018-05-05
Avoid use of root logger and fix close callback in python3

shenghaoyang:
- Using the root logger makes it non-obvious for users of the library
  to control the log messages from the library.

- By setting up a logger for ola and using it for all the logging
  calls, users of the library can simply do logging.getLogger('ola')
  and control the output of the logger.

- Note that any tools / tests depending on log output might need to 
  be changed due to the reasons above. Those changes can be done, 
  but I want to put this commit out first so we can discuss this approach, 
  or we could skip emulating past behavior and simply adopt 
  the convention for python libraries.

- Empty string object is no longer returned by ``recv()`` in python3 
  on connection close, but an empty bytes object is returned instead. 
  Comparing the return value to an empty string is insufficient. 
  Since both objects evaluate to false, check for that condition and 
  invoke close actions accordingly.

### DIFF
--- a/python/ola/__init__.py
+++ b/python/ola/__init__.py
@@ -1,0 +1,29 @@
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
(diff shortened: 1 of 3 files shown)


shenghaoyang:
``@maintainers``
Found another little bug in the python library - could I consolidate everything into this request?

peternewman:
Yeah go for it @shenghaoyang . Just retitle the PR to suit.

python/ola/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# __init__.py
+# Copyright (C) 2018 Shenghao Yang
+
+import logging
+
+"""Initialization code for the ola python library"""
```
peternewman: Codacy doesn't like this:
"String statement has no effect"
https://app.codacy.com/app/peternewman/ola/pullRequest?prid=1616358

I think it may need to be above the import, looking at some other real world examples.

shenghaoyang:
Yeah, my apologies, I messed up the docstring. It's supposed to be the
first actual statement. I'll get a fixup in tomorrow - was wondering why
codacy wasn't happy.

PR #1426, opened 2018-05-29
Switch to clang-10 on Travis

peternewman:


peternewman:
361ad2e14442e33023d9ecc3241605f0c6c65c63 Currently breaks the other non-clang builds...

peternewman:
The OS X builds currently seem to be broken for unrelated reasons, but this appears to be fine apart from that.

PR #1430, opened 2018-06-03
Debian fixes from wouter and based on latest lintian failures

peternewman:


peternewman:
A few bits are actually fresh changes from recent Debian Lintian warnings @yoe . I've also hacked about your commits a little, so apologies for the merge conflicts when you come to release!

Issue #1219, opened 2017-04-13
FTDI port


nbeAnomes:
Hello,

I would like to know why FTDI devices output port begins by 1, and not by 0 as others device.
From what I see, in the loop of the StartHook() function of the FtdiDmxDevice when the FtdiDmxOutputPort is created. Its port index is set with an offset of 1.

Anyway, thanks for your great work !

peternewman:
That's a fair point, feel free to open a PR. :)

The port ID is generally less important than the Description, although it looks like the current code names them all the same, which isn't very helpful:
https://github.com/OpenLightingProject/ola/blob/master/plugins/ftdidmx/FtdiDmxPort.h#L62
https://github.com/OpenLightingProject/ola/blob/master/plugins/ftdidmx/FtdiWidget.h#L182
https://github.com/OpenLightingProject/ola/blob/master/plugins/ftdidmx/FtdiWidget.h#L148

See the code here for the sort of thing we can do, although I suspect given different models of FTDI based unit, we probably want to keep it fairly generic:
https://github.com/OpenLightingProject/ola/pull/1072/files#diff-004f973378d58eab249e16bc9a237804R74

I've also been trying to move the info like the serial number that's common to the device, rather than the port, to the device description, rather than the port description. See for example the changes here:
https://github.com/OpenLightingProject/ola/pull/1039/files#diff-520883e6359feb30989879075897b7f0R66

So if you're doing some updates, doing that at the same time too would be great.

Have you got a multi-port FTDI based device then?

peternewman:
See my related issue regarding naming here https://github.com/OpenLightingProject/ola/issues/975

peternewman:
@Keeper-of-the-Keys I think this is what you were referring to regarding FTDI port naming?

Keeper-of-the-Keys:
@peternewman  Not quite, I have FTDI 2 boards with 4 ports, now when I create/assign a universe I can only chose which board(s) it goes to but not which of its' specific ports, they will get assigned in-order and if I for some reason did not assign the universes in-order or I named them instead of giving them numbers then currently it actually doesn't show anywhere (iirc the boards are at home so I can't check) to which specific port the universe is associated.

I have asked on the libftdi mailing list if there is a way to make the I/O LEDs of ports blink to "identify" ports, no answer on that yet (I tried something minor with that so far).

Is there currently a facility in OLA that the plugin just needs to export the port names/numbers properly or if this also not supported for any other devices?

[...]

PR #1435, opened 2018-06-12
FTDI DMX allow the same universe to be patched to multiple ports

peternewman:
Or patch port 2 before port 1

Fixes @Keeper-of-the-Keys 's comments in #1219, but not the core of that bug!


Keeper-of-the-Keys:
OK checked it and it makes 4 unnamed (no port X) instances of the board available.

peternewman:
Great thanks @Keeper-of-the-Keys . So it's not made it worse? But we probably need to work on the port naming so you know which of the specific ports to patch for a physical output?

And you can patch them in the order you want (e.g. selecting the second one first), or indeed more than one to the same universe?

Keeper-of-the-Keys:
@peternewman You just have a list of 4 instances, they seem to be in-order but of course if you choose to assign the second port first only the person who did the assigning will now that the second item listed currently is the 3rd port.

I think it is acceptable to have it like this but I/we should implement proper names of the ports asap, I'll be happy to cook up a patch if you give me some pointers.

Keeper-of-the-Keys:
Actually even if you assigned the 1st port first the second item should always be the 3rd port so that was a bad example, but in this case the first listed item would be the 1st and not the second port, you get the picture.

peternewman:
Okay great, merged as is then.

peternewman:
Further patch discussion to continue in #1219 .

PR #1439, opened 2018-07-06
Fix an incorrect response if we have the sensor PIDs but no sensors

peternewman:
(cherry picked from commit 68e33d942d6a1c3a52622d5211e19ac8148f222a)


peternewman:
It seems closing and re-opening also resolves this issue...

PR #1440, opened 2018-07-06
Fix an incorrect response if we have the sensor PIDs but no sensors

peternewman:
(cherry picked from commit 68e33d942d6a1c3a52622d5211e19ac8148f222a)


peternewman:
Was going to replaces #1439 as I targeted the wrong branch before, now fixed via other means.

PR #1442, opened 2018-07-07
Add support for libftdi1 and ftdidmx on Windows

peternewman:
Don't bother trying to build opendmx kernel plugin on Windows


peternewman:
I've also added it to NEWS @nomis52 .

Issue #1055, opened 2016-02-29
"readdir_r" deprecated in newer glibc versions


daveol:
In fedora's rawhide version of glibc `readdir_r` is deprecated ([build.log](https://kojipkgs.fedoraproject.org//work/tasks/9706/13179706/build.log)), the [patch](https://www.sourceware.org/ml/libc-alpha/2016-02/msg00093.html) or [bug](https://sourceware.org/bugzilla/show_bug.cgi?id=19056) don't really explain why or what to use else.

peternewman:
This seems like it might address why:
https://womble.decadent.org.uk/readdir_r-advisory.html

The patch explicitly says "It is recommended to use readdir and readdir64 instead."

nomis52:
Switching to readdir() is a step backwards.

I'd prefer that we use dirfd() to determine the size of the buffer for  readdir_r().

I've confirmed dirfd() is available on Mac & linux.

daveol:
> I'd prefer that we use dirfd() to determine the size of the buffer for readdir_r().

Okay, but that doesn't take away that readdir_r() is deprecated in the newer versions of glibc and will probably be removed in the future.

I see only one option, switching back to readdir()

nomis52:
Hmm, ok I'm sold. I'll look at the patch later today..

daveol:
The patch doesn't include thread handling at the moment (readdir is not threadsafe), which could be a problem if we use the function in two threads at the same time... do we?

peternewman:
Would it be worth using readdir_r where available (via configure) and only switching back to readdir on machines where the former has been deprecated, or is that overly complex?

peternewman:
We're getting other users hitting this in the wild now http://text.michd.me/isiqejojof.mel . @loansindi .

peternewman:
More issues:
https://groups.google.com/forum/#!topic/open-lighting/GYt4Ne3qQzM

daveol:
I now have written a different approach to a fix for this, it is however of very temporary nature. you can find that fix in PR #1167

peternewman:
@daveol put in a warnings skipping based fix as an interim workaround to this in https://github.com/OpenLightingProject/ola/pull/1168 . We'll still need to fix this properly for when readdir_r gets removed.

kripton:
There's a pretty good explanation on http://man7.org/linux/man-pages/man3/readdir_r.3.html now.
So I'd say that going back to `readdir` might not be a step backwards.

PR #1444, opened 2018-07-12
Add functionality to notify systemd on olad startup and config reload

shenghaoyang:
- This helps users who run olad under the management of systemd to
  order their services after olad startup, (i.e. starting up a daemon
  that depends on the olad RPC socket being available), avoiding
  the need for a manual delay.

- If anyone's versed in autotools magic, please help check the
  changes to `configure.ac`. I've only dabbled with autotools one
  other time in my life, and I'm sure if what I did was exactly correct.

- Tested on my local machine with a unit file running the olad binary
  as a notifying service. Reload messages tested as well. And I do
  see the systemd generated reload notifications bounding the
  output from olad during the reload process. 

- Not sure how we go about writing test cases for this, though. 
  Seems a  bit pathological to formally test something like this, 
  but if the CI environment has `systemd` and `libsystemd` installed, 
  we could cook up a test for the notification functionality using 
  `systemd-run` - if the notification didn't go through, then
  `systemd-run` should notify on an error starting the service.

Thanks for reviewing.

Shenghao

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -387,6 +387,13 @@ esac
 have_libftd2xx="no"
 AM_CONDITIONAL([HAVE_LIBFTD2XX], [test "x$have_libftd2xx" = xyes])
 
+# libsystemd
+AC_SEARCH_LIBS([sd_notify], [systemd], [have_libsystemd="yes"])
+AC_CHECK_HEADER([systemd/sd-daemon.h], [], [have_libsystemd="no"])
+
+AS_IF([test "x$have_libsystemd" = xyes],
+      [AC_DEFINE([HAVE_LIBSYSTEMD], [1], [define if libsystemd is installed])])
+
 # ncurses
 AC_CHECK_LIB([ncurses], [initscr], [have_ncurses="yes"])
 AM_CONDITIONAL([HAVE_NCURSES], [test "x$have_ncurses" = xyes])
--- a/olad/OlaServer.cpp
+++ b/olad/OlaServer.cpp
@@ -22,6 +22,10 @@
 #include <config.h>
 #endif  // HAVE_CONFIG_H
 
+#ifdef HAVE_LIBSYSTEMD
+#include <systemd/sd-daemon.h>
+#endif  // HAVE_LIBSYSTEMD
+
 #include <errno.h>
 #include <signal.h>
 #include <stdio.h>
@@ -472,9 +476,17 @@ bool OlaServer::InternalNewConnection(
 }
 
 void OlaServer::ReloadPluginsInternal() {
+#ifdef HAVE_LIBSYSTEMD
+  // Return value is intentionally not checked for both calls.
+  // See return value section under sd_notify(3).
+  sd_notify(0, "RELOADING=1\nSTATUS=Reloading plugins\n");
+#endif  // HAVE_LIBSYSTEMD
   OLA_INFO << "Reloading plugins";
   StopPlugins();
   m_plugin_manager->LoadAll();
+#ifdef HAVE_LIBSYSTEMD
+  sd_notify(0, "READY=1\nSTATUS=Plugin reload complete\n");
+#endif  // HAVE_LIBSYSTEMD
 }
 
 void OlaServer::UpdatePidStore(const RootPidStore *pid_store) {
--- a/olad/Olad.cpp
+++ b/olad/Olad.cpp
@@ -24,6 +24,10 @@
 #include <config.h>
 #endif  // HAVE_CONFIG_H
 
+#if HAVE_LIBSYSTEMD
+#include <systemd/sd-daemon.h>
+#endif  // HAVE_LIBSYSTEMD
+
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -171,6 +175,11 @@ int main(int argc, char *argv[]) {
       ola::NewCallback(olad->GetOlaServer(), &ola::OlaServer::ReloadPlugins));
 #endif  // _WIN32
 
+#if HAVE_LIBSYSTEMD
+  // Return value is intentionally not checked. See return value section
+  // under sd_notify(3).
+  sd_notify(0, "READY=1\nSTATUS=Startup complete\n");
+#endif  // HAVE_LIBSYSTEMD
   olad->Run();
   return ola::EXIT_OK;
 }


configure.ac
```diff
@@ -387,6 +387,13 @@ esac
 have_libftd2xx="no"
 AM_CONDITIONAL([HAVE_LIBFTD2XX], [test "x$have_libftd2xx" = xyes])
 
+# libsystemd
+AC_SEARCH_LIBS([sd_notify], [systemd], [have_libsystemd="yes"])
```
peternewman: If you see have_libftd2xx above, you should really initialise it to no before you use it here.

olad/OlaServer.cpp
```diff
@@ -472,9 +476,17 @@ bool OlaServer::InternalNewConnection(
 }
 
 void OlaServer::ReloadPluginsInternal() {
+#ifdef HAVE_LIBSYSTEMD
+  // Return value is intentionally not checked for both calls.
+  // See return value section under sd_notify(3).
+  sd_notify(0, "RELOADING=1\nSTATUS=Reloading plugins\n");
```
peternewman: Having read the man page, I understand why it's not checked, but for user sanity, we should add some logging. Probably both to signify sd_notify has been run/the return code and particularly a warning if it failed.

I'd probably suggest OLA_WARN if we get <0 return code, and then maybe OLA_INFO for 0 and OLA_DEBUG for >0, but that may be a bit excessive. Likewise others may argue those log levels should all be reduced by one...

olad/OlaServer.cpp
```diff
@@ -472,9 +476,17 @@ bool OlaServer::InternalNewConnection(
...
+#endif  // HAVE_LIBSYSTEMD
   OLA_INFO << "Reloading plugins";
   StopPlugins();
   m_plugin_manager->LoadAll();
+#ifdef HAVE_LIBSYSTEMD
+  sd_notify(0, "READY=1\nSTATUS=Plugin reload complete\n");
```
peternewman: Again, some logging to help the user.

olad/Olad.cpp
```diff
@@ -171,6 +175,11 @@ int main(int argc, char *argv[]) {
       ola::NewCallback(olad->GetOlaServer(), &ola::OlaServer::ReloadPlugins));
 #endif  // _WIN32
 
+#if HAVE_LIBSYSTEMD
+  // Return value is intentionally not checked. See return value section
+  // under sd_notify(3).
+  sd_notify(0, "READY=1\nSTATUS=Startup complete\n");
```
peternewman: Again, a bit of logging please.

olad/OlaServer.cpp
peternewman: What happens if the system is too old to support RELOADING? I assume they just get discarded as they're plain text?

olad/Olad.cpp
shenghaoyang: Hmm.. I don't think we need it for this one, though. It could be useful to tell the user that we're not running under `systemd`, or that we have failed to access the notification socket due to a configuration error.

However, if `olad` was launched as a systemd service of type `notify` (which is the intended use case): then `systemd` will report that the start of the service failed, anyway, which should alert the user to the error. There also shouldn't be an error accesing the notification socket, because `systemd` implicitly allows services with `type=notify` to access the notification socket.

configure.ac
shenghaoyang: No problem. I'll follow the examples from the above location. 

What do you think about making linking against `libsystemd` a command line option to the build system? I'm thinking build systems that have `libsystemd` may silently cause built binaries to link against `libsystemd`, and this could lead to a bit of a dependency problem, unless they're aware of this new change.

olad/OlaServer.cpp
shenghaoyang: Ditto.

olad/OlaServer.cpp
shenghaoyang: >What happens if the system is too old to support RELOADING? I assume they just get discarded 
>as they're plain text?

`sd_notify(3)` mentions this: `The string may contain any kind of variable assignments, but the following shall be considered well-known`.

I think that does indeed imply that if the service manager doesn't recognize the variable being assigned to, it'll either log that somewhere or ignore that assignment, so we shouldn't have a problem with old systems.

>  I understand why it's not checked, but for user sanity, we should add some logging. Probably 
>  both to signify sd_notify has been run/the return code and particularly a warning if it failed.

If that's the case, would a message at startup do? `libsystemd` provides capabilities for determining whether a binary has been setup under the control of `systemd`, and that could be used to indicate that notification messages are being used. As for the failures, though, we could log those as well, but I think conditionally - only when being run under `systemd`

shenghaoyang:
> Also should we add MAINPID https://www.freedesktop.org/software/systemd/man/sd_notify.html#MAINPID=%E2%80%A6 like the extended example here https://www.freedesktop.org/software/systemd/man/sd_notify.html#Examples

`MAINPID` is probably only useful if we run olad with `--daemon` and have it fork itself. However, when running under systemd, that's probably not required, and we can rely on `systemd` to fork us off and keep us happily tracked, which is what it's good at, anyway. Unless `olad` requires some special magic while forking, we shouldn't really need to bother with this. `MAINPID` without forking is considered to be the first process that is created when executing the `olad` binary.

If `olad` is going to adopt the creation of multiple processes, though, then we'll need to set this `MAINPID` accordingly - but I don't think this is the case...

olad/OlaServer.cpp
shenghaoyang: Scratch that, there's no library function to determine whether the service has been run under systemd, but the call shouldn't return a negative value if the socket was not passed, anyway. I'll log failures and log from main to indicate whether notifications are being passed.

configure.ac
peternewman: Yes, it may make sense, see e.g. https://github.com/OpenLightingProject/ola/blob/master/configure.ac#L494-L518 .

I was wondering, but at least it's only libsystemd, it doesn't require systemd itself.

shenghaoyang:
@peternewman is there a way to add some CFLAGS when compiling the sources that need to be linked with `libsystemd`? I did see some CFLAG variables set, but those were for `protobuf` and testing.

olad/NotifySystemd.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <string.h>
+#endif  // HAVE_LIBSYSTEMD
+
+#include "ola/Logging.h"
+
+#include "olad/Systemd.h"
+
+namespace ola {
+
+int notify_systemd(int unset_environment, const char *state) {
```
peternewman: We go CamelCase for our method names.

olad/NotifySystemd.cpp
```diff
@@ -0,0 +1,51 @@
...
+  int rtn = sd_notify(unset_environment, state);
+  if (rtn < 0) {
+    char buf[1024];
+    OLA_WARN << "Error sending notification to systemd: " <<
+      strerror_r(-rtn, buf, sizeof(buf));
+  }
+  return rtn;
+}
+
+bool notify_available() {
```
peternewman: Again

olad/NotifySystemd.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "olad/Systemd.h"
+
+namespace ola {
+
+int notify_systemd(int unset_environment, const char *state) {
+  int rtn = sd_notify(unset_environment, state);
+  if (rtn < 0) {
+    char buf[1024];
+    OLA_WARN << "Error sending notification to systemd: " <<
+      strerror_r(-rtn, buf, sizeof(buf));
```
peternewman: I think you'll only print the error code from strerror_r currently. Looking at https://linux.die.net/man/3/strerror_r it seems you should be able to just do `<< strerror(-rtn)`.

Can you also move the second << to the next line and align with the one above.

peternewman: Also I think you should add a configure check for strerror_r (see https://github.com/OpenLightingProject/ola/blob/master/configure.ac#L185 and e.g. https://github.com/OpenLightingProject/ola/blob/master/common/base/Credentials.cpp#L342 ), it's the first time it's being used in the codebase and we run on a variety of weird platforms. See also some similar fun and discussion with readdir_r https://github.com/OpenLightingProject/ola/issues/1055 .

configure.ac
```diff
@@ -748,6 +748,21 @@ AC_ARG_ENABLE(
...
+  [systemd],
+  [AS_HELP_STRING([--enable-systemd], [Enable systemd support])])
+
+have_libsystemd="no"
+AS_IF([test "x$enable_systemd" = "xyes"],
+  [PKG_CHECK_MODULES([libsystemd], [libsystemd >= 38], [have_libsystemd="yes"])
+   AC_CHECK_HEADER([systemd/sd-daemon.h], [], [have_libsystemd="no"])])
+
+AS_IF([test "x$have_libsystemd" = xyes],
+      [AC_DEFINE([HAVE_LIBSYSTEMD], [1], [define if systemd support wanted])])
```
peternewman: Nit: I'd probably say this is if systemd support present (or present and wanted).

configure.ac
```diff
@@ -748,6 +748,21 @@ AC_ARG_ENABLE(
   [AS_HELP_STRING([--disable-doxygen-version],
   [Substitute the Doxygen version with latest, for the website])])
 
+# systemd support
+AC_ARG_ENABLE(
+  [systemd],
+  [AS_HELP_STRING([--enable-systemd], [Enable systemd support])])
```
peternewman: You probably want to negate this, so it's turned on by default where available (or no-one will realise it exists/use it). See e.g. https://github.com/OpenLightingProject/ola/blob/master/configure.ac#L573-L589 and don't forget to update the AS_IF below too.

olad/Makefile.mk
```diff
@@ -42,6 +42,11 @@ ola_server_sources += olad/HttpServerActions.cpp \
 ola_server_additional_libs += common/http/libolahttp.la
 endif
 
+if HAVE_LIBSYSTEMD
+ola_server_sources += olad/NotifySystemd.cpp olad/Systemd.h
```
peternewman: We generally match .h and .cpp filenames.

olad/Systemd.h
```diff
@@ -0,0 +1,46 @@
...
+
+#ifndef OLAD_SYSTEMD_H_
+#define OLAD_SYSTEMD_H_
+
+namespace ola {
+
+/*
+ * @brief Notify systemd about daemon state changes.
+ *  
+ * This function logs on failures to queue notifications, but only if the
```
peternewman: SPaG "logs failures to queue" or "logs on failure to queue"

olad/OlaServer.cpp
```diff
@@ -472,9 +476,15 @@ bool OlaServer::InternalNewConnection(
 }
 
 void OlaServer::ReloadPluginsInternal() {
+#ifdef HAVE_LIBSYSTEMD
+  ola::notify_systemd(0, "RELOADING=1\nSTATUS=Reloading plugins\n");
```
peternewman: Could the socket have gone away by the time this gets fired though? Potentially minutes or hours/days later?

My previous comment was more along the lines of if I'm trying to troubleshoot why my reload isn't getting to systemd, I'd want to see (at info/debug level) the positive return code from calling sd_notify, so I know it's the systemd end I need to look at, and not OLA, or vice versa.

shenghaoyang: >Could the socket have gone away by the time this gets fired though? Potentially minutes or hours/days later?

Well, I'm not sure that this could _even_ happen, unless someone removes the environment variable. The socket is actually passed as a socket address (path name or name in abstract namespace, _a-la_ unix domain sockets), and `sd_notify` establishes a new socket sending data to that address, closing it on every return.

If `systemd` did somehow... fail, I think the whole system would be going up in flames, by then. But if it did happen and nobody is listening on the socket address, then the `sendmsg()` call in `sd_notify()` will fail and the error will be logged through `OLA_WARN`.

The only way for the `notify_systemd` call to fail silently is the removal of the `NOTIFY_SOCKET` environment variable.

configure.ac
shenghaoyang: Eek.. an automatic dependency feels a little weird to me, unless we put it out for people to know. If we put the existence of this option in `NEWS` (which is probably where it will end up in) and you suspect:
>  no-one will realise it exists/use it

then, if we put in a warning that `libsystemd` is going to become an automatic dependency, I kinda also think that
> no-one will realise it exists/use it

until bug reports come in regarding `libsystemd` not being found on non-build systems...

peternewman: Hmm, we have the best of both worlds with our plugins, a literal auto mode, so by default if the bits are there, you get your plugin and if they aren't you don't, but you can also force it on (so it complains if dependencies are missing) or force it off (so it doesn't use it even if present).

You may be right about bug reports, I guess I'd tend towards the nudge theory, if we turn it on, then people who read the docs will turn it off easily, those who don't will ideally search or file a bug/ask on the mailing list and there's a workaround.

If we leave it off, we should at least enable it in our Debian build config.

olad/OlaServer.cpp
peternewman: Okay, fair enough.

olad/NotifySystemd.cpp
peternewman: Ah yes, in which case can we just have a comment for our future selves if we switch away from the GNU extensions in future.

Well currently multiple plugins call strerror (and some do indeed spawn their own threads, at least one call is in a spawned thread), so we may have that issue already.

We should probably be moving to strerror_r throughout, but equally I wouldn't want this one usage to break the build on machines that don't have it, or by adding it to configure stop it building on those OSes.

olad/Olad.cpp
peternewman: It's those edge cases that can be frustrating to troubleshoot though. If I forget to put type=notify, I want olad to tell me the socket isn't available, so I know to tweak my systemd config. Likewise in that timeout case, I'd like to know olad did/didn't send the message so I can troubleshoot ifdef type stuff.

shenghaoyang: Yeah, I've added this in commit a5e57fe971e0a49f59bee5187923ef4bb5799894. A success/fail message everytime may clog up the logs if we want to use notify() a little more, though....

olad/NotifySystemd.cpp
shenghaoyang: > Ah yes, in which case can we just have a comment for our future selves if we switch away from the GNU extensions in future.

Sure, alright.

> We should probably be moving to strerror_r throughout, but equally I wouldn't want this one usage to break the build on machines that don't have it, or by adding it to configure stop it building on those OSes.

Any modern system with `systemd` should have this - deliberately using `sterrror()` feels pretty icky. Maybe we can take a page from `nginx`: they call `strerror()` in a section where it is known that no other thread can call `strerror()` and caches that for use.

Have a look at their implementation: https://github.com/nginx/nginx/blob/f8a9d528df92c7634088e575e5c3d63a1d4ab8ea/src/os/unix/ngx_errno.c

olad/Systemd.h
```diff
@@ -0,0 +1,46 @@
...
+ * @param state state block to pass to systemd.
+ * @return value returned from @ref sd_notify()
+ */
+int NotifySystemd(int unset_environment, const char *state);
+
+/* 
+ * @brief Tests whether the systemd notification socket is available.
+ * @return @c true if the socket is available, @c false if not.
+ */
+bool NotifyAvailable();
```
peternewman: It would probably be clearer if this had Systemd in it's name, perhaps SystemdNotifyAvailable for this function and then rename the other one to SystemdNotify. It's then a closer match to the sd_notify it's wrapping.

configure.ac
shenghaoyang: Hmm, alright. Let's default it to enabled. We're going to be playing devil's advocate (to some people out there) and _silently_ promote `systemd` adoption...

olad/Systemd.h
shenghaoyang: Fixed in commit b7afd380b0adf9764c5a1c507cedd1ac3fc587d0

configure.ac
shenghaoyang: Fixed in commit https://github.com/OpenLightingProject/ola/pull/1444/commits/83807d484ad4d66cc447a28ef395019fa13ff906

olad/NotifySystemd.cpp
shenghaoyang: Fixed in https://github.com/OpenLightingProject/ola/pull/1444/commits/469361d2b1c52d0475621cb97653511fa7d60201

shenghaoyang:
I've made a little too many commits - I'll squash a few when another round of review comes by.

olad/Systemd.h
```diff
@@ -0,0 +1,46 @@
...
+/*
+ * @brief Notify systemd about daemon state changes.
+ *  
+ * This function logs on any failure to queue notifications, but only if the
+ * notification socket environment variable is set.
+ *  
+ * @param unset_environment whether to unset the notification socket environment
+ *        variable so child processes cannot utilize it.
+ * @param state state block to pass to systemd.
+ * @return value returned from @ref sd_notify()
```
peternewman: I'm not sure this is actually the right Doxygen command, but I can't see a specific one either:
https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdref

olad/Systemd.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "olad/Systemd.h"
+
+namespace ola {
+
+int SystemdNotify(int unset_environment, const char *state) {
+  int rtn = sd_notify(unset_environment, state);
+  if (rtn < 0) {
+    char buf[1024];
+    OLA_WARN << "Error sending notification to systemd: ";
+    if (ola::Strerror_r(-rtn, buf, sizeof(buf))) {
```
peternewman: I think this if should all be within the function, rather than repeating it multiple times in each bit of code.

Also I think because you're doing multiple OLA_WARNs and it's a macro, I believe this would log multiple lines. Whereas a function returning a string would avoid those issues.

olad/Strerror_r.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef OLAD_STRERROR_R_H_
+#define OLAD_STRERROR_R_H_
+
+namespace ola {
+
+/*
+ * @brief XSI-compliant version of @ref strerror_r()
+ 
+ * See strerror(3) for more details.
```
peternewman: @see

olad/Olad.cpp
```diff
@@ -171,6 +175,18 @@ int main(int argc, char *argv[]) {
       ola::NewCallback(olad->GetOlaServer(), &ola::OlaServer::ReloadPlugins));
 #endif  // _WIN32
 
+#if HAVE_LIBSYSTEMD
+  if (ola::SystemdNotifyAvailable()) {
+    OLA_INFO << "Systemd notification socket address present. "
```
peternewman: This probably wants to be a comma not a full stop (then adjust the case on the line below.

olad/Olad.cpp
```diff
@@ -171,6 +175,18 @@ int main(int argc, char *argv[]) {
...
+             << "Sending notifications.";
+  } else {
+    OLA_WARN << "Systemd notification socket address not present. "
```
peternewman: Again

olad/Strerror_r.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Strerror_r.cpp
+ * Definition of strerror_r that is XSI-compliant.
```
peternewman: I'm generally slightly wary about the future maintainability of this, but I guess a pile of ifdef's to switch behaviour based on the detected version might not be any better.

Where did this come from, is there a link for credit/future reference?

olad/Strerror_r.cpp
```diff
@@ -0,0 +1,49 @@
+/*
```
peternewman: I wonder if these should be in include/ola and common/ola for future use by other bits of the codebase.

olad/Strerror_r.h
shenghaoyang: `@see` is supposed to link to some symbol, though, AFAIK. Not sure what doxygen will do when it tries to link to some man page...

olad/Strerror_r.cpp
shenghaoyang: Oops. Forgot the copyright. I'll add that.

olad/Strerror_r.cpp
shenghaoyang: Could do that, but I'd have to add `strerror_r` to the main list of functions checked. Probably not a problem?

peternewman: I can't remember, are they mandatory functions, or still optional?

olad/Strerror_r.h
peternewman: It says "Starts a paragraph where one or more cross-references to classes, functions, methods, variables, files or URL may be specified." So an URL isn't a symbol, although I don't know how it decides if something is a URL or not?

include/ola/base/Strerror_r.h
```diff
@@ -0,0 +1,57 @@
...
+ * @addtogroup strerror
+ * @{
+ */
+
+/*
+ * @brief XSI-compliant version of @ref strerror_r()
+ 
+ * See strerror(3) for more details.
+ */
+int Strerror_r(int errnum, char* buf, size_t buflen);
```
peternewman: Can we call this StrError_R or something, to match `ola::strings::StrNCopy` and others.

olad/Systemd.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "olad/Systemd.h"
+
+namespace ola {
+
+int SystemdNotify(int unset_environment, const char *state) {
+  int rtn = sd_notify(unset_environment, state);
+  if (rtn < 0) {
+    char buf[1024];
+    if (ola::Strerror_r(-rtn, buf, sizeof(buf))) {
+      OLA_WARN << "Error sending notification to systemd: errno = " << -rtn;
```
peternewman: You've now just had to duplicate the "Error sending notification to systemd" text. I still feel the whole return a string or return "errno" and a value should be in the function, as we'll always want that functionality. Or if you want to do a direct wrap of strerror_r, we should also ship a helper function which does what we actually want each time (and then calls the wrapped strerror_r).

Otherwise we have to keep writing this code:
```
    char buf[1024];
    if (ola::Strerror_r(-rtn, buf, sizeof(buf))) {
      OLA_WARN << "Error sending notification to systemd: errno = " << -rtn;
    } else {
      OLA_WARN << "Error sending notification to systemd: " << buf;
    }
```

Whereas we should just be able to write something like this (or whatever the function gets called) and get the same output as the above:
`    OLA_WARN << "Error sending notification to systemd: " << ola::StrError_rHelper(-rtn);`

olad/Strerror_r.h
shenghaoyang: Hmm, I guess I could put a link to a manual page in man7.

olad/Strerror_r.cpp
shenghaoyang: Well `AC_CHECK_FUNCS` wouldn't terminate if something is not found, so it's alright.

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,68 @@
...
+  // Pre-allocate to prevent re-allocation.
+  std::string out(StrError_R_BufSize, '\0');
+  if (StrError_R_XSI(errnum, &(out[0]), out.size())) {
+    out.assign("errno = ");
+    // Buffer size here is digits10 + 3 to account for:
+    // - Systems in which the int type cannot represent the full range
+    //   of digits in the position of the most significant base 10 digit.
+    //   (i.e. 0-2 representable in an 8-bit unsigned integer at the
+    //   hundredth's place instead of 0-9 because an u8 is clamped to
+    //   [0, 255])
```
peternewman: Can you just add a note on the end of this line to explicitly clarify that digits10 will therefore only return 2 for uint8. It was only when I read the docs that your comment made sense.

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // - Terminating NUL byte added by snprintf.
+    // - 1 additional unit of storage just in case a negative errnum is
+    //   passed, and somehow StrError_R_XSI() failed on that.
+    char errs[std::numeric_limits<int>::digits10 + 3];
+    int r(snprintf(errs, sizeof(errs), "%d", errnum));
+    if (r < 0) {
+      out.append("<error>");
+    // Ugly cast required to avoid compiler warning when comparing
+    // numbers of different signedness.
+    } else if (static_cast<unsigned int>(r) >= sizeof(errs)) {
```
peternewman: Would casting the sizeof to an int be better? Although there are downsides either way.

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // - Systems in which the int type cannot represent the full range
+    //   of digits in the position of the most significant base 10 digit.
+    //   (i.e. 0-2 representable in an 8-bit unsigned integer at the
+    //   hundredth's place instead of 0-9 because an u8 is clamped to
+    //   [0, 255])
+    // - Terminating NUL byte added by snprintf.
+    // - 1 additional unit of storage just in case a negative errnum is
+    //   passed, and somehow StrError_R_XSI() failed on that.
+    char errs[std::numeric_limits<int>::digits10 + 3];
+    int r(snprintf(errs, sizeof(errs), "%d", errnum));
```
peternewman: Our style is `int r = snprintf(...);`

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,68 @@
...
+    if (r < 0) {
+      out.append("<error>");
+    // Ugly cast required to avoid compiler warning when comparing
+    // numbers of different signedness.
+    } else if (static_cast<unsigned int>(r) >= sizeof(errs)) {
+      out.append("<truncated>");
+    } else {
+      out.append(errs);
```
peternewman: Shouldn't this have a resize too?

configure.ac
```diff
@@ -748,6 +751,28 @@ AC_ARG_ENABLE(
...
+have_libsystemd="no"
+AS_IF([test "x$enable_systemd" != "xno"],
+  [PKG_CHECK_MODULES([libsystemd], [libsystemd >= 38], [have_libsystemd="yes"], 
+                     [AC_MSG_WARN([disabling systemd support: can't find libsystemd])])
+   AC_CHECK_HEADER([systemd/sd-daemon.h], [],
+                   [have_libsystemd="no"
+                    AC_MSG_WARN([disabling systemd support: can't find systemd/sd-daemon.h])])
+   AS_IF([test "x$ac_cv_func_strerror_r" = "xno"],
+         [have_libsystemd="no"
+          AC_MSG_WARN([disabling systemd support: can't find strerror_r])])])
```
peternewman: The WARNs mean configure will continue. It might be nice to have a final bit of logic where if `"x$enable_systemd" != "xyes"` and `"x$have_libsystemd"=="xno"` it throws an error. So if you've explicitly added --enable-systemd to your configure, it bombs out, but if you've done nothing, it just warns and carries on.

include/ola/base/StrError_R.h
```diff
@@ -0,0 +1,78 @@
...
+ * If the length of the system-provided error description exceeds the length
+ * of this buffer minus one, then the output of @ref StrError_R() will only
+ * include the numerical error value provided.
+ */
+extern const int StrError_R_BufSize;
+
+/**
+ * @brief XSI-compliant version of \c strerror_r()
+ *
+ * See https://linux.die.net/man/3/strerror for more details.
```
peternewman: As mentioned, Doxygen explicitly mentions URLs with @see, so I think doing that and dropping the superflous words around it would make better use of their special styling.

include/ola/base/StrError_R.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+int StrError_R_XSI(int errnum, char* buf, size_t buflen);
+
+/**
+ * @brief Convenience function that wraps @ref StrError_R_XSI().
+ * 
+ * @param errnum error value to generate the textual description for.
+ * @return Textual description of the error value. If the description is
+ * trucnated due to an insufficient buffer length, the description will
```
peternewman: SPaG truncated

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+namespace ola {
+
+const int StrError_R_BufSize(1024);
+
+std::string StrError_R(int errnum) {
+  // Pre-allocate to prevent re-allocation.
+  std::string out(StrError_R_BufSize, '\0');
+  if (StrError_R_XSI(errnum, &(out[0]), out.size())) {
+    out.assign("errno = ");
```
peternewman: I'm assuming a stringstream doesn't allow access to a buffer to use it throughout this code? However what's the hit, aside from a bit of memory, with using an ostringstream in the error case, and then appending that to out, see e.g. https://github.com/OpenLightingProject/ola/blob/master/common/base/Version.cpp#L43-L47 . Which would completely avoid all the edge case handling you've had to write.

include/ola/base/StrError_R.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+/**
+ * @brief Length of the internal buffer used for @ref StrError_R().
+ * 
+ * If the length of the system-provided error description exceeds the length
+ * of this buffer minus one, then the output of @ref StrError_R() will only
+ * include the numerical error value provided.
+ */
+extern const int StrError_R_BufSize;
```
peternewman: We usually `static const`, is there a reason you need extern?

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <string.h>
+
+#include <limits>
+#include <string>
+
+#include "ola/base/StrError_R.h"
+
+namespace ola {
+
+const int StrError_R_BufSize(1024);
```
peternewman: Again our style is ...BufSize = 1024;

common/base/StrError_R.cpp
shenghaoyang: Hmm... that could be done, but r is known to be non-negative from this if onwards, but I guess this would be moot if we were to shift to using an `ostringstream`.

common/base/StrError_R.cpp
shenghaoyang: Could be done, just didn't feel like dragging the whole setup associated with the stream into something as trivial as representing an integer as text - really looking forward to the availability of `to_string` as we (hopefully) move towards `C++11` in the future.

include/ola/base/StrError_R.h
shenghaoyang: If we `static const` here - then every compilation unit that includes this header file would have their own copy of `StrError_R_BufSize`. `extern` here just lets me allocate storage for it in that one unit and have the other units referencing this variable link to it.

shenghaoyang: We could `#define` stuff.... but I'd want to avoid that route - maybe future refactoring to `constexpr` or `inline constexpr` in future?

common/base/StrError_R.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace ola {
+
+const int StrError_R_BufSize = 1024;
+
+std::string StrError_R(int errnum) {
+  // Pre-allocate to prevent re-allocation.
+  std::string out(StrError_R_BufSize, '\0');
+  if (StrError_R_XSI(errnum, &(out[0]), out.size())) {
+    std::ostringstream errs("errno = ");
```
peternewman: Sorry, I realised after I said this, I should just get you to use our existing function to do this ola::strings::IntToString :
http://docs.openlighting.org/ola/doc/latest/namespaceola_1_1strings.html#a1c6637c4498627e405cf8c95cef677c7

shenghaoyang:
Yikes. Wish I'd found that out earlier. Will push a commit in a tick.

peternewman:
Sorry @shenghaoyang I didn't think of it when I was commenting; TBH I'm used to more grown-up languages where you don't have to hand code such functions!

shenghaoyang:
> Sorry @shenghaoyang I didn't think of it when I was commenting; TBH I'm used to more grown-up languages where you don't have to hand code such functions!

Well... `C++11` is more grown up than `C++03` :). There's so much that can be torn out when that's switched to.

peternewman:
I was impressed to see the + string concatenation was in C++98!

configure.ac
```diff
@@ -234,6 +234,9 @@ AM_CONDITIONAL(HAVE_EPOLL, test "${ax_cv_have_epoll}" = "yes")
 AC_CHECK_FUNCS([kqueue])
 AM_CONDITIONAL(HAVE_KQUEUE, test "${ac_cv_func_kqueue}" = "yes")
 
+# strerror_r
+AM_CONDITIONAL([HAVE_STRERROR_R], [test "x$ac_cv_func_strerror_r" = "xyes"])
```
peternewman: I think you can drop the x's and square brackets here, looking at HAVE_KQUEUE just above, which will have had broader OS testing.

include/ola/base/StrError_R.h
peternewman: One for @nomis52 then as to why we've gone for static const over extern const?

shenghaoyang:
> Would you mind adding an example systemd file that uses this functionality too please?

Sure. I can use the unit file used for testing. What sort of systemd version are we targeting? Arch is around `systemd 239`, and I think some of the features I'm using _might_ not be available in mainstream distributions - a quick look on the Ubuntu package search service shows versions 204 - 237 in use.

peternewman:
I'd probably aim at Debian Stretch, as that's where our deb package will go first, which is on 232:
https://packages.debian.org/stretch/systemd

I guess you could add the latest stuff in another file somewhere or an issue or something (or comments in the file)? Assuming the features aren't backwards compatible/fail safe.

But there are Arch packages too, so I don't know how best we offer it to people like them too?

peternewman:
Indeed it will, e.g. https://wiki.debian.org/Teams/pkg-systemd/Packaging

There's probably something in Debian Policy too.

You shouldn't need any real privileges, aside from some permissions on the devices for USB dongles etc. That sounds about right, I don't know much about Arch/AUR at all TBH.

The udev uaccess stuff came in here:
https://github.com/OpenLightingProject/ola/pull/1430/commits/ce5d6db4f4dc17d43b622ec0717c49a46353a2e1
From:
https://github.com/OpenLightingProject/ola/commit/c679560b22555495a9e806707d13664411ab0b5c
Which is from:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=840389

But that's about all I really know about that bit too...

I think there's enough of a can of worms between init and systemd anyway, despite different bits of systemd. I don't think the systemd file needs to be completely gold plated, just lint clean for Debian/compliant with their minimum requirements and to allow people to actually make use of the functionality you've added in without them really needing to think for themselves!

shenghaoyang:
[RFC] I've added in a systemd unit file for a user instance of systemd. Try copying the unit file from `systemd/olad.service` to `~/.config/systemd/user/` and invoke:

- `systemd --user daemon-reload`
- `systemd --user start olad`

When you test this.

Hopefully the options I've chosen are compatible across most systems.

Cheers

Shenghao

systemd/olad.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=Ola Daemon (User mode)
+Documentation=https://github.com/OpenLightingProject/ola
```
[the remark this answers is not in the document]
shenghaoyang: No problems. Any other suggestions? I probably ought to write some kind of documentation article showing how this systemd integration is accomplished as well.

systemd/olad.service
```diff
@@ -0,0 +1,30 @@
...
+ProtectHome=tmpfs
+ProtectControlGroups=true
+RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6
+RestrictNamespaces=true
+LockPersonality=true
+MemoryDenyWriteExecute=true
+SystemCallArchitectures=native
+
+# Directory specifications
+ConfigurationDirectory="~/.ola/"
```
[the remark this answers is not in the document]
shenghaoyang: Yeah, I chose not to add that in. Since `olad` appears to default to the user configuration directory, it'd be safe to leave it as it is. If we're developing a unit file for a system wide daemon, then some modification would be required (as well distribution-specific group and user settings).

shenghaoyang:
> Does systemd allow you to depend on networks being up? I think I've been round this loop before with it, but I thought I'd ask.

Yes, you can, but this is a bit of a mixed bag. There are targets you can add dependencies on to wait for the network, but the loopback interface should be initialized by `systemd` directly before it brings up any units. Since the problems I've seen are from `olad` trying to listen on loopback before that is initialized (I think I've seen it in one of the issues), I don't think that's gonna be an issue.

You can take a look at some of the targets that can be added as dependencies for startup:
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/, but I'm not sure if they apply if we simply have a dependency on the loopback interface (for RPC). If we're listening on the wildcard address (for plugins), then we shouldn't even have to worry about other network interfaces except loopback. If we're not, however, then it's another can of worms.....

peternewman:
> > Does systemd allow you to depend on networks being up? I think I've been round this loop before with it, but I thought I'd ask.
> 
> Yes, you can, but this is a bit of a mixed bag. There are targets you can add dependencies on to wait for the network, but the loopback interface should be initialized by `systemd` directly before it brings up any units. Since the problems I've seen are from `olad` trying to listen on loopback before that is initialized (I think I've seen it in one of the issues), I don't think that's gonna be an issue.
> 
> You can take a look at some of the targets that can be added as dependencies for startup:
> https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/, but I'm not sure if they apply if we simply have a dependency on the loopback interface (for RPC). If we're listening on the wildcard address (for plugins), then we shouldn't even have to worry about other network interfaces except loopback. If we're not, however, then it's another can of worms.....
 
Some of the plugins (and indeed bits of the core), want our IP for various reasons. Firing off a SIGHUP will reload the plugins and should resolve most such issues.

> EDIT: I've noticed the build to be failing quite a bit due to the placement of the files in the `/systemd` directory. Is there another more appropriate location to place these files?

I assume you mean this: https://travis-ci.org/OpenLightingProject/ola/jobs/427037785#L7207

You need to either add the file to a Makefile (add one in systemd folder), or add it to the ignore list.

systemd/olad.service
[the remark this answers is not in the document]
peternewman: Not that I can immediately think of. Yeah some docs to put on the website or wiki would be good.

I guess also the man pages:
https://docs.openlighting.org/ola/man/

Actually reading the page I linked to, it says "It is a good idea to first reference documentation that explains what the unit's purpose is, followed by how it is configured, followed by any other related documentation.", so it should probably be the website, then the man pages, then conf, then GitHub.

PR #1448, opened 2018-07-15
Update some copyright

gobo-ws:


nomis52 (approved):

peternewman:
I was going to be a smart arse and say shouldn't we have code changes for that, but some JS has actually changed this year, although I don't think it would be visible to end users. Perhaps we should look at some JS magic to keep that year fresh?

gobo-ws:
Yeah,  using `Date().getFullYear` in JS for instance.
I noticed that there in no copyright notice in the "New UI" but maybe it is not needed.

peternewman:
Yes, I double checked that one too, and the RDM UI seems to have nothing at all (at least looking at the source).

What's your take @nomis52 ? I know in the source code we stopped doing date ranges some time ago, and just had a starting date. Can we cheat and just use JS? If so, how does that work if someone uses it on a PC with the clock set wrong, is the copyright invalid?

PR #1450, opened 2018-07-16
Detach kernel driver to allow claiming of USB interface

MarkMuth:
Hi I'm using a eurolite USB-DMX512-PRO Interface on Ubuntu 18.04 and ran into several problems.

First I had to create a new udev rule so that libusb_open wouldn't fail in LibUsbAdaptor.cpp. It looks like this:

SUBSYSTEM=="usb", ATTR{idVendor}=="04d8", ATTR{idProduct}=="fa63", MODE="0660", GROUP="dialout"

After creating this rule and rebooting /dev/ttyACM0 was created. So some kind of kernel driver was loaded for the eurolite interface.

This however caused libusb_claim_interface to fail. In this pull request I suggest to detach the kernel driver in case of an error and to retry claiming the USB interface.

Ideally the kernel driver should be reattached after closing olad. But that's not done yet. I thought maybe I'm missing something here.

### DIFF
--- a/libs/usb/LibUsbAdaptor.cpp
+++ b/libs/usb/LibUsbAdaptor.cpp
@@ -80,10 +80,22 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
   }
 
   int r = libusb_claim_interface(*usb_handle, interface);
-  if (r) {
+  if (r == LIBUSB_ERROR_BUSY) {
+	int error = libusb_detach_kernel_driver(*usb_handle, interface);
+	if (error) {
+		OLA_WARN << "Failed to detach kernel driver " << interface
+				 << " on device: " << usb_device;
+	} else {
+	  r = libusb_claim_interface(*usb_handle, interface);
+	}
+  }
+
+  if (r)
+  {
     OLA_WARN << "Failed to claim interface " << interface
              << " on device: " << usb_device << ": "
              << LibUsbAdaptor::ErrorCodeToString(r);
+
     libusb_close(*usb_handle);
     *usb_handle = NULL;
     return false;
--- a/libs/usb/LibUsbAdaptor.cpp
+++ b/libs/usb/LibUsbAdaptor.cpp
@@ -83,8 +83,8 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
   if (r == LIBUSB_ERROR_BUSY) {
     int error = libusb_detach_kernel_driver(*usb_handle, interface);
     if (error) {
-         OLA_WARN << "Failed to detach kernel driver for interface " << interface
-                  << " on device: " << usb_device;
+      OLA_WARN << "Failed to detach kernel driver for interface " << interface
+               << " on device: " << usb_device;
     } else {
       r = libusb_claim_interface(*usb_handle, interface);
     }
--- a/libs/usb/LibUsbAdaptor.cpp
+++ b/libs/usb/LibUsbAdaptor.cpp
@@ -81,21 +81,18 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
 
   int r = libusb_claim_interface(*usb_handle, interface);
   if (r == LIBUSB_ERROR_BUSY) {
-	int error = libusb_detach_kernel_driver(*usb_handle, interface);
-	if (error) {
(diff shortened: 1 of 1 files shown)


peternewman:
Hi Mark,

Briefly, our config for various devices is here:
https://wiki.openlighting.org/index.php/OLA_Device_Specific_Configuration

And our udev rules here:
https://github.com/OpenLightingProject/ola/blob/master/debian/ola.udev

MarkMuth:
Hi! Thank you for the links. Using the instructions udev rule + rmmod I can control the interface now with the ola version (0.10.5) shipped with Ubuntu 18.04.

Still I wonder, what is the point about instructing udev to create a device node for my interface if I later on just use rmmod to destroy it? Isn't there a way to instruct udev to ignore my eurolite from the beginning?

peternewman:
So you have to rmmod, even with the udev rule? I'd imagine there may be a way to use udev to tell it not to load a module, but I don't know off the top of my head.

libs/usb/LibUsbAdaptor.cpp
```diff
@@ -80,10 +80,22 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
   }
 
   int r = libusb_claim_interface(*usb_handle, interface);
-  if (r) {
+  if (r == LIBUSB_ERROR_BUSY) {
+	int error = libusb_detach_kernel_driver(*usb_handle, interface);
+	if (error) {
+		OLA_WARN << "Failed to detach kernel driver " << interface
+				 << " on device: " << usb_device;
```
peternewman: Can you align the << please

libs/usb/LibUsbAdaptor.cpp
```diff
@@ -80,10 +80,22 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
...
+	} else {
+	  r = libusb_claim_interface(*usb_handle, interface);
+	}
+  }
+
+  if (r)
+  {
```
peternewman: Can you move this to the line above please.

libs/usb/LibUsbAdaptor.cpp
```diff
@@ -80,10 +80,22 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
   }
 
   int r = libusb_claim_interface(*usb_handle, interface);
-  if (r) {
+  if (r == LIBUSB_ERROR_BUSY) {
+	int error = libusb_detach_kernel_driver(*usb_handle, interface);
+	if (error) {
+		OLA_WARN << "Failed to detach kernel driver " << interface
```
peternewman: Can you change the message to something like "Failed to detach kernal driver for interface" <<

libs/usb/LibUsbAdaptor.cpp
peternewman: The other tab/space lint errors will also need fixing before merge:
https://travis-ci.org/OpenLightingProject/ola/jobs/404585543#L2174

MarkMuth:
Should be better now ;)

libs/usb/LibUsbAdaptor.cpp
```diff
@@ -80,6 +80,15 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
   }
 
   int r = libusb_claim_interface(*usb_handle, interface);
+  if (r == LIBUSB_ERROR_BUSY) {
+    int error = libusb_detach_kernel_driver(*usb_handle, interface);
+    if (error) {
+         OLA_WARN << "Failed to detach kernel driver for interface " << interface
```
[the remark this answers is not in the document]
MarkMuth: Ok, there were too many leading white spaces. Removing them makes the line 79 characters long.

libs/usb/LibUsbAdaptor.cpp
```diff
@@ -80,6 +80,15 @@ bool OpenHandleAndClaimInterface(libusb_device *usb_device,
   }
 
   int r = libusb_claim_interface(*usb_handle, interface);
+  if (r == LIBUSB_ERROR_BUSY) {
+    int error = libusb_detach_kernel_driver(*usb_handle, interface);
```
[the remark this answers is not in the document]
MarkMuth: libusb_close() was called after libusb_claim_interface() failed, so I thought it should be called in any error situation.

PR #1456, opened 2018-07-31
Some changes for FreeBSD

RenZ0:
The first commit is for FreeBSD 10 (old one).
The second one is for FreeBSD 10.4 and 11.2, work in progress.

PR #1460, opened 2018-08-08
Changes to allow build on FreeBSD

RenZ0:


peternewman:
Based on #1456 .

The first commit is for FreeBSD 10 (old one).
The second one is for FreeBSD 10.4 and 11.2, work in progress.

common/testing/Makefile.mk
```diff
@@ -6,5 +6,6 @@ noinst_LTLIBRARIES += common/testing/libolatesting.la \
 common_testing_libolatesting_la_SOURCES = \
     common/testing/MockUDPSocket.cpp \
     common/testing/TestUtils.cpp
+common_testing_libolatesting_la_CXXFLAGS = $(COMMON_TESTING_PROTOBUF_FLAGS)
```
peternewman: We need @RenZ0 to check the exact message you were getting for this please and I'll see if we can use a more bespoke and appropriate set of flags? This will be in some time when he can restart the machine again.

peternewman:
Also blocked by upstream https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=230282 .

peternewman:
The upstream bug has now been merged!

common/testing/Makefile.mk
RenZ0: It is on FreeBSD 10.0-RELEASE. I can't reproduce that, I guess I had the message because I started from the release and I didn't run autoreconf. So this change is not needed.

peternewman: Okay, yeah great, lets revert this change then.

RenZ0:
Tests are now fixed as well.

ola/AutoStart.cpp
```diff
@@ -110,7 +110,11 @@ TCPSocket *ConnectToServer(unsigned short port) {
     // Try to start the server, we pass --daemon (fork into background) and
     // --syslog (log to syslog).
     execlp("olad", "olad", "--daemon", "--syslog",
-           reinterpret_cast<char*>(NULL));
+    #ifdef __FreeBSD__
+                    reinterpret_cast<char*>(0));
```
peternewman: Can we align this, and the line below with the ( above please.

RenZ0: Done, with some more occurences :)

RenZ0:
Indentation should be proper now.

common/network/PosixInterfacePicker.cpp
```diff
@@ -226,7 +226,7 @@ vector<Interface> PosixInterfacePicker::GetInterfaces(
         interface.index = ifrcopy.ifr_index;
 #else
         interface.index = ifrcopy.ifr_ifindex;
-#endif  // SIOCGIFINDEX
+#endif  // __FreeBSD__
```
peternewman: Good spot!

peternewman:
You could do @RenZ0 , but given it's unsupported it's probably not worth it.

neheb:
Same issue on musl 1.2.3 as well. Why not just make it 0 everywhere?

peternewman:
> Same issue on musl 1.2.3 as well. Why not just make it 0 everywhere?

Followed up in #1773 .

PR #1471, opened 2018-08-17
Fix typo and dead link in developer readme

kecramer:
http://www.aleph1.co.uk/chapter-10-arm-structured-alignment-faq has been hijacked and no longer sends the user to the desired destination. I believe the replaced link might work as a replacement.

### DIFF
--- a/README.developer
+++ b/README.developer
@@ -101,7 +101,7 @@ Structure packing syntax differs between platforms and compilers. If you declare
 a struct or class that needs to be packed, use the PACK() macro from
 ola/base/Macro.h . See below for an example of PACK() usage.
 
-Not that this macro doesn't work for enums. See plugins/espnet/EspNetPackets.h
+Note that this macro doesn't work for enums. See plugins/espnet/EspNetPackets.h
 for an example of how to pack an enum.
 
 Non x86 platforms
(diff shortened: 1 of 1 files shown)


peternewman:
Thanks @kecramer .

The other option is the Wayback Machine:
http://web.archive.org/web/20170608071911/http://www.aleph1.co.uk:80/chapter-10-arm-structured-alignment-faq

Curiously they were linking to the stress stuff even when the pages were still active...

Would you mind targeting our 0.10 branch instead with this fix, so we can get it out in our next release (or if not I can cherry-pick your commit)?

There's certainly a lot of commonality between the two documents, I wonder which came first?

kecramer:
Hi @peternewman, I'm not sure how to target the 0.10 branch. Is it just rebasing off the 0.10 and then requesting the merge into that branch instead of master? Only switching the target branch introduced a lot of commits into this PR which I don't think is intended.

The osuosl.org link appears to be nearly the same content (with added author info). I think it's unlikely for that link to break since it's from Oregon State University's Open Source Lab. Not sure what aleph1.co.uk used to be - maybe a personal site of a developer? Anyways, if the archive link is preferred, I can change it to that.

peternewman:
I think a rebase should work, but my git-fu is still rather weak. Basically we want your commit, or commits, on top of the 0.10 branch, rather than the master one (without all those other commits).

The new link is fine, having the author attribution on that page is probably better too, given the content seems to be identical in places.

kecramer:
@peternewman my git-fu was also letting me down, so I took the easy way out and just made a new PR where the patch is based on the 0.10 branch. It is available here: https://github.com/OpenLightingProject/ola/pull/1472

PR #1484, opened 2018-09-03
Use array.tobytes() instead of array.tostring() for Pyton 3.2 and higher

fschuetz04:
According to https://docs.python.org/3/library/array.html#array.array.tostring, `array.tostring()` is a deprecated alias for `array.tobytes()`, which was introduced in Python 3.2. With this change, `array.tobytes()` is used for Python 3.2 and higher.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.gcda
 *.gcno
 *.gcov
+.gch
 *.la
 *.lo
 *.log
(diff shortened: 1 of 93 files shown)


fschuetz04:
Is that correct like that?

peternewman:
Unfortunately not @fschuetz04 as it's also pulled in all the other changes different between the two. You'll need to either rebase or cherry-pick into a 0.10 branch.

PR #1486, opened 2018-09-07
Use array.tobytes() instead of array.tostring() for Pyton 3.2 and higher

fschuetz04:
According to https://docs.python.org/3/library/array.html#array.array.tostring, `array.tostring()` is a deprecated alias for `array.tobytes()`, which was introduced in Python 3.2. With this change, `array.tobytes()` is used for Python 3.2 and higher.

### DIFF
--- a/python/ola/OlaClient.py
+++ b/python/ola/OlaClient.py
@@ -18,6 +18,7 @@
 import array
 import socket
 import struct
+import sys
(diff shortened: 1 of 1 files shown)


peternewman:
Replaces https://github.com/OpenLightingProject/ola/pull/1484

PR #1495, opened 2018-09-22
Backport some Debian changes and add another Debian test

peternewman:
A mix of old stuff and a little new.

@yoe I notice your order is different here:
https://github.com/OpenLightingProject/ola/compare/0.10...yoe:master?expand=1#diff-d837a1c17de0268bcea321239ddbed47L16

Is yours preferred by policy, I'd assumed real stuff, then variable substitution would be best? I thought it would be nice to standardise to simplify future merges, but happy to go either way.

### DIFF
--- a/debian/Makefile.mk
+++ b/debian/Makefile.mk
@@ -31,4 +31,5 @@ EXTRA_DIST += \
     debian/source/lintian-overrides \
     debian/source/local-options \
     debian/tests/control \
+    debian/tests/hw.cc \
     debian/watch
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: ola
 Priority: optional
 Maintainer: Wouter Verhelst <wouter@debian.org>
 Uploaders: RenZO <renzo@imaginux.com>
-Build-Depends: debhelper (>= 9), autotools-dev, dh-autoreconf, bash-completion, libcppunit-dev, bison, flex, pkg-config, uuid-dev, python, python-protobuf, libprotobuf-dev, protobuf-compiler, libprotoc-dev, libusb-1.0-0-dev, libftdi-dev, liblo-dev, libmicrohttpd-dev, libncurses5-dev, libavahi-client-dev, python-numpy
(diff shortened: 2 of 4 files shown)


yoe:
You mean that depends line? It's not policy-preferred, it's just Wouter-preferred ;-) it's just cleaner that way, and also consistent with the other depends lines in the same file.

peternewman:
Oops, yeah I see the link didn't work. Yeah the depends line, I was thinking of the ola dependency (with only the version as a variable) as different from the fully variable based misc:depends. I've made upstream match yours now, to try and reduce the diff size when comparing in future.

yoe:
Sorry, I don't see what you're talking about. The only differences in `debian/control` between 0.10 and my master branch are:

- libftdi1-dev vs libftdi-dev in Build-Depends, for obvious reasons
- libmicrohttpd-dev version restriction in Build-Depends; I can probably drop that now, was for a bug in libmicrohttpd-dev that broke our build.
- Vcs-* headers pointing to my git repository rather than yours
- Depends: ordering of libola-dev where we add the libprotobuf-dev dependency at the end rather than in the middle (looks cleaner ;-)
- A Replaces: in ola-rdm-tests for old versions of the ola package (not sure why anymore, but I'm sure there's a good reason ;-)
- I add dependencies on javascript libraries since my packages don't ship them

So other than the reordering of the libprotobuf-dev thing, which I gather you didn't mean, there isn't anything that might be related. So what are we talking about? ;-)

peternewman:
Sorry @yoe I'm clearly not doing too well at expressing myself today! I was referring to this diff:
```
Depends: ola (= ${binary:Version}), libprotobuf-dev, ${misc:Depends}
Depends: ola (= ${binary:Version}), ${misc:Depends}, libprotobuf-dev
```

Anyway I've updated our order to match yours, so I'll shut up and merge this now. :laughing:

yoe:
Ah, heh, so it was the thing I thought it was at first :) glad we got that cleared up.

PR #1504, opened 2018-10-26
Fix a couple of RDM test bugs

peternewman:
Correct our behaviour if we get invalid interface indices.


peternewman:
Thanks @nomis52 . Merged after fixing some new flake8 warnings and silencing all deprecated declarations on Linux Travis due to log length issues.

PR #1512, opened 2018-11-13
Add a screenshot mode which hides private data, also more RDM test and model collector improvements

peternewman:


peternewman:
Blocked behind https://github.com/OpenLightingProject/ola/pull/1515

olad/OladHTTPServer.cpp
```diff
@@ -442,6 +443,15 @@ int OladHTTPServer::JsonServerStats(const HTTPRequest*,
 #endif  // _WIN32
 
   JsonObject json;
+#ifdef OLA_SCREENSHOT_MODE
+  json.Add("hostname", "olad-server");
+  json.Add("instance_name", OLA_DEFAULT_INSTANCE_NAME);
+  json.Add("config_dir", "/etc/ola/");
+  json.Add("ip", "10.79.76.65");
```
nomis52: We should use '***' or something similar rather than making up values.

peternewman:
Blocked behind https://github.com/OpenLightingProject/ola/pull/1518

olad/OladHTTPServer.cpp
peternewman: It spells 10.O.L.A in ASCII! I thought semi-genuine data would be more helpful in a screenshot, rather than lots of empty values. It would also mean you could look at the layout and styling realistically.

olad/OladHTTPServer.cpp
```diff
@@ -442,6 +443,15 @@ int OladHTTPServer::JsonServerStats(const HTTPRequest*,
...
+  json.Add("broadcast", "10.79.76.255");
+  json.Add("subnet", "255.255.255.0");
+  json.Add("hw_address", "01:ad:01:ad:01:ad");
```
peternewman: Hopefully this one is less subtle!

olad/OladHTTPServer.cpp
peternewman: Done

olad/OladHTTPServer.cpp
peternewman: Now made generic/obviously fake.

PR #1514, opened 2018-11-19
Update ClientWrapper.py

jimruxton:
Added the comparison to Event class so examples are compatible with python3 :  
 def __lt__(self,other):
    return self._run_at< other._run_at

### DIFF
--- a/python/ola/ClientWrapper.py
+++ b/python/ola/ClientWrapper.py
@@ -46,6 +46,9 @@ class _Event(object):
 
   def __cmp__(self, other):
     return cmp(self._run_at, other._run_at)
+  
+  def __lt__(self,other):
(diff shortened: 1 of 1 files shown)


peternewman:
Hi @jimruxton , are you interested in trying to fix the comments either for our master branch or ideally for our 0.10 branch? I think they should be fairly easy to do (I've not tried it, but I think in theory you can take the suggested fixes from my comments and easily integrate them, see https://help.github.com/articles/incorporating-feedback-in-your-pull-request/#applying-a-suggested-change ).

peternewman:
I'm closing this as it's been superceded by @brucelowekamp and others work in #1605 which is in 0.10 and will be in master shortly. Thanks for your efforts on it anyway @jimruxton .

PR #1520, opened 2018-12-06
Fix some RDM discovery bugs

peternewman:


peternewman:
@nomis52 I think the changes to DiscoveryAgent*.cpp from here should be fairly safe, in which case I'm inclined to backport them to 0.10, given it's a rather spectacular bug.

There's a slightly more optimised fix I'm working on too, which I guess should just go into master, which does the subsequent DUB either side of the phantom/non-muting responder, to attempt to avoid the issue.

peternewman:
Backporting done in #1521

PR #1535, opened 2019-02-04
Various improvements including fixing range validation

peternewman:
Also add some tests for the fix so it doesn't happen again


peternewman:
I've taken out the Protobuf 3.7 stuff for now, to get the other bits merged.

PR #1537, opened 2019-02-15
Fix comparison methods for python3

nils-van-zuijlen:
See https://portingguide.readthedocs.io/en/latest/comparisons.html#rich-comparisons for an explaination of why it was broken.

In particular, I was stuck with a bug in the `_Event` class because heaps are ordered and I wanted to add multiple events to it, which failed because they had no suitable comparison functions.

PR #1538, opened 2019-02-17
Fix comparison methods for python3

nils-van-zuijlen:
Rebased to merge in 0.10, original PR : #1537


python/ola/PidStore.py
```diff
@@ -160,6 +160,21 @@ def ValidateAddressing(self, args, command_class):
   def __cmp__(self, other):
     return cmp(self._value, other._value)
 
+  def __eq__(self, other):
```
peternewman: Related comment, this is only checking value, not other fields.

tools/rdm/TestDefinitions.py
```diff
@@ -6370,13 +6370,13 @@ def CheckFieldIsZero(self, fields, key):
 
   def CheckFieldIsBetween(self, fields, key, min_value, max_value):
     if fields[key] < min_value:
-          self.AddWarning(
```
peternewman: I'm guessing this is flake8 fixes?

peternewman:
You'll need to cherry-pick this commit to fix some of the Codespell issues:
https://github.com/OpenLightingProject/ola/commit/34c94331a5d89fe3102af34adc8e5f4aaf9a64d8

And https://github.com/OpenLightingProject/ola/pull/1532/commits/458fbb96d7af20b8ab099894450b24ce0019c054

And others which are probably in https://github.com/OpenLightingProject/ola/pull/1532, but likely jumbled with other bits unfortunately.

tools/rdm/TestState.py
```diff
@@ -15,6 +15,7 @@
 # TestState.py
 # Copyright (C) 2011 Simon Newton
 
+
```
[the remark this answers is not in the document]
nils-van-zuijlen: Yes it is.

AUTHORS
```diff
@@ -18,6 +18,8 @@ Contributors:
   Joshua Moyerman, ArtNet fixes.
   Masaki Muranaka, various patches
   Nicolas, for the win32 port of libartnet
+  Nicolas Bertrand, added ShowJockey DMX-U1 support
```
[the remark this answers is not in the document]
nils-van-zuijlen: Oops, yes.

python/ola/ClientWrapper.py
```diff
@@ -32,6 +33,7 @@
 __author__ = 'nomis52@gmail.com (Simon Newton)'
 
 
+@functools.total_ordering
```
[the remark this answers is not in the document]
nils-van-zuijlen: Yes, rebasing problems, I added it in the 1st commit, and removed it in the second, but cherry-pick did not like it, apparently.

python/ola/PidStore.py
nils-van-zuijlen: I copied what `__cmp__` did. Maybe not the best choice.

tools/rdm/TestDefinitions.py
nils-van-zuijlen: Yes.

nils-van-zuijlen:
But cherry picking commits from `master` will make horrible merge conflicts when you will merge master in 0.10 (or the opposite).

Maybe I should go back to merging it into master, no ?

python/ola/ClientWrapper.py
```diff
@@ -47,6 +49,12 @@ def __init__(self, time_ms, callback):
   def __cmp__(self, other):
     return cmp(self._run_at, other._run_at)
 
+  def __eq__(self, other):
+    return self._run_at == other._run_at and self._callback == other._callback
```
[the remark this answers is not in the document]
nils-van-zuijlen: Python sorting algorithms only checks with `<`. Cf https://docs.python.org/3/library/stdtypes.html#list.sort

peternewman:
> But cherry picking commits from `master` will make horrible merge conflicts when you will merge master in 0.10 (or the opposite).
> 
> Maybe I should go back to merging it into master, no ?

My git-fu is probably rubbish, but I normally seem to get away with it, I've done it a few times.

I'd like to get the bugfix into 0.10 if possible.

nils-van-zuijlen:
Normally I have addressed all the changes you requested. You may merge now that Travis is on a happy mood :smile:

python/ola/ClientWrapper.py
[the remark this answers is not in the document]
peternewman: I still generally feel __eq__ and __cmp__ should return the same, the easiest way may be something like `cmp([run_at, callback], [run_at, callback])` I think from https://docs.python.org/2.3/ref/comparisons.html .

From here, it looks like we should be either implementing hash, or making it unhashable, which I think was part of the initial reason this was raised:
https://docs.python.org/2/reference/datamodel.html#object.__hash__

python/ola/OlaClient.py
```diff
@@ -963,6 +964,10 @@ def RegisterUniverse(self, universe, action, data_callback, callback=None):
     Returns:
       True if the request was sent, False otherwise.
     """
+
+    if data_callback is None and action == self.REGISTER:
```
peternewman: Good spot, I guess this stops an uncaught exception occurring.

python/ola/PidStore.py
```diff
@@ -160,6 +160,25 @@ def ValidateAddressing(self, args, command_class):
...
+    return self.value == other.value \
```
peternewman: I think we favour behaving like C++, putting the and on the previous lines and aligning the self.values.

python/ola/ClientWrapper.py
```diff
@@ -47,6 +47,21 @@ def __init__(self, time_ms, callback):
...
+
+  def __lt__(self, other):
```
peternewman: Dunno how we should handle these bits then though...

python/ola/PidStore.py
```diff
@@ -160,6 +160,25 @@ def ValidateAddressing(self, args, command_class):
...
+      and self.name == other.name \
+      and self._requests == other._requests \
+      and self._responses == other._responses \
+      and self._validators == other._validators
+
+  def __lt__(self, other):
```
peternewman: Similar comment.

python/ola/ClientWrapper.py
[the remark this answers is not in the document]
nils-van-zuijlen: You are reading documentation for an old version of python. Current supported (by the Python Foundation) versions are 3.5-3.7 and 2.7.

As stated [here](https://docs.python.org/2.7/reference/datamodel.html?highlight=cmp#object.__cmp__), `__cmp__` is not required if rich comparison is implemented, so we should remove it.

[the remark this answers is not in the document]
nils-van-zuijlen: And `__hash__` should only be implemented on constant objects, otherwise it could mess up hash tables (python dict)

nils-van-zuijlen:
It works !
Does it do what you wanted ?

python/ola/PidStore.py
peternewman: Can we align the `self.` above each other, or does flake8 complain?

nils-van-zuijlen: No we can't.

python/ola/ClientWrapper.py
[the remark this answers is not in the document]
peternewman: Oops, well spotted. I assume the 2.3 note about cmp() still applies for the function in later versions. We want `__hash__` for the others though don't we?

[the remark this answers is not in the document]
peternewman: Looks like you're spot on, or have gone above and beyond:
https://github.com/python/cpython/blob/0baa72f4b2e7185298d09cf64c7b591efcd22af0/Lib/asyncio/events.py#L97-L145

python/ola/ClientWrapper.py
peternewman: Again this is above and beyond or matching Python:
https://github.com/python/cpython/blob/0baa72f4b2e7185298d09cf64c7b591efcd22af0/Lib/asyncio/events.py#L97-L145

python/ola/ClientWrapper.py
```diff
@@ -44,8 +44,30 @@ def __init__(self, time_ms, callback):
...
+    if other.__class__ is not self.__class__:
+      return NotImplemented
+    return self._run_at < other._run_at
+
+  def __le__(self, other):
+    if other.__class__ is not self.__class__:
+      return NotImplemented
+    return self._run_at <= other._run_at
```
peternewman: Can we change this to do the Python style, so < or `__eq__`:
https://github.com/python/cpython/blob/0baa72f4b2e7185298d09cf64c7b591efcd22af0/Lib/asyncio/events.py#L97-L145

python/ola/ClientWrapper.py
```diff
@@ -44,8 +44,30 @@ def __init__(self, time_ms, callback):
...
+
+  def __gt__(self, other):
+    if other.__class__ is not self.__class__:
+      return NotImplemented
+    return self._run_at > other._run_at
+
+  def __ge__(self, other):
```
peternewman: Again

python/ola/OlaClient.py
```diff
@@ -87,23 +87,33 @@ def __repr__(self):
...
+    if other.__class__ is not self.__class__:
+      return NotImplemented
     return self.id < other.id
 
   def __eq__(self, other):
+    if other.__class__ is not self.__class__:
+      return False
     return self.id == other.id
 
-  # These 4 could be replaced by functools.total_ordering when support
```
peternewman: Any reason for dropping the comment?

python/ola/OlaClient.py
```diff
@@ -179,25 +189,30 @@ def __repr__(self):
...
+    if other.__class__ is not self.__class__:
+      return NotImplemented
     return self.alias > other.alias
 
   def __ge__(self, other):
+    if other.__class__ is not self.__class__:
+      return NotImplemented
     return self.alias >= other.alias
 
-  def __ne__(self, other):
```
peternewman: How come this has been dropped?

python/ola/PidStore.py
```diff
@@ -157,8 +157,34 @@ def ValidateAddressing(self, args, command_class):
         return False
     return True
 
-  def __cmp__(self, other):
-    return cmp(self._value, other._value)
+  def __eq__(self, other):
+    if other.__class__ is not self.__class__:
+      return False
+    return self.value == other.value and \
```
peternewman: From https://github.com/python/cpython/blob/0baa72f4b2e7185298d09cf64c7b591efcd22af0/Lib/asyncio/events.py#L97-L145 it looks like round brackets means we can drop the backslash and fix the self. alignment.

python/ola/PidStore.py
```diff
@@ -157,8 +157,34 @@ def ValidateAddressing(self, args, command_class):
...
+
+  def __lt__(self, other):
+    if other.__class__ is not self.__class__:
+      return NotImplemented
+    return self._value < other._value
+
+  def __le__(self, other):
+    if other.__class__ is not self.__class__:
+      return NotImplemented
+    return self._value <= other._value
```
peternewman: Similar to the previous comment W.R.T. using `__eq__`

python/ola/MACAddress.py
```diff
@@ -51,18 +51,37 @@ def __hash__(self):
   def __repr__(self):
     return self.__str__()
 
-  def __cmp__(self, other):
```
nils-van-zuijlen: I wonder if this class should implement `__hash__` because anyone could write code like this which would change the value of the class :

```python
address = MACAddress(bytearray([0x01, 0x23, 0x45, 0x67, 0x89, 0xab]))  # creates a MACAddress
mac = address.mac_address  # gets its representation inside the class
print(hash(address))  # print the class' hash value, for example 1043151800894266584
mac[0] = 0xff  # changes the MAC address, shouldn't be possible but it is.
print(hash(address))  # prints the class' hash value, different from the first one, for example 6608185319733256019
```
And would get different results at the two print statements, without violating any python (implicit) rules.

python/ola/OlaClient.py
nils-van-zuijlen: No, by the way, do we still support python 2.6 ? The PSF does not support it anymore.

If we don't, I will happily remove the `le`, `gt` and `ge` methods from everywhere and put `@functools.total_ordering` on these classes.

python/ola/MACAddress.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, mac_address):
 
     Args:
       mac_address: The byte array representation of the MAC Address, e.g.
-      bytearray([0x01, 0x23, 0x45, 0x67, 0x89 0xab]).
+      bytearray([0x01, 0x23, 0x45, 0x67, 0x89, 0xab]).
```
peternewman: Good spot!

python/ola/OlaClient.py
nils-van-zuijlen: Oops, no, I remind now, functools.totalordering does only support `NotImplemented` since python 3.4. So we can't use it here until support for 2.7 is dropped.

nils-van-zuijlen: https://docs.python.org/3.7/library/functools.html#functools.total_ordering

python/ola/MACAddress.py
peternewman: Yep, this agrees it's bad:
https://stackoverflow.com/questions/13041352/python-quickly-hash-mutable-object

Going id based would solve our issue wouldn't it? Even if someone fiddles with the value, it's still the same object? Probably not ideal, but at least safer.

python/ola/OlaClient.py
peternewman: The 0.10 branch does, so I'd like to keep it, as this is only nominally a bug-fix. Please feel free to drop them from master and switch to functools.

peternewman: Dropping 2.7 isn't going to happen for a long time, Ubuntu Disco still has it:
https://packages.ubuntu.com/search?keywords=python&searchon=names&exact=1&suite=disco&section=all

As does 18.04:
https://packages.ubuntu.com/search?suite=bionic&exact=1&searchon=names&keywords=python

Which means it's live there until at least 2023!

Shame, given 2.7 supports it, but without NotImplemented:
https://docs.python.org/2.7/library/functools.html#functools.total_ordering

python/ola/MACAddress.py
peternewman: I don't suppose there's any equivalent to Doxygen code samples where we can validate that this example code compiles in Python land?

python/ola/OlaClient.py
nils-van-zuijlen: Inattention to detail. In python 3 its value is `not __eq__` by default, but it's not in python 2. My bad.

python/ola/MACAddress.py
```diff
@@ -51,19 +51,44 @@ def __hash__(self):
   def __repr__(self):
     return self.__str__()
 
-  def __cmp__(self, other):
-    if other is None:
-      return 1
-    return cmp(self.mac_address, other.mac_address)
+  def __eq__(self, other):
+    if other.__class__ is not self.__class__:
```
peternewman: Do we need to do:
`if other is None or other.__class__ is not self.__class__:`

To protect against other being None, or does None still have __class__?

python/ola/MACAddress.py
```diff
@@ -51,19 +51,44 @@ def __hash__(self):
...
 
   def __lt__(self, other):
+    if other is None:
+      return False
+    if other.__class__ is not self.__class__:
+      return NotImplemented
     return self.mac_address < other.mac_address
 
-  def __eq__(self, other):
+  # These 4 can not be replaced with functools.total_ordering because of the
```
peternewman: Would this be the same that they could "be replaced with functools.total_ordering when support for 2.7 is dropped because NotImplemented is not supported by it before 3.4"? Or even when we're on Python 3.4 will functools.total_ordering still not be sufficient?

python/ola/MACAddress.py
peternewman: We should still try and fix this please. It looks like a Python module just hashes the value:
https://github.com/drkjam/netaddr/blob/rel-0.7.x/netaddr/eui/__init__.py#L550-L552

But I'd agree we should do something or switch to "id".

python/ola/MACAddress.py
mvollrath: On the topic of `None.__class__`, anecdotally it has `__class__`:

```
$ python2
Python 2.7.15+ (default, Nov 27 2018, 23:36:35) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> None.__class__
<type 'NoneType'>
```

```
$ python3
Python 3.6.8 (default, Jan 14 2019, 11:02:34) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> None.__class__
<class 'NoneType'>
```

`isinstance(other, self.__class__)` would take care of this for you.

If `isinstance` isn't desired, it seems like a more bulletproof approach might be to wrap these comparisons in a `try` `except AttributeError` block, because it is ambiguous in [the documentation of `__class__`](https://docs.python.org/3/library/stdtypes.html#special-attributes) which instances have `__class__`.  In Python 3 a value not having a `__class__` attribute is an even more distant edge case.

Is there some reason this doesn't this return `NotImplemented` like the other operators?

nils-van-zuijlen: It doesn't return `NotImplemented` because if classes don't match, the two objects aren't equals.

python/ola/MACAddress.py
nils-van-zuijlen: No, we need to implement these functions ourselves because we have a special treatment applied to None that `functools.total_ordering` doesn't support.

python/ola/MACAddress.py
peternewman: Do you want to switch to isinstance @nils-van-zuijlen or is there some reason not to do so?

python/ola/MACAddress.py
peternewman: Is this because otherwise sorting None and a MAC you'd get stuck or something? Can we have a bit of a comment explaining why in the code please @nils-van-zuijlen ?

peternewman:
Sorry @nils-van-zuijlen I've been rubbish again. Any chance of some tests? Apart from that it essentially looks good to go thanks!

brucelowekamp:
I have tests for most of these written and should be able to submit a PR in a few days, though I'm not sure if I can add to this or need to open a new one. 

@peternewman is there a need to support python older than 2.7?  Quite a bit is simpler if we can assume 2.7 and recent 3. 

@nils-van-zuijlen 2.7 functools appears to have backports to support everything needed here

A couple more fixes and bugs found with the tests...

peternewman:
@nils-van-zuijlen or @brucelowekamp I think this can be closed now looking at the comparison:
https://github.com/OpenLightingProject/ola/compare/0.10...nils-van-zuijlen:0.10-python3

Would you agree this has all gone in via #1605?

nils-van-zuijlen:
Yes it can be closed, all meaningful commits are in #1605

peternewman:
Perfect and thanks again for your efforts @nils-van-zuijlen !

Issue #1344, opened 2017-12-17
Set Port IDs in OLA RDM messages


peternewman:
The standard says:
> 6.2.7.1 Port ID / Response Type field for Controller Generated Messages
> For Controller generated messages (GET_COMMAND, SET_COMMAND, and
> DISCOVERY_COMMAND), the Port ID field shall be set in the range of 1-255 identifying
> the Controller Port being used, such that the combination of Source UID and Port ID will uniquely
> identify the controller and port where the message originated.

It's not clear in our case if this should be per universe, or per plugin port (i.e. multiple ports per universe). I suspect it should be per plugin port, as a physical lighting desk with two ports would need two UARTs, and hence would likely have two port IDs.

Keeper-of-the-Keys:
I agree that each UART is supposed to have its' own port ID.

In the current setup on FTDI I think I'm not yet setting it but I can set it per widget as ong as they have unique UIDs there won't be collisions to confuse the core.

PR #1541, opened 2019-02-26
Initial FTDI RDM support

Keeper-of-the-Keys:
The support is not 100% according to the standard, it for instance does not check if after the minimum wait duration a full packet was actually received and if not waits longer.
This should of course only be a problem for devices that for some reason don't manage to send an RDM packet within normal/optimal timings.

I figure that adding that logic/better timings should be part of a separate later PR.


.gitignore
```diff
@@ -246,3 +246,4 @@ javascript/new-src/node_modules
 .cproject
 .settings
 .vscode/
+OLA.*
```
peternewman: I think you said this was for some IDE, can you make it more strict please, isn't OLA.* just based on your project name, so it won't actually work for other people?

plugins/ftdidmx/FtdiDmxPort.h
```diff
@@ -39,13 +43,14 @@ namespace ola {
 namespace plugin {
 namespace ftdidmx {
 
-class FtdiDmxOutputPort : public ola::BasicOutputPort {
+class FtdiDmxOutputPort
```
peternewman: Was this auto-formatting by the IDE? Can you remove it please? We've got some suitable editor configs: https://github.com/OpenLightingProject/editor-configs

Can you either use one of those, or tune yours and submit it if you're using an alternative editor.

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -42,7 +51,15 @@ FtdiDmxThread::FtdiDmxThread(FtdiInterface *interface, unsigned int frequency)
   : m_granularity(UNKNOWN),
     m_interface(interface),
     m_term(false),
-    m_frequency(frequency) {
+    m_frequency(frequency),
+    m_transaction_number(0),
+    m_discovery_agent(this),
+    m_uid(0x7a70, 0x12345678),
```
peternewman: Can we add a todo for this at least!

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +97,133 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
   }
 }
 
+void FtdiDmxThread::SendRDMRequest(ola::rdm::RDMRequest *request,
+                    ola::rdm::RDMCallback *callback) {
```
peternewman: Align with the (

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +97,133 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+
+/**
+ * Called when the discovery process finally completes
+ * @param callback the callback passed to StartFullDiscovery or
+ * StartIncrementalDiscovery that we should execute.
+ * @param status true if discovery worked, false otherwise
+ * @param uids the UIDSet of UIDs that were found.
+ */
+void FtdiDmxThread::DiscoveryComplete(ola::rdm::RDMDiscoveryCallback *callback,
+                                       bool,
```
peternewman: Can you switch to OLA_UNUSED

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +97,133 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+  if (callback) {
+    callback->Run(uids);
+  }
+}
+
+/**
+ * @brief Method called to cleanup any outstanding callbacks
+ * @param state
+ *
+ * All callbacks except the RDMCallback lack a way of reporting an error state to the caller.
```
peternewman: @note

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +97,133 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+
+void FtdiDmxThread::MuteDevice(const ola::rdm::UID &target,
+                               MuteDeviceCallback *mute_complete) {
+  ola::thread::MutexLocker locker(&m_rdm_mutex);
+  if(m_pending_request == nullptr) {
+    OLA_INFO << "Muting device";
+    m_mute_complete = mute_complete;
+    m_pending_request = ola::rdm::NewMuteRequest(m_uid, target, m_transaction_number += 1);
+  } else {
+      // Already pending request
```
peternewman: Can you fix the alignment

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +97,133 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+  }
+}
+
+void FtdiDmxThread::MuteDevice(const ola::rdm::UID &target,
+                               MuteDeviceCallback *mute_complete) {
+  ola::thread::MutexLocker locker(&m_rdm_mutex);
+  if(m_pending_request == nullptr) {
+    OLA_INFO << "Muting device";
+    m_mute_complete = mute_complete;
+    m_pending_request = ola::rdm::NewMuteRequest(m_uid, target, m_transaction_number += 1);
```
peternewman: Will m_transaction_number++ not work here?

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -105,6 +247,32 @@ void *FtdiDmxThread::Run() {
     }
 
     clock.CurrentTime(&ts1);
+    if(m_pending_request != nullptr) {
+
+      elapsed = ts1 - lastDMX;
+
+      if(elapsed.InMilliSeconds() < 500) {
```
peternewman: Can you put 500 as an enum or const or something please.

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -122,14 +290,79 @@ void *FtdiDmxThread::Run() {
...
+      if (!m_interface->Write(buffer)) {
+        goto framesleep;
+      } else {
+        clock.CurrentTime(&lastDMX);
+      }
+    } else {
+      if(m_interface->Write(&packetBuffer)) {
+          OLA_INFO << "RDM packet written to line";
+          if(m_pending_request->IsDUB()) {
+            usleep(58000); //min time before next packet broadcast allowed
```
peternewman: Again, no magic numbers

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -122,14 +290,79 @@ void *FtdiDmxThread::Run() {
...
+            readBytes = m_interface->Read(readBuffer, 258);
```
peternewman: Again

peternewman:
Travis lint errors will need fixing too:
https://travis-ci.org/OpenLightingProject/ola/jobs/498695471

peternewman:
And this compiler error:
```
libtool: compile:  ccache g++-8 -DHAVE_CONFIG_H -I. -I../.. -Wno-deprecated-declarations -I../../include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -pthread -Werror -Wno-error=deprecated-declarations -g -O2 -std=gnu++11 -pthread -MT plugins/ftdidmx/FtdiDmxThread.lo -MD -MP -MF plugins/ftdidmx/.deps/FtdiDmxThread.Tpo -c ../../plugins/ftdidmx/FtdiDmxThread.cpp -o plugins/ftdidmx/FtdiDmxThread.o >/dev/null 2>&1
../../plugins/ftdidmx/FtdiWidget.cpp: In member function â€˜bool ola::plugin::ftdidmx::FtdiInterface::Write(ola::io::ByteString*)â€™:
../../plugins/ftdidmx/FtdiWidget.cpp:373:61: error: invalid conversion from â€˜const unsigned char*â€™ to â€˜unsigned char*â€™ [-fpermissive]
   int bytesWritten = ftdi_write_data(&m_handle, packet->data(), packet->size());
                                                 ~~~~~~~~~~~~^~
In file included from ../../plugins/ftdidmx/FtdiWidget.cpp:42:
/usr/include/ftdi.h:406:67: note:   initializing argument 2 of â€˜int ftdi_write_data(ftdi_context*, unsigned char*, int)â€™
     int ftdi_write_data(struct ftdi_context *ftdi, unsigned char *buf, int size);
                                                    ~~~~~~~~~~~~~~~^~~
```

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Totally forgot that, didn't you say there was a built-in UID I should use?

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Absolutely.
I'm from the school of thought that says that using += is better because it prevents situations where you have 2 semi successive lines doing ++ (for bad reasons) which then go missed in debugging.

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: The server has a UID, which is probably the best bet, check the plugin adaptor docs.

peternewman:
It works though!!! I've just plugged in an FTDI RS485 lead and done some discovery and identify! :fireworks: :champagne: :100: :+1:

peternewman:
> And this compiler error:
> 
> ```
> libtool: compile:  ccache g++-8 -DHAVE_CONFIG_H -I. -I../.. -Wno-deprecated-declarations -I../../include -I./include -Wall -Wformat -W -fvisibility-inlines-hidden -pthread -Werror -Wno-error=deprecated-declarations -g -O2 -std=gnu++11 -pthread -MT plugins/ftdidmx/FtdiDmxThread.lo -MD -MP -MF plugins/ftdidmx/.deps/FtdiDmxThread.Tpo -c ../../plugins/ftdidmx/FtdiDmxThread.cpp -o plugins/ftdidmx/FtdiDmxThread.o >/dev/null 2>&1
> ../../plugins/ftdidmx/FtdiWidget.cpp: In member function â€˜bool ola::plugin::ftdidmx::FtdiInterface::Write(ola::io::ByteString*)â€™:
> ../../plugins/ftdidmx/FtdiWidget.cpp:373:61: error: invalid conversion from â€˜const unsigned char*â€™ to â€˜unsigned char*â€™ [-fpermissive]
>    int bytesWritten = ftdi_write_data(&m_handle, packet->data(), packet->size());
>                                                  ~~~~~~~~~~~~^~
> In file included from ../../plugins/ftdidmx/FtdiWidget.cpp:42:
> /usr/include/ftdi.h:406:67: note:   initializing argument 2 of â€˜int ftdi_write_data(ftdi_context*, unsigned char*, int)â€™
>      int ftdi_write_data(struct ftdi_context *ftdi, unsigned char *buf, int size);
>                                                     ~~~~~~~~~~~~~~~^~~
> ```

This fixes it, and works for me:
int bytesWritten = ftdi_write_data(&m_handle, const_cast<unsigned char*>(packet->data()), packet->size());

peternewman:
Also can you update the FTDI plugin's readme, probably with a general note, and a list of known good, and known bad devices, where it works or doesn't (e.g. the likely biasing issue on the Enttec Open, although I never could find the circuit diagram the other day to confirm).

.gitignore
Keeper-of-the-Keys: If I'm not mistaken OLA is the name that gets autogenerated when importing the project from existing source.
I'll change it to be specifically the 4 files generated?

plugins/ftdidmx/FtdiDmxPort.h
Keeper-of-the-Keys: No, at some point I inherited 2 classes so it was better looking on the next line.

I'll revert that.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: This is an exact copy from EnttecUSB
It seems that all the DiscoveryComplete functions completely ignore the status bool.

Either way I have added minor debug/warn output.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: I have not be able to find this UID, I'll try to change the logic to leverage the FTDI serial number for the device part of the UID while maintaining the Vendor part as is (like in a lot of other locations in the code, it may need to be defined in some RDM related enum).

plugins/ftdidmx/FtdiDmxPort.h
peternewman: Thanks

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: Done

peternewman:
It feels like it's got a bit more flakey in recent changes, more RDM timeouts than I used to have. I've also found a bug, if you send a vendorcast, it gets stuck and never returns:
```

olad/plugin_api/Universe.cpp:460: Universe 4, RDM request to 5168:ffffffff, SD: 0, CC 0x30, TN 16, PID 0x1000, PDL: 1
common/io/EPoller.cpp:306: ss process time was 0.000002
plugins/ftdidmx/FtdiDmxThread.cpp:297: OK To send RDM
plugins/ftdidmx/FtdiDmxThread.cpp:332: RDM packet written to line
plugins/ftdidmx/FtdiDmxThread.cpp:297: OK To send RDM
plugins/ftdidmx/FtdiDmxThread.cpp:332: RDM packet written to line
```

This just repeats forever, and the ola_rdm_set doesn't exit.

.gitignore
Keeper-of-the-Keys: Changed.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Changed to being based on the serial of the FTDI chip.

Whether UID should be set centrally for all interfaces and how is still under discussion.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: ?

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Unless you are opposed to this style I'm leaving it.

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -164,9 +496,23 @@ void *FtdiDmxThread::Run() {
 void FtdiDmxThread::CheckTimeGranularity() {
   TimeStamp ts1, ts2;
   Clock clock;
+  timespec req, rem;
```
peternewman: Can you look at integrating this with http://docs.openlighting.org/ola/doc/latest/classola_1_1_time_interval.html please.

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: Sorry @Keeper-of-the-Keys , I finally found it, https://github.com/OpenLightingProject/ola/blob/cb021b41860ab63a5cb3add27e7fa91ac32511bb/olad/OlaServer.cpp#L188-L190 although it doesn't seem to be exposed in APIs currently. It does indeed default to the IP when one is present, or zero if there isn't one. See also this if we do them centrally https://github.com/OpenLightingProject/ola/issues/1344 .

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: Don't feel you need to use it for some debug if it's ignored elsewhere. I suspect nothing bothers logging as the discovery controller will have logged the same error. It was more about moving from (void) to OLA_UNUSED to better show we don't process it. The Enttec code will get updated in due course when it next gets touched.

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: Sorry, I meant use the @note form of Doxygen flag, so it doesn't just disappear into the description.

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: I'd prefer to stick with ++ given it's what we've got across the whole of the rest of the code-base. I'm not really clear how += protects from having multiple lines doing the same thing (I assume you were referring to a copy/paste?).

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -25,31 +25,51 @@
...
-FtdiDmxThread::FtdiDmxThread(FtdiInterface *interface, unsigned int frequency)
+FtdiDmxThread::FtdiDmxThread(FtdiInterface *interface,
+                             unsigned int frequency,
+                             unsigned int serial)
   : m_granularity(UNKNOWN),
     m_interface(interface),
     m_term(false),
-    m_frequency(frequency) {
+    m_frequency(frequency),
+    m_transaction_number(0),
```
peternewman: I think this is now managed centrally due to what I did here: https://github.com/OpenLightingProject/ola/pull/1343/files

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+      m_pending_request = request;
+      m_rdm_callback = callback;
+  } else {
+    OLA_WARN << "Unable to queue RDM request, RDM operation already pending";
+  }
+}
+
+void FtdiDmxThread::RunFullDiscovery(ola::rdm::RDMDiscoveryCallback *callback) {
+  m_discovery_agent.StartFullDiscovery(
+        ola::NewSingleCallback(this,
```
peternewman: Indent by four, or alternatively lay it out like this:
```
void FtdiDmxThread::RunFullDiscovery(ola::rdm::RDMDiscoveryCallback *callback) {
  m_discovery_agent.StartFullDiscovery(ola::NewSingleCallback(
      this,
      &FtdiDmxThread::DiscoveryComplete,
      callback));
}
```

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+                               &FtdiDmxThread::DiscoveryComplete,
+                               callback));
+}
+
+void FtdiDmxThread::RunIncrementalDiscovery(rdm::RDMDiscoveryCallback *cb) {
+  m_discovery_agent.StartIncrementalDiscovery(
+        ola::NewSingleCallback(this,
```
peternewman: Again

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+    OLA_WARN << "FTDI discovery failed";
+  }
+  if (callback) {
+    callback->Run(uids);
+  }
+}
+
+/**
+ * @brief Method called to cleanup any outstanding callbacks
+ * @param state
```
peternewman: Please either document this parameter or drop the Doxygen line

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+ *
+ * All callbacks except the RDMCallback lack a way of reporting an error state
```
peternewman: Still Doxygen @note

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+
+void FtdiDmxThread::MuteDevice(const ola::rdm::UID &target,
+                               MuteDeviceCallback *mute_complete) {
+  ola::thread::MutexLocker locker(&m_rdm_mutex);
+  if (m_pending_request == nullptr) {
+    OLA_INFO << "Muting device";
+    m_mute_complete = mute_complete;
+    m_pending_request = ola::rdm::NewMuteRequest(m_uid,
+                                                 target,
+                                                 m_transaction_number += 1);
```
peternewman: I think this is overridden in olad/OlaServerServiceImpl.cpp, so we should probably make the argument to NewMuteRequest optional...

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+}
+
+void FtdiDmxThread::UnMuteAll(UnMuteDeviceCallback *unmute_complete) {
+  ola::thread::MutexLocker locker(&m_rdm_mutex);
+  if (m_pending_request == nullptr) {
+    OLA_INFO << "Sending UnMuteAll";
+    m_unmute_complete = unmute_complete;
+    m_pending_request = ola::rdm::NewUnMuteRequest(m_uid,
+                                                   ola::rdm::UID::AllDevices(),
+                                                   m_transaction_number += 1);
```
peternewman: Again

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+                           BranchCallback *callback) {
+  ola::thread::MutexLocker locker(&m_rdm_mutex);
+  if (m_pending_request == nullptr) {
+    OLA_INFO << "Sending branch";
+    m_branch_callback = callback;
+    m_pending_request =
+        ola::rdm::NewDiscoveryUniqueBranchRequest(m_uid,
+                                                  lower,
+                                                  upper,
+                                                  m_transaction_number += 1);
```
peternewman: Again

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -105,6 +280,31 @@ void *FtdiDmxThread::Run() {
...
+    if (m_pending_request != nullptr) {
+      elapsed = ts1 - lastDMX;
+
+      if (elapsed.InMilliSeconds() < HALF_SECOND_MS) {
+        if (!packetBuffer.empty()) {
+          packetBuffer.clear();
+        }
+
+        if (!ola::rdm::RDMCommandSerializer::PackWithStartCode(
+              *m_pending_request, &packetBuffer)) {
```
peternewman: Indent by 4 off !ola

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -122,14 +322,146 @@ void *FtdiDmxThread::Run() {
...
+              }
+              /*
+               * This section of code does minimal verification of the received
+               * frame.
+               * This assumes that the 4th byte in the buffer is the 3rd byte of
+               * the RDM Frame which defines length and no checksum has been done
+               * yet.
+               */
+              if (readBytes >= 4) {
+                if (readBuffer[0] == 0x00 && readBuffer[1] == 0xcc) {
```
peternewman: I assume 0x00 is the break? Use ola::rdm::START_CODE.

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -122,14 +322,146 @@ void *FtdiDmxThread::Run() {
...
+            usleep(MIN_WAIT_RDM_US);
+            readBytes = m_interface->Read(readBuffer, sizeof(readBuffer));
+
+            if (readBytes > 0) {
+              /*
+               * The following section of code tries to deal with replies that
+               * are being broadcast too slowly, I don't have equipment that
+               * malfunctions in this way so I have no way of testing this.
+               */
+              if (readBytes < 4) {
```
peternewman: Can you make the 4 a const, no magic numbers.

plugins/ftdidmx/FtdiDmxThread.h
```diff
@@ -26,21 +26,53 @@
...
+#include "ola/rdm/DiscoveryAgent.h"
+#include "ola/rdm/RDMResponseCodes.h"
 
 namespace ola {
 namespace plugin {
 namespace ftdidmx {
 
-class FtdiDmxThread : public ola::thread::Thread {
+enum {
+  HALF_SECOND_MS = 500,
```
peternewman: Why is it 500ms? Is this a time from the spec, if so name it as such, otherwise please name what it's the wait for.

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -186,6 +190,10 @@ void FtdiWidget::Widgets(vector<FtdiWidgetInfo> *widgets) {
           }
         }
 
+        /* TODO: Serial part of UID for RDM relies on serial,
```
peternewman: Can you make it TODO(name) or Doxygen style TODO (@todo I think).

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -364,14 +372,35 @@ bool FtdiInterface::Write(const ola::DmxBuffer& data) {
   }
 }
 
-bool FtdiInterface::Read(unsigned char *buff, int size) {
+
+bool FtdiInterface::Write(ola::io::ByteString *packet) {
+  int bytesWritten = ftdi_write_data(
+                       &m_handle,
```
peternewman: Indent by 4 from int

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -364,14 +372,35 @@ bool FtdiInterface::Write(const ola::DmxBuffer& data) {
...
+
+  int size = packet->size();
+  if (bytesWritten < 0) {
+    OLA_WARN << m_parent->Description() << " "
+             << ftdi_get_error_string(&m_handle);
+    return false;
+  } else if (bytesWritten == static_cast<int>(packet->size())) {
+    return true;
+  } else {
+    OLA_WARN << "Bytes Written: " << bytesWritten << " Packet Size: " << size;
```
peternewman: Can you make this more explicit, add a != in the log text.

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -364,14 +372,35 @@ bool FtdiInterface::Write(const ola::DmxBuffer& data) {
...
+}
+
+int FtdiInterface::Read(unsigned char *buff, int size) {
   int read = ftdi_read_data(&m_handle, buff, size);
+  OLA_INFO << "FtdiRead";
   if (read <= 0) {
     OLA_WARN << m_parent->Description() << " "
              << ftdi_get_error_string(&m_handle);
-    return false;
+    return read;
```
peternewman: The return doesn't need to be in the if now does it, just the log line in there?

plugins/ftdidmx/FtdiWidget.h
```diff
@@ -224,11 +224,15 @@ class FtdiInterface {
   /** @brief Toggle communications line BREAK condition on/off */
   bool SetBreak(bool on);
 
-  /** @brief Write data to a previously-opened line */
+  /** @brief Write data to a previously-opened line, DMX only */
```
peternewman: You could use Doxygen @pre to specify the line has already been opened.

plugins/ftdidmx/FtdiWidget.h
```diff
@@ -224,11 +224,15 @@ class FtdiInterface {
...
   bool Write(const ola::DmxBuffer &data);
 
+  /** @brief Write prepared packets to previously opened line,
```
peternewman: Again

plugins/ftdidmx/README.md
```diff
@@ -1,10 +1,34 @@
-FTDI USB Chipset DMX Plugin
-===========================
```
peternewman: Can we keep the ==== on the title to match all the other plugins please.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: I don't remember where I heard it but I'll try to fin a reference, I was told in the past that secure and clear coding should prefer += over ++

plugins/ftdidmx/FtdiDmxDevice.cpp
```diff
@@ -64,7 +65,15 @@ bool FtdiDmxDevice::StartHook() {
     FtdiInterface *port = new FtdiInterface(m_widget,
                                             static_cast<ftdi_interface>(i));
     if (port->SetupOutput()) {
-      AddPort(new FtdiDmxOutputPort(this, port, i, m_frequency));
+      AddPort(new FtdiDmxOutputPort(
+                    this,
+                    port,
+                    i,
+                    m_frequency,
+                    static_cast<unsigned int>(strtol(
```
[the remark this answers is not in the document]
Keeper-of-the-Keys: an 8 character base36 (the FTDI serial) string exceeds 32bits of data (it's 40-something bits) so it can have some collisions (though I expect that to be highly unlikely)

plugins/ftdidmx/FtdiDmxDevice.cpp
```diff
@@ -64,7 +65,15 @@ bool FtdiDmxDevice::StartHook() {
     FtdiInterface *port = new FtdiInterface(m_widget,
                                             static_cast<ftdi_interface>(i));
     if (port->SetupOutput()) {
-      AddPort(new FtdiDmxOutputPort(this, port, i, m_frequency));
+      AddPort(new FtdiDmxOutputPort(
+                    this,
```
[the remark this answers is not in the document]
Keeper-of-the-Keys: The things we do to keep within the 80 character limit of old terminals :D

plugins/ftdidmx/FtdiDmxPort.h
```diff
@@ -44,10 +48,11 @@ class FtdiDmxOutputPort : public ola::BasicOutputPort {
     FtdiDmxOutputPort(FtdiDmxDevice *parent,
                       FtdiInterface *interface,
                       unsigned int id,
-                      unsigned int freq)
-        : BasicOutputPort(parent, id),
+                      unsigned int freq,
+                      unsigned int serial)
+      : BasicOutputPort(parent, id, false, true),
```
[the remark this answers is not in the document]
Keeper-of-the-Keys: hmm I think this is legacy from when I did not have it working yet and stuff tended to blow up (at which time I wanted to be the clear trigger to blowing up). No reason not to change that.

plugins/ftdidmx/FtdiDmxDevice.cpp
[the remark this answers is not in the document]
Keeper-of-the-Keys: Changed, will push in a bit with a whole slew of other changes.

plugins/ftdidmx/FtdiDmxDevice.cpp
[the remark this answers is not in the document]
Keeper-of-the-Keys: Other changes obsoleted this.
Same as comment above.

plugins/ftdidmx/FtdiDmxThread.cpp
peternewman: Okay, I'd still suggest that being consistent within the code-base is probably more valuable, but perhaps updating everything to += should be saved for another PR. I know someone else prefers ++foo as it's more efficient apparently (assuming the compiler doesn't optimise it away).

plugins/ftdidmx/FtdiDmxDevice.cpp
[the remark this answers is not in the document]
peternewman: > The things we do to keep within the 80 character limit of old terminals :D

Think of it as allowing someone to view it on a phone/tablet in portrait mode now! :smile:

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Fixed

plugins/ftdidmx/FtdiWidget.cpp
Keeper-of-the-Keys: Fixed

plugins/ftdidmx/FtdiWidget.cpp
Keeper-of-the-Keys: Indeed

plugins/ftdidmx/FtdiWidget.h
Keeper-of-the-Keys: Not following you?

plugins/ftdidmx/README.md
Keeper-of-the-Keys: OK

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: still not sure what you are referring to here?

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: m_default_uid would need to be available through some getter function....
No reason to "calculate" a UID again for each interface.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: I switched to ++ but have them outside the new() operation as a distinct step so that mistakes don't happen.

common/utils/StringUtils.cpp
```diff
@@ -152,13 +152,16 @@ bool StringToBoolTolerant(const string &value, bool *output) {
   return false;
 }
 
-bool StringToInt(const string &value, unsigned int *output, bool strict) {
+bool StringToInt(const string &value,
+                 unsigned int *output,
+                 bool strict,
+                 uint8_t base) {
```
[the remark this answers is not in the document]
peternewman: Specifically, you're assuming, and the spec seems to imply, that it's case insensitive, but we should have a test of that so we catch issues on new platforms.

plugins/convert_README_to_header.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
peternewman: Does this only work with bash then? I'd be slightly worried about support challenges on other OSes if we have to switch away.

plugins/ftdidmx/FtdiDmxDevice.cpp
```diff
@@ -56,15 +56,19 @@ FtdiDmxDevice::~FtdiDmxDevice() {
 bool FtdiDmxDevice::StartHook() {
   unsigned int interface_count = m_widget->GetInterfaceCount();
   unsigned int successfully_added = 0;
+  unsigned int serial = 0;
 
   OLA_INFO << "Widget " << m_widget->Name() << " has " << interface_count
            << " interfaces.";
 
   for (unsigned int i = 1; i <= interface_count; i++) {
+    if (!StringToInt(m_widget->Serial(), &serial, false, 36)) {
```
peternewman: You're allowing trailing characters with strict=false, but if you buy two dongles at the same time, the leading ones are more likely to be the same, so truncating before passing to StringToInt would be more likely to generate unique UIDs.

plugins/ftdidmx/FtdiDmxDevice.cpp
```diff
@@ -56,15 +56,19 @@ FtdiDmxDevice::~FtdiDmxDevice() {
...
+      OLA_WARN << "StringToInt returned false, serial used: " << serial;
```
peternewman: Don't you want to show the serial passed in, rather than the result which should be the untouched zero?

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -364,15 +372,35 @@ bool FtdiInterface::Write(const ola::DmxBuffer& data) {
...
+  int size = packet->size();
+  if (bytesWritten < 0) {
     OLA_WARN << m_parent->Description() << " "
              << ftdi_get_error_string(&m_handle);
     return false;
-  } else {
+  } else if (bytesWritten == static_cast<int>(packet->size())) {
     return true;
+  } else {
+    OLA_WARN << "Bytes Written: " << bytesWritten \
```
peternewman: You don't need the trailing \ here.

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -364,15 +372,35 @@ bool FtdiInterface::Write(const ola::DmxBuffer& data) {
...
+                       const_cast<unsigned char *>(packet->data()),
+                       packet->size());
+
+  int size = packet->size();
+  if (bytesWritten < 0) {
     OLA_WARN << m_parent->Description() << " "
              << ftdi_get_error_string(&m_handle);
     return false;
-  } else {
+  } else if (bytesWritten == static_cast<int>(packet->size())) {
```
peternewman: It might be clearer to have all the good or bad cases together, rather than bad, good else bad. I'm guessing you've possibly put < 0 first as an optimisation, as it only needs one side, but perhaps doing < 0, != else true would be better?

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -364,15 +372,35 @@ bool FtdiInterface::Write(const ola::DmxBuffer& data) {
...
+    OLA_WARN << "Bytes Written: " << bytesWritten \
+             << " != Packet Size: " << size;
+    return false;
+  }
+}
+
+int FtdiInterface::Read(unsigned char *buff, int size) {
+  int read = ftdi_read_data(&m_handle, buff, size);
+  OLA_DEBUG << "ftdi_read_data() read: " << read;
+  if (read <= 0) {
```
peternewman: Is reading zero an error (in the ftdi_get_error_string sense)? What if the device just doesn't reply?

plugins/ftdidmx/README.md
```diff
@@ -1,10 +1,80 @@
...
+### Proper Line Biasing
+For simple DMX output (and input) all that is needed is a 130 Ohm resistor
+between data+ and data-.
+
+For RDM 2 additional resistors of 680 Ohm are needed:
+1. Pull-up connects between Data+ and VCC
+2. Pull-down between Data- and the common/ground.
+
+#### Diagram
+`
```
peternewman: Nice ASCII art, I think you need three backticks either end of it for it to format correctly in GitHub and in our new web UI (which uses a JS GFM formatter).

plugins/ftdidmx/README.md
```diff
@@ -1,10 +1,80 @@
...
+able to output and receive RDM packets.
+
+At this stage we can't guarantee that the plugin meets all timing specs since 
+I don't have any faulty RDM equipment to test with. It has however been tested
+with multiple responders on the line and discovery works correctly.
+
+*It is also critical that the FTDI local echo function is disabled*
+
+### Proper Line Biasing
+For simple DMX output (and input) all that is needed is a 130 Ohm resistor
```
peternewman: Section 2.5 of the RDM spec says 133 ohm, I appreciate 130 is probably good enough, but maybe a note to that effect?

plugins/ftdidmx/README.md
```diff
@@ -1,10 +1,80 @@
...
+between data+ and data-.
+
+For RDM 2 additional resistors of 680 Ohm are needed:
```
peternewman: Likewise 562 ohm, and 5.6 is available in E12 series upwards from the looks of things, so not even very niche.

plugins/ftdidmx/README.md
```diff
@@ -1,10 +1,80 @@
...
+2. Data+ (680 Ohm -> 9)
+3. 130 Ohm -> 1
+4. Not used
+5. GND (680 Ohm -> 1)
+6. Not used
+7. Short with 8 to disable echo
+8. Short with 7 to disable echo
+9. +5VDC (680 Ohm -> 2)
+
+The FTDI spec and the DMX/RDM spec disagree where the 130 Ohm resistor should
```
peternewman: What's on pin 3 of the FTDI then?

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -73,15 +99,164 @@ bool FtdiDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+  OLA_INFO << "Starting FtdiDmxThread";
+  TimeStamp ts1, ts2, ts3, lastDMX;
   Clock clock;
   CheckTimeGranularity();
   DmxBuffer buffer;
+  bool sendRDM = false;
+  TimeInterval elapsed, interval;
+  int readBytes;
+  unsigned int additionalWait = 0;
+  unsigned char readBuffer[258];
```
peternewman: Again where has this magic number come from? I'm guessing it's max RDM packet size + a bit?

common/utils/StringUtils.cpp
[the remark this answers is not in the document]
Keeper-of-the-Keys: What do you mean by tests?
Whoever uses the function and actively defines a base surely knows the base of the string they are sending to it?

plugins/convert_README_to_header.sh
Keeper-of-the-Keys: Works fine in sh too, while I was trying to fix it I changed it, will revert.

plugins/ftdidmx/FtdiDmxDevice.cpp
Keeper-of-the-Keys: You are right, I should just take the last 6 characters.

I probably also should now disallow dongles without a serial?

plugins/ftdidmx/FtdiDmxDevice.cpp
Keeper-of-the-Keys: Well I changed the function that it *does* change the output regardless of it's error state, I figured that those who care will be checking if(StringToInt) and not if(output !=0) (which would we immensely stupid). But I'll grep through the code to make sure.

Keeper-of-the-Keys: As far as I can tell all usages of StringToInt are inside if() statements.

plugins/ftdidmx/README.md
Keeper-of-the-Keys: I used the values given by Wayne Howell in Control Freak.

Don't mind writing other stuff, my test setup has these values though.

plugins/ftdidmx/README.md
Keeper-of-the-Keys: They call it "Input" and don't say anything other then "connect the 120R here if you are the first/last device on the line"

See page 16:
https://www.ftdichip.com/Support/Documents/DataSheets/Modules/DS_USB-COM485-PLUS4.pdf

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Yup

plugins/ftdidmx/README.md
Keeper-of-the-Keys: Thanks!
As far as I can tell it seems to be displaying OK without any backticks... 4 spaces at the beginning of the line did the trick.

plugins/ftdidmx/FtdiWidget.cpp
Keeper-of-the-Keys: Thanks, I never remember that.

plugins/ftdidmx/FtdiWidget.cpp
Keeper-of-the-Keys: You are absolutely right, changed.

plugins/ftdidmx/FtdiWidget.cpp
Keeper-of-the-Keys: 0 is OK but no response, good catch :)
(probably left over from when I was doing WriteAndRead)

Modified

plugins/ftdidmx/FtdiDmxDevice.cpp
Keeper-of-the-Keys: Did a change let me know what you think.

common/utils/StringUtils.cpp
```diff
@@ -168,10 +171,12 @@ bool StringToInt(const string &value, unsigned int *output, bool strict) {
   if (strict && *end_ptr != 0) {
     return false;
   }
+
+  *output = static_cast<unsigned int>(l);
```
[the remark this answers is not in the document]
Keeper-of-the-Keys: It's still returning false, a user who cares is anyhow checking for true/false.

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: It's actually - FTDI will always add 1 null byte at the beginning (not 100% sure but I think that is the break being treated as data) of DMX data (except for the DUB reply), the max packet length is 256 bytes and 1 more byte just in case.

peternewman:
I think you mean Codacy rather than Coverity. I won't worry too much about trying to keep it happy, it's just something we've been playing around with and isn't required for the PR to merge (only Travis is).

plugins/ftdidmx/FtdiDmxThread.cpp
Keeper-of-the-Keys: Does this mean I can remove all code and references to this?

At the moment for this pull I tend to think leave it and then I can remove and test again, but whatever you prefer.

plugins/ftdidmx/FtdiDmxThread.cpp
```diff
@@ -27,29 +27,58 @@
...
+    m_uid(OPEN_LIGHTING_ESTA_CODE, serial),
+    m_pending_request(nullptr),
+    m_rdm_callback(nullptr),
+    m_mute_complete(nullptr),
+    m_unmute_complete(nullptr),
+    m_branch_callback(nullptr) {
+  m_timer.setCaller("FtdiDmxThread " + m_interface->Description());
+
+  if (serial == 0) {
+    std::srand(std::time(nullptr));
```
peternewman: You should use our existing random class really, which is cross-platform etc:
https://github.com/OpenLightingProject/ola/blob/master/common/math/Random.cpp

common/utils/Clock.cpp
```diff
@@ -129,6 +131,11 @@ int64_t BaseTimeVal::InMilliSeconds() const {
           m_tv.tv_usec / ONE_THOUSAND);
 }
 
+int64_t BaseTimeVal::InMicroSeconds() const {
```
peternewman: This returns exactly the same value as BaseTimeVal::AsInt(), any reason we need the second function?

common/utils/Clock.cpp
```diff
@@ -272,4 +279,89 @@ void MockClock::CurrentTime(TimeStamp *timestamp) const {
...
+  req.tv_nsec = requested.MicroSeconds() * ONE_THOUSAND;
+
+  this->usleep(req);
+}
+
+void Sleep::usleep(uint32_t requested) {
+  timespec req;
+  req.tv_sec = requested / USEC_IN_SECONDS;
+  req.tv_nsec = (requested % USEC_IN_SECONDS) * ONE_THOUSAND;
+  req.tv_sec = 0;
```
peternewman: Why is this assigned and then reset?

common/utils/StringUtils.cpp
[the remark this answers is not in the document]
peternewman: Like the stuff in https://github.com/OpenLightingProject/ola/blob/master/common/utils/StringUtilsTest.cpp#L541-L564 for example.

So passing in something in base 36, and also validating that 123ABC converts to the same value as 123abc.

common/utils/StringUtils.cpp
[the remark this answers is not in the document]
peternewman: But if we've returned false we shouldn't bother setting the value.

plugins/ftdidmx/FtdiDmxThread.h
```diff
@@ -26,32 +26,94 @@
...
+  ALMOST_SECOND_MS = 900,
```
peternewman: Likewise

common/utils/Clock.cpp
Keeper-of-the-Keys: This is returning msec not usec...

common/utils/Clock.cpp
Keeper-of-the-Keys: I suspect an error on my part, though I also have a suspicion that maybe I did not want this function to exceed 1s

common/utils/Clock.cpp
peternewman: I guess AsInt is too then; I've just double-checked and the code is identical in both, only the method name changes.

CaptChrisD:
@Keeper-of-the-Keys I -- as well as others -- am also interested in getting this in. Any chance some effort could be put towards this.

Keeper-of-the-Keys:
> @Keeper-of-the-Keys I -- as well as others -- am also interested in getting this in. Any chance some effort could be put towards this.

I tried picking it back up but it turns out I broke one of my FTDI boards, I don't really have budget at the moment to buy new boards and thus it got snowed under other projects that pay my bills at the moment.

Specifically this PR is too large to get merged IMHO and my plan was to start working on the different components and get them merged one by one but as said need working hardware to develop software.

That being said I do still hope to get to it...

CaptChrisD:
@Keeper-of-the-Keys would you mind reaching out to me? chris at bigcove.io

PR #1561, opened 2019-06-14
Fix Getting Dimmer Max Value

kecramer:
Dimmer max is a single 16 bit value, not an 8 bit.

### DIFF
--- a/common/rdm/RDMAPI.cpp
+++ b/common/rdm/RDMAPI.cpp
@@ -4413,8 +4413,8 @@ void RDMAPI::_HandleGetDimmerMaximumLevel(
     if (data.size() != DATA_SIZE) {
       SetIncorrectPDL(&response_status, data.size(), DATA_SIZE);
     } else {
-      const uint16_t *ptr = reinterpret_cast<const uint16_t*>(data.data());
-      maximum_level = NetworkToHost(*ptr);
+      maximum_level = *(reinterpret_cast<const uint16_t*>(data.data()));
+      maximum_level = NetworkToHost(maximum_level);
     }
   }
   callback->Run(response_status, maximum_level);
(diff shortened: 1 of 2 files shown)


kecramer:
This is failing `make check` but so is master, and I believe it unrelated to this code.

common/rdm/RDMAPI.cpp
```diff
@@ -4413,7 +4413,8 @@ void RDMAPI::_HandleGetDimmerMaximumLevel(
     if (data.size() != DATA_SIZE) {
       SetIncorrectPDL(&response_status, data.size(), DATA_SIZE);
     } else {
-      maximum_level = data[0];
+      const uint16_t *ptr = reinterpret_cast<const uint16_t*>(data.data());
```
peternewman: Good catch, sorry we both missed it first time around. You can simply this slightly, if you look at how we do _HandleGetDMXAddress you don't need the ptr variable.

kecramer:
```common/utils/DmxBufferTest.cpp:197:10: error: explicitly assigning value of
      variable of type 'ola::DmxBuffer' to itself
      [-Werror,-Wself-assign-overloaded]
  buffer = buffer;
  ~~~~~~ ^ ~~~~~~
1 error generated.```

Is the error when I am up to date with origin/master (and for this branch as well)

common/rdm/RDMAPI.cpp
kecramer: Sounds good, I will tweak that today.

peternewman:
Thanks @kecramer . Sorry I missed it before, but could you switch SetDMXAddress to use the new GenericSetU16 you've very sensibly added.

kecramer:
Ah good point, definitely missed that one. Let me know if you have any other comments.

peternewman:
> Ah good point, definitely missed that one. Let me know if you have any other comments.

Well in fairness I didn't initially suggest even making a generic set as I'd missed the fact the brightness would have a set too.

PR #1563, opened 2019-06-25
Add DMXUSB usbpro device

DaAwesomeP:
This adds support for an Arduino-based DMXUSB Device with many universes. Basically, I extended the DMXKing UltraDMX Pro spec to allow for more universes and added a label so that the device can report how many universes it supports. All the info is in the [readme](https://github.com/DaAwesomeP/dmxusb/blob/749aa9329d05d4f63daa0e41a3163d5eabede4f8/README.md#dmxusb-device-specification-mode-2).

I used a prototype ETSA ID and picked an arbitrary device ID to try to avoid conflicts, but I suspect that there will be very few conflicts. I think that this should use the prototype ETSA IDs because of its DIY nature.

I don't think that Travis is failing because of this commit, but I can't really tell.

You may remember the first time I did this in [this thread](https://groups.google.com/forum/#!topic/open-lighting/SIMMzwRcxPY) where I simply emulated a DMXKing UltraDMX Pro with two universes. Now I can handle very large LED displays!


plugins/usbpro/WidgetDetectorThreadTest.cpp
```diff
@@ -337,6 +353,37 @@ void WidgetDetectorThreadTest::testDmxterWidget() {
 }
 
 
+/**
+ * Check that we can locate a DMXUSB widget.
+ */
+void WidgetDetectorThreadTest::testDMXUSBWidget() {
+  uint8_t serial_data[] = {0x78, 0x56, 0x34, 0x12};
+  uint8_t manufacturer_data[] = "\153\152DMXUSB";
```
DaAwesomeP: Ah, I remember what I couldn't get to work: Where do the 153 and 152 come from? What should they be for my device? (lines 360-362).

peternewman: Without looking at the code, that as octal is hex 0x6b 0x6a, which is the manufacturer ID for JPK, makers of the DMXKing:
http://rdm.openlighting.org/manufacturer/display?manufacturer=27243

DaAwesomeP:
So I have been testing this a lot, and the only issue I can see right now is that OLA doesn't save the patches to my DMXUSB device. I'm using the Debian `init.d` script to start and stop. It saves my OSC and E1.31 assignments.

plugins/usbpro/WidgetDetectorThreadTest.cpp
DaAwesomeP: Aha an octal! I'll push a commit.

peternewman:
Hi @DaAwesomeP ,

Looks interesting, you should probably also add some notes to:
https://wiki.openlighting.org/index.php/USB_Protocol_Extensions

I don't think we should merge this with a prototyping manufacturer ID, it's easy enough to register your own:
https://tsp.esta.org/tsp/working_groups/CP/mfctrIDs.php

As @mathertel has done:
https://github.com/mathertel/DmxSerial2/blob/master/src/DMXSerial2.cpp#L240-L247

peternewman:
> So I have been testing this a lot, and the only issue I can see right now is that OLA doesn't save the patches to my DMXUSB device after stopping and restarting OLA. I'm using the Debian `init.d` script to start and stop. It saves my OSC and E1.31 assignments.

Can we see some olad -l 4 logs for this to try and make sense of what's going wrong. Also the ola-usbserial.conf file and ola-port.conf files.

DaAwesomeP:
> Looks interesting, you should probably also add some notes to:
https://wiki.openlighting.org/index.php/USB_Protocol_Extensions

I'll definitely do so as soon as this is finalized! I have found that page *so* useful! Thank you and everyone who has put so much work into this project.

> I don't think we should merge this with a prototyping manufacturer ID, it's easy enough to register your own:
https://tsp.esta.org/tsp/working_groups/CP/mfctrIDs.php

I completely understand and respect your reasoning. I'm worried about my privacy and accountability. Registering my own ID requires listing my address and phone number, which is not something that I would like to do, as that would be my personal address and not my business address. Additionally, since my device is really just an open source library and not prepackaged, for-sale hardware, I don't want to tie my name to something that someone else could produce down the line. The only way to avoid that accountability is to use a prototype ID or to require the user to input an ID.

What if the ESTA ID was configurable in the conf file? I would then require the use to provide one in my library instead of always using a prototype one.

> Can we see some olad -l 4 logs for this to try and make sense of what's going wrong. Also the ola-usbserial.conf file and ola-port.conf files.

Yes! I'll provide them tomorrow.

DaAwesomeP:
Here you can find the log, `ola-usbserial.conf`, and `ola-port.conf`: https://gist.github.com/DaAwesomeP/9edc12c4dc0137faa149b42755edcd61.

I have a udev rule that maps my device (Teensy at `/dev/ttyACM*`) to `/dev/ttyDMX*`. This is because I have other non-lighting `ttyACM` devices plugged in. This issue persisted before and after I set this rule.

DaAwesomeP:
@peternewman do you have any idea why it isn't saving assignments.

Is it possible to load assignments from a file?

DaAwesomeP:
So what I think it may be is that I added an additional call after `GetParameters` called `GetExtendedParameters` that queries the device for how many DMX ports it supports. When this callback completes, `DMXUSBDevice::SetupPorts` is called which in turn calls `AddPort`. I think maybe the delay could be causing it? I figured my `GetExtendedParameters` would be blocking and prevent the constructor from returning but maybe not? Are the ports saved on each `AddPort` or when the constructor finishes?

DaAwesomeP:
Aha! I'm going to take a look at `ola::Port::UniqueId`...

DaAwesomeP:
Nope. `UniqueID` must be working properly because I see ` Patched 5-165165165165-O-0 to universe 1` in  the logs.

peternewman:
> 165165165165

So the port preferences only get saved in the DeviceManager destructor. They only get written in 
SavePortPatchings which is called from ReleaseDevice which itself is called by UnregisterDevice or UnregisterAllDevices. The latter is called by StopPlugins in OlaServer.

So I'd add some more debugging around some of that stuff, but I suspect they're not getting called due to a bug in your plugin not ending properly.

It looks like they're not getting written to the file for some reason. I'd try splitting the test in half, add some lines like this to ola_port.conf and see if it at least restores the patching:
`5-165165165165-O-1 = 2`

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * DMXUSBDevice.cpp
+ * A DMXUSB Device (based on DMX King Ultra DMX Pro code by Simon Newton)
+ * Copyright (C) 2019 Perry Naseck (DaAwesomeP)
+ *
+ * This device creates n ports, 0 input and n outputs (fetchs number from
```
peternewman: SPaG fetches

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+using std::string;
+
+/*
+ * Create a new device
+ * @param owner  the plugin that owns this device
+ * @param name  the device name
+ * @param dev_path  path to the pro widget
+ */
+DMXUSBDevice::DMXUSBDevice(ola::PluginAdaptor *plugin_adaptor,
+                                     ola::AbstractPlugin *owner,
```
peternewman: Fix the spacing so they align with just after the ( above, however see the general comment first.

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+    m_plugin_adaptor(plugin_adaptor),
+    m_dmxusb_widget(widget),
+    m_serial(),
+    m_got_parameters(false),
+    m_got_extended_parameters(false),
+    m_break_time(0),
+    m_mab_time(0),
+    m_rate(0),
+    m_out_ports(0),
+    m_in_ports(0),
```
peternewman: Isn't this always zero?

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+    m_fps_limit(fps_limit) {
+  m_str << std::setfill('0');
```
peternewman: There's also UsbProDevice::SerialToString, I wonder if it should be pulled out into a utils file and used more widely.

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+      done->Run();
+  }
+}
+
+
+/**
+ * Update the cached param values
+ */
+void DMXUSBDevice::UpdateParams(bool status,
+                                     const usb_pro_parameters &params) {
```
peternewman: Again alignment.

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+/*
+ * Handle device config messages
+ * @param controller An RpcController
+ * @param request the request data
+ * @param response the response to return
+ * @param done the closure to call once the request is complete
+ */
+void DMXUSBDevice::Configure(RpcController *controller,
+                                  const string &request,
```
peternewman: Again alignment.

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+    m_got_parameters = true;
+    m_break_time = params.break_time;
+    m_mab_time = params.mab_time;
+    m_rate = params.rate;
+  }
+}
+
+
+void DMXUSBDevice::UpdateExtendedParams(bool status,
+                                     const dmxusb_extended_parameters &params) {
```
peternewman: Again alignment.

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+    DMXUSBDevice::SetupPorts();
+  }
+}
+
+void DMXUSBDevice::SetupPorts() {
+  // add the out ports
+  for (int i = 0; i < static_cast<int>(m_out_ports); i++) {
+    std::ostringstream out_str;
+    out_str << " Output universe: " << (i + 1) << " of "
+        << static_cast<int>(m_out_ports) << ", " << m_str.str();
```
peternewman: Align with the << on the row above.

plugins/usbpro/DMXUSBDevice.h
```diff
@@ -0,0 +1,146 @@
...
+namespace plugin {
+namespace usbpro {
+
+/*
+ * An DMXUSB device
+ */
+class DMXUSBDevice: public UsbSerialDevice {
+ public:
+  DMXUSBDevice(ola::PluginAdaptor *plugin_adaptor,
+                    ola::AbstractPlugin *owner,
```
peternewman: Again alignment.

plugins/usbpro/DMXUSBDevice.h
```diff
@@ -0,0 +1,146 @@
...
+};
+
+
+/*
+ * The output ports
+ */
+class DMXUSBOutputPort: public BasicOutputPort {
+ public:
+  DMXUSBOutputPort(DMXUSBDevice *parent,
+                        DMXUSBWidget *widget,
```
peternewman: Again alignment.

plugins/usbpro/DMXUSBWidget.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+
+bool DMXUSBWidget::SendDMXPort(unsigned int port, const DmxBuffer &buffer) {
+  return SendDMXWithLabel(DMX_START_PORT + port, buffer);
+}
+
+
+bool DMXUSBWidget::SendDMXWithLabel(uint8_t label,
+                                         const DmxBuffer &data) {
```
peternewman: Again alignment.

plugins/usbpro/DMXUSBWidget.cpp
```diff
@@ -0,0 +1,127 @@
...
+    // failed
+    m_outstanding_extended_param_callbacks.pop_back();
+    dmxusb_extended_parameters params = {0, 0};
+    callback->Run(false, params);
+  }
+}
+
+
+void DMXUSBWidget::SpecificStop() {
+  // empty extedned_params struct
```
peternewman: Typo

plugins/usbpro/DMXUSBWidgetTest.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Check that we can send DMX on the all ports
+ */
+void DMXUSBWidgetTest::testAllSendDMX() {
+  // dmx data
+  DmxBuffer buffer;
+  buffer.SetFromString("0,1,2,3,4");
+
+  // test on 47 universes
```
peternewman: Why 47, is that when it hits the next label, perhaps add something explaining that?

plugins/usbpro/DMXUSBDevice.cpp
```diff
@@ -0,0 +1,269 @@
...
+    m_rate = params.rate;
+  }
+}
+
+
+void DMXUSBDevice::UpdateExtendedParams(bool status,
+                                     const dmxusb_extended_parameters &params) {
+  if (status) {
+    m_got_extended_parameters = true;
+    m_out_ports = params.out_ports;
```
peternewman: Given your 47 limit in the tests there should be some sanity checks here or at port creation time.

plugins/usbpro/DMXUSBDevice.cpp
DaAwesomeP: I haven't implemented input ports yet on the Arduino side, so this is really a placeholder until I do.

plugins/usbpro/DMXUSBDevice.cpp
DaAwesomeP: Yeah I just copied from the UltraDMX plugin.

plugins/usbpro/DMXUSBWidgetTest.cpp
DaAwesomeP: 47 is the maximum number that a Teensy can handle, so I figured that would be a good number to test with. I'll add a note.

DaAwesomeP:
> I think a different name than DMXUSB would be preferable if you haven't already used it too much, that sort of encompasses all devices and makes it very hard for users to differentiate exactly which device is being referred to.

Yeah I agree it is a bit generic. I have published it in the Arduino library list for some time now, so I can only change it on the OLA side, but I think that's ok. How about ArduinoDMXUSB?

plugins/usbpro/DMXUSBDevice.cpp
peternewman: It's probably better to comment out, or ideally remove the lines, rather than having them there without the functionality. It didn't look like the rest of the C++ code was present either from what I could see. If it was all in one commit you could always revert it in future when you have.

plugins/usbpro/DMXUSBDevice.cpp
peternewman: Okay, fancy attempting that bit of tidying up to both then (well all three including the UsbPro one too?

plugins/usbpro/DMXUSBWidgetTest.cpp
peternewman: What model of Teensy, how's it limited to that, available RAM? What are the refresh rate limitations of all this?

peternewman:
> Yeah I agree it is a bit generic. I have published it in the Arduino library list for some time now, so I can only change it on the OLA side, but I think that's ok. How about ArduinoDMXUSB?

That seems like a good compromise, and not too much of a diversion anyone finding it from the Arduino library list won't know what's going on.

DaAwesomeP:
At this point this pull is quite old and I have not touched it in some time. Closing and will reopen maybe sometime in the future.

DaAwesomeP:
These commits are now in the `DaAwesomeP-DMXUSB` branch of my fork.

PR #1568, opened 2019-07-08
0.10 g++ 9

peternewman:


peternewman:
Probably superseded by https://github.com/OpenLightingProject/ola/pull/1578 although that doesn't throw the pthread error and this does, so we probably need to pull in another change too.

peternewman:
Superseded by #1583 .

PR #1573, opened 2019-07-30
UsbPro Extended Device/Widget

rbalykov:
Extention to Ennec message labels
ESTA ID = 0x7FF7 Experimental
DEVICE ID =  i/o port count combined in uint16_t
LABELS = 100 + m_port_id

NOTE: m_port_id is private, had to make m_port_id_ext

### DIFF
--- a/plugins/uartdmx/UartDmxDevice.cpp
+++ b/plugins/uartdmx/UartDmxDevice.cpp
@@ -34,8 +34,10 @@ using std::string;
 
 const char UartDmxDevice::K_MALF[] = "-malf";
 const char UartDmxDevice::K_BREAK[] = "-break";
+const char UartDmxDevice::K_PADDING[] = "-padding";
 const unsigned int UartDmxDevice::DEFAULT_BREAK = 100;
 const unsigned int UartDmxDevice::DEFAULT_MALF = 100;
+const unsigned int UartDmxDevice::DEFAULT_PADDING = 24;
 
 
 UartDmxDevice::UartDmxDevice(AbstractPlugin *owner,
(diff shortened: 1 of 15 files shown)


plugins/uartdmx/UartDmxDevice.cpp
```diff
@@ -94,6 +103,9 @@ void UartDmxDevice::SetDefaults() {
   save |= m_preferences->SetDefaultValue(DeviceMalfKey(),
                                          UIntValidator(8, 1000000),
                                          DEFAULT_MALF);
+  save |= m_preferences->SetDefaultValue(DevicePaddingKey(),
+                                         UIntValidator(24, 512),
```
peternewman: Is this data slots, or including start code? I think in theory it can be less than 24 depending on frame rate etc.

plugins/uartdmx/UartWidget.cpp
```diff
@@ -54,8 +55,9 @@ namespace uartdmx {
 using std::string;
 using std::vector;
 
-UartWidget::UartWidget(const string& path)
+UartWidget::UartWidget(const std::string &path, unsigned int padding)
```
peternewman: You don't need the std:: as we're already using string.

plugins/uartdmx/UartWidget.cpp
```diff
@@ -54,8 +55,9 @@ namespace uartdmx {
...
     : m_path(path),
+	  m_padding (padding),
```
peternewman: Align with the line above and below (and drop the space before `(`)

plugins/uartdmx/UartWidget.h
```diff
@@ -80,7 +81,8 @@ class UartWidget {
     bool SetupOutput();
 
  private:
-  const std::string m_path;
+    const std::string m_path;
```
peternewman: Can you leave the indenting where it was, two in from the start of the line.

plugins/uartdmx/UartDmxDevice.cpp
rbalykov: This is data slots. Got this number somewhere on net.
Wikipedia:  However, it does require that packets be transmitted so that the leading edges of any two sequential BREAKs must be separated by at least 1204 μs, and receivers must be able to handle packets with break-to-break times as short as 1196 μs.[10] The minimum break-to-break transmit time can be achieved by sending packets that contain at least 24 slots

rbalykov:
> Can you separate the USBPro stuff out into a new PR (or the UART stuff)
I.e. make a new branch for new PR?

Issue #1553, opened 2019-04-11
New warnings with GCC9


yoe:
I received [Debian bug 925793](https://bugs.debian.org/925793), which produces a few new warnings of this type:

```
libs/acn/HeaderSetTest.cpp: In member function 'void HeaderSetTest::testTransportHeader()':
libs/acn/HeaderSetTest.cpp:78:29: error: implicitly-declared 'ola::acn::TransportHeader::TransportHeader(const ola::acn::TransportHeader&)' is deprecated [-Werror=deprecated-copy]
   78 |   TransportHeader header2 = header;
      |                             ^~~~~~
In file included from ./libs/acn/HeaderSet.h:30,
                 from libs/acn/HeaderSetTest.cpp:28:
../libs/acn/TransportHeader.h:58:8: note: because 'ola::acn::TransportHeader' has user-provided 'void ola::acn::TransportHeader::operator=(const ola::acn::TransportHeader&)'
   58 |   void operator=(const TransportHeader &other) {
      |        ^~~~~~~~
```

Basically, what happens here is that the compiler complains about a class which does have an assignment operator, but no copy constructor. You should probably add them.

peternewman:
Thanks @yoe. Looks like we won't be able to test on Travis easily, as this seems to be the only PPA with a suitable package and it's not in Travis' approved list: https://launchpad.net/~doko/+archive/ubuntu/toolchain

peternewman:
See also what looks to be the same thing on Fedora:
https://paste.fedoraproject.org/paste/yCwhpRW3hc3wALehI6sspw

yoe:
What's the status on this? The severity of this bug in Debian just got bumped, and it is likely that it will be bumped again soon.

peternewman:
> What's the status on this? The severity of this bug in Debian just got bumped, and it is likely that it will be bumped again soon.

Patches welcome! :stuck_out_tongue: 

I can try and have a go next week if no-one has done it before then if it's as simple as adding a copy constructor, but I'd need someone else to sanity test it.

peternewman:
See #1583

peternewman:
Done via #1578

PR #1578, opened 2019-08-21
GCC 9 fixes

yoe:
Upgrade to compiling with GCC 9.

If I've done everything right, this should deal with #1553

### DIFF
--- a/plugins/dummy/DummyPortTest.cpp
+++ b/plugins/dummy/DummyPortTest.cpp
@@ -668,11 +668,16 @@ void DummyPortTest::testParamDescription() {
 
  * Disable the warning to avoid it being an issue, but only for this
  * line.
+ *
+ * Also, work around clang producing an erorr on the
+ * "stringop-truncation" warning not existing there.
  */
+#ifndef __clang__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wstringop-truncation"
   strncpy(param_description.description, description.c_str(), str_len);
 #pragma GCC diagnostic pop
+#endif
 
   unsigned int param_data_length = (
       sizeof(param_description) - sizeof(param_description.description) +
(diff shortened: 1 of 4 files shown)


yoe:
NOTE: I'm not sure if the fix for the off-by-one error is the correct one; the chopping off of the trailing 0 may be on purpose. If so, the correct course of action would be to (locally) override that warning.

If not, then this should probably be merged (after review, of course).

plugins/dummy/DummyPortTest.cpp
```diff
@@ -661,7 +661,7 @@ void DummyPortTest::testParamDescription() {
 
   const string description("Code Version");
   size_t str_len = std::min(sizeof(param_description.description),
-                            description.size());
+                            description.size() + 1);
```
peternewman: Hi @yoe so this is intentional, the standard says:

> The Parameter Data Length shall accordingly be set to match the actual size of the text string
> being sent. Parsing of the string shall be length terminated corresponding to the Parameter Data
> Length field. Text fields shall terminate based on Parameter Data Length, however if a NULL is
> encountered then that shall also act as a terminator for the text field.

So 32 characters and no null terminator is valid, hence we don't copy it. What's the GCC error exactly? Anyway it needs overriding locally. I'm surprised and confused it's not triggering the same warning when we're doing exactly the same thing in common/rdm/ResponderHelper.cpp e.g.:
https://github.com/OpenLightingProject/ola/blob/master/common/rdm/ResponderHelper.cpp#L983-L985

yoe: So I changed this to use GCC pragma's instead, so that the relevant error will be ignored for that line only.

This should fix the issue.

peternewman:
I'm closing this as we'll take it into master via 0.10 hence #1583 is all we need.

Issue #691, opened 2015-03-31
Makefile lint target?


simark:
Could we add a "make lint" target that would run cpplint.py with the right parameters? It could also run pep8/flake8 for Python, and others for other languages.

nomis52:
That's a great idea.

We may also want to consider git pre-commit hooks to prevent untidy code from being submitted.

peternewman:
@Keeper-of-the-Keys This is a bare minimum prerequisite for dropping linting of generated files, if you're feeling keen, given you seem to keep mentioning it.

Keeper-of-the-Keys:
Cool, will see what I can do.

peternewman:
@stefansiegfried do you fancy tacking this one in another PR too?

Our cpplint.py config is already in Travis.

stefansiegfried:
@peternewman I would be happy to take this on

stefansiegfried:
@peternewman any thoughts on moving the cpplint command from .travis-ci.sh into its own bash script?
`
/cpplint.py \
    --filter=-legal/copyright,-readability/streams,-runtime/arrays \
    $(find ./ \( -name "*.h" -or -name "*.cpp" \) -and ! \( \
        -wholename "./common/protocol/Ola.pb.*" -or \
        -wholename "./common/rpc/Rpc.pb.*" -or \
        -wholename "./common/rpc/TestService.pb.*" -or \
        -wholename "./common/rdm/Pids.pb.*" -or \
        -wholename "./config.h" -or \
        -wholename "./plugins/*/messages/*ConfigMessages.pb.*" -or \
        -wholename "./tools/ola_trigger/config.tab.*" -or \
        -wholename "./tools/ola_trigger/lex.yy.cpp" \) | xargs)
  if [[ $? -ne 0 ]]; then
    exit 1;
  fi;
`
Otherwise, it seems a bit cumbersome to do from the Makefile itself

peternewman:
I'd probably rather not ship yet another .sh file if we can help it, it also feels a bit clunky to me if we have to do that.

I think you can probably drop the exit code check, I believe any non-zero code should stop the make command (older cpplint used to return the number of errors as the exit code, newer ones just exit 1 anyway I believe).

Ideally you can do some magic within the Makefile to replace all the find stuff given I'd assume we can get a list of files already as we're compiling the code and we can probably find some clever way to filter these bits out too (e.g. tagging them into custom groups which get included).

If it really isn't working have a look at this as an alternative option:
[...]

PR #1586, opened 2019-10-29
Logistics

Keeper-of-the-Keys:
As discussed by mail I'm going to do piece meal PRs for the FTDI stuff.

Kicking it off with a few logistical parts of the big PR:
* .gitignore
* travis lint blacklist of autogenerated files
* Fix of escaping for README files into C headers.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -11,6 +11,19 @@ COVERITY_SCAN_BUILD_URL="https://scan.coverity.com/scripts/travisci_build_coveri
 
 PYCHECKER_BLACKLIST="threading,unittest,cmd,optparse,google,google.protobuf,ssl,fftpack,lapack_lite,mtrand"
 
+LINT_BLACKLIST=$(cat <<EOLBL
+        -wholename "./common/protocol/Ola.pb.*" -or \
+        -wholename "./common/rpc/Rpc.pb.*" -or \
+        -wholename "./common/rpc/TestService.pb.*" -or \
+        -wholename "./common/rdm/Pids.pb.*" -or \
+        -wholename "./config.h" -or \
+        -wholename "./plugins/*/messages/*ConfigMessages.pb.*" -or \
+        -wholename "./tools/ola_trigger/config.tab.*" -or \
+        -wholename "./tools/ola_trigger/lex.yy.cpp" -or \
+        -wholename "./plugins/ftdidmx/FtdiDmxPluginDescription.h"
+EOLBL
+)
+
 SPELLINGBLACKLIST=$(cat <<-BLACKLIST
       -wholename "./.codespellignore" -or \
       -wholename "./.git/*" -or \
@@ -69,19 +82,12 @@ if [[ $TASK = 'lint' ]]; then
     echo "Found $nolints generic NOLINTs"
   fi;
   # then fetch and run the main cpplint tool
+  lintfiles=$(eval "find ./ \( -name "*.h" -or -name "*.cpp" \) -and ! \( $LINT_BLACKLIST \) | xargs")
   wget -O cpplint.py $CPP_LINT_URL;
   chmod u+x cpplint.py;
   ./cpplint.py \
     --filter=-legal/copyright,-readability/streams,-runtime/arrays \
-    $(find ./ \( -name "*.h" -or -name "*.cpp" \) -and ! \( \
-        -wholename "./common/protocol/Ola.pb.*" -or \
-        -wholename "./common/rpc/Rpc.pb.*" -or \
-        -wholename "./common/rpc/TestService.pb.*" -or \
-        -wholename "./common/rdm/Pids.pb.*" -or \
-        -wholename "./config.h" -or \
-        -wholename "./plugins/*/messages/*ConfigMessages.pb.*" -or \
-        -wholename "./tools/ola_trigger/config.tab.*" -or \
-        -wholename "./tools/ola_trigger/lex.yy.cpp" \) | xargs)
+     $lintfiles
   if [[ $? -ne 0 ]]; then
     exit 1;
   fi;
--- a/plugins/convert_README_to_header.sh
+++ b/plugins/convert_README_to_header.sh
@@ -30,8 +30,8 @@ outfilename=`basename $outfile`;
 # See http://stackoverflow.com/a/16576291
 # On Mac OS's sed, \n is not recognized as a newline character, but
 # \[actual newline] works
-desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\"/\\\"/g' -e 's/\n/\\\\n"\\
-"/g' "$path/README.md"`;
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's#\\\#\\\\\\\#g' -e 's#\"#\\\"#g' -e 's#\n#\\\\n"\\
+"#g' "$path/README.md"`;
 
 identifier=`echo "PLUGINS_${plugin}_${outfilename%.h}_H_" | tr '[:lower:]' '[:upper:]'`
 
--- a/.gitignore
+++ b/.gitignore
@@ -246,3 +246,8 @@ javascript/new-src/node_modules
 .cproject
 .settings
 .vscode/
+# QtCreator project files
+*.config
+*.creator*
+*.files
+*.includes


plugins/convert_README_to_header.sh
```diff
@@ -30,8 +30,8 @@ outfilename=`basename $outfile`;
 # See http://stackoverflow.com/a/16576291
 # On Mac OS's sed, \n is not recognized as a newline character, but
 # \[actual newline] works
-desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\"/\\\"/g' -e 's/\n/\\\\n"\\
-"/g' "$path/README.md"`;
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's#\\\#\\\\\\\#g' -e 's#\"#\\\"#g' -e 's#\n#\\\\n"\\
```
peternewman: This feels like the wrong number of backslashes to me, or is it one for the shell and one for C++?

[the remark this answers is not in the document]
Keeper-of-the-Keys: Just for the sake of clarity, too many slashes happening makes it tough to read.

Keeper-of-the-Keys: This was reached by trial and error, regex are black magic :P
But what it does is it changes a single \ in the README to a \\ (escaped \) in the header file.
\ also needs to be escaped in the regex and I think also for the bash script running it which is what ends up being 3...

Keeper-of-the-Keys: :O even here you need to escape you \\ interesting....
changes \\ to \\\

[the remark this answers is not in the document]
peternewman: Okay, given # is also the comment character in sed, and the slashes are going the other way, can we switch back just to ensure maximum portability, unless you've tested on Mac, Linux and the various BSDs?

.travis-ci.sh
```diff
@@ -11,6 +11,18 @@ COVERITY_SCAN_BUILD_URL="https://scan.coverity.com/scripts/travisci_build_coveri
...
+        -wholename "./common/rpc/TestService.pb.*" -or \
+        -wholename "./common/rdm/Pids.pb.*" -or \
+        -wholename "./config.h" -or \
+        -wholename "./plugins/*/messages/*ConfigMessages.pb.*" -or \
+        -wholename "./tools/ola_trigger/config.tab.*" -or \
+        -wholename "./tools/ola_trigger/lex.yy.cpp" -or \
+EOLBL
+)
+
 SPELLINGBLACKLIST=$(cat <<-BLACKLIST
```
peternewman: I know it's not your change, but can we SNAKE_CASE here to be consistent in the file please.

.travis-ci.sh
```diff
@@ -11,6 +11,18 @@ COVERITY_SCAN_BUILD_URL="https://scan.coverity.com/scripts/travisci_build_coveri
 
 PYCHECKER_BLACKLIST="threading,unittest,cmd,optparse,google,google.protobuf,ssl,fftpack,lapack_lite,mtrand"
 
+LINT_BLACKLIST=$(cat <<EOLBL
```
peternewman: Could we make the EOLBL and the one below more consistent, I guess being different may be safest, but consistent names would be nice.

Having just double-checked, the - at the start of the spelling one suppresses the leading tabs, although I'm not sure if we really need it:
http://tldp.org/LDP/abs/html/here-docs.html

.travis-ci.sh
```diff
@@ -11,7 +11,19 @@ COVERITY_SCAN_BUILD_URL="https://scan.coverity.com/scripts/travisci_build_coveri
...
-SPELLINGBLACKLIST=$(cat <<-BLACKLIST
+LINT_BLACKLIST=$(cat <<-EO_LINT_BL
+        -wholename "./common/protocol/Ola.pb.*" -or \
+        -wholename "./common/rpc/Rpc.pb.*" -or \
+        -wholename "./common/rpc/TestService.pb.*" -or \
+        -wholename "./common/rdm/Pids.pb.*" -or \
+        -wholename "./config.h" -or \
+        -wholename "./plugins/*/messages/*ConfigMessages.pb.*" -or \
+        -wholename "./tools/ola_trigger/config.tab.*" -or \
+        -wholename "./tools/ola_trigger/lex.yy.cpp" -or \
```
peternewman: This trailing `-or \` needs removing.

Keeper-of-the-Keys: very embarrassing :|

.travis-ci.sh
```diff
@@ -69,19 +81,12 @@ if [[ $TASK = 'lint' ]]; then
     echo "Found $nolints generic NOLINTs"
   fi;
   # then fetch and run the main cpplint tool
+  lintfiles=$(eval "find ./ \( -name "*.h" -or -name "*.cpp" \) -and ! \( $LINT_BLACKLIST \) | xargs")
```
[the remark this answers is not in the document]
Keeper-of-the-Keys: I'm not sure what you're getting at...

.travis-ci.sh
```diff
@@ -11,6 +11,19 @@ COVERITY_SCAN_BUILD_URL="https://scan.coverity.com/scripts/travisci_build_coveri
...
+LINT_BLACKLIST=$(cat <<EOLBL
+        -wholename "./common/protocol/Ola.pb.*" -or \
+        -wholename "./common/rpc/Rpc.pb.*" -or \
+        -wholename "./common/rpc/TestService.pb.*" -or \
+        -wholename "./common/rdm/Pids.pb.*" -or \
+        -wholename "./config.h" -or \
+        -wholename "./plugins/*/messages/*ConfigMessages.pb.*" -or \
+        -wholename "./tools/ola_trigger/config.tab.*" -or \
+        -wholename "./tools/ola_trigger/lex.yy.cpp" -or \
+        -wholename "./plugins/ftdidmx/FtdiDmxPluginDescription.h"
```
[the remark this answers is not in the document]
Keeper-of-the-Keys: BTW why drop it? Even now it is auto-generated and there is really no reason to check it, just not yet with the contents that annoys the lint checks....

.travis-ci.sh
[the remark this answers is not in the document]
peternewman: Just in terms of generating lintfiles as a separate variable, rather than just doing it in-line in the other code.

peternewman:
Please can you fix the issues codespell caught, I appreciate they aren't yours (sorry). I'm genuinely not sure if the OutputStream one should be struct-based or structured, I'm beginning to suspect the latter.

https://travis-ci.org/OpenLightingProject/ola/jobs/605241223#L1061-L1062

.travis-ci.sh
peternewman: Heh, no worries.

.travis-ci.sh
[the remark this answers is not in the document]
peternewman: There's a bit of me thinks we should hopefully be able to fix it failing the lint and that generally files we self auto-generate we should be able to make lint clean (compared to say the output of the lexer where we don't have any control of that).

Plus if we're going to go down that route, we should ignore ALL the plugin descriptions, not just one, then this needs implementing too really https://github.com/OpenLightingProject/ola/issues/691 (otherwise random developers will find lint-dirty code). I'd also want to double-check what we ship in terms of tarballs, but I think this is built locally so that's not relevant.

plugins/convert_README_to_header.sh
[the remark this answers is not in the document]
Keeper-of-the-Keys: I don't care strongly but I do think that using a character that is not a slash will be more readable for everyone....

.travis-ci.sh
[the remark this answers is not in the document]
Keeper-of-the-Keys: That's what is also done in the other tasks as far as I can tell.

Keeper-of-the-Keys:
@peternewman Is anything still blocking this (the compile failures are I assume due to the race conditions since there is no reason to assume compilation would be affected by a commit that changes no code)?

plugins/convert_README_to_header.sh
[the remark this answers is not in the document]
peternewman: I'd agree in principle, but given how rarely this expression gets changed, and the (presumed) lack of cross-platform testing, I feel the pragmatic solution is to go for the slightly less readable option.

COPYING
```diff
@@ -7,8 +7,8 @@
  of this license document, but changing it is not allowed.
 
 [This is the first released version of the Lesser GPL.  It also counts
- as the successor of the GNU Library Public License, version 2, hence
- the version number 2.1.]
+ as the successor of the GNU Library General Public License, version 2,
```
peternewman: Thanks. I'd always been a bit nervous about whether this counted as a licence change and hadn't ever managed to find a straight answer.

Keeper-of-the-Keys: I'm not a lawyer, but even if it is a license change it is just from LGPL to LGPL....
That being said the official text of LGPL v 2.1 does not feature the word General and TBH these files can probably just be ignored by spellintian since they aren't supposed to change and are "correct"....

(The main reason I did it was I was looking for a change that would cause travis to run again and hopefully pass this time)

peternewman: > I'm not a lawyer, but even if it is a license change it is just from LGPL to LGPL....

GNU LPL to GNU LGPL surely?

> That being said the official text of LGPL v 2.1 does not feature the word General and TBH these files can probably just be ignored by spellintian since they aren't supposed to change and are "correct"....

I'd rather only ignore stuff we can't change, not just stuff we choose not to.

> (The main reason I did it was I was looking for a change that would cause travis to run again and hopefully pass this time)

Don't worry about that. I can re-run them if necessary, and we've currently got bigger issues on OS X and With GCC8 anyway...

Keeper-of-the-Keys: > > I'm not a lawyer, but even if it is a license change it is just from LGPL to LGPL....
> 
> GNU LPL to GNU LGPL surely?
I looked up the language of LGPL and it is what we have.
https://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
> 
> > That being said the official text of LGPL v 2.1 does not feature the word General and TBH these files can probably just be ignored by spellintian since they aren't supposed to change and are "correct"....
> 
> I'd rather only ignore stuff we can't change, not just stuff we choose not to.
Yeah will push an undo.
> 
> > (The main reason I did it was I was looking for a change that would cause travis to run again and hopefully pass this time)
> 
> Don't worry about that. I can re-run them if necessary, and we've currently got bigger issues on OS X and With GCC8 anyway...

peternewman: Thanks @Keeper-of-the-Keys . While I'd looked at the GNU website I'd not thought to check their version of the licence!

peternewman:
> @peternewman Is anything still blocking this (the compile failures are I assume due to the race conditions since there is no reason to assume compilation would be affected by a commit that changes no code)?

I'm hoping to fix the Linux builds at least in #1589 just to check this passes that, it's probably a bit belt and braces.

Keeper-of-the-Keys:
So nothing is blocking this from being merged anymore?

peternewman:
> So nothing is blocking this from being merged anymore?

Correct, well not of yours anyway.

Keeper-of-the-Keys:
@peternewman I took the liberty of purging .travis.yml of all legacy tests to get an idea if this would lead to more successful CI builds and also speed up things...

If you oppose I will push another commit reinstating these checks however as I stated in https://github.com/OpenLightingProject/ola/pull/1589 I believe that gcc-8 should not be part of legacy tests since it is not part of their toolchains, if we want to test gcc-8 we can either add bionic tests to the existing matrix or we can say that "hey master deals with something that will be released and has no relevance on Ubuntu Xenial or OS X from 5 years ago lets treat it that way"

peternewman:
> @peternewman I took the liberty of purging .travis.yml of all legacy tests to get an idea if this would lead to more successful CI builds and also speed up things...
> 
> If you oppose I will push another commit reinstating these checks however as I stated in #1589 I believe that gcc-8 should not be part of legacy tests since it is not part of their toolchains, if we want to test gcc-8 we can either add bionic tests to the existing matrix or we can say that "hey master deals with something that will be released and has no relevance on Ubuntu Xenial or OS X from 3 years ago lets treat it that way"

Can you revert them for now (at least in this PR, as it's essentially ready to be merged). I'll try and find some time soon to respond more fully to your comments in the other PR, although an issue might be a more effective place to do it, currently the discussion is scattered around a number of PRs and other places (probably some on the mailing list too).

peternewman:
@Keeper-of-the-Keys if you can revert the last three commits and either pull in the changes from #1590 or wait for @nomis52 to approve them and me to merge them then we can merge this in.

Keeper-of-the-Keys:
Done and done

common/rdm/DiscoveryAgent.cpp
```diff
@@ -119,6 +119,7 @@ void DiscoveryAgent::InitDiscovery(
   // push the first range on to the branch stack
   UID lower(0, 0);
   m_uid_ranges.push(new UIDRange(lower, UID::AllDevices(), NULL));
+  //m_uid_ranges.push(new UIDRange(UID(4096, 0), UID(8192, 0), NULL));
```
peternewman: Can you remove this please, not sure where it's snuck in from?

common/rdm/DiscoveryAgent.cpp
```diff
@@ -130,6 +131,7 @@ void DiscoveryAgent::InitDiscovery(
  * devices (incremental only).
  */
 void DiscoveryAgent::UnMuteComplete() {
+  OLA_INFO << "UnMuteComplete Callback!";
```
peternewman: Likewise.

Keeper-of-the-Keys:
Whoops 1 second forgot why I was explicitly defining files instead of commit -a because I had that debug stuff and hadn't thrown it out yet.

Keeper-of-the-Keys:
Done

plugins/convert_README_to_header.sh
```diff
@@ -30,8 +30,8 @@ outfilename=`basename $outfile`;
 # See http://stackoverflow.com/a/16576291
 # On Mac OS's sed, \n is not recognized as a newline character, but
 # \[actual newline] works
-desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\"/\\\"/g' -e 's/\n/\\\\n"\\
-"/g' "$path/README.md"`;
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\\\/\\\\\\\/g' -e 's/\"/\\\"/g' -e 's/\n/\\\\n"\\
+"#g' "$path/README.md"`;
```
peternewman: This last hash needs fixing too. I'm not quite sure why Travis isn't picking it up.

Keeper-of-the-Keys: As you can see I moved the whole regex to next line that will prevent such oversights in the future.

peternewman: You've actually just broken it on OS X. Please see the comment a line or two above in the code:
https://github.com/OpenLightingProject/ola/pull/1586/files#diff-e27975dbf242f99283db8699d57acd1cL30-L32

Keeper-of-the-Keys: On Linux we are also not inserting newlines.... we're inserting "\n" into a C++ string that is actually one of the reasons lint doesn't like these files....
For a "true" newline to be insterted it needs to not be escaped....

Keeper-of-the-Keys: This goes back to should we care whether or not autogenerated content goes over 80 chars, if we don't we don't need to insert "true" newlines and just adding "\n" is sufficient for the C++.

If we do then it needs to be dine right all the way on both Linux and OS X.

To me option 1 is the more convincing one, we're "reading" these files only through the C++ code that is rendering the "\n" for us, the original README from which they are generated is what we care about.

Either way a possible way for this to look on Linux is:
-e 's/\n/\\\\n\" \+ \n\"/g'
Where you should also add a separate case if the OS is OS X - 
if [ OS == 'OS X']
desc = [sed for OSX]
else
desc = [sed]

Keeper-of-the-Keys: I'm pushing the Linux change because of backslash rendering on github

Keeper-of-the-Keys:
@peternewman the functionality you requested was not present as is and would have actually broken the C++ string...
I now added it for Linux but have no way to check it for OS X.

plugins/convert_README_to_header.sh
peternewman: > On Linux we are also not inserting newlines.... we're inserting "\n" into a C++ string

No, it's replacing an actual newline with a \n **and** a newline, so we keep the formatting of the original Markdown readme file both within the *PluginDescription.h file and in the char array the file returns.

The original (unmodified) sed expression works correctly on both Linux and OS X. I'd be surprised if your escaping of backslashes has broken it too, but would you like to revert your most recent change and retest it on your machine.

peternewman: The below is a snippet from plugins/dummy/DummyPluginDescription.h generated on my machine with the original sed expression by removing that file and then re-running `make builtfiles`. Are you saying this produces different output on your machine? Please can you give us an example?

```
 * This file has been autogenerated by convert_README_to_header.sh, DO NOT EDIT.
 */
#ifndef PLUGINS_DUMMY_DUMMYPLUGINDESCRIPTION_H_
#define PLUGINS_DUMMY_DUMMYPLUGINDESCRIPTION_H_

namespace ola {
namespace plugin {
namespace dummy {

const char plugin_description[] = "Dummy Plugin\n"
"============\n"
"\n"
"The plugin creates a single device with one port. When used as an output\n"
"port it prints the first two bytes of dmx data to stdout.\n"
```

Keeper-of-the-Keys: On Linux that doesn't work:
`plugins/convert_README_to_header.sh` plugins/dummy/ /tmp/master.out

The above is the version on master, results in (less with line numbers):
afbeelding

Keeper-of-the-Keys: I did a push that adds newlines for Linux (but probably not for OS X) however I think we are doing backflips in the air for functionality that is irrelevant.

peternewman: > On Linux that doesn't work:
> `plugins/convert_README_to_header.sh` plugins/dummy/ /tmp/master.out

What you actually mean is it doesn't work on your version of Linux, so what version of sed and what flavour of Linux are you on?

> The above is the version on master

Just to clarify 100%, you mean on the ola master with no other changes? i.e. https://github.com/OpenLightingProject/ola/blob/master/plugins/convert_README_to_header.sh

@FloEdelmann may have some thoughts too.

I'm tempted to say we should probably look at fixing this issue before we get involved in your other change to the sed command too.

Keeper-of-the-Keys: Ubuntu 19.04

`~$ sed --version
`sed (GNU sed) 4.7

Ran git checkout master which by me is at merge of #1579 then ran the script.

Keeper-of-the-Keys: As pointed out before this behavior has probably been this way at least since Ubuntu 16.04 (xenial) on which the lint check are run and fail on these files.

peternewman: > As pointed out before this behavior has probably been this way at least since Ubuntu 16.04 (xenial) on which the lint check are run and fail on these files.

The lint checks in master run fine on these files, as per the Travis run against that specific commit:
https://travis-ci.org/OpenLightingProject/ola/jobs/576292203

Unfortunately my bionic change (18.04) is only against 0.10 so far which doesn't have this way of generating the plugin code.

peternewman: Although your earlier bionic run did run, and fail:
https://travis-ci.org/OpenLightingProject/ola/jobs/609809927#L974-L977

```
sh ./plugins/convert_README_to_header.sh ./plugins/artnet ./plugins/artnet/ArtNetPluginDescription.h
sh ./plugins/convert_README_to_header.sh ./plugins/dummy ./plugins/dummy/DummyPluginDescription.h
sed: -e expression #6, char 14: unterminated `s' command
sed: -e expression #6, char 14: unterminated `s' command
```

Is that the same as you get on your personal machine with 19.04?

Keeper-of-the-Keys: Yup

Keeper-of-the-Keys: > > As pointed out before this behavior has probably been this way at least since Ubuntu 16.04 (xenial) on which the lint check are run and fail on these files.
> 
> The lint checks in master run fine on these files, as per the Travis run against that specific commit:
> https://travis-ci.org/OpenLightingProject/ola/jobs/576292203
> 
> Unfortunately my bionic change (18.04) is only against 0.10 so far which doesn't have this way of generating the plugin code.

The convert script is new in master? That explains some of it then....

peternewman: > The convert script is new in master? That explains some of it then....

Yeah the convert script is only in master.

I've created a new issue to track some of the issues arising from this https://github.com/OpenLightingProject/ola/issues/1592 .

plugins/convert_README_to_header.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
peternewman: Please can you stop making sweeping changes on chunks of this PR. The whole point of splitting them up was to be able to get small bits in at a time.

If you revert the sh/bash change, and all the changes to sed **except** the one to escape slashes or whatever the initial change was, then I believe this PR can be merged in it's current state. Yes it's not perfect, and there are known issues, but Travis still passes and it doesn't make things any worse. Then we can look at fixing the other bits in a separate PR. Likewise the base one is almost there apart from some trivial failures which Travis is already catching and flagging.

plugins/convert_README_to_header.sh
```diff
@@ -30,8 +30,14 @@ outfilename=`basename $outfile`;
...
 # On Mac OS's sed, \n is not recognized as a newline character, but
 # \[actual newline] works
-desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\"/\\\"/g' -e 's/\n/\\\\n"\\
-"/g' "$path/README.md"`;
+desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\\\/\\\\\\\/g' -e 's/\"/\\\"/g'\
+ -e 's/\n/\\\\n\"\
+\"/g' "$path/README.md"`;
+
+if (( $? != 0 )); then
+	echo 'Sed failed to generate $desc'
```
peternewman: In another PR please, I suspect returning the desc might not be helpful as a flood of text, unless you redirect stderr to stdout too, which has its own risks.

Similar stuff what's doing to all the other backticks now we know of the issue.

Keeper-of-the-Keys: It's not returning the content fo $desc, single quotes don't resolve variables.

Keeper-of-the-Keys:
reverted.
I love how it needs to be tested on every distro out there but Linux.... (because until I pointed out that on Linux it's **not** inserting newlines no one knew or cared)

Keeper-of-the-Keys:
See #1593 for the other improvement.

PR #1588, opened 2019-11-03
Add user chosable base support to StringToInt()

Keeper-of-the-Keys:
Needed for FTDI plugin to create a unique ID for RDM packets based on base36 serial stored in the EEPROM.

### DIFF
--- a/common/utils/StringUtilsTest.cpp
+++ b/common/utils/StringUtilsTest.cpp
@@ -539,7 +539,7 @@ void StringUtilsTest::testStringToBoolTolerant() {
 
 
 void StringUtilsTest::testStringToUInt() {
-  unsigned int value;
+  unsigned int value, value2;
   OLA_ASSERT_FALSE(StringToInt("", &value));
   OLA_ASSERT_FALSE(StringToInt("-1", &value));
   OLA_ASSERT_TRUE(StringToInt("0", &value));
@@ -561,6 +561,14 @@ void StringUtilsTest::testStringToUInt() {
   OLA_ASSERT_FALSE(StringToInt("1 bar baz", &value, true));
   OLA_ASSERT_FALSE(StringToInt("65537cat", &value, true));
   OLA_ASSERT_FALSE(StringToInt("4294967295bat bar", &value, true));
+
+  //Base36 test
+  OLA_ASSERT_TRUE(StringToInt("12FZ9A", &value, false, 36));
+  OLA_ASSERT_TRUE(StringToInt("12fz9a", &value2, false, 36));
+  OLA_ASSERT_EQ(value, value2);
+  OLA_ASSERT_TRUE(StringToInt("STRICT1", &value, true, 36));
+  OLA_ASSERT_TRUE(StringToInt("strict1", &value2, true, 36));
+  OLA_ASSERT_EQ(value, value2);
 }
 
 void StringUtilsTest::testStringToUIntOrDefault() {
--- a/common/utils/StringUtils.cpp
+++ b/common/utils/StringUtils.cpp
@@ -152,13 +152,16 @@ bool StringToBoolTolerant(const string &value, bool *output) {
   return false;
 }
 
-bool StringToInt(const string &value, unsigned int *output, bool strict) {
+bool StringToInt(const string &value,
+                 unsigned int *output,
+                 bool strict,
+                 uint8_t base) {
   if (value.empty()) {
     return false;
   }
   char *end_ptr;
   errno = 0;
-  long long l = strtoll(value.data(), &end_ptr, 10);  // NOLINT(runtime/int)
+  long long l = strtoll(value.data(), &end_ptr, base);  // NOLINT(runtime/int)
   if (l < 0 || (l == 0 && errno != 0)) {
     return false;
   }
@@ -175,9 +178,12 @@ bool StringToInt(const string &value, unsigned int *output, bool strict) {
   return true;
 }
 
-bool StringToInt(const string &value, uint16_t *output, bool strict) {
+bool StringToInt(const string &value,
+                 uint16_t *output,
+                 bool strict,
+                 uint8_t base) {
   unsigned int v;
-  if (!StringToInt(value, &v, strict)) {
+  if (!StringToInt(value, &v, strict, base)) {
     return false;
   }
   if (v > UINT16_MAX) {
@@ -187,9 +193,12 @@ bool StringToInt(const string &value, uint16_t *output, bool strict) {
   return true;
 }
 
-bool StringToInt(const string &value, uint8_t *output, bool strict) {
+bool StringToInt(const string &value,
+                 uint8_t *output,
+                 bool strict,
+                 uint8_t base) {
   unsigned int v;
-  if (!StringToInt(value, &v, strict)) {
+  if (!StringToInt(value, &v, strict, base)) {
     return false;
   }
   if (v > UINT8_MAX) {
@@ -199,13 +208,16 @@ bool StringToInt(const string &value, uint8_t *output, bool strict) {
   return true;
 }
 
-bool StringToInt(const string &value, int *output, bool strict) {
+bool StringToInt(const string &value,
+                 int *output,
+                 bool strict,
+                 uint8_t base) {
   if (value.empty()) {
     return false;
   }
   char *end_ptr;
   errno = 0;
-  long long l = strtoll(value.data(), &end_ptr, 10);  // NOLINT(runtime/int)
+  long long l = strtoll(value.data(), &end_ptr, base);  // NOLINT(runtime/int)
   if (l == 0 && errno != 0) {
     return false;
   }
@@ -222,9 +234,12 @@ bool StringToInt(const string &value, int *output, bool strict) {
   return true;
 }
 
-bool StringToInt(const string &value, int16_t *output, bool strict) {
+bool StringToInt(const string &value,
+                 int16_t *output,
+                 bool strict,
+                 uint8_t base) {
   int v;
-  if (!StringToInt(value, &v, strict)) {
+  if (!StringToInt(value, &v, strict, base)) {
     return false;
   }
   if (v < INT16_MIN || v > INT16_MAX) {
@@ -234,9 +249,12 @@ bool StringToInt(const string &value, int16_t *output, bool strict) {
   return true;
 }
 
-bool StringToInt(const string &value, int8_t *output, bool strict) {
+bool StringToInt(const string &value,
+                 int8_t *output,
+                 bool strict,
+                 uint8_t base) {
   int v;
-  if (!StringToInt(value, &v, strict)) {
+  if (!StringToInt(value, &v, strict, base)) {
     return false;
   }
   if (v < INT8_MIN || v > INT8_MAX) {
--- a/include/ola/StringUtils.h
+++ b/include/ola/StringUtils.h
@@ -328,7 +328,8 @@ bool StringToInt(const std::string &value,
  */
 bool StringToInt(const std::string &value,
                  int16_t *output,
-                 bool strict = false);
+                 bool strict = false,
+                 uint8_t base = 10);
 
 /**
  * @brief Convert a string to a int8_t.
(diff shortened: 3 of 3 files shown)


common/utils/StringUtilsTest.cpp
```diff
@@ -561,6 +561,14 @@ void StringUtilsTest::testStringToUInt() {
   OLA_ASSERT_FALSE(StringToInt("1 bar baz", &value, true));
   OLA_ASSERT_FALSE(StringToInt("65537cat", &value, true));
   OLA_ASSERT_FALSE(StringToInt("4294967295bat bar", &value, true));
+
+  //Base36 test
+  OLA_ASSERT_TRUE(StringToInt("12FZ9A", &value, false, 36));
+  OLA_ASSERT_TRUE(StringToInt("12fz9a", &value2, false, 36));
+  OLA_ASSERT_EQ(value, value2);
```
peternewman: Can we also have a test that they evaluate to the numeric value you're expecting please.

peternewman: Perhaps add a false one too and test octal or hex.

common/utils/StringUtilsTest.cpp
```diff
@@ -561,6 +561,14 @@ void StringUtilsTest::testStringToUInt() {
   OLA_ASSERT_FALSE(StringToInt("1 bar baz", &value, true));
   OLA_ASSERT_FALSE(StringToInt("65537cat", &value, true));
   OLA_ASSERT_FALSE(StringToInt("4294967295bat bar", &value, true));
+
+  //Base36 test
```
peternewman: Please add a space:
`// Base36 test`

common/utils/StringUtilsTest.cpp
```diff
@@ -561,6 +561,14 @@ void StringUtilsTest::testStringToUInt() {
...
+  OLA_ASSERT_TRUE(StringToInt("12FZ9A", &value, false, 36));
+  OLA_ASSERT_TRUE(StringToInt("12fz9a", &value2, false, 36));
+  OLA_ASSERT_EQ(value, value2);
+  OLA_ASSERT_TRUE(StringToInt("STRICT1", &value, true, 36));
```
peternewman: This is currently returning false on Travis:
https://travis-ci.org/OpenLightingProject/ola/jobs/606897230#L21660-L21676

Keeper-of-the-Keys: Indeed, I didn't check numeric values ahead of time STRICT1 overflows an int...

I just ran make on my end I guess that doesn't run the tests because I did not have issues.

common/utils/StringUtilsTest.cpp
Keeper-of-the-Keys: Added octal, hex and evaluation of expected values.

I did not add false tests, adding "false" test cases to other bases seems like double testing to me, we already tested that the function returns false for overflow and for garbage, there is no reason to expect different behavior under a different base (except for "what is garbage?")

Keeper-of-the-Keys:
@peternewman Unless there is something more you want me to add can you re-trigger travis?

common/utils/StringUtilsTest.cpp
peternewman: Thanks.

I'd just like to see one or two false ones, given especially bases other than octal/hex probably aren't frequently used, so it would be good to know it works as expected on the more obscure OSes please.

common/utils/StringUtilsTest.cpp
peternewman: Yeah you need to do "make check" to actually run the tests.

peternewman:
> @peternewman Unless there is something more you want me to add can you re-trigger travis?

As per the other I'd like to get #1589 in so at least some compilation is tested.

common/utils/StringUtilsTest.cpp
Keeper-of-the-Keys: Done I think

common/utils/StringUtilsTest.cpp
Keeper-of-the-Keys: Did that before pushing, none of "my" tests are failing now :)

common/utils/StringUtilsTest.cpp
peternewman: Thanks

Keeper-of-the-Keys:
> Please can we switch the strtoul/strtoll bits of HexStringToInt to use this new function?

Leave HexStringToInt() intact or replace the calls to it with StringToInt()?
Also should it be strict because it does not contain such an argument?

Keeper-of-the-Keys:
@peternewman I took the liberty to replace HexStringToInt hope you don't mind.

Also looking at its' usage it seemed to me that strict mode was most appropriate (though I guess the way it used to work was non-strict, but the data delivered to it should not be longer then the wanted number string)

peternewman:
> > Please can we switch the strtoul/strtoll bits of HexStringToInt to use this new function?
> 
> Leave HexStringToInt() intact or replace the calls to it with StringToInt()?
> Also should it be strict because it does not contain such an argument?
> 
> Edit: according to a quick grep HexStringToInt() is not used very much (5 usages outside of unit tests and function definitions) so I think just replacing these 5 instances with StringToInt() is the way to go.
> 
> Still leaves me with the question are they "strict" or not?

I meant just replace the bit of the call within HexStringToInt() that currently calls strtoul/strtoll with your new options to StringToInt(). You've dropped the protection that the check for ABCDEFabcdef0123456789 offered, or is that still covered by the base argument to strtoul/strtoll etc?

Also we can't just remove the function entirely, we need to add a wrapper version of it and mark it as deprecated with a date in the Doxygen in case other people are using it via our libraries.

Keeper-of-the-Keys:
> > > Please can we switch the strtoul/strtoll bits of HexStringToInt to use this new function?
> > 
> > 
> > Leave HexStringToInt() intact or replace the calls to it with StringToInt()?
> > Also should it be strict because it does not contain such an argument?
> > Edit: according to a quick grep HexStringToInt() is not used very much (5 usages outside of unit tests and function definitions) so I think just replacing these 5 instances with StringToInt() is the way to go.
> > Still leaves me with the question are they "strict" or not?
> 
> I meant just replace the bit of the call within HexStringToInt() that currently calls strtoul/strtoll with your new options to StringToInt().
I actually tried that first but the compiler had a really hard time (ie. failed) to select the right versions of the function and thus failed to compile.
> You've dropped the protection that the check for ABCDEFabcdef0123456789 offered, or is that still covered by the base argument to strtoul/strtoll etc?
I'm not sure what you are asking here? base = 16 covers hex in all its' permutations as far as I understand, I copied all the HexStringToInt() unit tests and replaced them to be StringToInt() with base = 16  and as far as I can tell it passes them.
> 
> Also we can't just remove the function entirely, we need to add a wrapper version of it and mark it as deprecated with a date in the Doxygen in case other people are using it via our libraries.
Understood but I have no solution for the problem I mentioned above...
> 
> Personally I'd be tempted to say that converting a hex string to an int is important enough we could keep the wrapper function which makes it a lot more obvious what it's doing without needing to know what the arguments are. I'm rather surprised we only use it five times, although I think that's because the actual calls are wrapped within a UID/MAC FromString function.

So maybe have a single instance of HexStringToInt() wrapping StringToInt() instead of the 6 different versions we used to have?

Keeper-of-the-Keys:
@peternewman I see what you mean by asking about the "ABCDEFabcdef0123456789" protection, this would be strict = true in StringToInt() which causes a false to be returned if strtoul/strtoll finished to consume the string anywhere before the end of the string.

peternewman:
> > I meant just replace the bit of the call within HexStringToInt() that currently calls strtoul/strtoll with your new options to StringToInt().
> 
> I actually tried that first but the compiler had a really hard time (ie. failed) to select the right versions of the function and thus failed to compile.

Was that with one function or six?

The different versions are to deal with two things, firstly the max value part, but also how it needs to cast things back and forth, see also various other functions in this file.

peternewman:
> > You've dropped the protection that the check for ABCDEFabcdef0123456789 offered, or is that still covered by the base argument to strtoul/strtoll etc?
> 
> I'm not sure what you are asking here? base = 16 covers hex in all its' permutations as far as I understand, I copied all the HexStringToInt() unit tests and replaced them to be StringToInt() with base = 16 and as far as I can tell it passes them.

How are you running the tests your end? Via `make check`? If you look on Travis it's failing them on both compilers.

Keeper-of-the-Keys:
> How are you running the tests your end? Via `make check`? If you look on Travis it's failing them on both compilers.

Yup as far as I understood the output there are failed checks but they are from a different section, but now that I look at the logs I see that this is not the case :man_facepalming: 

The following tests fail:
  OLA_ASSERT_FALSE(StringToInt("0xf", &value, true, 16));
and
  OLA_ASSERT_TRUE(PrefixedHexStringToInt("0xffffffff", &value));

It is unclear to me why the first test should fail... 0xf is valid hex so why should the function fail?

The second one I need to be more awake to be able to analyze.

Keeper-of-the-Keys:
> Was that with one function or six?

Six.
> 
> The different versions are to deal with two things, firstly the max value part, but also how it needs to cast things back and forth, see also various other functions in this file.

Yeah but for some reason even though the same 6 overloaded versions of StringToInt() exist using them in the HexStringToInt() function caused errors that the compiler could not select the right version needed :man_shrugging:

Keeper-of-the-Keys:
OK after looking at this a bit more here are my observations on the 2 test results:

1. This used to be an input resulting in false for the HexStringToInt() function since it only allows 0..F in either case so 0**x** would fail, however strtol() actually accepts 0x, 0X and 0 as prefixes for base 16 and base 8, thus I think we can remove this test.

2. Since strtol() and thus StringToInt() actually **accepts** the prefix this whole function can be deprecated, the problem itself seems to be that the wrong overloaded function is selected since earlier tests of the "ffffffff" string have returned -1 (unless the checks stop instantly on the first fail).

Keeper-of-the-Keys:
PrefixedHexStringToInt() is only used twice:
* examples/ola-rdm.cpp:      (ola::PrefixedHexStringToInt(pid_name, &pid_value) ||
* common/rdm/StringMessageBuilder.cpp:      ola::PrefixedHexStringToInt(input, &int_value) ||

Also excellent candidate for deprecation & replacement with StringToInt()

Keeper-of-the-Keys:
:man_facepalming: When a test fails further tests aren't being run... I'll be pushing an update once I figure out what is failing here...

Keeper-of-the-Keys:
@peternewman I may be wrong here but I'm pretty sure that the current implementation of StringToInt() has a few mistakes in implementation the most glaring seeming to be that in the "int" output version it is doing a static_cast to unsigned int on the output.
https://github.com/OpenLightingProject/ola/blob/90c2691bd9abaf897c962695dc9cb9792eb5ca45/common/utils/StringUtils.cpp#L221

Or is this on purpose?

Keeper-of-the-Keys:
After reviewing the original code and my code and also looking at the way things function it seems to me that the error in the -1 tests comes from the fact that strtol and strtoll are storing in and dealing with variables that are much larger then any of the ints we are dealing with so when it gets for instance `0xff` that is treated as `0x00000000000000ff `and not as `0xffffffffffffffff` now if the function is checking for `> INT8_MAX` it will always be true for negative numbers since their sign is currently not a sign bit but the MSB.

The solution seems to be that we should check for `> UINT8_MAX, UINT16_MAX` etc.

It seems that strtol will basically read the whole sequence of characters as numbers and only a -/+ as the sign for the number from my tests and also from the documentation.

Keeper-of-the-Keys:
The reason why in the "old" code HexStringToInt() worked just fine and dandy while using StringToInt() to take over its' job fails is that HexStringToInt() never bothers to check whether the resulting value is within bounds or not, it just returns the result of strtoll() or strtoul() respectively and lets the runtime handle type conversion etc.

StringToInt() OTOH is checking whether the resulting value is within bounds or not, then since strtoll() treats the string as a whole as a number and only the minus (-) sign is a sign a negative number provided in an encoding that has a sign bit will always exceed INT*_MAX

@peternewman @nomis52 I don't know how to resolve this we seem to have 3*3*2 different wrappers of strtol/strtoul that implement different logic surrounding the function.

Keeper-of-the-Keys:
@peternewman @nomis52 I think I resolved most/all of the logic there is one case of behavior that changes as reflected in https://github.com/OpenLightingProject/ola/pull/1588/commits/7aef2672cec87b1236fff62bc138f9a1c56c31df however I believe the old behavior was wrong if we wanted `0xffff` to be treated as `-1` for a `int16_t` then `0xffffffff` should also be treated as such, StringToInt() is not here to determine whether the original string value is of a similar length in it's binary representation but does answer the question whether the number represented by the string would fit in the variable we chose.

Keeper-of-the-Keys:
Also travis is blowing up on some xenial apt packages...

PR #1589, opened 2019-11-04
Try and fix the g++8 builds on Travis

peternewman:
Mac issues still need looking at in #1584


Keeper-of-the-Keys:
This may be considered cussing but I think we should strongly consider the amount of checks that are being run for OS X, we run 6 different tests for OS X while for Linux we are happy with a two tests (gcc/clang). OS X should be no different test 2 compilers and be done with it or alternatively we add another hour runtime to travis and test the different versions of libftdi cppunit for Linux too.

Keeper-of-the-Keys:
All OS X compiles seem to have failed on protobuf version 3.1 which doesn't seem supported anymore - https://formulae.brew.sh/formula/protobuf

peternewman:
> All OS X compiles seem to have failed on protobuf version 3.1 which doesn't seem supported anymore - https://formulae.brew.sh/formula/protobuf

Yeah unfortunately I hit that in #1584 . Given it's not proved easy to fix yet, I was going to try and get the bits here in so the Ubuntu build is working at least.

> This may be considered cussing but I think we should strongly consider the amount of checks that are being run for OS X, we run 6 different tests for OS X while for Linux we are happy with a two tests (gcc/clang). OS X should be no different test 2 compilers and be done with it or alternatively we add another hour runtime to travis and test the different versions of libftdi cppunit for Linux too.

I appreciate that, I don't believe we can easily install both versions of cppunit at least on Linux and probably libftdi too, hence testing them on OS X, but we should only need to do them on one to confirm the various configure and macro logic works correctly.

Keeper-of-the-Keys:
> > All OS X compiles seem to have failed on protobuf version 3.1 which doesn't seem supported anymore - https://formulae.brew.sh/formula/protobuf
> 
> Yeah unfortunately I hit that in #1584 . Given it's not proved easy to fix yet, I was going to try and get the bits here in so the Ubuntu build is working at least.
> 
> 
> I appreciate that, I don't believe we can easily install both versions of cppunit at least on Linux and probably libftdi too, hence testing them on OS X, but we should only need to do them on one to confirm the various configure and macro logic works correctly.

The question is do we need to support that much? Did we commit to supporting CPPUnit 1.13 or libftdi0? We don't have Linux builds happening for every single Ubuntu release since xenial even though there have been quite a few.

Personally I'm running Ubuntu 19.04 and may end up switching back to Debian or some other Debian based distro due to snap-creep in Ubuntu 19.10, I do understand we need to test some backwards compatibility but when looking at the testing matrix it just seems blown way out of proportion.

peternewman:
> The question is do we need to support that much? Did we commit to supporting CPPUnit 1.13 or libftdi0? We don't have Linux builds happening for every single Ubuntu release since xenial even though there have been quite a few.

16.04 is supported until 2021 (without paid extended support) and is CPPUnit 1.13:
https://packages.ubuntu.com/search?keywords=cppunit&searchon=names&suite=xenial&section=all

So I'd suggest the answer is yes to the older CPPUnit. I'd broadly rather build the latest Ubuntu and OS X versions, so we hit the problems as soon as we can, giving us the most time to fix issues before they impact too many users. We added the CPPUnit/libftdi stuff when they came up, and given they're easy to test I figure we may as well keep testing them while they're still required.

> I do understand we need to test some backwards compatibility but when looking at the testing matrix it just seems blown way out of proportion.

Maybe it is too big, but it shouldn't have much real impact on us, you only really need to care about the OS X specific bits on your final build. #1584 should improve things as it will put the basic quick builds first, like lint and do some other tweaks.

Ideally we'd have broader coverage via buildbot, but people don't seem too engaged with that unfortunately to run slaves:
https://buildbot.openlighting.org/waterfall

Keeper-of-the-Keys:
But we do feel it, when we commit something we're literally waiting hours for travis to finish.
For 0.10 which is packaged in these old distros it is relevant to test however master (0.11) will only be in *future distros and should definitely have a more lean testing profile again (even if it can actually run on older distros there is no need for us to commit to it and promise compatibility)

Keeper-of-the-Keys:
To circle back to my previous comment, we want gcc-8 support, gcc-8 is *not* available in standard Ubuntu Xenial, there does not seem to be a valid reason to bend over backwards and graft it into the xenial test environment.

0.10 -- gets no gcc-8

0.11/master -- for now get an Ubuntu 18.04 test environment in which gcc-8 is available. Upon release the test env will probably be Ubuntu 20.04 unless ola 0.11 gets backported to 18.04

Keeper-of-the-Keys:
Just noticed - according to travis output the compiler used here is gcc-5.4 on bionic it's gcc-7.4 so we're still missing something....

PR #1594, opened 2019-11-17
fix gpio off-by-one in dmx loop

brucelowekamp:
Loop conditional used 1-based dmx start_address compared against array size.  (although actual Get() code below correctly subtracted 1 when referencing the array)

Result was if you set up GPIO with exactly 3 ports and then write exactly [0, 0, 0] to it in code it would ignore the last element when writing to the gpio device.

### DIFF
--- a/plugins/gpio/GPIODriver.cpp
+++ b/plugins/gpio/GPIODriver.cpp
@@ -168,7 +168,7 @@ bool GPIODriver::UpdateGPIOPins(const DmxBuffer &dmx) {
   };
 
   for (uint16_t i = 0;
-       i < m_gpio_pins.size() && (i + m_options.start_address < dmx.Size());
+       i < m_gpio_pins.size() && (i + m_options.start_address - 1u < dmx.Size());
        i++) {
     Action action = NO_CHANGE;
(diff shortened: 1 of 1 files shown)


plugins/gpio/GPIODriver.cpp
```diff
@@ -168,7 +168,7 @@ bool GPIODriver::UpdateGPIOPins(const DmxBuffer &dmx) {
   };
 
   for (uint16_t i = 0;
-       i < m_gpio_pins.size() && (i + m_options.start_address < dmx.Size());
+       i < m_gpio_pins.size() && (i + m_options.start_address - 1u < dmx.Size());
```
peternewman: Looking at some other equivalent code in the SPI plugin, we calculate a `const unsigned int first_slot = m_options.start_address - 1;` then use that throughout, which should be slightly more optimal in the for loop, as well as clearer throughout generally. Would you mind making the same change to your PR?

brucelowekamp:
I don't know of any way to change the base of an existing PR so submitted a new one.

Will update to reflect comment.

peternewman:
See #1595

PR #1595, opened 2019-11-19
fix gpio off-by-one in dmx loop

brucelowekamp:
Loop conditional used 1-based dmx start_address compared against array size. (although actual Get() code below correctly subtracted 1 when referencing the array)

Result was if you set up GPIO with exactly 3 ports and then write exactly [0, 0, 0] to it in code it would ignore the last element when writing to the gpio device.

(resubmitted based against 0.10)


plugins/gpio/GPIODriver.cpp
```diff
@@ -168,7 +168,7 @@ bool GPIODriver::UpdateGPIOPins(const DmxBuffer &dmx) {
   };
 
   for (uint16_t i = 0;
-       i < m_gpio_pins.size() && (i + m_options.start_address < dmx.Size());
+       i < m_gpio_pins.size() && (i + m_options.start_address - 1u < dmx.Size());
```
peternewman: This line is currently too long, it needs to be 80 characters or less.

peternewman: Looking at some other equivalent code in the SPI plugin, we calculate a `const unsigned int first_slot = m_options.start_address - 1;` then use that throughout, which should be slightly more optimal in the for loop, as well as clearer throughout generally. Would you mind making the same change to your PR?

brucelowekamp:
@peternewman Thanks for the feedback.  Probably should have squashed the second change in since I had to create a new PR anyway, but then triggered a full rebuild with all the rebasing and took a few hours on the RPi :)

Do you normally squash PRs as you merge, or should I update that in the PR?

Was just luck I found this by creating exactly the configuration to trigger it in testing my relay setup.   Only took a few minutes to realize it was an off-by-one in the driver when I realized my code updated the DMX monitor but not the relay, but the DMX console did both.  So I knew pretty much what I was looking for :)

plugins/gpio/GPIODriver.cpp
```diff
@@ -48,6 +48,7 @@ using std::vector;
 
 GPIODriver::GPIODriver(const Options &options)
     : m_options(options),
+      m_first_slot(options.start_address - 1u),
```
peternewman: I don't believe you'll need the u after the 1 now. If you do, can you switch to casting it instead which is how we normally do this.

peternewman:
> Do you normally squash PRs as you merge, or should I update that in the PR?

We don't worry too much about the squash so you can just leave it as is if you'd like.
 
> Was just luck I found this by creating exactly the configuration to trigger it in testing my relay setup. Only took a few minutes to realize it was an off-by-one in the driver when I realized my code updated the DMX monitor but not the relay, but the DMX console did both. So I knew pretty much what I was looking for :)

Thanks for the patch anyway!

brucelowekamp:
> Just one minor comment on your code.
> 
> I know it's not yours, but would you mind fixing this typo please:
> ./common/rdm/RDMHelper.cpp:789: semaphone ==> semaphore

no problem.  Been a bit busy getting the light show ready...

plugins/gpio/GPIODriver.cpp
```diff
@@ -48,6 +48,7 @@ using std::vector;
 
 GPIODriver::GPIODriver(const Options &options)
     : m_options(options),
+      m_first_slot(options.start_address - 1),
```
peternewman: Can we stick to calculating this each time UpdateGPIOPins is called please. I know it's not 100% optimised, but it means when we finally get stuff like dynamic config this won't need fixing and it just generally reduces the risk of weird edge cases. Other plugins like SPI do it as follows:
`const unsigned int first_slot = m_options.start_address - 1;`

brucelowekamp: Makes sense.  I'm traveling this week so may not be able to get to it until the weekend.

peternewman: That's fine, it may take me a few days after that to review it too.

brucelowekamp: @peternewman pushed the change.  It will work when m_options is updated dynamically, though right now it's copied as a constant.  But I didn't see a pattern elsewhere for non-constant options or passing in option updates, so I'm assuming you plan on a refactoring to support that across all components in the future.

brucelowekamp: If I understand the latest build output, it's complaining that it doesn't like "acn" anywhere because it thinks it's a misspelling?  :) Assume you will add that to the whitelist, or do I need to rebase to pick up that config change somehow?

peternewman: Thanks @brucelowekamp . Yeah it's more just if we can avoid creating this issue in the first place, it would make future updates a lot simpler if/when someone gets a chance to add that feature. We've got fake RDM control of some stuff like the SPI pixels, so you can configure them from our GUI or over Art-Net etc.

peternewman: > If I understand the latest build output, it's complaining that it doesn't like "acn" anywhere because it thinks it's a misspelling? :) Assume you will add that to the whitelist, or do I need to rebase to pick up that config change somehow?

Indeed that is the issue. I've pushed a change to fix it somewhere but I'm damned if I can find it. We just need to add a -L acn to the codespell command (although it's probably worth making it a variable as it's repeated twice currently):
https://github.com/codespell-project/codespell/blob/master/README.rst#usage

brucelowekamp: @peternewman I pushed a fix that moves the ignore words to a separate file, then renamed the ignore lines file for clarity.  I thought it would be cleaner to keep that out of the build script.

PR #1596, opened 2019-11-24
Usleep

Keeper-of-the-Keys:
Adds a usleep implementation, also makes logic for timer granularity logic available for everyone (instead of just ftdi plugin).

Also adds a wrapper for AsInt() for naming consistency and clarity.


Keeper-of-the-Keys:
Branched from the wrong place :x: closing & redoing later today.

PR #1597, opened 2019-11-24
Usleep

Keeper-of-the-Keys:
Adds InMicroSeconds() as wrapper for AsInt() for naming consistency.
Adds Sleep class with implementation of usleep() and granularity detection
Renames/adds some constants for consistency and use.

### DIFF
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -125,8 +125,8 @@ void BaseTimeVal::AsTimeval(struct timeval *tv) const {
 }
 
 int64_t BaseTimeVal::InMilliSeconds() const {
-  return (m_tv.tv_sec * static_cast<int64_t>(ONE_THOUSAND) +
-          m_tv.tv_usec / ONE_THOUSAND);
+  return (m_tv.tv_sec * static_cast<int64_t>(MSEC_IN_SEC) +
+          m_tv.tv_usec / MSEC_IN_SEC);
 }
 
 int64_t BaseTimeVal::AsInt() const {
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -39,6 +39,8 @@
 #include <sstream>
 #include <string>
 
+#include "ola/Logging.h"
+
 namespace ola {
 
 using std::string;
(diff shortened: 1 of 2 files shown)


common/utils/Clock.cpp
```diff
@@ -272,4 +274,87 @@ void MockClock::CurrentTime(TimeStamp *timestamp) const {
...
 }
+
+Sleep::Sleep(std::string caller) :
+  m_caller(caller) {
+}
+
+/**
+ * @brief Set wanted granularity for usleep and check it.
+ * @note does not check at the nanosecond level,
+ * since internal sturtures use usecs.
```
peternewman: SPaG structures

common/utils/Clock.cpp
```diff
@@ -272,4 +274,87 @@ void MockClock::CurrentTime(TimeStamp *timestamp) const {
...
+ *
+ * @param wanted wanted/needed granularity in usecs
+ * @param maxDeviation max deviation in usecs tolerated by calling thread.
+ *
+ * @attention the granularity of sleep is highly fluctuating depending on the
+ * load of the system, a prior GOOD state is no guarantee for future proper
+ * timing.
+ */
+bool Sleep::CheckTimeGranularity(uint64_t wanted, uint64_t maxDeviation) {
```
peternewman: Please can you change maxDeviation to snake_case.

include/ola/Clock.h
```diff
@@ -40,8 +40,10 @@
 
 namespace ola {
 
+static const uint64_t NSEC_IN_SEC = 1000000000;
 static const int USEC_IN_SECONDS = 1000000;
-static const int ONE_THOUSAND = 1000;
+static const int MSEC_IN_SEC = 1000;
+static const int ONE_THOUSAND = MSEC_IN_SEC;
```
peternewman: Please add a deprecation notice with the date to ONE_THOUSAND and ensure it's not used anywhere else in the codebase any more (common/thread/Mutex.cpp).

include/ola/Clock.h
```diff
@@ -40,8 +40,10 @@
 
 namespace ola {
 
+static const uint64_t NSEC_IN_SEC = 1000000000;
```
peternewman: Can we be consistent please and make these all _IN_SECONDS.

peternewman: Can we reverse the order they're defined and define them based on each other, i.e.:
MSEC_IN_SECONDS = 1000
USEC_IN_SECONDS = MSEC_IN_SECONDS * 1000
etc

include/ola/Clock.h
```diff
@@ -257,5 +266,33 @@ class MockClock: public Clock {
  private:
   TimeInterval m_offset;
 };
+
+enum TimerGranularity { UNKNOWN, GOOD, BAD };
+
+/**
+ * @brief The Sleep class implements usleep with some granualtiry detection.
```
peternewman: SPaG granualtiry->granularity

include/ola/Clock.h
```diff
@@ -257,5 +266,33 @@ class MockClock: public Clock {
...
+ */
+class Sleep {
```
peternewman: I think this should probably be called GranularSleep or something should we ever have to do something like wrap/override usleep or similar.

include/ola/Clock.h
```diff
@@ -257,5 +266,33 @@ class MockClock: public Clock {
...
+ public:
+  explicit Sleep(std::string caller);
+
+  void setCaller(std::string caller) { m_caller = caller; }
+
+  void usleep(TimeInterval requested);
+  void usleep(uint32_t requested);
+  void usleep(timespec requested);
+
+  TimerGranularity getGranularity() { return m_granularity; }
```
peternewman: This should be CamelCase (i.e. GetGranularity)

include/ola/Clock.h
```diff
@@ -257,5 +266,33 @@ class MockClock: public Clock {
...
+enum TimerGranularity { UNKNOWN, GOOD, BAD };
+
+/**
+ * @brief The Sleep class implements usleep with some granualtiry detection.
+ */
+class Sleep {
+ public:
+  explicit Sleep(std::string caller);
+
+  void setCaller(std::string caller) { m_caller = caller; }
```
peternewman: CamelCase again please.

include/ola/Clock.h
```diff
@@ -257,5 +266,33 @@ class MockClock: public Clock {
...
+
+  void usleep(TimeInterval requested);
```
peternewman: Can you uppercase the first letter of these please.

include/ola/Clock.h
peternewman: Actually looking at the code more, ONE_THOUSAND is still in regular use, so scrap that.

include/ola/Clock.h
peternewman: Better yet, please define them based on ONE_THOUSAND.

common/utils/Clock.cpp
peternewman: This shouldn't be called Check... given it's actually usleeping. Should there not also be one that uses the defined wanting and max deviation times? Test maybe, or ideally a similar synonym.

common/utils/Clock.cpp
```diff
@@ -272,4 +274,87 @@ void MockClock::CurrentTime(TimeStamp *timestamp) const {
...
+
+  timespec t;
+  t.tv_sec = wanted / USEC_IN_SECONDS;
+  t.tv_nsec = (wanted % USEC_IN_SECONDS) * ONE_THOUSAND;
+
+  clock.CurrentTime(&ts1);
+  this->usleep(1);
+  clock.CurrentTime(&ts2);
+  TimeInterval interval = ts2 - ts1;
+  m_clock_overhead = interval.InMicroSeconds();
```
peternewman: Surely this should be interval -1 given you've asked it to sleep for one microsecond?

include/ola/Clock.h
```diff
@@ -160,6 +168,7 @@ class TimeInterval {
   int32_t MicroSeconds() const { return m_interval.MicroSeconds(); }
 
   int64_t InMilliSeconds() const { return m_interval.InMilliSeconds(); }
+  int64_t InMicroSeconds() const { return m_interval.InMicroSeconds(); }
```
peternewman: Please can you correct the CamelCase of this and it's siblings to InMicroseconds (and throughout the codebase). We'll just add a note to the NEWS file like when we renamed DMXRecieved(sic) to DMXReceived.

common/utils/Clock.cpp
```diff
@@ -272,4 +274,87 @@ void MockClock::CurrentTime(TimeStamp *timestamp) const {
...
+void Sleep::usleep(TimeInterval requested) {
+  timespec req;
+  req.tv_sec = requested.Seconds();
+  req.tv_nsec = requested.MicroSeconds() * ONE_THOUSAND;
+
+  this->usleep(req);
+}
+
+void Sleep::usleep(uint32_t requested) {
+  timespec req;
```
peternewman: This should be using the existing maths in TimeInterval rather than rewriting that code here.

include/ola/Clock.h
Keeper-of-the-Keys: renamed all

include/ola/Clock.h
Keeper-of-the-Keys: done

include/ola/Clock.h
Keeper-of-the-Keys: The whole point of this class is to do usleep without the various test systems screaming at us that usleep shouldn't be used.

common/utils/Clock.cpp
Keeper-of-the-Keys: It's usleeping but it's only meant to be used to get a rough idea of how accurate usleep is after which further sleeps should be done with Usleep() 

(which still remains a total crapshoot no matter what unless we/I find out if there is a way to trigger the program to be run in a real time fashion but afaik that is changing the kernel scheduler settings, so that would just be adding docs that "if the user wants really acurate timings they should do ....")

common/utils/Clock.cpp
Keeper-of-the-Keys: You are correct, but in the grand scheme of timer inaccuracy bouncing all over the place I don't think it matters much (but I will fix it :grin: )

include/ola/Clock.h
Keeper-of-the-Keys: Did this, should I add a note to NEWS or will you once you merge?

common/utils/Clock.cpp
Keeper-of-the-Keys: I'm not sure I follow, we have a TimeIntervalToTimeSpec function somewhere?

Keeper-of-the-Keys: (It does make sense to add such a function but I did not see any such thing listed in the header)

Issue #1506, opened 2018-10-27
Python 3 Compatibility


peternewman:
> if i try the simple fade example this does not work:
> stefan:~/ola/python/examples[master]$ python3 ./ola_simple_fade.py
> Traceback (most recent call last):
>   File "./ola_simple_fade.py", line 69, in <module>
>     wrapper.AddEvent(SHUTDOWN_INTERVAL, wrapper.Stop)
>   File "/home/stefan/ola/python/ola/ClientWrapper.py", line 287, in AddEvent
>     self._ss.AddEvent(time_in_ms, callback)
>   File "/home/stefan/ola/python/ola/ClientWrapper.py", line 215, in AddEvent
>     heapq.heappush(self._events, event)
> TypeError: unorderable types: _Event() < _Event()

See:
https://groups.google.com/d/msg/open-lighting/7TuvU0T1CAo/0hBbXHw8BgAJ
https://groups.google.com/d/msg/open-lighting/7TuvU0T1CAo/oVIYBMk8BgAJ

peternewman:
See also https://github.com/OpenLightingProject/ola/pull/1514

peternewman:
And #1537

brucelowekamp:
I ran into this problem b/c my first attempt to use ola was on a mac with homebrew (btw, thanks, having great fun with ola), and mac homebrew builds it against python 3.  

I actually considered whether it would be possible to fix the formula to use python2, but it would require a py2 build of protobuf3.6, which given python 2 is going unsupported (https://pythonclock.org/) etc seemed like it might be hard to get approved (though just copying the python code over works, of course)

I have a local build with #1537 applied that resolved the problems, though as I'm deploying on a RPi I forced everything back to py2 after awhile anyway.  Regardless +1 for getting the py3 compatibility shipped.  Do agree with the comments on #1537 about the need for more tests especially if intention is to ship both py2 and py3 for awhile.   I might be able to devote a bit of time to some of that in the next couple of months, if needed.

peternewman:
> I ran into this problem b/c my first attempt to use ola was on a mac with homebrew (btw, thanks, having great fun with ola), and mac homebrew builds it against python 3.

[...]

PR #1583, opened 2019-09-22
Gcc 9 for 0.10 branch

yoe:
This gets all the changes that are also on #1578, but on top of the 0.10 branch.

It should actually work, as opposed to #1568...

PR #1605, opened 2020-01-03
Py3 cmp

brucelowekamp:
This picks up on @nils-van-zuijlen work on fixing cmp for python3 compatibility in PR #1538   My client code works in python3 after these changes, so might be enough for #1506 though would like to get it built with python3 as well.

Note that Nils had added comments that functools requires dropping 2.7.  As far as I can tell the features in 3.x were backported to 2.7, so dropping 2.6 would let a lot of the boilerplate here be ignored.  Also, sorting None elements in with UID and MACAddress appear to work (judging by the tests) with the current implementation when functools.total_ordering is used.  I updated comments to reflect that.

Changes in here on top of #1538 
- move to isinstance rather than comparing class
- add tests for all the comparisons as well as ClientWrapper events and a simple DMX send test
- implement all other comparisons in terms of eq and lt
- ensure that hash are == when eq == True by not including extra fields in hash (alternatively if the other fields should be in hash, they should be included in cmp)
- ResponderTest implements cmp by comparing names, which I think matches the original __cmp__.  The only use I can come up for this is sorting the subclass test implementations, so this functionality is preserved in the comparisons, though I didn't find it used anywhere.  

minor random changes:
- support creating ClientWrapper events with TimeDelta instead of ms
- exit ClientWrapper loop faster when stopping, though check sockets on our way out
- 9010 was defined as a constant, but not used in code
- the code seems to categorically ignore two other flake8 tests, so added them to exclude docs


peternewman:
Firstly thanks for this!

> Note that Nils had added comments that functools requires dropping 2.7. As far as I can tell the features in 3.x were backported to 2.7, so dropping 2.6 would let a lot of the boilerplate here be ignored.

So taking Ubuntu as a good basemark, if we ignore the paid support option then 16.04 is the oldest supported OS, it's Python is still only 2.7:
https://packages.ubuntu.com/search?keywords=python&searchon=names&exact=1&suite=xenial&section=all

Although even 14.04 had 2.7...

I think 0.10.8 will be the final 0.10 branch release as there are some RDM changes which become too much of a new feature for a patch release, so that will help things too. If this PR becomes too awkward we can just push it into 0.11 anyway.

README.developer
```diff
@@ -241,7 +241,7 @@ flake8 can be installed following the instructions here:
 
 Run it with:
   --max-line-length 80 --exclude *_pb2.py,.git,__pycache \
-  --ignore E111,E121,E127,E129 data/rdm include/ola python scripts \
+  --ignore E111,E114,E121,E127,E129,W504 data/rdm include/ola python scripts \
```
peternewman: This should match https://github.com/OpenLightingProject/ola/blob/0.10/.travis-ci.sh#L247

Would you also mind adding a note to the Travis script reminding editors to keep README.developer updated, at least until something equivalent to #691 is in.

python/ola/MACAddress.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, mac_address):
 
     Args:
       mac_address: The byte array representation of the MAC Address, e.g.
-      bytearray([0x01, 0x23, 0x45, 0x67, 0x89 0xab]).
+      bytearray([0x01, 0x23, 0x45, 0x67, 0x89, 0xab]).
```
peternewman: Good spot!

.travis.yml
```diff
@@ -287,6 +287,7 @@ before_cache:
 install:
 # Match the version of protobuf being installed via apt
   - pip install --user protobuf==3.1.0
+  - pip install --user timeout-decorator
```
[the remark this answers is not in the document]
brucelowekamp: That's used by the Event and OlaClient tests in ClientWrapperTest.  Not used outside the tests.  Honestly it could be dropped if it were preferable to let the tests hang if something breaks and let the build system kill them, but I hate tests that do that :)

Actually, I realized I forgot to add something to disable those tests in Windows in Py2.  It relies on socketpair that wasn't supported natively in Win until 3.5, and IMHO it wouldn't be worth adding one of the various implementations at this point.  Will add that.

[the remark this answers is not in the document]
brucelowekamp: Checked and the python code is already not windows compatible so not changing anything in that test.

README.developer
brucelowekamp: I added the comment.

In the build with py3 change I actually added a testcase for make check that runs python compileall to make sure all of the py at least compiles.  Would be pretty easy to do the same for flake8 since it's available via pypi.

brucelowekamp:
> Can you add these new test scripts into debian/tests/control so they are tested as part of the Debian CI too.

Haven't worked with debuild before but will try to get it added in the next day or two.

brucelowekamp:
@peternewman I got debuild working to the point toward the end that it noticed that my protobuf install wasn't from a deb and quit (I have a hand-built protobuf since I couldn't find a package for 3.1.0).  There doesn't seem to be a way to bypass that check.  Do you know the location of a package I could use instead or another way to get around that?

README.developer
peternewman: Sounds good although I suspect the flake8 thing should be part of the lint target rather than check, as per https://github.com/OpenLightingProject/ola/issues/691 because it's nice to encourage people to run make check, but we don't want to force them to have flake8 installed. Although for bonus points configure could handle it all seamlessly. Although I suspect it should still be a separate target, I often write rather messy code while developing, but still running make check, which I wouldn't want to fail on that, before tidying up ready for committing.

common/web/SchemaParserTest.cpp
```diff
@@ -159,6 +159,7 @@ void JsonSchemaParserTest::ReadTestCases(const string& filename,
   const string comment_prefix = "//";
   string line;
   while (getline(in, line)) {
+    line.erase(line.find_last_not_of("\r")+1);
```
peternewman: Can we have a comment to our future selves explaining why this is here.

peternewman:
> I think leaving 2.6 supported until the next major release makes sense. Having put the boilerplate code in, it doesn't really cost anything to leave it. If someone needs to refactor again they could remove the extra comparisons and do some things like shift the protocol classes to use the attrs package rather than tons of boilerplate, but it's fine now.

Sounds very sensible

> I don't believe 2.7 is going anywhere :)

Indeed I suspect you're right.

> IMHO this change ought to be in 10.8 for those who want to use the OLA client API with a py3 project. It would require them to do a manual install, but it would unblock them (and those who try to use the mac brew build that forced py3).

Yeah we always seem to be playing catchup now with new releases.

My thinking is release 0.10.8 probably as soon as this is in, possibly after one or two other minor bugs. Then I suspect release 0.11.0 almost immediately as it's had a queue of shiny new features appearing since 2016 (gulp), but no-one seems to offer to test it thoroughly when I ask, branch that off for 0.11.1 if there are any bugfixes, then master becomes what will be 0.12.0 and people can start culling or breaking things if they like, such as C++11 and Python 2, rather than throw away a lot of known good compatibility for no good reason. Then people have a choice of a bug fix on what they're using (0.10.8), latest stable (0.11.0) or bleeding edge (master).

> I have more changes that are mostly finished to make the whole thing build and install with python3. There are quite a few changes but they're mostly mechanical and not in the client sdk (Nils got most of them there). I don't know that there's a reason not to put the additional changes into 10.8 either, but as there's one non-trivial change (short but I will need help validating it), need for a new builder, and as there's no test-suite I don't really know if it runs, I felt like I would put it into a separate PR.

Thanks @brucelowekamp sounds sensible, we can make a judgement on if it goes into 0.10.8 or gets pushed back.

peternewman:
> @peternewman I got debuild working to the point toward the end that it noticed that my protobuf install wasn't from a deb and quit (I have a hand-built protobuf since I couldn't find a package for 3.1.0). There doesn't seem to be a way to bypass that check. Do you know the location of a package I could use instead or another way to get around that?

That's great news.

So the official Debian release gets round this by cherry-picking our hack that's only in our 0.10 branch to make Protobuf work with a newer version, I'm still on an older release which hence has an old protobuf and doesn't have these issues.

I suspect your best bet might be to test the debuild stuff against @yoe 's branch which is where the Debian package, and hence our debuild usage comes from:
https://github.com/OpenLightingProject/ola/compare/0.10...yoe:master

Then just cherry-pick your debuild changes back from a fork of that into this PR or something? He can probably advise either way or possibly even give you some useful CI or something, he's certainly our guru when it comes to Debian!

peternewman:
To clarify here, do you think this PR is ready to go aside from debuild? I think while tests are nice, if it looks like it will be a long haul or you don't get too far with what it and you've tested what you've committed (which I'm guessing you have), we could get it merged and leave debuild as a nice to have and leave the issue open (or raise a new one) to track that.

brucelowekamp:
> To clarify here, do you think this PR is ready to go aside from debuild? I think while tests are nice, if it looks like it will be a long haul or you don't get too far with what it and you've tested what you've committed (which I'm guessing you have), we could get it merged and leave debuild as a nice to have and leave the issue open to track that as well as general Python code python3 compatibility.

I've used it with my code in both 2.7 and 3.7.  Now my code only sends and doesn't receive or do any sort of management, but as far as I can tell the changes are consistent.  As I said in the other thread I'm not sure the new tests will pass in 2.6, but I believe that would be it.

Having said that, I went to look again and found that @nils-van-zuijlen original change altered the semantics of the comparison in PidStore.py.  I'm not sure why (I tried to avoid that and made sure the hashes and comparisons were consistent, but I missed this one).  Will ask if he remembers why.

nils-van-zuijlen:
I added checks for the other fields in https://github.com/OpenLightingProject/ola/pull/1538/commits/3d9d8d81e3c04f0dc20bd900b2c049a05d237e51 from @peternewman 's question in https://github.com/OpenLightingProject/ola/pull/1538/commits/3d9d8d81e3c04f0dc20bd900b2c049a05d237e51#r257505257

brucelowekamp:
> > I added checks for the other fields in [3d9d8d8](https://github.com/OpenLightingProject/ola/commit/3d9d8d81e3c04f0dc20bd900b2c049a05d237e51) from @peternewman 's question in [3d9d8d8#r257505257](https://github.com/OpenLightingProject/ola/commit/3d9d8d81e3c04f0dc20bd900b2c049a05d237e51#r257505257)
> 
> Thanks. I agree they should be there, but I realized I missed PidStore.py entirely, including testing. Hope to get some tests in later this week.

@peternewman @nomis52 In doing a PidStore test (pulling tests from PidStoreTest.cpp), I discovered there is no override support in the python PidStore (loading common/rdm/testdata/pids fails).  Doesn't appear hard to implement, but a comment in PidStore.h says higher version is a newer pid, but the PidStoreTest.cpp appears to assume the lower version will be used.   It looks relatively easy to add version to the store and pids, but wanted to check that the "lower version wins" behavior is correct before I implement it.

peternewman:
> @peternewman @nomis52 In doing a PidStore test (pulling tests from PidStoreTest.cpp), I discovered there is no override support in the python PidStore (loading common/rdm/testdata/pids fails). Doesn't appear hard to implement, but a comment in PidStore.h says higher version is a newer pid, but the PidStoreTest.cpp appears to assume the lower version will be used. It looks relatively easy to add version to the store and pids, but wanted to check that the "lower version wins" behavior is correct before I implement it.

Oops, good spot!

Where's the comment and test about versions, the only stuff I can find is here:
https://github.com/OpenLightingProject/ola/blob/master/include/ola/rdm/PidStore.h#L64-L66

I think the version comment is prior to overrides existing and I don't believe we actually check versions anywhere. The idea though is the override file gets loaded regardless of how versions might compare.

One thing, if we add this to the Python code, which as you say we should, it should probably be disabled by default in the RDM tests/model collector because I think by default it should probably use the "official" data.

peternewman:
> I added checks for the other fields in [3d9d8d8](https://github.com/OpenLightingProject/ola/commit/3d9d8d81e3c04f0dc20bd900b2c049a05d237e51) from @peternewman 's question in [3d9d8d8#r257505257](https://github.com/OpenLightingProject/ola/commit/3d9d8d81e3c04f0dc20bd900b2c049a05d237e51#r257505257)

@nils-van-zuijlen what's the link to my question, it just seems to point to the commit?

nils-van-zuijlen:
It's https://github.com/OpenLightingProject/ola/pull/1538#discussion_r257505257 

Sorry, I didn't check if the previous link was working as intended.

brucelowekamp:
Ah, I found a description of what the version number was, but not the overrides.proto comment (or the code corresponding to it).

Should be pretty easy to implement.   Hope to get it in over the weekend and rebase the code on top of current 0.10.  

Bruce

python/ola/PidStore.py
```diff
@@ -851,7 +895,7 @@ def GetDescription(self, indent=0):
       names.append('<%s>' % atom.name)
       output.append(atom.GetDescription(indent=2))
 
-    return ' '.join(names), '\n'.join(output)
+    return ' '.join(names) + ':\n' + '\n'.join(output)
```
brucelowekamp: I want to call out this change b/c it's an interface change that changes the return type of GetDescription from a tuple to a string.  I couldn't find anything else in the code base that would depend on that and doubt it was deliberate, but I could be wrong.  However, in addition to making more sense to return a string, the tuple implementation broke for Groups inside a Group (b/c it wound up trying to convert a tuple to a string) so needed to have a string here.

brucelowekamp:
> It's [#1538 (comment)](https://github.com/OpenLightingProject/ola/pull/1538#discussion_r257505257)
> 
> Sorry, I didn't check if the previous link was working as intended.

I have the comparison checking name, value, requests, and responses.  It's a bit of a hack b/c doing a true deep comparison would be a huge effort, but comparing GetDescription seems reasonable and mostly correct.

brucelowekamp:
Unfortunately the last docker desktop update entirely broke volume sharing (specifically tar) so I can't currently run distcheck in my normal build environment.  Will see if I can fix it or create another somewhere else.  Build had a problem finding the Pid test files, which I was afraid of.

common/rdm/PidStoreTest.cpp
```diff
@@ -409,7 +409,7 @@ void PidStoreTest::testPidStoreDirectoryLoad() {
 
   vector<const PidDescriptor*> all_pids;
   esta_store->AllPids(&all_pids);
-  OLA_ASSERT_EQ(static_cast<size_t>(4), all_pids.size());
+  OLA_ASSERT_EQ(static_cast<size_t>(6), all_pids.size());
```
brucelowekamp: This change and the test data change due to PidStore.py requiring DEVICE_INFO, QUEUED_MESSAGES, and SUPPORTED_PARAMETERS to be defined when loading the pidstore

brucelowekamp:
@peternewman I'm sure you're aware, but there are quite a few changes in master but not in 0.10 that you made to the python code (particularly pidstore) back in 2016-17.  One of the bugs I found was also fixed there (and another bug I didn't find) so I pulled those changes in, but not all.  Hopefully it will automatically merge back into master now.

README.developer
peternewman: I'll resolve this comment as I've opened #1619 to track that feature request.

common/web/SchemaParserTest.cpp
peternewman: LGTM

peternewman:
Sorry for being so rubbish @brucelowekamp .

Some easy bits first, these two have both changed permissions from 644 to 755, could you revert that please:
- [ ] common/web/SchemaParserTest.cpp
- [ ] libs/usb/LibUsbAdaptor.h

python/ola/MACAddressTest.py
```diff
@@ -27,6 +28,15 @@
...
 class MACAddressTest(unittest.TestCase):
 
+  def allNotEqual(self, t):
+    for pair in itertools.combinations(t, 2):
+      self.assertNotEqual(pair[0], pair[1])
+
+  def allHashNotEqual(self, t):
+    h = map(hash, t)
+    for pair in itertools.combinations(h, 2):
+      self.assertNotEqual(pair[0], pair[1])
```
peternewman: Given these two functions are used in lots of the cases, can you pull them out into a TestUtils or something file and just import that each time, like we do in C++ land.

python/ola/PidStore.py
```diff
@@ -1065,11 +1132,22 @@ def NameToValue(self, pid_name, esta_id=None):
...
+    """A helper method to convert a manufacturer ID to a name
+
+    Args:
+      esta_id: The 2-byte esta / manufacturer ID.
+
+    Returns:
+      The name of the manufacturer, or None if it wasn't found.
+    """
+    return self._manufacturer_id_to_name.get(esta_id, None)
+
```
peternewman: For review info, this is just ported from master.

python/ola/PidStore.py
```diff
@@ -981,14 +1044,18 @@ def LoadFile(self, pid_file_name, validate):
           manufacturer.manufacturer_name)
 
       for pid_pb in manufacturer.pid:
+        if override:  # if processing overrides delete any conflicting entry
+          old = pid_dict.pop(pid_pb.value, None)
+          if old is not None:
+            del name_dict[old.name]
```
peternewman: Presumably you're not just doing pid_pb.value in pid_dict because you need to handle the name_dict deletion?

python/ola/PidStore.py
```diff
@@ -928,17 +973,32 @@ def __init__(self):
...
+
+    raw_list = pid_files
+    pid_files = []
+    override_file = None
+
+    for f in raw_list:
+      if os.path.basename(f) == "overrides.proto":
+        override_file = f
+      else:
+        pid_files.append(f)
```
peternewman: I think you may want to skip manufacturer_names.proto here too, as per the C++:
https://github.com/OpenLightingProject/ola/blob/e90be831d23aad074cc7efd7ebc75f9d6daf7b75/common/rdm/PidStoreLoader.cpp#L93-L95

python/ola/PidStore.py
```diff
@@ -157,14 +157,58 @@ def ValidateAddressing(self, args, command_class):
...
+
+  def __ne__(self, other):
+    return not self == other
 
   def __str__(self):
     return '%s (0x%04hx)' % (self.name, self.value)
 
   def __hash__(self):
-    return self._value
+    return hash((self._name, self._value))
```
peternewman: I think value is the key bit (and manufacturer where present, but I think that's handled in a different way here). I think if we have two PIDs with the same numeric value and different names, we probably WANT to have a collision, or flag an exception or whatever.

python/ola/PidStore.py
```diff
@@ -157,14 +157,58 @@ def ValidateAddressing(self, args, command_class):
...
+                       for f in [RDM_GET, RDM_SET, RDM_DISCOVERY]])
+    ax = setToDescription(a)
+    bx = setToDescription(b)
+    return (ax > bx) - (ax < bx)  # canonical for cmp(a,b)
+
+  def __eq__(self, other):
+    if not isinstance(other, self.__class__):
+      return False
+    return (self.value == other.value and
+            self.name == other.name and
```
peternewman: As below, I think name is probably irrelevant. I guess requests/responses might be for some merge scenario.

python/ola/PidStore.py
```diff
@@ -157,14 +157,58 @@ def ValidateAddressing(self, args, command_class):
...
+            self._GroupCmp(self._requests, other._requests) == 0 and
+            self._GroupCmp(self._responses, other._responses) == 0)
+
+  def __lt__(self, other):
+    if not isinstance(other, self.__class__):
+      return NotImplemented
+    if self.value != other.value:
+      return self._value < other._value
+    if self.name != other.name:
+      return self.name < other.name
```
peternewman: As other comments.

python/ola/PidStoreTest.py
```diff
@@ -0,0 +1,221 @@
...
+# PidStoreTest.py
+# Copyright (C) 2020 Bruce Lowekamp
+
+import itertools
+import os
+import unittest
+import ola.PidStore as PidStore
+
+"""Test cases for the PidStore class.
+   Relies on the PID data from rdm tests,
```
peternewman: Nit SPaG RDM

common/rdm/PidStoreTest.cpp
peternewman: Do you actually mean PidStoreTest.py? i.e. you've written your tests to test against real PIDs and then added them to the test data?

peternewman: Also PROXIED_DEVICE_COUNT, PROXIED_DEVICES and COMMS_STATUS from the looks of things?

.travis.yml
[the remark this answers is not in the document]
peternewman: If it's required, we should add it to the Debian, configure script etc. Can I suggest taking out the decorators for the moment until we've got this in, and then looking at adding them back in later to simplify the merge, given if the tests are passing it will work fine without the decorators.

common/rdm/PidStoreTest.cpp
brucelowekamp: yes, I copied some of the real pid data into the pid test data so the python sanity checks were satisfied.  I made that change to the cpp tests so they accepted the new test data, but didn't add the same sanity checks to the cpp code.

tools/rdm/TestState.py
```diff
@@ -15,6 +15,7 @@
 # TestState.py
 # Copyright (C) 2011 Simon Newton
 
+
```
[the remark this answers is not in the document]
brucelowekamp: I can't think why else I would have done that, but I'll take it out and see if flake8 complains...

tools/rdm/ResponderTestTest.py
```diff
@@ -0,0 +1,60 @@
...
+
+class ZTestFixture(TestFixture):
+  pass
+
+
+class TestFixtureTest(unittest.TestCase):
+
+  def testCmp(self):
+    base = TestFixture({}, 2, 123, None)
+    base2 = TestFixture({}, 2, 123, None)
```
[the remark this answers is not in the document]
brucelowekamp: changing this to test that behavior and leaving a and z to test that the classname comparison is working

python/ola/UIDTest.py
```diff
@@ -27,16 +27,22 @@
...
     uid = UID(0x707a, 0x12345678)
     self.assertEqual(0x707a, uid.manufacturer_id)
     self.assertEqual(0x12345678, uid.device_id)
     self.assertEqual('707a:12345678', str(uid))
 
-    # Python 3 does not allow sorting of incompatible types.
-    # We don't use sys.version_info.major to support Python 2.6.
-    if sys.version_info[0] == 2:
-        self.assertTrue(uid > None)
+    self.assertTrue(uid > None)
```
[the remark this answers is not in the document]
brucelowekamp: There is a comment in the code that comparison with None was needed for at least one of the types, I think for lists that would have None.  I haven't tried to figure out if it's needed everywhere but preserved it where it is.

strings/notImplemented tested with "hello" further down

tools/rdm/ResponderTest.py
```diff
@@ -76,8 +76,34 @@ def __str__(self):
...
 
-  def __cmp__(self, other):
-    return cmp(self.__class__.__name__, other.__class__.__name__)
+  # original __cmp__ compared __class__.__name__.  This sorts
+  # subclasses, so this implementation preserves that behavior.
+  def __eq__(self, other):
+    return other.__class__ is self.__class__
+
+  def __lt__(self, other):
+    if not isinstance(other, TestFixture):
```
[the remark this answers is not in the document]
brucelowekamp: Updated the comment and switched to issubclass (agree if someone doesn't read the comment, it makes it a bit clearer what's going on)

tools/rdm/ResponderTest.py
```diff
@@ -76,8 +76,34 @@ def __str__(self):
   def __repr__(self):
     return self.__class__.__name__
 
-  def __cmp__(self, other):
-    return cmp(self.__class__.__name__, other.__class__.__name__)
+  # original __cmp__ compared __class__.__name__.  This sorts
+  # subclasses, so this implementation preserves that behavior.
+  def __eq__(self, other):
+    return other.__class__ is self.__class__
```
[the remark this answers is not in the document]
brucelowekamp: So I think that would have a different behavior if someone were to write two TestFixtures with the same name in different scopes.  I would believe those are different classes and so should be !=, i.e. the current implementation.  Technically the __lt__ only gives partial ordering in this case and could do total ordering if we compared class id if the names are equal, but this strikes me as not really needed.

python/ola/OlaClientTest.py
```diff
@@ -0,0 +1,128 @@
...
+
+  def allHashNotEqual(self, t):
+    h = map(hash, t)
+    for pair in itertools.combinations(h, 2):
+      self.assertNotEqual(pair[0], pair[1])
+
+  def testPlugin(self):
+    # hash and eq only on id (1)
+    a = Plugin(1, "a", False, False)
+    aeq = Plugin(1, "a", False, True)
```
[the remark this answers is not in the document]
brucelowekamp: Funny, I think I did a2 first and later decided aeq was clearer :)

Regardless, updated so the other fields are all different.

python/ola/ClientWrapper.py
```diff
@@ -36,16 +36,48 @@ class _Event(object):
   """An _Event represents a timer scheduled to expire in the future.
 
   Args:
-    time_ms: the number of ms before this event fires
+    delay: timedelta or number of ms before this event fires
```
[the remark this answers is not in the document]
brucelowekamp: it's clearer

python/ola/ClientWrapper.py
```diff
@@ -186,8 +218,10 @@ def Run(self):
       sleep_time = 1
       now = datetime.datetime.now()
       self._CheckTimeouts(now)
+      if self._quit:
+        sleep_time = 0
```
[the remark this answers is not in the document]
brucelowekamp: yeah, otherwise there was a 1 second delay, which probably isn't ever an issue in real life, but was very noticeable in the tests

python/ola/ClientWrapperTest.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class ClientWrapperTest(unittest.TestCase):
+
+  def testEventCmp(self):
+    a = _Event(0.5, a_func)
+    b = _Event(1, b_func)
+    b2 = _Event(1, b_func)
+    b2._run_at = b._run_at  # tests rely on timestamps being actually equal
+    b3 = _Event(1, a_func)
```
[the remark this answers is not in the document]
brucelowekamp: I updated the comment just to make it clear what b, b2, b3, and c are.  I'm not quite sure c is actually needed but left it there anyway.

python/ola/OlaClient.py
```diff
@@ -972,6 +1027,10 @@ def RegisterUniverse(self, universe, action, data_callback, callback=None):
     Returns:
       True if the request was sent, False otherwise.
     """
+
+    if data_callback is None and action == self.REGISTER:
```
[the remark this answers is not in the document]
brucelowekamp: @nils-van-zuijlen did that https://github.com/OpenLightingProject/ola/commit/889c00c458e5e5940c49649fe3955013c937914d to enforce there is a callback on register but not a callback on unregister.

python/ola/MACAddressTest.py
brucelowekamp: Yeah, I think I even had it that way initially and got tripped up by the distcheck setup.  I think I figured that out later, so will move back to that setup.

python/ola/PidStore.py
brucelowekamp: There's no guarantee that entities in the overrides are actually already defined, so this handles deleting if there cleanly.

python/ola/PidStore.py
brucelowekamp: Interesting.  That's in main but not 0.10.  Put it in and will see if anything changes...

python/ola/PidStore.py
brucelowekamp: So I honestly haven't learned enough about the protocol to have opinions about those aspects.  In truth this is a bit weaker than it ought to be, as a == b => hash(a) == hash(b) and this only tests name and value (wimping out b/c traversing requests etc would be a lot of code) whereas eq uses GroupCmp to compare requests and responses.  I thought that was unlikely to cause a problem, though.  I can change it to something else, but IMHO hash and eq/lt should match behavior here.

My understanding is the loader actually enforces not loading duplicate PIDs unless override is set, so in practice it shouldn't be a problem.

brucelowekamp:
Unfortunately docker has broken distcheck again, so wasn't able to fully test the TestUtils build, so naturally it broke.  Will take a look, though I think the push addresses everything else.

brucelowekamp:
Eventually got Docker to work by pointing builddir into the container instead of the share and running make single-threaded.  Took awhile....

@peternewman please take another look

tools/rdm/TestStateTest.py
```diff
@@ -0,0 +1,51 @@
...
+      self.assertNotEqual(pair[0], pair[1])
+
+  def testCmp(self):
+    self.assertEqual(TestState.PASSED, TestState.PASSED)
+
+    states = sorted([TestState.PASSED, TestState.FAILED,
+                     TestState.BROKEN, TestState.NOT_RUN])
+    self.assertEqual(states,
+                              [TestState.BROKEN, TestState.FAILED,
+                               TestState.NOT_RUN, TestState.PASSED])
```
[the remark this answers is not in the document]
peternewman: Done

tools/rdm/ResponderTestTest.py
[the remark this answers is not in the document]
peternewman: Done

tools/rdm/TestState.py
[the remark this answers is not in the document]
peternewman: Done

configure.ac
```diff
@@ -966,7 +966,16 @@ AC_CONFIG_LINKS([python/ola/PidStore.py:python/ola/PidStore.py
...
+                 python/ola/__init__.py:python/ola/__init__.py
+                 python/ola/ClientWrapper.py:python/ola/ClientWrapper.py
+                 python/ola/OlaClient.py:python/ola/OlaClient.py
+                 python/ola/rpc/SimpleRpcController.py:python/ola/rpc/SimpleRpcController.py
+                 python/ola/rpc/StreamRpcChannel.py:python/ola/rpc/StreamRpcChannel.py
+                 python/ola/rpc/__init__.py:python/ola/rpc/__init__.py
+                 tools/rdm/ExpectedResults.py:tools/rdm/ExpectedResults.py
+                 tools/rdm/TestCategory.py:tools/rdm/TestCategory.py
+                 tools/rdm/TestState.py:tools/rdm/TestState.py
+                 tools/rdm/TimingStats.py:tools/rdm/TimingStats.py])
```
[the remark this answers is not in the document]
peternewman: Done

python/ola/ClientWrapperTest.py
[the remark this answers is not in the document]
peternewman: Thanks.

python/ola/ClientWrapper.py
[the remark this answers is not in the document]
peternewman: Thanks

python/ola/OlaClient.py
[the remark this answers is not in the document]
peternewman: Ah fine, very sensible. We don't seem to do it in C++ land, but I think the registration works in a slightly different way there.

python/ola/PidStore.py
peternewman: Yeah 0.10 doesn't have the concept of manufacturer names.

common/rdm/PidStoreTest.cpp
peternewman: Ah great, thanks for confirming.

python/ola/PidStoreTest.py
peternewman: @brucelowekamp this still needs fixing case wise please.

python/ola/PidStore.py
peternewman: Very good point!

peternewman: I just need to check what happens in C++ if we merge an override with set only onto a normal PID with get only, do we get both or set only, as the Python should match this.

python/ola/PidStore.py
peternewman: Nothing should change, but if you grabbed the manufacturer_names.proto file I think it would fail to load as it would see duplicates.

python/ola/MACAddressTest.py
peternewman: Moved to python/TestUtils.py

python/ola/OlaClientTest.py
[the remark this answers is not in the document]
peternewman: Done

python/ola/OlaClientTest.py
```diff
@@ -0,0 +1,128 @@
...
+    s = sorted([c, a, b])
+    self.assertEqual([a, b, c], s)
+
+    self.assertEqual(a.__lt__("hello"), NotImplemented)
+    self.assertNotEqual(a, "hello")
+
+  def testDevice(self):
+    # only eq on alias (2)
+    a = Device(1, 2, "a", 4, [1, 2], [3, 4])
+    aeq = Device(0, 2, "a", 4, [1, 2, 3], [3, 4])
```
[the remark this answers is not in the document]
peternewman: Done

python/ola/OlaClientTest.py
```diff
@@ -0,0 +1,128 @@
...
+    s = sorted([b, a, c])
+    self.assertEqual([a, b, c], s)
+
+    self.assertEqual(a.__lt__("hello"), NotImplemented)
+    self.assertNotEqual(a, "hello")
+
+  def testPort(self):
+    # hash and eq only on id (1)
+    a = Port(1, 1, False, "a", False)
+    aeq = Port(1, 2, False, "a", True)
```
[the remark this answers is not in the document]
peternewman: Done.

python/ola/OlaClientTest.py
```diff
@@ -0,0 +1,128 @@
...
+    s = sorted([c, a, b])
+    self.assertEqual([a, b, c], s)
+
+    self.assertEqual(a.__lt__("hello"), NotImplemented)
+    self.assertNotEqual(a, "hello")
+
+  def testUniverse(self):
+    # universe doesn't have hash and implements eq and < only on id
+    a = Universe(1, 2, Universe.LTP, [1, 2], [3, 4])
+    aeq = Universe(1, 2, Universe.HTP, [1, 2], [3, 4])
```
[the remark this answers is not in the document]
peternewman: Done.

python/ola/OlaClientTest.py
```diff
@@ -0,0 +1,128 @@
...
+    s = sorted([c, b, a])
+    self.assertEqual([a, b, c], s)
+
+    self.assertEqual(a.__lt__("hello"), NotImplemented)
+    self.assertNotEqual(a, "hello")
+
+  def testRDMNack(self):
+    # hash and eq only on value (1)
+    a = RDMNack(1, "a")
+    aeq = RDMNack(1, "also a")
```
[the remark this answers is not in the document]
peternewman: Now all consistent so leaving as is.

tools/rdm/ResponderTest.py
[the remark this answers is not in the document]
peternewman: All sounds very sensible and wise, resolving.

tools/rdm/ResponderTest.py
```diff
@@ -76,8 +76,35 @@ def __str__(self):
   def __repr__(self):
     return self.__class__.__name__
 
-  def __cmp__(self, other):
-    return cmp(self.__class__.__name__, other.__class__.__name__)
+  # TestFixture and its subclasses are equal/ordered based on
+  # class name.  Two TestFixtures are equal if they are the same class,
+  # and ordering is based comparing __class__.__name__
```
peternewman, line 81: Nit "based on comparing"

tools/rdm/ResponderTest.py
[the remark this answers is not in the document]
peternewman: Did I miss the comment, or do you mean the bit above `__eq__`? If so it looks fine when it's not flooded with review comments! :man_facepalming:

python/ola/UIDTest.py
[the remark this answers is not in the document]
peternewman: Ah yes, I wonder if it's worth merging all the comparison bits into one test function, but I'm happy to leave it as is.

tools/rdm/ResponderTestTest.py
```diff
@@ -0,0 +1,60 @@
...
+
+    a = ATestFixture({}, 2, 123, None)
+    a2 = ATestFixture({}, 2, 123, None)
```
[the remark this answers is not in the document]
peternewman: You seem to have dropped the subclass version of these tests @brucelowekamp ?

python/ola/PidStore.py
peternewman: So broadly the value is the parameter ID sent on the wire, the name is just a friendly label we display for it within the software, so if they're both 0x0042, but one is called foo and one bar, the actual data on the wire and response from the responder will be the same, just how we display part of it will change, hence wanting a collision.

python/ola/Makefile.mk
```diff
@@ -58,16 +58,42 @@ python/ola/Version.py: python/ola/Makefile.mk configure.ac config/ola_version.m4
...
+	echo "export PYTHONPATH=${top_builddir}/python:${top_srcdir}/python; TESTDATADIR=$(srcdir)/common/rdm/testdata $(PYTHON) ${srcdir}/python/ola/PidStoreTest.py; exit \$$?" > $(top_builddir)/python/ola/PidStoreTest.sh
+	chmod +x $(top_builddir)/python/ola/PidStoreTest.sh
+
 dist_check_SCRIPTS += \
     python/ola/DUBDecoderTest.py \
+    python/ola/ClientWrapperTest.py \
     python/ola/MACAddressTest.py \
+    python/ola/OlaClientTest.py \
+    python/ola/PidStoreTest.py \
+    python/TestUtils.py \
```
peternewman, line 82: I think it would be good if we can leave this in python/ola and fix the other bits to avoid some other potential challenges.

peternewman: Won't adding this file to AC_CONFIG_LINKS avoid these issues?

brucelowekamp: You're right, forgot about that hack.  Will revert that one.

tools/rdm/ResponderTest.py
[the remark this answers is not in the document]
brucelowekamp: Yeah, I have the one comment above eq that applies to both eq and lt (which are the only two non-boilerplate comparisons used)

.travis.yml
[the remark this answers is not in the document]
brucelowekamp: For posterity, the decorators are now commented out

python/ola/PidStore.py
```diff
@@ -161,7 +161,11 @@ def __cmp__(self, other):
     return cmp(self._value, other._value)
 
   def __eq__(self, other):
-    return self._value == other._value
+    return self.value == other.value \
+      and self.name == other.name \
+      and self._requests == other._requests \
+      and self._responses == other._responses \
+      and self._validators == other._validators
```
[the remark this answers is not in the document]
brucelowekamp: @peternewman would like to make sure you're OK with Pid being == if value, name, requests, and responses are equal as that's a change in functionality here.

tools/rdm/ResponderTestTest.py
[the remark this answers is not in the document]
brucelowekamp: comparing a and a2 doesn't really exercise any different code than comparing base and base2.  All of the ordering comparisons between subclasses etc are still there.

python/ola/PidStore.py
brucelowekamp: There's that other conversation I opened on the matching eq/lt where Nils added the other fields beyond value.  It sounds like you would rather eq, lt, and hash all consider only value?

brucelowekamp: I'm going to make that change in the next changeset.

yoe:
fwiw, I've just merged this, with minimal changes, into my Debian branch, and pushed the result to https://salsa.debian.org/wouter/ola/-/tree/py3

I'll pass any fixes to make things work on Debian here.

peternewman:
> fwiw, I've just merged this, with minimal changes, into my Debian branch, and pushed the result to https://salsa.debian.org/wouter/ola/-/tree/py3
> 
> I'll pass any fixes to make things work on Debian here.

Great thanks @yoe . Does that mean OLA should stay in Debian for the next release? I guess subject to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925793 AKA #1553 AKA #1583 ?

Any idea what we need to do to nudge the Ubuntu people to go the same way?

tools/rdm/ResponderTest.py
peternewman: Done.

python/ola/Makefile.mk
peternewman: Done.

tools/rdm/ResponderTestTest.py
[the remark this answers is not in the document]
peternewman: Okay fair enough.

tools/rdm/ResponderTest.py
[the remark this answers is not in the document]
peternewman: Okay great

python/ola/PidStoreTest.py
peternewman: Done.

peternewman:
I started removing some double spaces while fixing some other minor typos @brucelowekamp , then I Googled and found out why, https://www.python.org/dev/peps/pep-0008/#comments (learning something new every day).

python/ola/PidStore.py
```diff
@@ -212,7 +204,7 @@ def __str__(self):
     return '%s (0x%04hx)' % (self.name, self.value)
 
   def __hash__(self):
-    return hash((self._name, self._value))
+    return hash(self._name)
```
peternewman, line 207: This should be value not name surely?

python/ola/PidStore.py
peternewman: Done.

yoe:
> > fwiw, I've just merged this, with minimal changes, into my Debian branch, and pushed the result to https://salsa.debian.org/wouter/ola/-/tree/py3
> > I'll pass any fixes to make things work on Debian here.
> 
> Great thanks @yoe . Does that mean OLA should stay in Debian for the next release? I guess subject to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925793 AKA #1553 AKA #1583 ?

As soon as the Python3 build works in something that I can upload to Debian, and the code compiles with a compiler that's available in Debian, I'll reupload it. If that happens a few weeks before the relevant freeze for Bullseye (starting next January according to https://release.debian.org/) then it will still be in that release, yes.

I haven't had much time to look at finalizing the gcc9 issues, mainly because of https://jonathancarter.org/2020/03/02/free-software-activities-for-2020-02/, but hopefully I'll be able to pick that up again soon.

> Any idea what we need to do to nudge the Ubuntu people to go the same way?

Unfortunately, it is too late for Focal (20.04) by now, since they're already in deep freeze in preparation for their release late this month. TTBOMK, it is not possible to add packages to Ubuntu after their freeze, let alone after the release. I could perhaps look at creating a PPA once everything is back in order in Debian if there is sufficient demand?

python/ola/PidStore.py
peternewman: Name should definitely be dropped, request/response is TBC a little bit of testing of the C++ and Protobuf behaviour.

brucelowekamp:
> I started removing some double spaces while fixing some other minor typos @brucelowekamp , then I Googled and found out why, https://www.python.org/dev/peps/pep-0008/#comments (learning something new every day).

Every time I've pushed without running flake8, I've regretted it...

python/ola/PidStore.py
brucelowekamp: And embarrassingly no tests around that.  Checked and found one other case in uid where hash == wasn't checked (added an additional test around hash != b/c it made it look better though it was actually run via the helper function later)

brucelowekamp:
> > > fwiw, I've just merged this, with minimal changes, into my Debian branch, and pushed the result to https://salsa.debian.org/wouter/ola/-/tree/py3
> > > I'll pass any fixes to make things work on Debian here.
> > 
> > 
> > Great thanks @yoe . Does that mean OLA should stay in Debian for the next release? I guess subject to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925793 AKA #1553 AKA #1583 ?
> 
> As soon as the Python3 build works in something that I can upload to Debian, and the code compiles with a compiler that's available in Debian, I'll reupload it. If that happens a few weeks before the relevant freeze for Bullseye (starting next January according to https://release.debian.org/) then it will still be in that release, yes.
> 
> I haven't had much time to look at finalizing the gcc9 issues, mainly because of https://jonathancarter.org/2020/03/02/free-software-activities-for-2020-02/, but hopefully I'll be able to pick that up again soon.
> 
> > Any idea what we need to do to nudge the Ubuntu people to go the same way?
> 
> Unfortunately, it is too late for Focal (20.04) by now, since they're already in deep freeze in preparation for their release late this month. TTBOMK, it is not possible to add packages to Ubuntu after their freeze, let alone after the release. I could perhaps look at creating a PPA once everything is back in order in Debian if there is sufficient demand?

The other challenge is that this is only (I think the most important) part of the python3 work.  I have another set of changes that got most of the tools/rdm working (or at least passing sanity checks), and there was another patch submitted a couple weeks ago for part of the API.  Given how long it's taken getting these patches through, I thought it best to wait to put in a PR for more.  The new protobuf code should help as well, though.  But ultimately there's more work, and it will require testing from some folks who use that part of the code.  (I only use the sending API)

Bruce

yoe:
Another option (certainly not my favourite) if we are not able to get this done in time for the Ubuntu 20.10 release (which would require that this is all resolved by July-August if I'm not mistaken) is to upload a version of ola with all python support disabled at compile time.

That would definitely fit the bill as far as Debian is concerned, but is obviously not a very good long-term strategy, and disabling the python-based tools would have some downside. Still, it would get *something* in for Ubuntu 20.10, which is better than what's going to happen to 20.04.

python/ola/PidStore.py
peternewman: And equally embarrassing, I didn't even consider the fact it wasn't caught by any tests, and we should add one! :man_facepalming:

python/ola/PidStoreTest.py
```diff
@@ -172,6 +172,11 @@ def testCmp(self):
     self.assertNotEqual(p1, p2)
     self.assertNotEqual(p1b, p2)
 
+    self.assertEqual(hash(p1), hash(p1a))
+    self.assertEqual(hash(p1), hash(p1b))
```
peternewman, line 176: Can we have an equality test with different PID names please?

python/ola/PidStoreTest.py
```diff
@@ -172,6 +172,11 @@ def testCmp(self):
...
+    self.assertNotEqual(hash(p1), hash(p2))
+    self.assertNotEqual(hash(p1b), hash(p2))
```
peternewman, line 178: Likewise an inequality test.

python/ola/UIDTest.py
```diff
@@ -109,6 +110,7 @@ def testCmp(self):
     self.assertFalse(u3 == u2)
 
     self.assertNotEqual(u3, u4)
+    self.assertNotEqual(hash(u3), hash(u4))
```
peternewman, line 113: Can we also test hash comparison of u2 and u3, to ensure the manufacturer ID is in the hash correctly.

peternewman:
> > > fwiw, I've just merged this, with minimal changes, into my Debian branch, and pushed the result to https://salsa.debian.org/wouter/ola/-/tree/py3
> > > I'll pass any fixes to make things work on Debian here.
> > 
> > 
> > Great thanks @yoe . Does that mean OLA should stay in Debian for the next release? I guess subject to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925793 AKA #1553 AKA #1583 ?
> 
> As soon as the Python3 build works in something that I can upload to Debian, and the code compiles with a compiler that's available in Debian, I'll reupload it. If that happens a few weeks before the relevant freeze for Bullseye (starting next January according to https://release.debian.org/) then it will still be in that release, yes.

Ah great thanks @yoe , so we won't have dropped out of Debian, that's good. Your fix for GCC9 #1583 means it compiles with GCC at least (if not Clang currently).

> I haven't had much time to look at finalizing the gcc9 issues, mainly because of https://jonathancarter.org/2020/03/02/free-software-activities-for-2020-02/, but hopefully I'll be able to pick that up again soon.

Amazing, congratulations to both of you! :confetti_ball: :bride_with_veil: :man_in_tuxedo: 

> > Any idea what we need to do to nudge the Ubuntu people to go the same way?
> 
> Unfortunately, it is too late for Focal (20.04) by now, since they're already in deep freeze in preparation for their release late this month. TTBOMK, it is not possible to add packages to Ubuntu after their freeze, let alone after the release. I could perhaps look at creating a PPA once everything is back in order in Debian if there is sufficient demand?

Damn, I hoped they might have an exception for packages which had previously been in Ubuntu and it's a particular shame it's an LTS release. @VoltVisionFrenchy had previously made an OLA PPA, perhaps if we ask really nicely he'll be able to update that (or possibly automate it so it just pulls from Git and builds in future.

> The other challenge is that this is only (I think the most important) part of the python3 work. I have another set of changes that got most of the tools/rdm working (or at least passing sanity checks), and there was another patch submitted a couple weeks ago for part of the API.

Yeah hopefully the other stuff is simpler when we've got the core changes in and should need less checking for example as it won't interact with other bits so much or be as complicated hopefully.

> Given how long it's taken getting these patches through, I thought it best to wait to put in a PR for more. The new protobuf code should help as well, though. But ultimately there's more work, and it will require testing from some folks who use that part of the code. (I only use the sending API)

Yeah sorry I've been rather slow with some of these PR reviews @brucelowekamp . Various bits of other life going on. Hopefully given the current situation things will quieten down a bit! :mask: 

I'm sure we can cover off the testing too.

> Another option (certainly not my favourite) if we are not able to get this done in time for the Ubuntu 20.10 release (which would require that this is all resolved by July-August if I'm not mistaken) is to upload a version of ola with all python support disabled at compile time.

Technically we could just release the ola deb couldn't we (without ola-python or ola-rdm-tests)?

> That would definitely fit the bill as far as Debian is concerned, but is obviously not a very good long-term strategy, and disabling the python-based tools would have some downside. Still, it would get _something_ in for Ubuntu 20.10, which is better than what's going to happen to 20.04.

Yeah definitely. I guess we missed a trick there and should have done just ola (without Python) for 20.04.

python/ola/UIDTest.py
brucelowekamp: that's actually handled by allHashNotEqual below

python/ola/PidStoreTest.py
peternewman: Done.

python/ola/PidStoreTest.py
peternewman: Done.

python/ola/UIDTest.py
peternewman: Ah yes, thanks.

yoe:
> > Another option (certainly not my favourite) if we are not able to get this done in time for the Ubuntu 20.10 release (which would require that this is all resolved by July-August if I'm not mistaken) is to upload a version of ola with all python support disabled at compile time.
> 
> Technically we could just release the ola deb couldn't we (without ola-python or ola-rdm-tests)?

Yes, we could (and we'd have to, otherwise those packages would be empty, and that makes no sense)

> > That would definitely fit the bill as far as Debian is concerned, but is obviously not a very good long-term strategy, and disabling the python-based tools would have some downside. Still, it would get _something_ in for Ubuntu 20.10, which is better than what's going to happen to 20.04.
> 
> Yeah definitely. I guess we missed a trick there and should have done just ola (without Python) for 20.04.

Yup, but it's too late now, unfortunately.

python/ola/PidStore.py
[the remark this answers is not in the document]
peternewman: Having finally tested the C++, overrides.proto just overrides the entire PID, so if you have a set only and overrides.proto has a get only of the same PID, the result is get only. Hence we should probably just compare on value I think and disregard the other fields, as in theory only one PID should ever exist, and two would be a duplicate if the value matches (i.e. same ID).

python/ola/PidStore.py
peternewman: I think this is equivalent to this in C++ land? https://github.com/OpenLightingProject/ola/blob/master/common/rdm/PidStoreHelper.cpp#L227-L235

A group within a group would certainly be at the edge of expected behaviour based on this:
https://wiki.openlighting.org/index.php/RDM_PID_Definitions#Parameter_Definitions_and_OLA

The inner group would need to have a defined length to allow the outer group length to be calculated. I guess if you had say a set of curves, with each curve having five points, it might make sense.

From a technical perspective, why have you added a : before the newline? Can I see some example output from this?

Actually looking further, I think this will break a command, but due to some admittedly rather lacking documentation. If you run the following command:
`./python/examples/ola_rdm_get.py -u 0 --uid 7a70:ffffff00 sensor_definition`

What output do you get? I get the following, I don't think yours will match anymore:
```
Usage: get sensor_definition <sensor_number>
  sensor_number: <[0, 254]> 

*** Missing argument for sensor_number
```

peternewman: Perhaps we should add a test for this too?

python/ola/PidStore.py
peternewman: So I did check and in C++ it override the entire PID if it exists, so any pre-existing functionality is lost and hence this is the correct behaviour.

python/ola/PidStore.py
```diff
@@ -1065,11 +1130,22 @@ def NameToValue(self, pid_name, esta_id=None):
     Returns:
       The value for this PID, or None if it wasn't found.
     """
-    pid = self.GetName(pid_name)
+    pid = self.GetName(pid_name, esta_id)
```
peternewman, line 1133: We should probably add a test for this bug too?

python/ola/PidStoreTest.py
```diff
@@ -0,0 +1,192 @@
...
+      "<protocol_major> <protocol_minor> <device_model> "
+      "<product_category> <software_version> <dmx_footprint> "
+      "<current_personality> <personality_count> <dmx_start_address> "
+      "<sub_device_count> <sensor_count>:\n  "
+      "protocol_major: <[0, 255]> \n  protocol_minor: <[0, 255]> \n  "
+      "device_model: <[0, 65535]> \n  product_category: <[0, 65535]> \n  "
+      "software_version: <[0, 4294967295]> \n  dmx_footprint: <[0, 65535]> \n  "
+      "current_personality: <[0, 255]> \n  personality_count: <[0, 255]> \n  "
+      "dmx_start_address: <[0, 65535]> \n  sub_device_count: <[0, 65535]> \n  "
+      "sensor_count: <[0, 255]> ")
```
peternewman, line 65: As mentioned, I think this should still be a tuple for the one place it's used.

peternewman, line 65: Nit, but it would be nice if there wasn't a space between the `>` and `\n`.

python/ola/PidStoreTest.py
```diff
@@ -0,0 +1,192 @@
...
+    self.assertEqual(pid.value, 32768)
+    self.assertEqual(pid, store.GetPid(32768, OPEN_LIGHTING_ESTA_CODE))
+    self.assertEqual(pid.name, "FOO_BAR")
+    self.assertIsNotNone(pid.GetRequest(PidStore.RDM_GET))
+    self.assertIsNotNone(pid.GetResponse(PidStore.RDM_GET))
+    self.assertIsNone(pid.GetRequest(PidStore.RDM_SET))
+    self.assertIsNone(pid.GetResponse(PidStore.RDM_SET))
+
+    self.assertEqual(pid.GetResponse(PidStore.RDM_GET).GetDescription(),
+                     "<baz>:\n  baz: <[0, 4294967295]> ")
```
peternewman, line 106: A repeat of my earlier comments for this test case too.

python/ola/Makefile.mk
```diff
@@ -58,16 +58,42 @@ python/ola/Version.py: python/ola/Makefile.mk configure.ac config/ola_version.m4
 # TESTS
 ##################################################
 
+python/ola/ClientWrapperTest.sh: python/ola/Makefile.mk
+	mkdir -p $(top_builddir)/python/ola
+	echo "export PYTHONPATH=${top_builddir}/python:${top_srcdir}/python; $(PYTHON) ${srcdir}/python/ola/ClientWrapperTest.py; exit \$$?" > $(top_builddir)/python/ola/ClientWrapperTest.sh
```
peternewman, line 63: I don't think you need the export if you drop the `;`
```suggestion
	echo "PYTHONPATH=${top_builddir}/python:${top_srcdir}/python $(PYTHON) ${srcdir}/python/ola/ClientWrapperTest.py; exit \$$?" > $(top_builddir)/python/ola/ClientWrapperTest.sh
```

See also PidStoreTest.sh and https://github.com/OpenLightingProject/ola/blob/master/data/rdm/Makefile.mk#L27

This also avoids corrupting the shell this is run in by overwriting things with export.

Is the top_srcdir bit still needed now you've added all those symlinks via configure?

python/ola/Makefile.mk
```diff
@@ -58,16 +58,42 @@ python/ola/Version.py: python/ola/Makefile.mk configure.ac config/ola_version.m4
...
+	chmod +x $(top_builddir)/python/ola/ClientWrapperTest.sh
+
+python/ola/OlaClientTest.sh: python/ola/Makefile.mk
+	mkdir -p $(top_builddir)/python/ola
+	echo "export PYTHONPATH=${top_builddir}/python:${top_srcdir}/python; $(PYTHON) ${srcdir}/python/ola/OlaClientTest.py; exit \$$?" > $(top_builddir)/python/ola/OlaClientTest.sh
```
peternewman, line 68: As above.

tools/rdm/Makefile.mk
```diff
@@ -55,7 +55,24 @@ launcher_files = \
 
 EXTRA_DIST += $(launcher_files)
 
-CLEANFILES += tools/rdm/*.pyc
+tools/rdm/ResponderTestTest.sh: tools/rdm/Makefile.mk
+	mkdir -p $(top_builddir)/python/ola
+	echo "export PYTHONPATH=${top_builddir}/python:${top_srcdir}/python; $(PYTHON) ${srcdir}/tools/rdm/ResponderTestTest.py; exit \$$?" > $(top_builddir)/tools/rdm/ResponderTestTest.sh
```
peternewman, line 60: Same query about export and top_srcdir.

tools/rdm/Makefile.mk
```diff
@@ -55,7 +55,24 @@ launcher_files = \
...
+	chmod +x $(top_builddir)/tools/rdm/ResponderTestTest.sh
+
+dist_check_SCRIPTS += \
+   tools/rdm/TestStateTest.py \
+   tools/rdm/ResponderTestTest.py
```
peternewman, line 65: Can we sort these two please.

tools/rdm/Makefile.mk
```diff
@@ -55,7 +55,24 @@ launcher_files = \
...
+
+if BUILD_PYTHON_LIBS
+test_scripts += \
+   tools/rdm/TestStateTest.py \
+   tools/rdm/ResponderTestTest.sh
```
peternewman, line 70: Again can we sort please.

python/ola/PidStore.py
```diff
@@ -65,7 +69,7 @@ class UnpackException(Error):
 
 
 class MissingPLASAPIDs(Error):
```
peternewman, line 71: I wonder if we should really have renamed this (or kept the old one for backwards compatibility and subclass or re-raise or something)?

python/ola/PidStore.py
brucelowekamp: yeah, fair.   added

tools/rdm/TestStateTest.py
```diff
@@ -0,0 +1,51 @@
...
+class TestStateCmpTest(unittest.TestCase):
+  def allNotEqual(self, t):
+    for pair in itertools.combinations(t, 2):
+      self.assertNotEqual(pair[0], pair[1])
+
+  def allHashNotEqual(self, t):
+    h = map(hash, t)
+    for pair in itertools.combinations(h, 2):
+      self.assertNotEqual(pair[0], pair[1])
+
```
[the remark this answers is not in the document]
brucelowekamp: yup, missed that

python/ola/PidStore.py
brucelowekamp: The name hasn't changed since it was introduced, AFAICT.  The comment change was one of the changes brought in from master.

python/ola/PidStore.py
brucelowekamp: Ah, I didn't know the use case for this.  In some ideal world I think I would expect a usage message to ask separately for the summary of parameters and then details, and also for the python to display e.g. uint32 like the C++ does instead of the numbers, but I'd rather not dive into that sort of refactor at this point :)

python/ola/PidStoreTest.py
brucelowekamp: Interestingly that was a bug in how the increment format was added.

python/ola/Makefile.mk
brucelowekamp: Amusingly the whole export and top_srcdir thing was copied from an existing test, but I wasn't even consistent in using it in all of the cases.

I don't think there's actually any harm in it, as it's being used in shell scripts that I believe are exec'ed from the build process, so it wouldn't inherit elsewhere.  But it's still poor form (particularly on a line with a command) so better to fix that and the top_srcdir.

(and annoyingly I pushed before I realized there was another use of the pattern in tools/rdm...)

python/ola/PidStore.py
peternewman: Done.

tools/rdm/TestStateTest.py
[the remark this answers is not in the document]
peternewman: Done.

python/ola/PidStoreTest.py
peternewman: Done.

python/ola/PidStoreTest.py
peternewman: Done.

tools/rdm/Makefile.mk
peternewman: Done.

tools/rdm/Makefile.mk
peternewman: Done.

python/ola/PidStore.py
```diff
@@ -406,7 +406,7 @@ def GetDescription(self, indent=0):
     if self._multiplier:
       increment = ', increment %s' % (10 ** self._multiplier)
 
-    return ('%s%s: <%s> %s' % (indent, self.name, self._GetAllowedRanges(),
+    return ('%s%s: <%s>%s' % (indent, self.name, self._GetAllowedRanges(),
                                increment))
```
peternewman, line 410: De-indent this by one.

python/ola/PidStoreTest.py
peternewman: Done thanks.

python/ola/PidStore.py
peternewman: Ah okay, one for after this mammoth PR is in then!

python/ola/Makefile.mk
peternewman: Heh, it's okay, we seem to often be rather inconsistent.

Thanks for fixing.

python/ola/PidStore.py
```diff
@@ -892,7 +892,7 @@ def GetDescription(self, indent=0):
       names.append('<%s>' % atom.name)
       output.append(atom.GetDescription(indent=2))
 
-    return ' '.join(names) + ':\n' + '\n'.join(output)
+    return ' '.join(names), '\n'.join(output)
```
peternewman: Thanks, would you mind correcting the comment docs in the base class to explain this.

python/ola/PidStore.py
```diff
@@ -157,14 +161,50 @@ def ValidateAddressing(self, args, command_class):
         return False
     return True
 
-  def __cmp__(self, other):
-    return cmp(self._value, other._value)
+  def _GroupCmp(self, a, b):
+    def setToDescription(x):
+      def descOrNone(y):
+        return y.GetDescription() if y is not None else "None"
```
peternewman, line 167: I think this will break with the most recent changes. I don't think it's currently used anywhere but could be useful in the future, so up to you if you want to drop it or fix it?

peternewman:
So, I think there is a deletion, an updated of a comment and a minor fix or deletion listed above. Then we're good to go! :tada: 

Can you confirm, have you re-run a few basic real world tests since all the changes?

Fixes for later (in a new PR, by one of us):
- [ ] Indenting of groups (and groups of groups), this seems to behave a bit oddly (the indent by 2 everywhere), and I think is broken for groups of groups, as they don't indent themselves.
- [ ] MissingPLASAPIDs exception should become MissingESTAPIDs, possibly one for just 0.11.0 as a technically breaking change, although I doubt it's explicitly caught anywhere - https://github.com/OpenLightingProject/ola/pull/1605#discussion_r414944047

python/ola/PidStore.py
peternewman: I only found it after a bit of hunting through the code.

Do you mean have two separate methods which essentially both walk the tree? So in simple terms add two methods which both call GetDescription and return the two arguments individually?

Yeah definitely another one for a separate PR! The Python is returning the valid range (so in the case of sensor_number it's not uint8, as you can't get 0xFF for that particular PID; as it means ALL_SENSORS, and getting all doesn't make sense). But we could look at those other changes and making them more similar.

python/ola/PidStore.py
brucelowekamp: I fixed it, though as you said it doesn't seem to be referenced anywhere.  Put in a test case that I have no idea if it is what the intended use is, but at least confirmed it broke before the fix and works now.

brucelowekamp:
> So, I think there is a deletion, an updated of a comment and a minor fix or deletion listed above. Then we're good to go! 🎉
> 
> Can you confirm, have you re-run a few basic real world tests since all the changes?
> 
> Fixes for later (in a future PR, by one of us):
> 
> * [ ]  Indenting of groups (and groups of groups), this seems to behave a bit oddly (the indent by 2 everywhere), and I think is broken for groups of groups, as they don't indent themselves.
> * [ ]  MissingPLASAPIDs exception should become MissingESTAPIDs, possibly one for just 0.11.0 as a technically breaking change, although I doubt it's explicitly caught anywhere - [#1605 (comment)](https://github.com/OpenLightingProject/ola/pull/1605#discussion_r414944047)
> * [ ]  Refactor GetDescription - [#1605 (comment)](https://github.com/OpenLightingProject/ola/pull/1605#discussion_r415332991)

I put (part of) my lightshow back up when quarantine began and checked today that my code still works with the python lib (in both 2.7 and 3.7).   

Groups didn't look like it would be hard to fix the indenting on given an example pid and deciding what it should actually produce :)

python/ola/PidStore.py
peternewman: Done.

python/ola/PidStore.py
peternewman: Done.

python/ola/PidStore.py
peternewman: Thanks. It's only come in during this PR, I think in one of the previous iterations to fully compare two PIDs. But given it's written it seemed a shame to drop it.

peternewman:
> Groups didn't look like it would be hard to fix the indenting on given an example pid and deciding what it should actually produce :)

Yeah agreed, I think we'd just need to have a ponder exactly what we wanted to do, and to generate some test examples.

python/ola/PidStore.py
peternewman: As mentioned elsewhere this is just id which you've done.

python/ola/PidStore.py
peternewman: Done.

yoe:
I gave advice on python stuff? Oh dear. I do have a reputation to consider ;-P

PR #1608, opened 2020-01-21
Simplify the sed command used to generate plugin header files, closes #1592 (hopefully)

### LINKED ISSUE #1592: plugins/convert_README_to_header.sh doesn't work in bionic (and possibly others after xenial)
Although your earlier bionic run did run, and fail:
https://travis-ci.org/OpenLightingProject/ola/jobs/609809927#L974-L977

```
sh ./plugins/convert_README_to_header.sh ./plugins/artnet ./plugins/artnet/ArtNetPluginDescription.h
sh ./plugins/convert_README_to_header.sh ./plugins/dummy ./plugins/dummy/DummyPluginDescription.h
sed: -e expression #6, char 14: unterminated `s' command
sed: -e expression #6, char 14: unterminated `s' command
```

Is that the same as you get on your personal machine with 19.04?

_Originally posted by @peternewman in https://github.com/OpenLightingProject/ola/pull/1586_

See the rest of the discussion in https://github.com/OpenLightingProject/ola/pull/1586#pullrequestreview-314609405

We should start by making the script fail so make fails if the sed command fails, and then try and fix the command.

peternewman:
An alternative option to #1593 although without the required error checking too.


plugins/convert_README_to_header.sh
```diff
@@ -27,11 +27,11 @@ fi
 plugin=`basename "$path"`;
 outfilename=`basename $outfile`;
 
-# See http://stackoverflow.com/a/16576291
-# On Mac OS's sed, \n is not recognized as a newline character, but
-# \[actual newline] works
-desc=`sed -e ':a' -e 'N' -e '$!ba' -e 's/\"/\\\"/g' -e 's/\n/\\\\n"\\
-"/g' "$path/README.md"`;
+#s/\"/\\\"/g - replace " with \" throughout
+#1!s/^/\"/g - Apart from the first line, add a \" to the start of each line
```
FloEdelmann: Why do we need this? I thought the conversion here should look something like this:

```md
First line with "quotes"
Second line
```

```c
const char plugin_description[] = "First line with \"quotes\"\nSecond line";
```

peternewman: Readability/debugability/maintainability, for example if there are issues with the escaping within plugin_description its a lot easier to find them if it isn't one long continuous line. It also removes any potential for a line length limitation within a pre-processor/compiler that we're unaware of. It also means the file passes lint which I appreciate some people feel isn't necessarily important.

The current output, which my tweaked sed command matches, looks like this:
`const char plugin_description[] = "First line with \"quotes\"\n"
"Second line";`

Obviously the contents of the variable ends up like your example when interpreted by the compiler, and then when sent to the GUI it can be correctly rendered with the new lines to break up the text.

As far as I can see from here, everything used in my proposed sed command is part of the standard POSIX sed feature-set:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

FloEdelmann: Ah, I see. I was fooled by the comment in the next line, since it doesn't mention that a `"` is added after the `\n`. Other than that, I think it looks fine then.

peternewman: Oh yes, very true!

peternewman: Done, although I see there is also some inconsistency about escaping on the RHS of the substitution, I'll try and find out what the correct behaviour there should be.

peternewman:
> This looks like a better alternative to #1592 now.

Thanks @FloEdelmann, I think you mean #1593 though.

I think we still want the `set -e/set +e` bit and we ought to keep the escaping of backslashes from that too if it's needed.

Keeper-of-the-Keys:
Please note that #1593 also resolves the backslash character properly something which the original sed did not (see ASCII art in #1541) iirc.

As far as whether or not to use a long line or multiple lines... My understanding is we're trying to keep the lint beast at bay without a blacklist, but @peternewman is correct that it does make debugging easier.

peternewman:
> Please note that #1593 also resolves the backslash character properly something which the original sed did not (see ASCII art in #1541) iirc.

Agreed @Keeper-of-the-Keys . Do you want to pull this fix into #1593, or would you rather I take the rest of #1593 and pull that into this PR?

Keeper-of-the-Keys:
Whatever works :)
I'll try to test this over the weekend.

peternewman:
> I'll try to test this over the weekend.

Thanks, FYI aside from observation and compilation, I've been testing changes don't affect the output using this snippet:
```
#!/bin/sh

for file in */README.md; do
  echo "Generating $file";
  sh convert_README_to_header.sh `dirname $file` testmakereadme/orig/`dirname $file`.h
  sh convert_README_to_header.new.sh `dirname $file` testmakereadme/new/`dirname $file`.h
done
diff testmakereadme/orig/ testmakereadme/new/
```

So far it works fine on Ubuntu 16.04/sed 4.2.2 and Raspbian Buster/sed 4.7.

Keeper-of-the-Keys:
Finally had time to check this' it does not as I suspected take care of the backslash character.
However this is easily remedied add the following before and regex that adds any backslashes:
`-e 's/\\\/\\\\\\\/g'`

As far as the count goes it was arrived at through trial and error, the first can be either 3 or 4 bachslashes and the second 7 or 8 but not 6.

peternewman:
> @peternewman Finally had time to check this, it does not as I suspected take care of the backslash character.
> However this is easily remedied add the following before and regex that adds any backslashes:
> `-e 's/\\\/\\\\\\\/g'`
> 
> As far as the count goes it was arrived at through trial and error, the first can be either 3 or 4 bachslashes and the second 7 or 8 but not 6.

Great, thanks for testing @Keeper-of-the-Keys . I might get this into 0.10 as a bug fix, and then we can have yours, with the set -e and the backslash stuff in master for your extra slashes in your new docs.

peternewman:
> Great, thanks for testing @Keeper-of-the-Keys . I might get this into 0.10 as a bug fix

Of course I'm talking nonsense, this functionality isn't in 0.10 at all, buggy or otherwise! :man_facepalming: 

Can you update #1593 with the changes from here please @Keeper-of-the-Keys ?

PR #1609, opened 2020-01-22
Updated Nanoleaf plugin Readme with better directions

miniman13:
Updated Readme with better directions on how to use nanoleaf plugin, enabling udp mode and api to do so

### DIFF
--- a/plugins/nanoleaf/README.md
+++ b/plugins/nanoleaf/README.md
@@ -7,10 +7,10 @@ This plugin uses the streaming external control effects functionality of
 the Aurora OpenAPI.
 
 Currently only the [Nanoleaf
-Aurora](https://nanoleaf.me/en/consumer-led-lighting/products/
-smarter-series/nanoleaf-light-panels-smarter-kit/)
+Aurora](https://nanoleaf.me/en/consumer-led-lighting/products/smarter-series/nanoleaf-light-panels-smarter-kit/)
 is supported.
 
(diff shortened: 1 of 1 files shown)


plugins/nanoleaf/README.md
```diff
@@ -7,10 +7,10 @@ This plugin uses the streaming external control effects functionality of
 the Aurora OpenAPI.
 
 Currently only the [Nanoleaf
-Aurora](https://nanoleaf.me/en/consumer-led-lighting/products/
-smarter-series/nanoleaf-light-panels-smarter-kit/)
+Aurora](https://nanoleaf.me/en/consumer-led-lighting/products/smarter-series/nanoleaf-light-panels-smarter-kit/)
```
peternewman: We can also simplify the URL as they've dropped part of it:
https://nanoleaf.me/en/products/smarter-series/nanoleaf-light-panels-smarter-kit/

plugins/nanoleaf/README.md
```diff
@@ -24,5 +27,6 @@ one controller by adding multiple `controller =` lines
 The port to stream to on the controller, range is 1 - 65535.
 
 `<controller IP>-panels = [int]`
-The list of panel IDs to control, each panel is mapped to three DMX512 slots
-(for Red, Green and Blue).
+The list of panel IDs to control, each panel is mapped to three DMX512
+slots (for Red, Green and Blue). 
+Panel IDs are comma separated eg: `321,789,123`
```
peternewman, line 32: I've just noticed off the back of #1730 these example values are invalid for the v1 protocol. We'd want say (to show order is irrelevant):
```suggestion
Panel IDs are comma separated eg: `132,178,54`
```

PR #1612, opened 2020-01-30
First attempt at LLRP header and PDU

peternewman:
All there apart from flags should always be 0xF0 rather than 0x70


kripton:
Cherry-picked into #1939

PR #1571, opened 2019-07-21
Update HTTPServer.cpp to allow CORS

GertOdB:
This enables Cross-Origin Resource Sharing (CORS) as requested in here : https://github.com/OpenLightingProject/ola/issues/1547

PR #1625, opened 2020-03-27
Allow all origin in HTTPServer responses (CORS)

quadbyte:
Similar to https://github.com/OpenLightingProject/ola/pull/1571 but with the requested changes.

I'm not really used to PR, so please bear with me.

### DIFF
--- a/common/http/HTTPServer.cpp
+++ b/common/http/HTTPServer.cpp
@@ -375,6 +375,13 @@ void HTTPResponse::SetNoCache() {
 
 
 /**
+ * @brief Set the appropriate headers so this response is accessible from any origin
+ */
+void HTTPResponse::SetAllowAllOrigin() {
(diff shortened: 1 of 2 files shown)


common/http/HTTPServer.cpp
```diff
@@ -374,6 +374,13 @@ void HTTPResponse::SetNoCache() {
 }
 
 
+/**
+ * @brief Set the appropriate headers so this response is accessible from any origin
+ */
+void HTTPResponse::SetAllowAllOrigin() {
```
peternewman: I'm wondering if we should have a slightly clearer name, perhaps `SetAccessControlAllowAllOrigin` or `SetAccessControlAllowOriginAll`

quadbyte: Make sense, will change it to SetAccessControlAllowOriginAll.

quadbyte:
renamed helper method to SetAccessControlAllowOriginAll and commited/pushed.

PR #1627, opened 2020-04-02
usbdmx: Add support for having multiple inflight transfers

DanielG:
In developing a custom USB->DMX widget for OLA I've noticed that packet latency isn't as consistent as possible when only using a single libusb transfer at a time. Hence here I add support for more than one and also fix a nasty use after free hazard.

Commit messages copied below:

##### `usbdmx: Add support for having multiple inflight transfers`
Currently we only use a single active transfer for both sending and
reveiving usb packets.

This is suboptimal in terms of latency, USB actually guarantees certain
turnaround times for interrupt transfers for example but that only really
works if there is always an IRP ready to go in the USB host controller.

According to my measurments interrupt transfers with bInterval=1 can be
delayed by some tens of milliseconds instead of happening evey millisecond
on the dot if there are periods with no transfers ready to go.

In order to support multiple transfers being inflight I replace the
transfer state machinery in AsyncUsbTransceiverBase with a set of inflight
transfers and a queue of idle transfers.

The bookkeeping happens in a new TransferCompleteInternal method which
wraps the TransferComplete() method provided by subclasses. This split also
allowed moving some duplicated logic for LIBUSB_TRANSFER_NO_DEVICE to the
superclass.

The number of transfers to allow is passed to AsyncUsb{Sender,Receiver} as
a constructor argument since supporting more than one needs some
cooperation from the derived classes. Currently everything just defaults to
one transfer which shouldn't change the behaviour one bit.

----

##### `usbdmx: Remove redundant CancelTransfer() calls`:

The superclass destructor already calls CancelTransfer, no need to do it in
the derived classes also. To prevent this in the future this also make
CancelTransfer private.

----

##### `usbdmx: Fix use after free hazard in destructor`

(see diff for commentary)


DanielG:
Changes:
- Actually {fire off,allow} up to the maximum number of transfers in the {receiver,sender}. Since PerformTransfer() is handled in the AsyncUsb{Sender,Receiver} classes we have to do the transfer count handling there.

In theory this should mean widget specific classes shouldn't even be able to notice what the transfer queue depth actually is, so maybe we can just turn this on across all widgets.

DanielG:
Changes:
- Fix the use-after-free fix :)
AsyncUsb{Sender,Receiver} have to call CancelTransfer, not the base class, because they close the device handle and we must also not free the device before the transfers complete!

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -28,13 +28,16 @@ namespace usbdmx {
 
 AsyncUsbReceiver::AsyncUsbReceiver(ola::usb::LibUsbAdaptor *adaptor,
                                    libusb_device *usb_device,
-                                   PluginAdaptor *plugin_adaptor)
-    : AsyncUsbTransceiverBase(adaptor, usb_device),
+                                   PluginAdaptor *plugin_adaptor,
+                                   unsigned int n_transfers)
```
peternewman, line 32: I think num_transfers would probably be more common here based on equivalent code we've got elsewhere.

plugins/usbdmx/AsyncUsbReceiver.cpp
```diff
@@ -58,25 +61,20 @@ bool AsyncUsbReceiver::Start() {
     return false;
   }
   ola::thread::MutexLocker locker(&m_mutex);
-  return PerformTransfer();
+  bool success = true;
+  while (!m_idle.empty() && success) {
+    success = success && PerformTransfer();
+  }
+  return success;
```
peternewman, line 68: Wouldn't this work just the same? As soon as success becomes false, because a transfer fails, it won't enter into the next loop and would then return false:
```suggestion
  bool success = true;
  while (!m_idle.empty() && success) {
    success = PerformTransfer();
  }
  return success;
```

plugins/usbdmx/AsyncUsbReceiver.h
```diff
@@ -50,10 +50,12 @@ class AsyncUsbReceiver: public AsyncUsbTransceiverBase {
    * @param adaptor the LibUsbAdaptor to use.
    * @param usb_device the libusb_device to use for the widget.
    * @param plugin_adaptor the PluginAdaptor to use for the widget.
+   * @param n_transfers maximum number of inflight transfers.
```
peternewman, line 53: Again

peternewman: Or based on the Doxygen it could also be max_transfers

plugins/usbdmx/AsyncUsbSender.cpp
```diff
@@ -30,12 +30,15 @@ namespace usbdmx {
 using ola::usb::LibUsbAdaptor;
 
 AsyncUsbSender::AsyncUsbSender(LibUsbAdaptor *adaptor,
-                               libusb_device *usb_device)
-    : AsyncUsbTransceiverBase(adaptor, usb_device),
+                               libusb_device *usb_device,
+                               unsigned int n_transfers)
```
peternewman, line 34: Likewise

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -29,36 +31,68 @@ namespace usbdmx {
...
+  {
+    ola::thread::MutexLocker locker(&m_mutex);
+    inflight_set::iterator i = m_inflight.find(transfer);
+    if (i == m_inflight.end()) { // not found
+      OLA_WARN << "Mismatched libusb transfer: "
+               << transfer << " not in inflight set";
+      return;
+    }
+
+    m_inflight.erase(i);
```
peternewman, line 57: Would this be easier replacing with our ola::STLLookupAndRemove? http://docs.openlighting.org/ola/doc/latest/group__stl.html#ga7a896801adfcea56e630bd5007c135e1

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -29,36 +31,68 @@ namespace usbdmx {
...
+    m_idle.push(transfer);
+  }
+
+  if (transfer->status == LIBUSB_TRANSFER_NO_DEVICE)
+    m_device_disconnected = true;
```
peternewman, line 62: Can we have some { } to avoid dangling else type stuff.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -29,36 +31,68 @@ namespace usbdmx {
...
+    if (m_inflight.empty())
+      m_cond.Signal();
+  }
 }
-}  // namespace
 
 AsyncUsbTransceiverBase::AsyncUsbTransceiverBase(LibUsbAdaptor *adaptor,
-                                                 libusb_device *usb_device)
+                                                 libusb_device *usb_device,
+                                                 unsigned int n_transfers)
```
peternewman, line 75: Likewise

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -67,62 +101,69 @@ bool AsyncUsbTransceiverBase::Init() {
...
 
 void AsyncUsbTransceiverBase::CancelTransfer() {
-  if (!m_transfer) {
-    return;
+  m_suppress_continuation = true;
+
+  inflight_set::iterator i;
+  for (i = m_inflight.begin(); i != m_inflight.end(); ++i) {
+    int ret = m_adaptor->CancelTransfer(*i);
+    if (ret != 0 && ret != LIBUSB_ERROR_NOT_FOUND) {
```
peternewman, line 109: When are we getting LIBUSB_ERROR_NOT_FOUND?

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -67,62 +101,69 @@ bool AsyncUsbTransceiverBase::Init() {
...
+    /* Wait for cancelled transfers to complete. Not waiting for the
+     * callbacks to happen is undefined behaviour according to libusb
+     * docs. */
+    m_cond.Wait(&m_mutex);
   }
+}
+
+struct libusb_transfer *AsyncUsbTransceiverBase::CurrentTransfer() {
+  if (m_idle.empty())
+    throw std::out_of_range("AsyncUsbTransceiverBase::m_idle is empty");
```
peternewman, line 125: We've not got any exception throwing in the code so far, returning NULL and then checking for that elsewhere would match the behaviour of the rest of the code better, possibly combined with a local log message. Or switching to assert().

plugins/usbdmx/AsyncUsbTransceiverBase.h
```diff
@@ -42,9 +44,11 @@ class AsyncUsbTransceiverBase {
    * @brief Create a new AsyncUsbTransceiverBase.
    * @param adaptor the LibUsbAdaptor to use.
    * @param usb_device the libusb_device to use for the widget.
+   * @param n_transfers maximum number of inflight transfers.
```
peternewman, line 47: Likewise

peternewman (changes requested):
Some initial comments

DanielG:
Changes:
- Fix typo emtpy -> empty

DanielG:
Changes:
- Fix cpplint complaint about two spaces after code, before comment.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -29,36 +31,69 @@ namespace usbdmx {
...
+  if (transfer->status == LIBUSB_TRANSFER_NO_DEVICE) {
+    m_device_disconnected = true;
+  }
+
+  TransferComplete(transfer);
+
+  {
+    ola::thread::MutexLocker locker(&m_mutex);
+    if (m_inflight.empty())
+      m_cond.Signal();
```
peternewman, line 70: Can we have some { } again

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -67,62 +102,83 @@ bool AsyncUsbTransceiverBase::Init() {
...
+     * state anymore.
+     *
+     * Even if we were to check for this before cancelling it could still
+     * happen that the transfer completes between the check and the
+     * cancellation taking effect.
+     */
+    if (ret != 0 && ret != LIBUSB_ERROR_NOT_FOUND) {
+      OLA_WARN << "libusb_cancel_transfer returned "
+               << m_adaptor->ErrorCodeToString(ret);
+    }
```
peternewman, line 123: I wondered if we should have some debug here, but I guess if we're only calling cancel during shutdown/disconnect it's not very exciting.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
```diff
@@ -67,62 +102,83 @@ bool AsyncUsbTransceiverBase::Init() {
...
+    /* Wait for cancelled transfers to complete. Not waiting for the
+     * callbacks to happen is undefined behaviour according to libusb
+     * docs. */
+    m_cond.Wait(&m_mutex);
   }
+}
+
+struct libusb_transfer *AsyncUsbTransceiverBase::CurrentTransfer() {
+  /* This should never happen. In AsyncUsbSender::SendDMX we only ask the
+   * perform superclass to perform a transfer m_idle is not empty and in
```
peternewman, line 136: Typo?
```suggestion
   * perform superclass to perform a if transfer m_idle is not empty and in
```

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
peternewman: Apologies, wrong command, after looking more carefully ola::STLRemove will do what you want:
http://docs.openlighting.org/ola/doc/latest/group__stl.html#ga6bab5568a6d6ef14d4faccea9eddd41b

Or is there a risk of multiple equal transfers in the set (I assume they have a unique ID, but maybe not)?

peternewman (changes requested):
I still can't get use to C++'s pop not being like this:
https://perldoc.perl.org/functions/pop

A few more, mostly minor, comments.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
peternewman: > Or is there a risk of multiple equal transfers in the set (I assume they have a unique ID, but maybe not)?

Reading more carefully, they have to be unique, so no risk there.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: I don't get this obsession with `{ }`, but sure. GCC has a warning now for when a dodgy multi-statement macro is expanded in a bracket-less `if` you know ;)

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: The actual typo is 'ask the perform superclass' to perfom' -> 'ask the superclass to perform'.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: What sort of debug prints are you thinking of beyond the error string?

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
peternewman: :facepalm: Oops, yeah that makes more sense!

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
peternewman: Just logging the LIBUSB_ERROR_NOT_FOUND, but perhaps at OLA_DEBUG level, so we can see all the failures.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
peternewman: Just I've been tripped up by writing something like this a few times in the past:
```suggestion
    if (m_inflight.empty())
      m_cond.Signal();
      OLA_DEBUG << "Inflight empty";
```

Would GCC catch that too now?

DanielG: Yup, that's `-Wmisleading-indentation` which is part of `-Wall`:

```
void foo() {}
int main() {
        if (true)
                foo();
                foo();
}
```

`gcc -Wall test.cc`:
```
test.cc: In function ‘int main()’:
test.cc:3:9: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
         if (true)
         ^~
test.cc:5:17: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
                 foo();
```

There's also `-Wmultistatement-macros`, also part of `-Wall` which is the case I meant:

```
void foo() {}
#define dodgy() foo(); foo();
int main() {
        if (true)
                dodgy()
}
```

This will give:

```
test.cc: In function ‘int main()’:
test.cc:2:17: warning: macro expands to multiple statements [-Wmultistatement-macros]
 #define dodgy() foo(); foo();
                 ^~~
test.cc:5:17: note: in expansion of macro ‘dodgy’
                 dodgy()
                 ^~~~~
test.cc:4:9: note: some parts of macro expansion are not guarded by this ‘if’ clause
         if (true)
         ^~
```

peternewman: Ah very fancy. I'd probably still favour the brackets too though, as they don't really hurt.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: Sorry to push back on this, but honestly I don't see STLRemove having any advantage over callind `.erase()` here. All this change would do is invalidate my testing and perform another unnecessary traversal to find the right element again. A combined lookup and remove would have cleaned up the code a bit but for sure, but just with just the removal I don't see the point.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: Done.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: Done, log level now depends on where the error is NOT_FOUND or not.

plugins/usbdmx/AsyncUsbTransceiverBase.cpp
DanielG: Done

PR #1635, opened 2020-04-28
Master resync

peternewman:


peternewman:
Merging with a broken pychecker check, to separate the fix for that into another PR to avoid confusion and simplify the reviews later.

PR #1636, opened 2020-05-07
Added a doxyfile specifically for the python api

TFA-N:
I discussed adding a doxygen documentation with Peter on IRC, so here is my first attempt at it. It still needs a bit of work, as some of the pages generated are rather ugly.


peternewman:
Thanks @TFA-N . To fix the Travis issues (aside from Pychecker which I need to look at), you need to add the file to the tarball or exclude it from our check:
https://travis-ci.org/github/OpenLightingProject/ola/jobs/684473397#L6674-L6679

In this case, looking at the existing Doxyfile, we should add it here:
https://github.com/OpenLightingProject/ola/blob/master/Makefile.am#L209

peternewman:
I've just been looking at our Ja-rule/OLE code and we generate separate user manuals and developer manuals there in two folders and run two separate commands, which might be another option for this:
https://github.com/OpenLightingProject/ja-rule/tree/master/user_manual

PR #1637, opened 2020-05-12
Add a flake8 annotation action

peternewman:


peternewman:
I've no idea why this isn't showing up in the PR, but it looks great here:
https://github.com/peternewman/ola/actions?query=branch%3A0.10-c11-compat

Issue #1640, opened 2020-05-16
README.developer: Update comment about C++11 usage


kripton:
According to https://github.com/OpenLightingProject/ola/blob/master/README.developer#L83, C++11 should not be used since it was not completely supported in Ubuntu 12.04. According to https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions, Ubuntu 12.04 was EOL 2019-04 so that piece of documentation might need an update.

PR #1647, opened 2020-07-01
ola_rdm_get/set should only display actionable PIDs

### LINKED ISSUE #1632: ola_rdm_get/set --list-pids should only show PIDs you can get/set rather than all of them
So if there is a set only PID (e.g. RESET_DEVICE) it shouldn't show with ola_rdm_get --list-pids.

Make sure we update the docs to make this clear too.

DragoonBoots:
Fixes #1632 

Side note - README.developer requires using C++03, but I've noticed #1640 exists.  This change can be cleaned up a bit using C++11 features (i.e. lambdas) and I'd like to update if C++11 support has become A Thing.


examples/ola-rdm.cpp
```diff
@@ -207,16 +207,22 @@ void DisplayHelpAndExit(const options &opts) {
...
 void DisplayPIDsAndExit(uint16_t manufacturer_id,
-                        const PidStoreHelper &pid_helper) {
+                        const PidStoreHelper &pid_helper,
+                        bool set_mode) {
+  vector<const ola::rdm::PidDescriptor *> pids;
+  pid_helper.SupportedPids(manufacturer_id, &pids);
+  // Remove PIDs that don't support the proper command class (GET with ola_rdm_get, SET with ola_rdm_set)
   vector<string> pid_names;
-  pid_helper.SupportedPids(manufacturer_id, &pid_names);
-  sort(pid_names.begin(), pid_names.end());
```
peternewman, line 218: You seem to have dropped the sort in the rewrite. It would be good to get that back, either as a sort of the strings you're accumulating, or perhaps create a sort function (if we don't have one already), sort the descriptors beforehand then print directly.

examples/ola-rdm.cpp
```diff
@@ -207,16 +207,22 @@ void DisplayHelpAndExit(const options &opts) {
...
-
+  vector<const ola::rdm::PidDescriptor *>::const_iterator it_pids = pids.begin();
+  for (; it_pids != pids.end(); ++it_pids) {
+    if ((set_mode && (*it_pids)->SetRequest() != NULL) || (!set_mode && (*it_pids)->GetRequest() != NULL)) {
+      pid_names.push_back((*it_pids)->Name());
```
peternewman, line 223: I.e. I think this and the cout stuff can be one operation if they are sorted on the way into this for loop.

examples/ola-rdm.cpp
```diff
@@ -207,16 +207,22 @@ void DisplayHelpAndExit(const options &opts) {
...
 /*
  * Dump the list of known pids
  */
 void DisplayPIDsAndExit(uint16_t manufacturer_id,
-                        const PidStoreHelper &pid_helper) {
+                        const PidStoreHelper &pid_helper,
+                        bool set_mode) {
+  vector<const ola::rdm::PidDescriptor *> pids;
+  pid_helper.SupportedPids(manufacturer_id, &pids);
+  // Remove PIDs that don't support the proper command class (GET with ola_rdm_get, SET with ola_rdm_set)
```
peternewman, line 218: This line is too long, they need to be 80 characters or less, so wrap it onto the next line.

examples/ola-rdm.cpp
```diff
@@ -207,16 +207,22 @@ void DisplayHelpAndExit(const options &opts) {
...
   vector<string> pid_names;
-  pid_helper.SupportedPids(manufacturer_id, &pid_names);
-  sort(pid_names.begin(), pid_names.end());
-
+  vector<const ola::rdm::PidDescriptor *>::const_iterator it_pids = pids.begin();
```
peternewman, line 220: And here.

examples/ola-rdm.cpp
```diff
@@ -207,16 +207,22 @@ void DisplayHelpAndExit(const options &opts) {
...
+  for (; it_pids != pids.end(); ++it_pids) {
+    if ((set_mode && (*it_pids)->SetRequest() != NULL) || (!set_mode && (*it_pids)->GetRequest() != NULL)) {
```
peternewman, line 222: Here

examples/ola-rdm.cpp
```diff
@@ -615,7 +621,7 @@ int main(int argc, char *argv[]) {
   delete opts.uid;
 
   if (opts.list_pids)
-    DisplayPIDsAndExit(dest_uid.ManufacturerId(), controller.PidHelper());
+    DisplayPIDsAndExit(dest_uid.ManufacturerId(), controller.PidHelper(), opts.set_mode);
```
peternewman, line 624: And here.

common/rdm/PidStoreHelper.cpp
```diff
@@ -286,6 +288,15 @@ void PidStoreHelper::SupportedPids(
   if (store) {
     store->AllPids(descriptors);
   }
+
+  // Sort PIDs by name
+  sort(descriptors->begin(), descriptors->end(), &CompPidsForSort);
```
peternewman, line 293: Given something else may want the PIDs sorted by ID for example, or might not care, I think it would be better to do the sort in the consuming function.

common/rdm/PidStoreHelper.cpp
```diff
@@ -286,6 +288,15 @@ void PidStoreHelper::SupportedPids(
...
+}
+
+
+bool PidStoreHelper::CompPidsForSort(const PidDescriptor *a,
```
peternewman, line 297: Perhaps ComparePidsByName would be clearer?

Do you fancy adding just a basic @brief Doxygen like some of the other bits have too?

include/ola/rdm/PidStoreHelper.h
```diff
@@ -100,6 +100,8 @@ class PidStoreHelper {
     MessageDeserializer m_deserializer;
     RDMMessagePrinter m_message_printer;
     ola::messaging::SchemaPrinter m_schema_printer;
+
+    static bool CompPidsForSort(const PidDescriptor* a, const PidDescriptor* b);
```
peternewman, line 104: Obviously make this public if it's going to be used in the consuming code.

examples/ola-rdm.cpp
```diff
@@ -212,14 +213,18 @@ void DisplayHelpAndExit(const options &opts) {
...
+                        const PidStoreHelper &pid_helper,
+                        bool set_mode) {
+  vector<const PidDescriptor *> pids;
+  pid_helper.SupportedPids(manufacturer_id, &pids);
+  // Remove PIDs that don't support the proper command class (GET with
+  // ola_rdm_get, SET with ola_rdm_set)
+  vector<const PidDescriptor *>::const_iterator it_pids = pids.begin();
+  for (; it_pids != pids.end(); ++it_pids) {
+    if ((set_mode && (*it_pids)->SetRequest() != NULL)
+        || (!set_mode && (*it_pids)->GetRequest() != NULL)) {
```
peternewman, line 225: We generally prefer the || at the end of the line:
```suggestion
    if ((set_mode && (*it_pids)->SetRequest() != NULL) ||
        (!set_mode && (*it_pids)->GetRequest() != NULL)) {
```

examples/ola-rdm.cpp
```diff
@@ -615,7 +620,8 @@ int main(int argc, char *argv[]) {
   delete opts.uid;
 
   if (opts.list_pids)
-    DisplayPIDsAndExit(dest_uid.ManufacturerId(), controller.PidHelper());
+    DisplayPIDsAndExit(dest_uid.ManufacturerId(), controller.PidHelper(),
+                       opts.set_mode);
```
peternewman, line 624: Nit, can you put each argument on a new line please (i.e. move the PidHelper one onto it's own line). Sorry I should have suggested this in the earlier comment.

common/rdm/PidStoreHelper.cpp
DragoonBoots: I think I misinterpreted your previous comment on this one.  Will move sorting to the consumer.

AUTHORS
```diff
@@ -30,3 +30,4 @@ Contributors:
   Stefan Krüger, added APA102 support to the SPI Plugin
   Tobi Schäfer, for the MacPort files
   Tor Håkon Gjerde, added AVLdiy D512 support
+  Dan Keenan, various patches
```
peternewman, line 33: I think we're sorted by first name.

common/rdm/PidStoreHelper.cpp
```diff
@@ -287,5 +289,16 @@ void PidStoreHelper::SupportedPids(
...
   }
 }
+
+
+/**
+ * @brief Compare PIDs by name
+ *
+ * Suitable for use as a sort comparison function
+ */
+bool PidStoreHelper::ComparePidsByName(const PidDescriptor *a,
```
peternewman, line 299: There's a tiny bit of me wonders if this should be under PidDescriptor, as you could use it directly via the PidStore for example (rather than the helper), but I'm happy either way.

If you want to add a basic test for it would be great, although this is pretty trivial, so no worries if not. Actually looking further it seems we've got no testing for PidStoreHelper currently!

DragoonBoots: I agree with moving it to the PidDescriptor (it's what's being compared, after all...); I suppose this made more sense when the sorting was happening in the same class.

common/rdm/PidStoreHelper.cpp
```diff
@@ -32,6 +33,7 @@ namespace rdm {
 
 using std::string;
 using std::vector;
+using std::sort;
```
peternewman, line 36: Again

examples/ola-rdm.cpp
```diff
@@ -212,14 +213,20 @@ void DisplayHelpAndExit(const options &opts) {
...
-  sort(pid_names.begin(), pid_names.end());
-
-  vector<string>::const_iterator iter = pid_names.begin();
-  for (; iter != pid_names.end(); ++iter) {
-    cout << *iter << endl;
+                        const PidStoreHelper &pid_helper,
+                        bool set_mode) {
+  vector<const PidDescriptor *> pids;
+  pid_helper.SupportedPids(manufacturer_id, &pids);
+  std::sort(pids.begin(), pids.end(), &PidDescriptor::CompareByName);
```
peternewman, line 220: If you're using std::sort, you can just go for sort here. If it's only used once, I generally drop the using and keep the namespace on it.

common/rdm/PidStore.cpp
```diff
@@ -208,6 +208,17 @@ bool PidDescriptor::IsSetValid(uint16_t sub_device) const {
 }
 
 
+/**
+ * @brief Compare PIDs by name
+ *
+ * Suitable for use as a sort comparison function
+ */
+bool PidDescriptor::CompareByName(const PidDescriptor *a,
```
peternewman, line 216: I did wonder about a more consistent name, but we currently have a mix of lt and LessThan suffixes and prefixes as well as ordering, and snake and camel case, so one for another day I think! Although if you wanted to touch it, perhaps order/ordering rather than compare given it's not the cmp operator underneath.

examples/ola-rdm.cpp
DragoonBoots: I usually prefer to leave the namespace in across the board (probably why I missed removing the `using` before), and it's what [Google likes to do](https://google.github.io/styleguide/cppguide.html#Namespaces).

peternewman: The Google stuff looks like its about using a namespace, rather than an class in a namespace as we're doing here. Certainly cpplint doesn't complain about this setup.

peternewman:
Thanks again @DragoonBoots . The other stuff finally got in so this could be merged.

AUTHORS
peternewman: Ooh we should fix this at some point somewhere!

PR #1422, opened 2018-05-24
Support libftdi1 too in the ftdidmx plugin

Keeper-of-the-Keys:
libftdi1 support

PR #1648, opened 2020-07-04
Add in/out point support to ola_recorder

### LINKED ISSUE #1604: Add the ability for ola_recorder to record for a certain amount of time
As raised in https://groups.google.com/forum/#!topic/open-lighting/IYniIKUXvI4

DragoonBoots:
Fixes #1604.

This seemed like a good way to learn more about OLA's event handling.


peternewman:
Just for info, I started on #1607 which I think actually covers the specifics of #1604 but then got sidetracked with various stuff. It looks like we've actually worked on different features, which is great, there's just a small chance of a merge conflict when they both go in. I'll try and get mine merged in soon to simplify that.

The other features look good and are certainly likely to be wanted by people.

DragoonBoots:
Worst case I can try to rebase this once #1607 is merged.  I did a bit of refactoring to clean up how show files are read; I imagine that's the most likely place for a merge conflict.

examples/ShowLoader.cpp
```diff
@@ -141,6 +141,24 @@ ShowLoader::State ShowLoader::NextFrame(unsigned int *universe,
...
+  State state = NextFrame(&entry->universe, &entry->buffer);
+  if (state != State::OK) {
+      return state;
+  }
+  state = NextTimeout(&entry->next_wait);
+  if (state != State::OK) {
+      return state;
+  }
+
+  return State::OK;
```
peternewman, line 158: Any reason to not just return state here? Which also makes the if above redundant for now, aside from my other comment.

examples/ShowLoader.cpp
```diff
@@ -141,6 +141,24 @@ ShowLoader::State ShowLoader::NextFrame(unsigned int *universe,
...
+/**
+ * Read the next show file entry
+ * @param entry a ShowEntry to fill with data
+ */
+ShowLoader::State ShowLoader::NextEntry(ShowEntry *entry) {
+  State state = NextFrame(&entry->universe, &entry->buffer);
+  if (state != State::OK) {
+      return state;
+  }
+  state = NextTimeout(&entry->next_wait);
```
peternewman, line 153: Technically, if a show file only has one entry, there will be no timeout, just:
`1 1,2,3`

examples/ShowPlayer.cpp
```diff
@@ -85,55 +92,84 @@ int ShowPlayer::Playback(unsigned int iterations,
...
+    m_client.GetSelectServer()->Terminate();
+    return;
   }
+  SendFrame(entry);
+}
 
-  OLA_INFO << "Registering timeout for " << timeout << "ms";
+
+/**
+ * Send @p frame and wait for next
```
peternewman, line 158: Does the Doxygen not want to be `@p entry` from looking at https://www.doxygen.nl/manual/commands.html#cmdp

examples/ola-recorder.cpp
```diff
@@ -117,20 +123,14 @@ int VerifyShow(const string &filename) {
   map<unsigned int, unsigned int> frames_by_universe;
   uint64_t total_time = 0;
 
-  unsigned int universe;
-  ola::DmxBuffer buffer;
-  unsigned int timeout;
+  ShowEntry entry;
   ShowLoader::State state;
   while (true) {
```
peternewman, line 128: I think it would probably good if verify also respected, and printed your in and out times. So you can find out how many frames per universe within your time range.

man/ola_recorder.1
```diff
@@ -32,6 +32,11 @@ Print
 version information
 .IP "--duration <uint32_t>"
 The length of time (seconds) to run for.
+.IP "--start <uint_32_t>"
+Time (milliseconds) in show file to start playback from.
+.IP "--stop <uint32_t>"
+Time (milliseconds) in show file to stop playback at.  If the show file is
+shorter, this option is ignored.
```
peternewman, line 39: It would be nice if it printed a warning if this was the case.

examples/ShowPlayer.cpp
```diff
@@ -85,55 +92,84 @@ int ShowPlayer::Playback(unsigned int iterations,
...
-    case ShowLoader::INVALID_LINE:
-      m_client.GetSelectServer()->Terminate();
-      return;
-    default:
-      {}
+/**
+ * Seek to @p time in the show file
+ * @param time the time (in milliseconds) to seek to
+ */
+void ShowPlayer::SeekTo(const unsigned int time) {
```
peternewman, line 108: Time is a used function:
http://www.cplusplus.com/reference/ctime/time/

So it would be good to use another variable name, perhaps seek_time?

examples/ShowPlayer.cpp
DragoonBoots: Yup, that's what I get for trusting CLion's rename tools...

examples/ShowLoader.cpp
DragoonBoots: This may actually be a bug in the playback post-refactor - since the last entry never has a timeout, the last frame will never be sent.  I'll have to rework this a bit to accommodate this.

DragoonBoots: Would you say it's safe to assume a show file with a trailing timeout is invalid?  Assuming a showfile will always end with a frame makes this less messy.  It looks like verify didn't care before, but the recorder won't output a trailing timeout.

examples/ShowPlayer.cpp
peternewman: Perhaps it's time for a bug report...

examples/ShowLoader.cpp
peternewman: There's a chance it might have existed before too, I've not reviewed the code too closely.

I added this TODO note in my other PR:
https://github.com/OpenLightingProject/ola/pull/1607/files#diff-5970e4574d66fe51e161810176e44cadR110-R111

I think there are lots of edge cases, but given we're reliant on incoming data that's not under our control to get the frames, I think the best case would be if it could handle leading and trailing timeouts too. Especially as the format is so simple, I've certainly suggested generating show files, I'm not sure I've actually done it. If you're recording a duration, do you want 10s from now, or from the first DMX frame or what? We could also theoretically grab the current data when we start recording, but that potentially changes the frequency of the frames giving a lumpy fade.

If you're doing the duration stuff, it's conceivable you want say a 10s, or an hour long recording, and you don't want that to drift over a week, so the fancy display at the top of the hour is always at that time.

I think at a minimum verify should throw a warning/error and report statistics that match what it will actually do in terms of timing. We could always leave fixing the complicated bit for the future with a TODO or an issue.

DragoonBoots: Alright, it's a bit clunky, but I think I've found a way to handle both cases (trailing timeout/trailing frame).

I generated a few test show file by having sACNView run a chase across two universes; having about 10 seconds of data to play with made testing easier.

I might make a separate PR that cleans up output - as I've been testing this I've used my desk calculator to add up the timeouts to verify that start/stop times are what they should be.  For DMX output, I feel something along the lines of "<time>: Univ <#>: <dmx values> would be more helpful, but that also feels a bit out of scope here.  As you said, if you use this for something long-running it's conceivable the clock might drift quite a bit (especially on something nasty without an actual RTC like a Raspberry Pi).

examples/ShowPlayer.cpp
```diff
@@ -85,68 +92,111 @@ int ShowPlayer::Playback(unsigned int iterations,
...
   }
+
+  // Keep reading through the show file until desired time is reached.
+  ShowEntry entry;
+  unsigned int playhead_time = m_playback_pos;
+  do {
+    ShowLoader::State state = m_loader.NextEntry(&entry);
+    switch (state) {
+      case ShowLoader::END_OF_FILE:
+        HandleEndOfFile();
```
peternewman, line 124: Shouldn't this just bomb out and not call this function during seek, otherwise if your file is 10s, your start is 20s and infinite loops you would get:
1. SeekTo(time after end of file)
1. Hit EOF
1. m_loader.reset()
1. SeekTo(time after end of file)
etc

I think it should just say it overran the file looking for the start and then:
`m_client.GetSelectServer()->Terminate();`

examples/ShowPlayer.cpp
```diff
@@ -85,68 +92,111 @@ int ShowPlayer::Playback(unsigned int iterations,
...
+        return;
+      case ShowLoader::INVALID_LINE:
+        m_client.GetSelectServer()->Terminate();
```
peternewman, line 127: Now we've got numbers, perhaps say why we've given up, with a line number.

examples/ShowPlayer.cpp
```diff
@@ -85,68 +92,111 @@ int ShowPlayer::Playback(unsigned int iterations,
...
+        return;
+      default: {}
+    }
+    playhead_time += entry.next_wait;
+  } while (playhead_time < seek_time);
```
peternewman, line 132: Doesn't this want to be <=? Imagine the following file with a seek time of 5:
```
1 1,2,3
0
2 11,12,13
5
1 4,5,6
0
2 14,15,16
1
1 7,8,9
```

We want to output
`1 4,5,6`

As the first frame don't we? I think it will currently output 7,8,9?

examples/ShowPlayer.cpp
```diff
@@ -85,68 +92,111 @@ int ShowPlayer::Playback(unsigned int iterations,
...
+             << m_playback_pos << " ms)";
+  }
   m_iteration_remaining--;
   if (m_infinite_loop || m_iteration_remaining > 0) {
     m_loader.Reset();
-    m_client.GetSelectServer()->RegisterSingleTimeout(
-        m_loop_delay,
-        ola::NewSingleCallback(this, &ShowPlayer::SendNextFrame));
+    // Move to start point and send the frame
+    SeekTo(m_start);
```
peternewman, line 199: We should still honour any loop delay that's been set, why don't you just call the new Start function?

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,51 @@ int VerifyShow(const string &filename) {
...
+    if (FLAGS_stop > 0 && playback_pos >= FLAGS_stop) {
+      // Compensate for overshooting the stop time
+      playback_pos -= (playback_pos - FLAGS_stop);
       break;
-    total_time += timeout;
+    }
+    if (!playing && playback_pos >= FLAGS_start) {
+      // Found the start point
+      playing = true;
+      entry.next_wait = playback_pos - FLAGS_start;
```
peternewman, line 142: Isn't this irrelevant here? Or otherwise you're double subtracting FLAGS_start if it is used.

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,51 @@ int VerifyShow(const string &filename) {
...
+    }
+    if (playing) {
+      frames_by_universe[entry.universe]++;
+    }
   }
 
+  const unsigned int total_time = playback_pos - FLAGS_start;
```
peternewman, line 149: Was there a reason for dropping this, and playback_pos from being uint64_t? That's less than a month in ms!

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,51 @@ int VerifyShow(const string &filename) {
...
   map<unsigned int, unsigned int>::const_iterator iter;
   unsigned int total = 0;
   cout << "------------ Summary ----------" << endl;
+  if (FLAGS_start > 0) {
+    cout << "Starting at: " << FLAGS_start / 1000.0 << " second(s)" << endl;
+    if (FLAGS_stop > 0) {
```
peternewman, line 155: We don't have to have a start to have a stop point.

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,51 @@ int VerifyShow(const string &filename) {
...
        ++iter) {
     cout << "Universe " << iter->first << ": " << iter->second << " frames" <<
       endl;
     total += iter->second;
   }
+  std::cout << endl;
   cout << "Total frames: " << total << endl;
-  cout << "Playback time: " << total_time / 1000 << "." << total_time % 10 <<
-    " seconds" << endl;
+  cout << "Playback time: " << total_time / 1000.0 << " second(s)" << endl;
```
peternewman, line 168: Yes, I'd noticed this when testing duration and getting confused. It might be worth us looking at setw to make something nicer output wise.

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,51 @@ int VerifyShow(const string &filename) {
...
+    }
+    std::cout << endl;
+  }
   for (iter = frames_by_universe.begin(); iter != frames_by_universe.end();
        ++iter) {
     cout << "Universe " << iter->first << ": " << iter->second << " frames" <<
       endl;
     total += iter->second;
   }
+  std::cout << endl;
```
peternewman, line 166: Again.

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,51 @@ int VerifyShow(const string &filename) {
...
+  const unsigned int total_time = playback_pos - FLAGS_start;
   map<unsigned int, unsigned int>::const_iterator iter;
   unsigned int total = 0;
   cout << "------------ Summary ----------" << endl;
+  if (FLAGS_start > 0) {
+    cout << "Starting at: " << FLAGS_start / 1000.0 << " second(s)" << endl;
+    if (FLAGS_stop > 0) {
+      cout << "Stopping at: " << FLAGS_stop / 1000.0 << " second(s)" << endl;
+    }
+    std::cout << endl;
```
peternewman, line 158: This can just be cout as we're already `using std::cout;`

examples/ola-recorder.cpp
```diff
@@ -154,6 +175,20 @@ int VerifyShow(const string &filename) {
...
 }
 
+
+/**
+ * Playback a recorded show
+ */
+int PlaybackShow() {
+  ShowPlayer player(FLAGS_playback.str());
+  int status = player.Init();
+  if (!status)
```
peternewman, line 185: Can we have some {}

examples/ShowPlayer.cpp
DragoonBoots: That's a good point - there really isn't a valid case where it should hit EOF here.

examples/ShowPlayer.cpp
DragoonBoots: Stepping through with the debugger, I've got 11,12,13 immediately followed by 4,5,6 (0 ms delay).
I interpreted the requirements as "starting with the thing 5ms in," which would be 11,12,13.  Thinking about it now though, "start after 5ms has passed" (i.e. in this case 4,5,6) does seem more natural.  Will change.

peternewman: I wasn't expecting that result. I took it that the first line of the file is output at time zero, and then delay ms later the next line is output. I think 4,5,6 is the first data that is 5ms from the start (along with 14,15,16). i.e. the delay is attached to the following line (with an unlisted 0ms delay before the first line).

Remember in my slightly contrived (although fully supported) example, they are two universes, so 1,2,3 and 11,12,13 are happening at the same time.

peternewman: This looks to be what the old code did to me too:
Read frame
Register timeout to repeat
Output frame
Timeout expires, repeat process

DragoonBoots: With some more testing, it seems there's a special case here.  If the start time lands at the end of a frame's timeout (as 5 does with your example) and the following frame has a timeout of 0 (as does 4,5,6), that following frame is skipped.  Enough to be noticeable?  Probably not, since timeout 0 means the next frame is broadcast near instantly, but still technically a problem.

peternewman: If they are two separate universes though, then that's potentially a bigger problem as one universe gets no data for 2 x timeperiods, which could also vary in length. With some protocols and few data changes, I think you could be talking 2+ seconds.

examples/ola-recorder.cpp
DragoonBoots: I can definitely restore it - because all of the time-related flags are uint32_t (the largest type the flag macros allow), it seemed logical to use the same type.  My thought was one could either define a show with no time bounds that lasts until the heat death of the universe, or use bounds and have the show last up to about a month.  No strong feelings either way.

examples/ola-recorder.cpp
DragoonBoots: I thought there was a bug when `--start 1000 --end 1500` resulted in some weird output... :grin:

DragoonBoots:
Just saw your new comments, I'll work on it a bit more.

examples/ShowPlayer.cpp
DragoonBoots: This seems to be a bigger mess than I expected - it still follows the old process once playback has started, but finding the start point seems fraught with danger.

DragoonBoots: I think the solution here will have to be:
- Track the value each universe is at while seeking
- Once the start point is found, send what the current state should be; this may be many lines ago for some universes.

DragoonBoots: Alright, I think I've untangled this.
Using this file:
```
1 1,2,3
0
2 11,12,13
5
1 2,3,4
1
1 4,5,6
0
2 14,15,16
3
1 7,8,9
1
2 17,18,19
```
`--start 5` begins with `1: 2,3,4` `2: 11,12,13`
`--start 6` begins with `1: 4,5,6` `2: 11,12,13`

DragoonBoots:
I appreciate your working with me on this @peternewman - It's so nice to receive good feedback like this!
Should be good to go now.

examples/ShowPlayer.cpp
peternewman: > This seems to be a bigger mess than I expected - it still follows the old process once playback has started, but finding the start point seems fraught with danger.

Why do you think I didn't do the complicated bit? :laughing: 

> I think the solution here will have to be:
> 
>     * Track the value each universe is at while seeking
>     * Once the start point is found, send what the current state should be; this may be many lines ago for some universes.

Yeah, sounds reasonable, it potentially means we'd generate additional data, which we don't normally do, but I can't think of a better solution.



> Alright, I think I've untangled this.
> Using this file:

Would you agree your file maps to this table?

Time | Universe 1 | Universe 2
------------ | ------------- |  ------------- 
0 | 1,2,3 | 11,12,13
5 | 2,3,4 |
6 | 4,5,6 | 14,15,16
9 | 7,8,9 |
10 | | 17,18,19

> `--start 5` begins with `1: 2,3,4` `2: 11,12,13`
So that one looks fine.
> `--start 6` begins with `1: 4,5,6` `2: 11,12,13`
Is that going to output 14,15,16 immediately afterwards on universe 2?

The perfect solution when seeking is probably to look ahead and find the last point in time where you're still in the start time asked for (updating the stored values as you go), then output those values, then go into normal mode. Although I appreciate that's probably even more complicated. Both Art-Net and E1.31 (sACN) have a sync packet, so this would possibly help in making those synchronous.

peternewman:
> I appreciate your working with me on this @peternewman 
Thanks for doing the hard work and writing and testing the code!

> It's so nice to receive good feedback like this!
No worries, that's good to hear, I'm always a bit worried I'm being a bit too fussy.

examples/ShowPlayer.cpp
DragoonBoots: That table looks correct to me, and seems to match output when using `--start <time>`

`--start 6` gives:
```
examples/ShowPlayer.cpp:217: Universe: 2: 11,12,13
examples/ShowPlayer.cpp:217: Universe: 1: 4,5,6
examples/ShowPlayer.cpp:206: Registering timeout for 0ms
examples/ShowPlayer.cpp:217: Universe: 2: 14,15,16
examples/ShowPlayer.cpp:206: Registering timeout for 3ms
examples/ShowPlayer.cpp:217: Universe: 1: 7,8,9
examples/ShowPlayer.cpp:206: Registering timeout for 1ms
examples/ShowPlayer.cpp:217: Universe: 2: 17,18,19
```

Do you think it's worth adding another special case such that it would skip outputting 11,12,13 on U2 and send 14,15,16 instead?  I would have it not complete seeking (and therefore not output) until it encounters a non-0 timeout.

examples/ola-recorder.cpp
peternewman: Answering your comments in reverse:

> or use bounds and have the show last up to about a month.

I think that's the key, if it was over a month I wouldn't have said anything (I don't realistically expect a show to run for over a month), but monthly showfiles and a Christmas special don't seem too unlikely for example.

> I can definitely restore it - because all of the time-related flags are uint32_t (the largest type the flag macros allow), it seemed logical to use the same type.

I think in reality that suggests we possibly need an additional flag type, but that's probably one for another PR!

DragoonBoots: The latest commit replaces time variables with a typedef for uint64_t.

examples/ShowPlayer.cpp
DragoonBoots: Avoid that minor flaw turned out to be simpler than I expected.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
+     timeout,
+     ola::NewSingleCallback(this, &ShowPlayer::SendNextFrame));
+}
+
+
+/**
+ * Send data contained in @p entry
+ */
+void ShowPlayer::SendFrame(const ShowEntry &entry) const {
+  OLA_INFO << "Universe: " << entry.universe << ": " << entry.buffer.ToString();
```
peternewman, line 218: This probably wants to be debug level to avoid flooding the log with every frame.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
+  m_client.GetClient()->SendDMX(entry.universe, entry.buffer, args);
 }
 
 
 /**
  * Handle the case where we reach the end of file
  */
 void ShowPlayer::HandleEndOfFile() {
+  if (m_stop > m_playback_pos) {
+    OLA_WARN << "Show file ends before the stop time (Actual length "
```
peternewman, line 229: You could possibly delay until m_stop here to do what was asked for timing wise. I'm not sure if it's a good idea or can be handled with other options or flags.

examples/ola-recorder.cpp
```diff
@@ -57,6 +57,12 @@ DEFINE_uint32(duration, 0, "The length of time (seconds) to run for.");
 // 0 means infinite looping
 DEFINE_s_uint32(iterations, i, 1,
                 "The number of times to repeat the show, 0 means unlimited.");
+DEFINE_uint32(start, 0,
+              "Time (milliseconds) in show file to start playback from.");
+DEFINE_uint32(stop, 0,
+              "Time (milliseconds) in show file to stop playback at.  If"
```
peternewman, line 63: I think just one space at the end of the sentence, but I think I got caught out making this comment elsewhere, or that may be Python...

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,62 @@ int VerifyShow(const string &filename) {
...
     if (state != ShowLoader::OK)
       break;
-    frames_by_universe[universe]++;
-
-    state = loader.NextTimeout(&timeout);
-    if (state != ShowLoader::OK)
+    playback_pos += entry.next_wait;
+    if (FLAGS_stop > 0 && playback_pos >= FLAGS_stop) {
+      // Compensate for overshooting the stop time
+      playback_pos -= (playback_pos - FLAGS_stop);
```
peternewman, line 136: Isn't this just FLAGS_stop?

examples/ola-recorder.cpp
```diff
@@ -115,36 +121,62 @@ int VerifyShow(const string &filename) {
...
       break;
-    total_time += timeout;
+    }
+    if (!playing && playback_pos > FLAGS_start) {
+      // Found the start point
+      playing = true;
+    }
+    if (playing) {
+      frames_by_universe[entry.universe]++;
```
peternewman, line 144: This won't match the frames output if it's pre-cached frames? Perhaps increment during seek and then reset to 1 if non-zero when you currently set playing=true?

examples/ShowPlayer.cpp
```diff
@@ -35,20 +34,24 @@
 #include <fstream>
 #include <iostream>
 #include <string>
+#include <unordered_map>
```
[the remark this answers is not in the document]
DragoonBoots: Whoops - I've been working on a greenfield project at the same time.

examples/ShowPlayer.h
```diff
@@ -34,6 +34,8 @@
  */
 class ShowPlayer {
  public:
+  typedef uint64_t PlaybackTime;
```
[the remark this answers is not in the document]
DragoonBoots: I've always been a fan of typedefs when types have shared meaning (e.g. positions, start/stop times), but don't have any strong feelings.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
+    switch (state) {
+      case ShowLoader::END_OF_FILE:
+        OLA_FATAL << "Show file ends before the start time (Actual length "
+                  << m_playback_pos << " ms)";
+        return state;
+      case ShowLoader::INVALID_LINE:
+        HandleInvalidLine();
+        return state;
+      default: {
+      }
```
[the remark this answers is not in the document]
DragoonBoots: It's true - I seem to recall a similar construct in the original code for sending frames; I guess I removed the rest of them.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
+    playhead_time += entry.next_wait;
+    entries[entry.universe] = entry;
+    if (!found && playhead_time == seek_time) {
+      // Use the next frame if landing on the trailing edge of a frame's timeout
+      found = true;
+      continue;
+    }
+    if (found || playhead_time > seek_time) {
+      break;
+    }
```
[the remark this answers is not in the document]
DragoonBoots: I noticed this as well - this should be fixed in the latest pushed commit

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
-  unsigned int timeout;
-  ShowLoader::State state = m_loader.NextTimeout(&timeout);
-  if (state != ShowLoader::OK) {
-    return state;
+void ShowPlayer::SendNextFrame() {
+  ShowEntry entry;
+  ShowLoader::State state = m_loader.NextEntry(&entry);
+  if (state == ShowLoader::END_OF_FILE ||
+      (m_stop > 0 && m_playback_pos >= m_stop)) {
+    if (entry.buffer.Size() > 0) {
```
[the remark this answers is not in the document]
DragoonBoots: Which logic?  Having SendFrame not send empty buffers makes sense; that I can move.  SendNextFrame is the regular callback when running the show, so I felt it made more sense for the stop checking to happen there.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
+                  << m_playback_pos << " ms)";
+        return state;
+      case ShowLoader::INVALID_LINE:
+        HandleInvalidLine();
+        return state;
+      default: {
+      }
+    }
+    playhead_time += entry.next_wait;
+    entries[entry.universe] = entry;
```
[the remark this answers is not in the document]
DragoonBoots: Good point about different numbers of slots - the buffers would have to get merged instead of replaced which might murder seek performance.  I imagine that could be visible on looping shows.

examples/ShowPlayer.cpp
DragoonBoots: It was info level before this PR - If this changes it may also make sense to change the log level for the timeouts.  Thoughts?

examples/ShowPlayer.cpp
DragoonBoots: Hmmmm.  If I have a 10 second show file and want it to last 20 seconds, right now I can set delay to 10 and avoid the stop flag at all.  However, that also requires I do some math when computers are quite good at doing that for me.  It would also require first knowing the actual length of the show file using verify.

I feel like adding an extra option would sow more confusion, especially since the warning will always be seen with the default log level as warn.

examples/ola-recorder.cpp
DragoonBoots: My grade school typing teacher is sneering at me...

Checking elsewhere, the convention seems to be one space.

examples/ola-recorder.cpp
DragoonBoots: I think this showed up when I was trying to figure out what craziness was causing `start 1` `stop 2` to become `length 1.7`

Yup.  Math is still math.

examples/ola-recorder.cpp
DragoonBoots: That's true - now that multiple universes are handled properly, this needs some fixing.

I originally wanted to implement this by running the ShowPlayer in a sort of simulate mode to avoid these issues, but it's too closely coupled to having olad running for that to be workable...

examples/ShowPlayer.cpp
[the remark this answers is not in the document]
DragoonBoots: Found why - we compile with -Werror and not having the default creates a warning:

```
examples/ShowPlayer.cpp:131:12: error: enumeration value ‘OK’ not handled in switch [-Werror=switch]
  131 |     switch (state) {
      |            ^
```

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,184 @@ int ShowPlayer::Init() {
...
+ * @param seek_time the time (in milliseconds) to seek to
+ */
+ShowLoader::State ShowPlayer::SeekTo(PlaybackTime seek_time) {
+  // Seeking to a time before the playhead's position requires moving from the
+  // beginning of the file.  This could be optimized more if this happens
+  // frequently.
+  if (seek_time < m_playback_pos) {
+    m_loader.Reset();
+    m_playback_pos = 0;
+  }
```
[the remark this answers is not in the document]
DragoonBoots: There be dragons...

This code is only hit when looping, and I can't imagine any other situation where it would be used right now, so this is really just an acknowledgement of a potential performance issue if a hero decides to add jumps to support their Christmas light show.

examples/ShowPlayer.cpp
DragoonBoots: The more I think about the more I'm convinced waiting is the correct behavior.  Reusing the same scenario, keeping it as is could mean a show with start=1000 and end=5000 lasts for 2 seconds, for example.  With my end-user hat on, that feels like a bug.

peternewman: Yeah, I think we should log a warning during verify if you've passed in a stop flag which is bigger than the show file, but I think the reality will be you have a 9.9 second showfile and you want it to run for 10s and loop type of scenarios.

examples/ShowPlayer.cpp
peternewman: Agreed that the timeouts should match the same log level.

Well you've been using it, how do you feel, is it too noisy or about right? Looking at the current live source, there aren't as many logging bits specifically in the ola_recorder stuff as I thought there might be, so it might not be an issue. Certainly there seem to be very few other bits at OLA_INFO level.

examples/ShowPlayer.cpp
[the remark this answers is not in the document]
peternewman: Yeah sorry I meant not sending an empty buffer.

examples/ShowPlayer.cpp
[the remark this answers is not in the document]
peternewman: It should only need to deal with combining a longer and shorter buffer pair, so should just be a memcpy or something, but probably one for another day.

examples/ola-recorder.cpp
peternewman: We do PIMPL ( https://en.cppreference.com/w/cpp/language/pimpl ) in a few other places. I think the only olad interfaces that needs are connect and send data and register a timeout, it ought to be possible to mock those up, but it might not be worth it.

examples/ShowPlayer.cpp
```diff
@@ -170,9 +173,9 @@ ShowLoader::State ShowPlayer::SeekTo(PlaybackTime seek_time) {
...
   ShowLoader::State state = m_loader.NextEntry(&entry);
-  if (state == ShowLoader::END_OF_FILE ||
-      (m_stop > 0 && m_playback_pos >= m_stop)) {
-    if (entry.buffer.Size() > 0) {
+  const bool stop_point_hit = m_stop > 0 && m_playback_pos >= m_stop;
+  if (state == ShowLoader::END_OF_FILE || stop_point_hit) {
+    if (stop_point_hit) {
       // Send the last frame before looping/exiting
       SendFrame(entry);
     }
```
peternewman: I think it should only output this last frame if m_playback_pos == m_stop, otherwise you're potentially sending a frame after m_stop, which isn't what you've asked for in the CLI options.

examples/ShowPlayer.cpp
```diff
@@ -225,17 +231,19 @@ void ShowPlayer::SendFrame(const ShowEntry &entry) const {
  * Handle the case where we reach the end of file
  */
 void ShowPlayer::HandleEndOfFile() {
+  uint64_t loop_delay = m_loop_delay;
   if (m_stop > m_playback_pos) {
     OLA_WARN << "Show file ends before the stop time (Actual length "
```
peternewman, line 236: Nit, the a of actual should be lower case.

examples/ShowPlayer.cpp
```diff
@@ -225,17 +231,19 @@ void ShowPlayer::SendFrame(const ShowEntry &entry) const {
...
              << m_playback_pos << " ms)";
```
peternewman, line 237: Perhaps extend the log line to say you're pausing for x ms to catch up or something?

examples/ola-recorder.cpp
```diff
@@ -60,7 +60,7 @@ DEFINE_s_uint32(iterations, i, 1,
 DEFINE_uint32(start, 0,
               "Time (milliseconds) in show file to start playback from.");
 DEFINE_uint32(stop, 0,
-              "Time (milliseconds) in show file to stop playback at.  If"
+              "Time (milliseconds) in show file to stop playback at. If"
               " the show file is shorter, this option is ignored.");
```
peternewman: Nit, I normally try and put the space at the end of the line so you don't have to remember to keep it in place, like so:
```suggestion
              "Time (milliseconds) in show file to stop playback at. If "
              "the show file is shorter, this option is ignored.");
```

examples/ola-recorder.cpp
```diff
@@ -112,6 +112,21 @@ int RecordShow() {
...
+ * This allows frames that would be cached during playback to be counted.
+ */
+void ClampVerifyFrameCount(map<unsigned int, unsigned int> &frames) {
+  map<unsigned int, unsigned int>::iterator iter;
+  for (iter = frames.begin(); iter != frames.end(); ++iter) {
+    if (iter->second > 1) {
+      iter->second = 1;
+    }
+  }
+}
```
peternewman, line 127: Does this actually need to be a whole function, it only seems to get used once in one place, wouldn't a simple loop through the frames when we go into playback mode be just as easy (without needing to pass the frame counter in etc?

examples/ShowPlayer.cpp
DragoonBoots: I asked a non-programmer friend about this because sometimes it's nice to hear from people who don't care about the low-level details.  He thought knowing actual output was less important ("Want to know what the output is?  Look at the lights! Or open the web browser and see what olad says.").

I'm inclined to drop the DMX output and timeouts down to debug and open with a summary of how the file will be played (e.g. Playing from ##ms to ##ms, loop delay, etc) so the user knows it's alive.

The other things that log at info level should probably stay there, i.e. the looping notice.

examples/ShowPlayer.cpp
```diff
@@ -111,49 +111,52 @@ void ShowPlayer::Loop() {
  * Seek to @p seek_time in the show file
  * @param seek_time the time (in milliseconds) to seek to
  */
-ShowLoader::State ShowPlayer::SeekTo(PlaybackTime seek_time) {
+ShowLoader::State ShowPlayer::SeekTo(uint64_t seek_time) {
   // Seeking to a time before the playhead's position requires moving from the
   // beginning of the file.  This could be optimized more if this happens
   // frequently.
-  if (seek_time < m_playback_pos) {
+  if (seek_time <= m_playback_pos) {
```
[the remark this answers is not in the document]
DragoonBoots: Kind of - when the first and last frame end up being the same (could be a long static look, for example), it would only be sent on the first iteration and never again.  This forces it to be loaded again.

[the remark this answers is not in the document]
DragoonBoots: I noticed this while testing seeking to exactly EOF.

examples/ShowPlayer.cpp
```diff
@@ -111,49 +111,52 @@ void ShowPlayer::Loop() {
...
-        OLA_FATAL << "Show file ends before the start time (Actual length "
-                  << m_playback_pos << " ms)";
-        return state;
-      case ShowLoader::INVALID_LINE:
-        HandleInvalidLine();
-        return state;
-      default: {
+    if (state == ShowLoader::END_OF_FILE) {
+      if (playhead_time == seek_time) {
+        // Send the only frame(s) we have and loop
```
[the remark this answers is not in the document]
DragoonBoots: This should only happen when seeking to the last frame in the file - the end result is sending that frame (or frames, for multiple universes) and going through the normal EOF process.  I can add a message noting that start == EOF.

examples/ola-recorder.cpp
DragoonBoots: Not really, just trying to stop VerifyShow from becoming any more unwieldy.

examples/ShowPlayer.cpp
peternewman: So stop_point_hit could now be just put directly in line `if (state == ShowLoader::END_OF_FILE || stop_point_hit) {` just above.

examples/ShowPlayer.cpp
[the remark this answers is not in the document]
peternewman: Perhaps just add a comment in the code explaining this for our future selves?

examples/ShowPlayer.cpp
peternewman: For the summary, what about running verify at the start, which also stops playback getting half way through and then failing. Although perhaps add a --no-verify option to playback for cases where you're e.g. running schedules on demand from an external system and don't want the delay (i.e. assuming you've pre-verified all your shows).

I was chatting to another user and they said similar, things like loop status at the start end end (I don't think we do a countdown of iterations currently) are what they care about. Which also allows for external application wrapping.

We could also use cout to write to stdout, although my feeling is that's probably not necessarily the right thing to do given this is almost a daemon, you don't interact with it or expect output from it, so the strength of being able to redirect all logs to syslog is probably higher.

examples/ola-recorder.cpp
```diff
@@ -115,36 +122,70 @@ int VerifyShow(const string &filename) {
...
+
+      // Clamp the frames count between 0 and 1
+      map<unsigned int, unsigned int>::iterator clamp_iter;
+      for (clamp_iter = frames_by_universe.begin();
+           clamp_iter != frames_by_universe.end();
+           ++clamp_iter) {
+        if (clamp_iter->second > 1) {
+          clamp_iter->second = 1;
+        }
+      }
```
peternewman, line 153: This still didn't smell quite right and I've realised why, see the comment above.

examples/ola-recorder.cpp
```diff
@@ -115,36 +122,70 @@ int VerifyShow(const string &filename) {
...
-    state = loader.NextFrame(&universe, &buffer);
+    state = loader.NextEntry(&entry);
     if (state != ShowLoader::OK)
       break;
-    frames_by_universe[universe]++;
-
-    state = loader.NextTimeout(&timeout);
-    if (state != ShowLoader::OK)
+    playback_pos += entry.next_wait;
+    frames_by_universe[entry.universe]++;
```
peternewman, line 135: Wouldn't this make the clamping routine redundant, and is probably clearer to read? Subject to the `playback_pos > FLAGS_start` bit?

```suggestion
    if (playing)
      frames_by_universe[entry.universe]++;
    } else {
      // Clamp the frame count to 1 as we haven't actually started playing yet
      frames_by_universe[entry.universe] = 1;
    }
```

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,205 @@ int ShowPlayer::Init() {
...
 
 /**
  * Handle the case where we reach the end of file
  */
 void ShowPlayer::HandleEndOfFile() {
-  m_iteration_remaining--;
-  if (m_infinite_loop || m_iteration_remaining > 0) {
-    m_loader.Reset();
+  if (m_iteration_remaining > 0) {
+    m_iteration_remaining--;
```
peternewman, line 237: Do we print the remaining iteration count elsewhere? It would be nice to see somewhere.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +69,205 @@ int ShowPlayer::Init() {
...
+  OLA_INFO << "Universe: " << entry.universe << ": " << entry.buffer.ToString();
+  ola::client::SendDMXArgs args;
+  m_client.GetClient()->SendDMX(entry.universe, entry.buffer, args);
 }
 
 
 /**
  * Handle the case where we reach the end of file
  */
 void ShowPlayer::HandleEndOfFile() {
```
peternewman, line 235: Is this actually EndOfFile now, or also end of show if stop point < EOF?

examples/ola-recorder.cpp
DragoonBoots: That's a really great way of simplifying this!

examples/ShowPlayer.cpp
DragoonBoots: I like the idea of re-using verify like that.
I hadn't considered daemonizing the program like that, although it sounds nice - if run as a cron job or similar to playback at a certain time or something.

DragoonBoots: > We could also use cout to write to stdout, although my feeling is that's probably not necessarily the right thing to do given this is almost a daemon, you don't interact with it or expect output from it, so the strength of being able to redirect all logs to syslog is probably higher.

My instinct is to output `--verify` to stdout (it's intentional and likely run interactively) and the pre-playback summary to log level info (as the rest of playback output already ends up in a logging target).  If we go this route, I'd add a louder note to the man page/help text about the `-l` options as otherwise it's entirely possible for a user to see no output at all, including summary, loop notices, etc. while running the program because the default log level is warn.

The only other place the logging level options are mentioned is in the docs for olad; because this seems to be the example program that makes heaviest use of the logging system it feels worth a mention.

DragoonBoots:
I've switched `--verify` to simulate the ShowPlayer instead of attempting to have it match the ShowPlayer's behavior, but poorly.  This should clear up some of your comments about the original routine becoming a bit of a mess with the added features, as well as simplifying integrating it with playback.

examples/ola-recorder.cpp
```diff
@@ -200,9 +174,24 @@ int VerifyShow(const string &filename) {
...
+  if (!FLAGS_no_verify) {
+    // Verify the show and print a summary before running
+    string summary;
+    const int verified = VerifyShow(filename, &summary);
+    // Printing a newline first makes this output look better in interactive
+    // terminal logs.
+    OLA_INFO << endl << summary;
+    if (verified != ola::EXIT_OK) {
+      // Show did not pass verification
+      return verified;
```
peternewman, line 187: We should print an error here, or do we get one from verify?

DragoonBoots: Verify handles printing errors (it's the same method used when running with `--verify`); I'll add a comment about that.

examples/ola-recorder.cpp
```diff
@@ -154,6 +170,41 @@ int VerifyShow(const string &filename) {
...
+ */
+int PlaybackShow() {
+  const string filename = FLAGS_playback.str();
+  if (FLAGS_verify_playback) {
+    // Verify the show and print a summary before running
+    string summary;
+    const int verified = VerifyShow(filename, &summary);
+    // Printing a newline first makes this output look better in interactive
+    // terminal logs.
+    OLA_INFO << endl << summary;
```
peternewman, line 185: Do you mean because it spaces each iteration? This will give some weird effects like a line stamp with no output. Can you give some example output to explain what you mean, or do you mean in a loop, in which case a ---- separator might look better (but perhaps only when looping)? Or printing a cout << endl but only if we're not logging to syslog.

examples/ola-recorder.cpp
```diff
@@ -154,6 +170,41 @@ int VerifyShow(const string &filename) {
...
+    if (verified != ola::EXIT_OK) {
+      // Show did not pass verification
+      // The Verify method has already informed the user, so fail with that
+      // method's result.
```
peternewman, line 189: Because we've merged the output, and force it down to info level, at the default warn level we won't get any output, it will just fail. It would be good to either get an error message here, or just log a generic verify failed message.

examples/ola-recorder.cpp
```diff
@@ -164,16 +215,20 @@ int main(int argc, char *argv[]) {
                "Record a series of universes, or playback a previously "
                "recorded show.");
 
+  if (FLAGS_stop > 0 && FLAGS_stop < FLAGS_start) {
+    OLA_FATAL << "Stop time must be later than start time.";
```
peternewman, line 219: Did I ask before, what happens if stop==start, or do you just get that one frame played out?

man/ola_recorder.1
```diff
@@ -32,6 +34,11 @@ Print
...
```
peternewman, line 41: It would be nice to be able to do exactly a 30s loop (from 5 to 35), even if your showfile is a bit short (say 34, but then just warn that it's run out of showfile).

man/ola_recorder.1
```diff
@@ -32,6 +34,11 @@ Print
 version information
 .IP "--duration <uint32_t>"
 The length of time (seconds) to run for.
```
peternewman, line 36: As it's just tripped me up, can we reword this a bit to make it clear it's the total time it will playback for, what takes priority with loops and stop start etc.

examples/ola-recorder.cpp
DragoonBoots: In the case of stop == start, only the final frame is used.  This is what the weirdness at the beginning of `ShowPlayer::SeekTo()` is supposed to handle.

examples/ShowPlayer.cpp
```diff
@@ -66,90 +70,268 @@ int ShowPlayer::Init() {
...
+    m_iteration_remaining--;
+  }
+
+  const bool loop = m_infinite_loop || m_iteration_remaining > 0;
+  uint64_t loop_delay = m_loop_delay;
+  if (m_stop > m_playback_pos) {
+    OLA_WARN << "Show file ends before the stop time (actual length "
+             << m_playback_pos << " ms)";
+    if (loop) {
+      const uint64_t remaining_time = m_stop - m_playback_pos;
```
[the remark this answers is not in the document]
DragoonBoots: It doesn't look like it - this is only called from `SendNextFrame()`.  The cases this message is not shown are EOF without a stop point specified or a stop point specified within the bounds of the show file.

Not forcing the stop point to be within bounds allows for an iteration duration longer than the show file, holding last look until the user wants it to loop.  I think this situation is what you're referring to with a 30 second loop in your comment below?

man/ola_recorder.1
DragoonBoots: I'll double check; this should already be possible.  Just missed updating docs.

examples/ola-recorder.cpp
```diff
@@ -108,44 +117,51 @@ int RecordShow() {
...
-    cout << "Universe " << iter->first << ": " << iter->second << " frames" <<
-      endl;
-    total += iter->second;
+    summary->assign(out.str());
   }
-  cout << "Total frames: " << total << endl;
-  cout << "Playback time: " << total_time / 1000 << "." << total_time % 10 <<
-    " seconds" << endl;
-
   if ((state == ShowLoader::OK) || (state == ShowLoader::END_OF_FILE)) {
```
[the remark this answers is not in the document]
DragoonBoots: Looks like it's both in this function depending on where you are - I'll clean it up a bit.

examples/ola-recorder.cpp
```diff
@@ -108,44 +117,51 @@ int RecordShow() {
...
+
+  if (state == ola::EXIT_OK && summary != NULL) {
+    map<unsigned int, uint64_t> frames_by_universe = player.GetFrameCount();
+    const uint64_t total_time = player.GetRunTime();
+    std::stringstream out;
+    map<unsigned int, uint64_t>::const_iterator iter;
+    uint64_t total = 0;
+    out << "------------ Summary ----------" << endl;
+    if (FLAGS_start > 0) {
+      out << "Starting at: " << FLAGS_start / 1000.0 << " second(s)" << endl;
```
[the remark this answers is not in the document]
DragoonBoots: After adding info about other flags passed these are a bit more phrasal (it also helps establish start/stop times as relative to the recorded show file, not playback times

examples/ola-recorder.cpp
DragoonBoots: The summary is only ever printed for the entire run, not per iteration (since each iteration is the same, and total frame counts are likely more helpful than frame counts per iteration).

Example:
```
ola_recorder -l3 --playback test.show --iterations 5
examples/ola-recorder.cpp:206: 
------------ Summary ----------
For all (5) iterations:
Universe 1: 20 frames
Universe 2: 15 frames

Total frames: 35
Total playback time: 0.05 seconds
examples/ShowPlayer.cpp:330: ----- 4 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
examples/ShowPlayer.cpp:330: ----- 3 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
examples/ShowPlayer.cpp:330: ----- 2 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
examples/ShowPlayer.cpp:330: ----- 1 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
```

Without leading newline:
```
ola_recorder -l3 --playback test.show --iterations 5
examples/ola-recorder.cpp:206: ------------ Summary ----------
For all (5) iterations:
Universe 1: 20 frames
Universe 2: 15 frames

Total frames: 35
Total playback time: 0.05 seconds
examples/ShowPlayer.cpp:330: ----- 4 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
examples/ShowPlayer.cpp:330: ----- 3 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
examples/ShowPlayer.cpp:330: ----- 2 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
examples/ShowPlayer.cpp:330: ----- 1 iteration(s) remain -----
examples/ShowPlayer.cpp:333: ----- Waiting 0 ms before looping -----
```
...which seems a bit ghastly IMO

examples/ola-recorder.cpp
DragoonBoots: The summary info is at info level, but there are still log messages from ShowPlayer that get sent to fatal, error, or warn that will have been shown.

After mashing my keyboard into the show file to make it invalid, I get:
```
ola_recorder --verify test.show
examples/ShowLoader.cpp:139: Line 15 invalid: sergtesdrgtsd
examples/ShowPlayer.cpp:349: Invalid data at line 15
```

man/ola_recorder.1
DragoonBoots: I'll work on it.  I also want to change this to milliseconds, as it's a bit confusing for this to be the only time argument that takes seconds, not milliseconds.

peternewman: Hmm, actually I'm not sure we should change it, as that breaks existing scripts...

examples/ola-recorder.cpp
peternewman: I'm not too bothered either way, if you move the summary down so it's nicely formatted, it would be good to put a line saying "verification of FILENAME:" or something maybe, to give the log line some context. That's probably the only/one of the only multi-line log messages, hopefully it doesn't upset the syslog option.

examples/ShowPlayer.cpp
```diff
@@ -81,13 +86,13 @@ int ShowPlayer::Playback(unsigned int iterations,
...
   m_stop = stop;
-  m_status = ola::EXIT_OK;
+  m_status = ola::EXIT_SOFTWARE;
 
   if (!m_simulate) {
     ola::io::SelectServer *ss = m_client.GetSelectServer();
     if (duration != 0) {
       ss->RegisterSingleTimeout(
-          duration * 1000,
+          duration,
```
peternewman: As mentioned, this isn't good for backwards compatibility, and on a potentially key option. I can't really think of a good solution though, replacing the argument seems quite clunky.

examples/ShowPlayer.cpp
```diff
@@ -130,16 +135,21 @@ int ShowPlayer::Playback(unsigned int iterations,
...
+    // Success conditions
+    case ShowLoader::END_OF_FILE:
+    case ShowLoader::OK:
       m_status = ola::EXIT_OK;
-    }
-    if (!m_simulate) {
-      m_client.GetSelectServer()->Terminate();
-    }
+      break;
+      // All other states are considered errors of some sort
```
peternewman, line 145: I think this should be de-dented to match the word case shouldn't it, as it applies to the following block.

examples/ola-recorder.cpp
```diff
@@ -55,15 +55,19 @@ DEFINE_default_bool(verify_playback, true,
 DEFINE_s_string(universes, u, "",
                 "A comma separated list of universes to record");
 DEFINE_s_uint32(delay, d, 0, "The delay in ms between successive iterations.");
-DEFINE_uint32(duration, 0, "The length of time (seconds) to run for.");
+DEFINE_uint32(duration, 0, "Total playback time (milliseconds); the program "
+                           "will close after this time has elapsed. This "
+                           "option overrides the iteration option.");
```
peternewman, line 60: I probably should have commented here, I'm worried about this breaking change.

examples/ola-recorder.cpp
```diff
@@ -55,15 +55,19 @@ DEFINE_default_bool(verify_playback, true,
...
 // 0 means infinite looping
 DEFINE_s_uint32(iterations, i, 1,
-                "The number of times to repeat the show, 0 means unlimited.");
+                "The number of times to repeat the show, 0 means unlimited. "
+                "The duration option overrides this option.");
```
peternewman, line 64: I know we don't always verify, but will we get the log message if it does verify and the two choices conflict?

examples/ShowPlayer.cpp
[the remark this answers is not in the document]
peternewman: I think I was more worried from a maths perspective of having 2 - 5 = trouble!

examples/ShowPlayer.cpp
DragoonBoots: When do BC breaks usually happen in this project?  I've no strong feelings about this particular change, but if a 0.11 (for instance) release is already on the roadmap this feels like a good combination of feature+cleanup to add to that.

Having said that, this is an example program and by no means justifies a new BC breaking release on its own.

examples/ola-recorder.cpp
DragoonBoots: Adding a warning message for this case

examples/ShowPlayer.cpp
```diff
@@ -126,6 +126,11 @@ int ShowPlayer::Playback(unsigned int iterations,
     }
   }
 
+  if (m_iteration_remaining > 0) {
+    OLA_WARN << "More iterations requested than can fit in the duration "
+                "(" << m_iteration_remaining << " iteration(s) remain)";
+  }
```
peternewman: Should this not go in verify instead?

examples/ShowPlayer.cpp
peternewman: We try hard to avoid them, as we get so much pain from upstream when we have to deal with their breaking changes. For example #1651 and #1641 and #1422 and a whole pile of libusb cruft.

I'm going to put the rest of my comments about this change in the other discussion on it.

examples/ShowPlayer.cpp
DragoonBoots: Given that other usage-type problems (e.g. stop point beyond show file end) are warned during playback, this seemed a fitting place for the warning.

man/ola_recorder.1
peternewman: So initially I was sort of sold, but then I realised why I really don't like this. So the only "safe" way to do it would be to rename the argument, and then seamlessly auto-translate it and then throw an error if you try and use both with conflicting values.

To try and give an analogy, you setup a cron job running your ola_recorder apart from this one deals with money and pays you each time it's run, the payment amount is set to $1 (given you're American). An excellent programmer called Dan comes along and adds some more features to it, like tax calculations which take cents, he wants to standardise and so sets all the arguments to use cents. You see a new version and upgrade, next month you just get paid one cent a few months later you notice and you're rather upset.

That's why I'm reluctant. People don't read changelogs, despite our best efforts. We could compromise on a warning if your time is less than 2s say (i.e. you've fed in a time in seconds into a millisecond argument), which would at least make finding the cause of your issue easy, but it would be nice to get rid of it entirely.

I pondered floats/doubles, but it seems we're into another world of pain there if we want a certain amount of accuracy so still not ideal. Something decimal based feels like the right sort of way to go though anyway. Realistically the smallest time you want is 1ms, also most people want stuff to run for a number of seconds, and probably don't care about the ms, or only will in the sense of half a second. So if we moved to milliseconds we're also getting people to multiply by 1000 all the time.

Something decimal based feels like a win, a bit like this, but perhaps just simplified and as part of our string library:
https://github.com/vpiotr/decimal_for_cpp

Realistically, milliseconds is likely to be the main place we use it, or possibly the odd other one, so it could possibly just be a string split and then two strtol. I think we then win all round, people can keep just using 1s, or can enter 1.5/1.25/1.125 or whatever if they want a specific ms duration, but it's always clear what's going on. If you want it to run for hours or days, you don't end up with an unmanageable number of zeros after it either.

Thoughts? But like I say in the other one, shall we get the rest of it in (with seconds for the duration) and then look at this?

DragoonBoots: I agree with you re. BC needs, just trying to gauge how best to handle it.  Sounds like the preferred solution is to leave duration as it was and handle the need for decimals across the code base at a later time.

examples/ShowPlayer.cpp
peternewman: Hmm yeah, can I say both? :smile: It would be nice to spot config issues during validation, rather than needing to wait for a 30 minute show to run, but it would also be good to see them during show run...

DragoonBoots: Because the verify function is the player without output, it appears in both places.  For instance, with a recorded show file about 11 seconds long:

```
ola_recorder --verify recorded.show --iterations 3 --duration 1
------------ Summary ----------
For all (3) iterations:
After playing for 1 second(s) total:
Universe 1: 4 frames
Universe 2: 40 frames

Total frames: 44
Total playback time: 1 seconds
examples/ShowPlayer.cpp:130: More iterations requested than can fit in the duration (3 iteration(s) remain)
```

peternewman: Ah, I forgot, and got confused by your comment about it being during playback!

peternewman (approved):
Thanks again for all your work on this @DragoonBoots , sorry it became a bit of a beast!

PR #1655, opened 2020-07-19
python3 rdm compat work from sdbbs 1615

brucelowekamp:
This takes the patches from @sdbbs required for the python rdm support to be py3 compatible.  Just a minor formatting change, remove the comparison fix, and fixing the rdm_get example for py3.

This has been manually tested using the python/examples code and the dummy plugin with olad.  Looking at code coverage, it still didn't exercise some of the RDM Frame code that I found concerning (discussed in #1615) but I don't actually know what paths would trigger that code, and don't know that I have the right devices to exercise it anyway.

Ideally this would have tests included.   I looked a bit at mocking, and I suspect it's possible by supplying a socket and sending packets manual, but pretty ugly and brittle.  My preferred way to automate it would be to stand up olad with the dummy plugin and use the python examples/python tests against the running olad.  AFAIK there aren't any automated tests that do that at present (except the ola client tests that use the olad library internally), and I decided against doing that without discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -358,8 +358,8 @@ before_install:
 after_failure:
 # Disabled as otherwise the logfile is too big
 #  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
-#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
-#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
+  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
+  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,8 +90,9 @@ matrix:
     - os: linux
       dist: xenial
       compiler: clang
-      env: TASK='compile'
-      python: '3.8'
+      env:
+      - TASK='compile'
+      - PYTHON='python3'
       addons:
         apt:
           packages:
@@ -104,7 +105,6 @@ matrix:
       dist: xenial
       compiler: gcc
       env: TASK='compile'
-      python: '2.7'
       addons:
         apt:
           packages:
@@ -306,12 +306,12 @@ before_cache:
 
 install:
 # Match the version of protobuf being installed via apt
-  - if [[ "$PROTOBUF" == "latest" ]]; then pip install --user protobuf; fi
-  - if [[ "$PROTOBUF" != "latest" ]]; then pip install --user protobuf==3.1.0; fi
+  - if [[ "$PROTOBUF" == "latest" ]]; then pip install --user protobuf; pip3 install --user protobuf; fi
+  - if [[ "$PROTOBUF" != "latest" ]]; then pip install --user protobuf==3.1.0; pip3 install --user protobuf==3.1.0; fi
   # disable until can be added to all build variants
-  #- pip install --user timeout-decorator
+  #- pip install --user timeout-decorator; pip3 install --user timeout-decorator
 # We need to use pip rather than apt on Xenial
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then pip install --user numpy; fi
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then pip install --user numpy; pip3 install --user numpy; fi
   - if [ "$TASK" = "coverage" ]; then pip install --user cpp-coveralls; fi
   - if [ "$TASK" = "flake8" ]; then pip install --user flake8; fi
   - if [ "$TASK" = "codespell" ]; then pip3 install --user git+https://github.com/codespell-project/codespell.git; fi
@@ -344,9 +344,10 @@ before_install:
  - if [ "$TRAVIS_OS_NAME" == "osx" -a "$CPPUNIT" == "1.14" ]; then brew install cppunit; fi # install the latest cppunit, which needs C++11
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then PATH=/usr/local/opt/ccache/libexec:$PATH; fi # Use ccache on Mac too
 #Put back the old pip numpy we need to work
- - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then pip install --upgrade --no-deps --force-reinstall --user numpy; fi
+ - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then pip install --upgrade --no-deps --force-reinstall --user numpy; pip3 install --upgrade --no-deps --force-reinstall --user numpy; fi
 #Coverity doesn't work with g++ 5 or 6, so only upgrade to g++ 4.9 for that
  - if [ "$TRAVIS_OS_NAME" == "linux" -a \( "$TASK" = "compile" -o "$TASK" = "coverage" -o "$TASK" = "doxygen" \) -a "$CXX" = "g++" ]; then export CXX="ccache g++-9" CC="ccache gcc-9"; fi
+ - if [ "$PYTHON" == "python3" ]; then export PYTHON="python3"; fi
  - if [ "$TASK" = "coverity" -a "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="gcc-4.9"; fi
 #Use the latest clang if we're compiling with clang
  - if [ "$TRAVIS_OS_NAME" == "linux" -a "$CXX" = "clang++" ]; then export CXX="clang++-6.0" CC="clang-6.0"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -178,7 +178,7 @@ matrix:
       dist: xenial
       env:
        - TASK='codespell'
-       - PATH=/opt/python/3.7.1/bin:$PATH
+       - PYTHON='python3'
       addons:
         apt:
           packages:
@@ -314,7 +314,7 @@ install:
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then pip install --user numpy; fi
   - if [ "$TASK" = "coverage" ]; then pip install --user cpp-coveralls; fi
   - if [ "$TASK" = "flake8" ]; then pip install --user flake8; fi
-  - if [ "$TASK" = "codespell" ]; then pip3 install --user git+https://github.com/codespell-project/codespell.git; fi
+  - if [ "$TASK" = "codespell" ]; then pip install --user git+https://github.com/codespell-project/codespell.git; fi
   - if [ "$TASK" = "jshint" ]; then npm install -g grunt-cli; fi
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ matrix:
       dist: xenial
       compiler: clang
       env: TASK='compile'
-      python: '2.7'
+      python: '3.8'
       addons:
         apt:
           packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -353,8 +353,8 @@ before_install:
 after_failure:
 # Disabled as otherwise the logfile is too big
 #  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
-  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
-  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
+#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
+#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -358,8 +358,7 @@ before_install:
 after_failure:
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log; fi
-# Disabled as otherwise the logfile is too big
-#  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
+  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -356,12 +356,12 @@ before_install:
  - if [ "$TASK" == "spellintian" -o "$TASK" == "spellintian-duplicates" ]; then wget "http://old-releases.ubuntu.com/ubuntu/pool/main/l/lintian/lintian_2.5.104_all.deb"; sudo dpkg -i lintian_*.deb; sudo apt-get install -f -y; fi # Install a later lintian
 
 after_failure:
+  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log; fi
+  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log; fi
 # Disabled as otherwise the logfile is too big
 #  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
-  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log; fi
-  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log; fi
 
 after_success:
   - if [ "$TASK" = "coverage" ]; then coveralls --gcov /usr/bin/gcov-8 -b . -E '.*Test\.cpp$' -E '.*\.pb\.cc$' -E '.*\.pb\.cpp$' -E '.*\.pb\.h$' -E '.*\.yy\.cpp$' -E '.*\.tab\.cpp$' -E '.*\.tab\.h$' -E '.*/doxygen/examples.*$' --gcov-options '\-lp' > /dev/null; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -306,18 +306,19 @@ before_cache:
 
 install:
 # Match the version of protobuf being installed via apt
-  - if [[ "$PROTOBUF" == "latest" ]]; then pip install --user protobuf; pip3 install --user protobuf; fi
-  - if [[ "$PROTOBUF" != "latest" ]]; then pip install --user protobuf==3.1.0; pip3 install --user protobuf==3.1.0; fi
+  - if [[ "$PROTOBUF" == "latest" ]]; then pip install --user protobuf; fi
+  - if [[ "$PROTOBUF" != "latest" ]]; then pip install --user protobuf==3.1.0; fi
   # disable until can be added to all build variants
-  #- pip install --user timeout-decorator; pip3 install --user timeout-decorator
+  #- pip install --user timeout-decorator
 # We need to use pip rather than apt on Xenial
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then pip install --user numpy; pip3 install --user numpy; fi
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then pip install --user numpy; fi
   - if [ "$TASK" = "coverage" ]; then pip install --user cpp-coveralls; fi
   - if [ "$TASK" = "flake8" ]; then pip install --user flake8; fi
   - if [ "$TASK" = "codespell" ]; then pip3 install --user git+https://github.com/codespell-project/codespell.git; fi
   - if [ "$TASK" = "jshint" ]; then npm install -g grunt-cli; fi
 
 before_install:
+ - if [ "$PYTHON" == "python3" ]; then pyenv global 3.7.1 ; fi
 #Fix permissions for unbound (and possibly others)
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then if [ ! -d /usr/local/sbin ]; then sudo mkdir -p /usr/local/sbin && sudo chown -R $(whoami) /usr/local/sbin; fi; fi
 #Add a missing gnupg folder
@@ -344,10 +345,9 @@ before_install:
  - if [ "$TRAVIS_OS_NAME" == "osx" -a "$CPPUNIT" == "1.14" ]; then brew install cppunit; fi # install the latest cppunit, which needs C++11
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then PATH=/usr/local/opt/ccache/libexec:$PATH; fi # Use ccache on Mac too
 #Put back the old pip numpy we need to work
- - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then pip install --upgrade --no-deps --force-reinstall --user numpy; pip3 install --upgrade --no-deps --force-reinstall --user numpy; fi
+ - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then pip install --upgrade --no-deps --force-reinstall --user numpy; fi
 #Coverity doesn't work with g++ 5 or 6, so only upgrade to g++ 4.9 for that
  - if [ "$TRAVIS_OS_NAME" == "linux" -a \( "$TASK" = "compile" -o "$TASK" = "coverage" -o "$TASK" = "doxygen" \) -a "$CXX" = "g++" ]; then export CXX="ccache g++-9" CC="ccache gcc-9"; fi
- - if [ "$PYTHON" == "python3" ]; then export PYTHON="python3"; fi
  - if [ "$TASK" = "coverity" -a "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="gcc-4.9"; fi
 #Use the latest clang if we're compiling with clang
  - if [ "$TRAVIS_OS_NAME" == "linux" -a "$CXX" = "clang++" ]; then export CXX="clang++-6.0" CC="clang-6.0"; fi
--- a/Makefile.am
+++ b/Makefile.am
@@ -149,6 +149,9 @@ built_sources =
 # Test scripts are run if BUILD_TESTS is true.
 test_scripts =
 
+# directories with python code that should be test compiled during the build
+PYTHON_BUILD_DIRS =
+
 # The includes
 # -----------------------------------------------------------------------------
 
@@ -173,6 +176,7 @@ include olad/Makefile.mk
 include protoc/Makefile.mk
 include python/Makefile.mk
 include tools/Makefile.mk
+include scripts/Makefile.mk
 
 # -----------------------------------------------------------------------------
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -175,8 +175,8 @@ include plugins/Makefile.mk
 include olad/Makefile.mk
 include protoc/Makefile.mk
 include python/Makefile.mk
-include tools/Makefile.mk
 include scripts/Makefile.mk
+include tools/Makefile.mk
 
 # -----------------------------------------------------------------------------
 
--- a/python/Makefile.mk
+++ b/python/Makefile.mk
@@ -3,10 +3,10 @@ include python/ola/Makefile.mk
 
 python/PyCompileTest.sh: python/Makefile.mk
 	mkdir -p $(top_builddir)/python
-# 	restore this line when py3 compat is done for whole tree
-#	echo "$(PYTHON) -m compileall -f tools scripts python include data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
-	echo "$(PYTHON) -m compileall python data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	echo "$(PYTHON) -m compileall -f $(PYTHON_BUILD_DIRS); exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
 	chmod +x $(top_builddir)/python/PyCompileTest.sh
+
+PYTHON_BUILD_DIRS += python
 
 if BUILD_PYTHON_LIBS
 test_scripts += \
--- a/python/Makefile.mk
+++ b/python/Makefile.mk
@@ -5,7 +5,7 @@ python/PyCompileTest.sh: python/Makefile.mk
 	mkdir -p $(top_builddir)/python
 # 	restore this line when py3 compat is done for whole tree
 #	echo "$(PYTHON) -m compileall -f tools scripts python include data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
-	echo "$(PYTHON) -m compileall -f python data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	echo "$(PYTHON) -m compileall python data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
 	chmod +x $(top_builddir)/python/PyCompileTest.sh
 
 if BUILD_PYTHON_LIBS
--- a/python/Makefile.mk
+++ b/python/Makefile.mk
@@ -1,2 +1,17 @@
 include python/examples/Makefile.mk
 include python/ola/Makefile.mk
+
+python/PyCompileTest.sh: python/Makefile.mk
+	mkdir -p $(top_builddir)/python
+# 	restore this line when py3 compat is done for whole tree
+#	echo "$(PYTHON) -m compileall -f tools scripts python include data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	echo "$(PYTHON) -m compileall -f python data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	chmod +x $(top_builddir)/python/PyCompileTest.sh
+
+if BUILD_PYTHON_LIBS
+test_scripts += \
+    python/PyCompileTest.sh
+endif
+
+CLEANFILES += \
+    python/PyCompileTest.sh
--- a/python/Makefile.mk
+++ b/python/Makefile.mk
@@ -3,7 +3,7 @@ include python/ola/Makefile.mk
 
 python/PyCompileTest.sh: python/Makefile.mk
 	mkdir -p $(top_builddir)/python
-	echo "$(PYTHON) -m compileall -f $(PYTHON_BUILD_DIRS); exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	echo "$(PYTHON) -m compileall $(PYTHON_BUILD_DIRS); exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
 	chmod +x $(top_builddir)/python/PyCompileTest.sh
 
 PYTHON_BUILD_DIRS += python
--- a/python/ola/ClientWrapperTest.py
+++ b/python/ola/ClientWrapperTest.py
@@ -17,6 +17,7 @@
 # Copyright (C) 2019 Bruce Lowekamp
 
 import array
+import binascii
 import datetime
 import socket
 # import timeout_decorator
@@ -208,7 +209,13 @@ class ClientWrapperTest(unittest.TestCase):
 
     def DataCallback(self):
       data = sockets[1].recv(4096)
-      self.assertTrue(len(data) > 100)
+      expected = binascii.unhexlify(
+        "7d000010080110001a0d557064617465446d784461746122680801126400000"
+        "000000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000000"
+        "000000")
+      self.assertEqual(data, expected)
       results.gotdata = True
       wrapper.AddEvent(0, wrapper.Stop)
 
--- a/python/ola/ClientWrapperTest.py
+++ b/python/ola/ClientWrapperTest.py
@@ -215,7 +215,10 @@ class ClientWrapperTest(unittest.TestCase):
         "000000000000000000000000000000000000000000000000000000000000000"
         "000000000000000000000000000000000000000000000000000000000000000"
         "000000")
-      self.assertEqual(data, expected)
+      self.assertEqual(data, expected,
+                       msg="Regression check failed. If protocol change "
+                       "was intended set expected to: " +
+                       str(binascii.hexlify(data)))
       results.gotdata = True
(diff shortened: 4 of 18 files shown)


brucelowekamp:
@sdbbs hadn't heard from you for awhile and finally got time to play with it and couldn't generate any additional test cases.  This is just your code changes with the formatting change and the example chang.  More than happy to abandon this is you want to rebase your PR on 0.10 with the other two changes.

brucelowekamp:
@peternewman mac build is broken at present?  I don't think the failure had anything to do with this PR

sdbbs:
Hi @brucelowekamp ,

> hadn't heard from you for awhile 

Sorry about that - I've been assigned to slightly different work these days (weeks, rather), so haven't had much time to respond - I only noticed this notification because it had my pseudonym there :)

>  More than happy to abandon this is you want to rebase your PR on 0.10 with the other two changes.

I would have loved to do that, unfortunately what I'm assigned to at work now, does not allow me the time to take a look into it deeper - I might have some more time in a month or so, but it's just a wild guess. So I'm more than happy this python3 compatibility is looked into regardless.

>  Looking at code coverage, it still didn't exercise some of the RDM Frame code that I found concerning (discussed in #1615) 

Oh, that thread has advanced - I'd love to take a look at it in more detail, but as I mentioned, I don't really have the time for it these days ...

Just wanted to mention something - I think since I posted the #1615, I found that the last hunk (in function Pack, I guess), also needs a `if sys.version >= '3.2':` - I am attaching the patch where I recorded that here. 

Note the attached patch has an additional string in the warning in the Unpack function, that is probably just some leftover from my debugging, so you can ignore it. Also, the patch was done in respect to OLA commit f69e4a61 :

[ola0.10.7-1.py3_06_f69e4a61.patch.txt](https://github.com/OpenLightingProject/ola/files/4947143/ola0.10.7-1.py3_06_f69e4a61.patch.txt)

So, yeah - sorry about not being able to contribute more at this time, but my work at this time does not allow me the time for it; hope things clear up a bit soon, so I can go back to contributing a bit more ... And thanks again for looking a bit more into this!

python/ola/OlaClient.py
```diff
@@ -1357,7 +1357,11 @@ def _RDMMessage(self, universe, uid, sub_device, param_id, callback, data,
     request.uid.device_id = uid.device_id
     request.sub_device = sub_device
     request.param_id = param_id
-    request.data = data
+    if sys.version >= '3.2':
+      request.data = eval(data)[0] if data else bytes(data, 'utf-8')
+    else:
+      # eval(data)[0] if data else data - broke (MultiDim), 2019-10-31
+      request.data = data  # works, 2019-10-31
```
peternewman, line 1364: I think we can remove these comments can't we? The old code only had `request.data = data`.

I'm guessing this isn't necessary/required as one was an internal structure and one is a Protobuf call:
https://github.com/OpenLightingProject/ola/commit/255df6a6d86b030545384d79dce62d2509b8215a

python/ola/PidStore.py
```diff
@@ -458,7 +458,7 @@ def _AccountForMultiplierPack(self, value):
         raise ArgsValidationError(
             'Conversion will lose data: %d -> %d' %
             (new_value, (new_value / multiplier * multiplier)))
-      new_value = new_value / multiplier
+      new_value = int(new_value / multiplier)
```
peternewman, line 461: Good catch!

python/ola/OlaClient.py
sdbbs: > I think we can remove these comments can't we?

Of course! (and sorry about that, did not carefully review that patch before posting; those comments were only intended to be internal)

peternewman:
> fixing the rdm_get example for py3.

We should probably start with compilation checks for all the python example code. I assume there is a way to easily check it's valid at the CLI?

It doesn't look like we currently have Pack/Unpack tests for Python. See equivalents in common/rdm/MessageSerializerTest.cpp and common/rdm/RDMCommandSerializerTest.cpp for example.

> Ideally this would have tests included. I looked a bit at mocking, and I suspect it's possible by supplying a socket and sending packets manual, but pretty ugly and brittle.

Can we not go half way and generate the packet it would send to the socket and validate that the generated packet matches the expected manually generated data?

> My preferred way to automate it would be to stand up olad with the dummy plugin and use the python examples/python tests against the running olad. AFAIK there aren't any automated tests that do that at present (except the ola client tests that use the olad library internally), and I decided against doing that without discussion.

I think this would be great, although perhaps one for another PR. If we also ran the RDM responder tests (or perhaps run them as a first thing, as they're easy to automate), it would also mean we'd never end up with our responders breaking the tests and vice versa, which is a great place to be. I think I've half looked in the past but never quite got round to it.

brucelowekamp:
> Hi @brucelowekamp ,
> 
> > hadn't heard from you for awhile
> 
> Sorry about that - I've been assigned to slightly different work these days (weeks, rather), so haven't had much time to respond - I only noticed this notification because it had my pseudonym there :)

No worries, I don't use ola routinely either, so have spent a few months before getting back to it.  But would also like it to work well with py3.

> Just wanted to mention something - I think since I posted the #1615, I found that the last hunk (in function Pack, I guess), also needs a `if sys.version >= '3.2':` - I am attaching the patch where I recorded that here.

Thanks, grabbed that as well.


> > fixing the rdm_get example for py3.
> 
> We should probably start with compilation checks for all the python example code. I assume there is a way to easily check it's valid at the CLI?

That part was easy.  I actually had it written from my local branch that has (almost all) the python3 changes done for the rest of the tree.  Those changes aren't in (I think we discussed doing that after 0.10.8) so I commented out the parts of the tree that aren't updated and just run the compilation test on the sdk (and the test in data that does compile).

> 
> It doesn't look like we currently have Pack/Unpack tests for Python. See equivalents in common/rdm/MessageSerializerTest.cpp and common/rdm/RDMCommandSerializerTest.cpp for example.

agree, there aren't tests for Pack/Unpack.  I could possibly look at it at some point but didn't have time now.

> 
> > Ideally this would have tests included. I looked a bit at mocking, and I suspect it's possible by supplying a socket and sending packets manual, but pretty ugly and brittle.
> 
> Can we not go half way and generate the packet it would send to the socket and validate that the generated packet matches the expected manually generated data?

Yeah, that can be done.  In fact in ClientWrapperTest I used a socketpair to confirm data is going out, so the framework is even there, it's a matter of working out the test cases.


In the meantime I've pushed the comment fix, additional fix, and the compile test.

brucelowekamp:
@peternewman I used the socketpair to add a regression test for rdm_get that I think effectively tests the functionality.  Assuming that approach looks reasonable, I can pretty easily abstract it a bit and add some more test cases in there.  Obviously it's possible to do unit tests for pack/unpack, though I think this tests a bit more of the codepath.

The mac build is currently failing on the python compile test.  I'd like to see what that error is but I don't find test-suite.log in the build output.  What am I missing?  

On linux, at least, distcheck passes on my machine using PYTHON=python3.  The parts of the tree that aren't python compatible aren't exercised by distcheck.  Any objection if I try to switch just one of the builds to py3 just to catch regressions?

Other than the mac error, just simple misses on license and flake8 (that I broke fixing the vpath) that I can fix.

peternewman:
> @peternewman I used the socketpair to add a regression test for rdm_get that I think effectively tests the functionality.

Looks good.

> Assuming that approach looks reasonable, I can pretty easily abstract it a bit and add some more test cases in there.

Personally I'd be tempted to try and do a lot of the other tests at a higher level, that hex data is basically garbage (I think it's actually two levels of protobuf wrapped). So diagnosing would be a pain.

> Obviously it's possible to do unit tests for pack/unpack, though I think this tests a bit more of the codepath.

I think one test like this is useful for the codepath, but testing the other pack/unpack separately like we do in C++ is easier to see what's going on and add new tests for new things.

> The mac build is currently failing on the python compile test. I'd like to see what that error is but I don't find test-suite.log in the build output. What am I missing?

It should be one of these (unless it has another path, in which case just add another line):
https://github.com/OpenLightingProject/ola/blob/0.10/.travis.yml#L358-L359

> Any objection if I try to switch just one of the builds to py3 just to catch regressions?

I'd be tempted to add another Travis check, then we can clearly see if it's a Python 2 or three issue. Adding one to Linux would be pretty easy and quick I'd think. Or I guess on Linux the Clang and G++ tests could do one Python 2 and one 3, on the assumption there shouldn't be interactions between them.

peternewman:
> > The mac build is currently failing on the python compile test. I'd like to see what that error is but I don't find test-suite.log in the build output. What am I missing?
> 
> It should be one of these (unless it has another path, in which case just add another line):
> https://github.com/OpenLightingProject/ola/blob/0.10/.travis.yml#L358-L359

Edit: try commenting out the config.log logging, I think all the logging is too long and so we can't see the test suite log.

python/Makefile.mk
```diff
@@ -1,2 +1,17 @@
 include python/examples/Makefile.mk
 include python/ola/Makefile.mk
+
+python/PyCompileTest.sh: python/Makefile.mk
+	mkdir -p $(top_builddir)/python
+# 	restore this line when py3 compat is done for whole tree
+#	echo "$(PYTHON) -m compileall -f tools scripts python include data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	echo "$(PYTHON) -m compileall -f python data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
```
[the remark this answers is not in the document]
brucelowekamp: I acknowledge the point, but it also seems a bit odd to have that cruft added to include/Makefile.mk just to compile that one script.  The various coverage/flake8 etc tests generally run once from the root dir; in theory it could be put in Makefile.am, but that seems odd.  IMHO this was the best compromise, but if you feel strongly about it, I could put it in the various subdirs that need it.  (after I check the fix for the race condition that's causing the mac build to fail right now)

python/ola/ClientWrapperTest.py
```diff
@@ -208,7 +209,13 @@ class results:
 
     def DataCallback(self):
       data = sockets[1].recv(4096)
-      self.assertTrue(len(data) > 100)
+      expected = binascii.unhexlify(
+        "7d000010080110001a0d557064617465446d784461746122680801126400000"
+        "000000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000000"
+        "000000")
```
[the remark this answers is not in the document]
brucelowekamp: yes, this is basically another regression test with the output captured via debug from a real run

python/ola/rpc/StreamRpcChannel.py
```diff
@@ -170,6 +171,7 @@ def _SendMessage(self, message):
     data = message.SerializeToString()
     # combine into one buffer to send so we avoid sending two packets
     data = self._EncodeHeader(len(data)) + data
+    logging.debug("send->" + str(binascii.hexlify(data)))
```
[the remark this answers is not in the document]
brucelowekamp: This is the standard python library logging.  Curiously the only other place it's ever used is in the rdm_discovery.py example.   This wouldn't generally be enabled (and definitely not at debug level unless someone really wants something verbose), but I think is likely to be a bit useful in detailed debugging (and definitely with any desire to generate more regression tests).  However, looking at how it's called right now, it does the hexlify and then throws it away, so while I'd like to keep it, I think I will push a change to guard the line so it only prepares the log message if it's going to be used.

brucelowekamp:
@peternewman agree, I'll try to add maybe one or two more regression tests, but then a few at a higher level.  The regression tests are useful for their purpose, but as you say, not for debugging.

For some reason one of the mac builds had plugins/usbpro/WidgetDetectorThreadTester fail.   

I tried to switch the clang build to python 3.8, but for some reason the obvious travis.yml change was ignored a build time.  I'll have to look at that more later.   (and I could only see what happened a run time, as a clean run doesn't save the files or enough output.

peternewman:
> @peternewman agree, I'll try to add maybe one or two more regression tests, but then a few at a higher level. The regression tests are useful for their purpose, but as you say, not for debugging.

Great, sounds good.

> For some reason one of the mac builds had plugins/usbpro/WidgetDetectorThreadTester fail.

Yeah that and HealthCheckedConnectionTester have always been a bit flaky on Travis Mac. I think it's a race condition of some sort, but as they only show on the VM it's a bit hard to track down.
 
> I tried to switch the clang build to python 3.8, but for some reason the obvious travis.yml change was ignored a build time. I'll have to look at that more later. (and I could only see what happened a run time, as a clean run doesn't save the files or enough output.

If I was to guess, I think you'd need to possibly do brew link to make python->python3, but I've no idea if that upsets other stuff. Or somehow otherwise tell it to use the python3 binary.

What did config.log say about what it was using/had found?

peternewman:
Hi @brucelowekamp ,

@yoe would like us to release 0.10.8 by Nov 14th at the latest, to give time to get it into Debian before their deadline. How's this going, anything I can do to help? It would be great to get this in by then if possible, but no worries if not.

brucelowekamp:
> Hi @brucelowekamp ,
> 
> @yoe would like us to release 0.10.8 by Nov 14th at the latest, to give time to get it into Debian before their deadline. How's this going, anything I can do to help? It would be great to get this in by then if possible, but no worries if not.

I'll see if I can get some time in the next week to get a couple tests in here.

python/ola/RDMTest.py
```diff
@@ -0,0 +1,107 @@
...
+      self.assertEqual(data["product_category"], 32767)
+      self.assertEqual(data["dmx_start_address"], 1)
+      self.assertEqual(data["sub_device_count"], 0)
+      self.assertEqual(data["sensor_count"], 3)
+      wrapper.AddEvent(0, wrapper.Stop)
+
+    wrapper._ss.AddReadDescriptor(sockets[1], lambda: DataCallback(self))
+
+    uid = UID.FromString("7a70:ffffff00")
+    pid = pid_store.GetName("DEVICE_INFO", uid.manufacturer_id)
```
peternewman, line 93: You can just do this here as it's not a manufacturer specific PID:
```suggestion
    pid = pid_store.GetName("DEVICE_INFO")
```

peternewman (changes requested):
A few minor comments

brucelowekamp:
@peternewman You started to review just as I was finishing up another push with some major refactoring...

So figured out how to set up pack/unpack tests yesterday and then realized that the new code was broken when there were actually parameters in the structures.  Dove in a bit more and realized that the biggest problem was that default empty messages were being set to '', which is a string in py3.  By changing those to b'' almost everything just worked the way it originally was, and b'' literals were added in py2.6, so should be entirely safe.

New tests now exercise both empty requests and requests with params.  I've tested both with command line, though I didn't find a rdm_get I could run against the dummy plugin that would return a non-NACK response, but for the test that seems sufficient.  

I've run the tests and commands manually in py2 and py3 now.  

AFAIK this completes the intention of the original pull request, and the only outstanding item would be to figure out how to get a python3 build to run in travis.  I haven't had a chance to look at that again and need to get some other work done right now.  If a python3 build was enabled, need to NOT run flake8 check with python3 because it fails except for the python directory, though can easily change the Makefile to only run flake8 there.

Assuming the checks pass, though, this is ready other than that.  Should fix the travis file in here if it's merged before getting a python3 build to run.

brucelowekamp:
@sdbbs could you check if the new version works for you?  I think I tested a reasonable set of scenarios, but I wouldn't claim to have them all tested....

python/ola/rpc/StreamRpcChannel.py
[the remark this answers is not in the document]
brucelowekamp: The latest PR changes the behavior here to be guarded by a variable.  Only enabled via a code change, but IMHO that's sufficient given the level of debugging involved here.

python/ola/PidStoreTest.py
```diff
@@ -198,6 +198,79 @@ def testCmp(self):
...
+    with self.assertRaises(PidStore.ArgsValidationError):
+      args = ["2020", "255", "20", "20", "20", "20"]
+      blob = pid.Pack(args, PidStore.RDM_SET)
+
+    with self.assertRaises(PidStore.ArgsValidationError):
+      args = ["2020", "-1", "20", "20", "20", "20"]
+      blob = pid.Pack(args, PidStore.RDM_SET)
+
+    pid = store.GetName("LANGUAGE_CAPABILITIES")
+    args = ["Aa"]
```
peternewman, line 252: These should technically be lower case, so perhaps test with "en"?

python/ola/RDMTest.py
```diff
@@ -101,6 +102,61 @@ def ResponseCallback(self, response, data, unpack_exception):
...
+      results.gotrequest = True
+      response = binascii.unhexlify(
+        "2e000010080210002228080010021800220200062851300038004a0908f0f401"
+        "1500ffffff520908f0f40115ac107de05811")
+      sent_bytes = sockets[1].send(response)
+      self.assertEqual(sent_bytes, len(response))
+
+    def ResponseCallback(self, response, data, unpack_exception):
+      results.gotresponse = True
+      self.assertEqual(response.response_type, client.RDM_NACK_REASON)
```
peternewman, line 142: This is NACKing because it's not manufacturer specific. If you test this instead it should ACK:
http://rdm.openlighting.org/pid/display?manufacturer=31344&pid=32769

peternewman:
> @peternewman You started to review just as I was finishing up another push with some major refactoring...

Heh, oh well.

> So figured out how to set up pack/unpack tests yesterday and then realized that the new code was broken when there were actually parameters in the structures. Dove in a bit more and realized that the biggest problem was that default empty messages were being set to '', which is a string in py3. By changing those to b'' almost everything just worked the way it originally was, and b'' literals were added in py2.6, so should be entirely safe.

Ah great news.

> New tests now exercise both empty requests and requests with params.

Perfect!

> I've tested both with command line, though I didn't find a rdm_get I could run against the dummy plugin that would return a non-NACK response, but for the test that seems sufficient.

I thought you meant you wanted a NACK. So as mentioned in the other review, querying this PID will get an ACK and a description:
http://rdm.openlighting.org/pid/display?manufacturer=31344&pid=32769
 
Getting DMX personality description for ID 0 will NACK and 1 with ACK on that responder too.

> I've run the tests and commands manually in py2 and py3 now.

Excellent.

> AFAIK this completes the intention of the original pull request, and the only outstanding item would be to figure out how to get a python3 build to run in travis.

I'll try and have a look at that.

> If a python3 build was enabled, need to NOT run flake8 check with python3 because it fails except for the python directory, though can easily change the Makefile to only run flake8 there.

That's fine. Our flake8 run is separate so could stick to just being run as Python 2. Or we could add a more configured GitHub action one to hit the Python3 clean bits.

> Assuming the checks pass, though, this is ready other than that. Should fix the travis file in here if it's merged before getting a python3 build to run.

Great.

peternewman (changes requested):
A few more comments

python/Makefile.mk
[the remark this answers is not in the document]
brucelowekamp: @peternewman I haven't changed this code.  Thoughts?

python/ola/ClientWrapperTest.py
```diff
@@ -208,7 +209,16 @@ class results:
...
+      self.assertEqual(data, expected,
+                       msg="Regression check failed. If protocol change "
```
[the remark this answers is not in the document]
brucelowekamp: Personally I don't find the standard output at all useful but in theory at least it shows where there's a difference.  But I thought printing out the "change the test to AAAA if you meant to do this" was the most helpful and leave the standard in case it helps something.

python/ola/PidStoreTest.py
```diff
@@ -198,6 +198,79 @@ def testCmp(self):
...
+    store.Load([os.path.join(path, "test_pids.proto")])
+
+    pid = store.GetName("REAL_TIME_CLOCK")
+
+    args = ["2020", "6", "20", "20", "20", "20"]
+    blob = pid.Pack(args, PidStore.RDM_SET)
+    self.assertTrue(len(blob) > 1)
+
+    with self.assertRaises(PidStore.ArgsValidationError):
+      args = ["2000", "6", "20", "20", "20", "20"]
```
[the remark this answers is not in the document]
brucelowekamp: I commented those cases for what they're testing in case someone needs to look at them again.  I sort of liked that one b/c the valid range isn't 0- or 1- so it seemed a bit more interesting to check boundaries....

python/ola/RDMTest.py
brucelowekamp: In some ways it wasn't bad to have a NACK, but I wanted to have one that returned a response just to make sure the request actually made sense.  Needed one that was both in the dummy plugin and in the test_pids.proto and realized DMX_PERSONALITY_DESCRIPTION is, so switched to that.

brucelowekamp:
@peternewman it looks like the failures I see are not related to any changes in here, so I consider this ready.  Please let me know if you would like any other changes.

Just to state clearly, this changes the linux clang build to use python3.  The direct "python" setting in travis apparently only impacts code run directly by travis, but xenial is using pyenv to manage python installations so it turned out to be a simple matter of setting the pyenv version and then all the pip commands worked fine.  I left in a PYTHON=python3 b/c it seemed better, though I think it was unnecessary.  

I dislike that it requires pinning pyenv to a specific release (3.7.1).  There is a feature request to support wildcards (3.7*) but the pyenv developers seem unmoved.  Though travis.yml already pins PATH to that for codespell, so it isn't the only thing that needs to be changed.

python/ola/Makefile.mk
```diff
@@ -104,4 +104,5 @@ CLEANFILES += \
     python/ola/ClientWrapperTest.sh \
     python/ola/OlaClientTest.sh \
     python/ola/PidStoreTest.sh \
-    python/ola/RDMTest.sh
+    python/ola/RDMTest.sh \
+    python/ola/__pycache__/*
```
peternewman, line 108: .gitignore too?

Does our other python code in tools/rdm need the same addition too?

python/ola/PidStoreTest.py
```diff
@@ -228,45 +228,53 @@ def testPackRanges(self):
     blob = pid.Pack(args, PidStore.RDM_SET)
     self.assertTrue(len(blob) > 1)
 
+    # invalid year (2002 < 2003)
```
peternewman, line 231: Nice work, I'd meant literally the year 1 but this works fine too!

python/ola/PidStoreTest.py
```diff
@@ -228,45 +228,53 @@ def testPackRanges(self):
...
       blob = pid.Pack(args, PidStore.RDM_SET)
 
+    # invalid month > 12
     with self.assertRaises(PidStore.ArgsValidationError):
       args = ["2020", "13", "20", "20", "20", "20"]
       blob = pid.Pack(args, PidStore.RDM_SET)
 
+    # invalid month > 255
     with self.assertRaises(PidStore.ArgsValidationError):
       args = ["2020", "255", "20", "20", "20", "20"]
```
peternewman, line 248: This is >= 255, do you mean to set the test value to 256?

peternewman (changes requested):
Just a few minor comments

peternewman:
> @peternewman it looks like the failures I see are not related to any changes in here, so I consider this ready.

Agreed.

> Please let me know if you would like any other changes.

Review with a few minor bits on the latest changes. I clearly need to fix all that Codespell stuff!

> Just to state clearly, this changes the linux clang build to use python3. The direct "python" setting in travis apparently only impacts code run directly by travis, but xenial is using pyenv to manage python installations so it turned out to be a simple matter of setting the pyenv version and then all the pip commands worked fine. I left in a PYTHON=python3 b/c it seemed better, though I think it was unnecessary.
> 
> I dislike that it requires pinning pyenv to a specific release (3.7.1). There is a feature request to support wildcards (3.7*) but the pyenv developers seem unmoved. Though travis.yml already pins PATH to that for codespell, so it isn't the only thing that needs to be changed.

Do you think pyenv is a better/more pythonic way to do it than I'm doing with the path variable in Codespell? I think it would probably make sense to be consistent between the two as they both have to run as Python 3 of some version.

python/ola/PidStoreTest.py
brucelowekamp: Heh, guess that looks silly.  I'll set it to 256.

python/ola/Makefile.mk
brucelowekamp: They are just pyc files in there so gitignore already covers them.  

tools/rdm will need a similar change, but it doesn't vaguely run under py3, so not an issue at present, and I think should be left for that change.  I had probably most of those changes done with the earlier pass, but no way to test them.

brucelowekamp:
> > I dislike that it requires pinning pyenv to a specific release (3.7.1). There is a feature request to support wildcards (3.7*) but the pyenv developers seem unmoved. Though travis.yml already pins PATH to that for codespell, so it isn't the only thing that needs to be changed.
> 
> Do you think pyenv is a better/more pythonic way to do it than I'm doing with the path variable in Codespell? I think it would probably make sense to be consistent between the two as they both have to run as Python 3 of some version.

I'm not sure if it's more pythonic, but given that travis/xenial seems to have adopted pyenv to manage their python versions it's probably the better way to go.  I glanced at it and I think it's a trivial change, so pushing it now.

brucelowekamp:
> I'm not sure if it's more pythonic, but given that travis/xenial seems to have adopted pyenv to manage their python versions it's probably the better way to go. I glanced at it and I think it's a trivial change, so pushing it now.

Looks like it worked.  Looking again, I'm a bit unclear if the travis variable PYTHON=python3 exports all the way down to the build commands.  I think it might, in which case that second command I added to set it based on travis env was unnecessary.  But I don't think it hurts anything, either.

python/ola/PidStoreTest.py
peternewman: Done.

python/ola/Makefile.mk
peternewman: I didn't know if we needed to ignore the folder too, or will git ignore it if it's ignoring all the files so it's empty (hence .gitkeep elsewhere)?

Ah I didn't realise tool/rdm was really broken too. We should be able to test against the dummy responders mostly (to prove it runs and hopefully gives correct results). I suspect it can wait until 0.11.0, TBH I generally suggest people run the tests against master for all the latest and greatest anyway.

python/ola/RDMTest.py
peternewman: Perfect. Yeah perhaps do some magic merge/revert or manually copy/paste so we get a NACK test too?

python/ola/PidStoreTest.py
peternewman: Done

python/ola/PidStoreTest.py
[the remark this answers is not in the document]
peternewman: Yeah sounds good. I had to double-check the standard to confirm the minimum year was indeed 2003!

python/ola/RDMTest.py
peternewman: Actually better yet, do a GET on DMX_PERSONALITY_DESCRIPTION 0 which the responder should NACK as out of range, or even better, do a SET on DMX_PERSONALITY 0 as you're not testing that yet!

python/ola/PidStoreTest.py
```diff
@@ -198,6 +198,79 @@ def testCmp(self):
     self.assertNotEqual(hash(p1b), hash(p2))
     self.assertNotEqual(hash(p1a), hash(p3))
 
+  def testPackUnpack(self):
+    store = PidStore.PidStore()
+    store.Load([os.path.join(path, "test_pids.proto")])
+
+    pid = store.GetName("DMX_PERSONALITY_DESCRIPTION")
+
+    # Pid.Pack only packs requests and Pid.Unpack responses
```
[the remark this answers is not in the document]
peternewman: Done.

python/ola/RDMTest.py
peternewman: Done.

python/ola/ClientWrapperTest.py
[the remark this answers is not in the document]
peternewman: I suspect it won't be generally, but you should be able to paste it (or part of it) into http://rdm.openlighting.org/tools/packet-extractor for example to see what's going on I think. Plus it's always nice to have a bit of a sanity check of values; it's not actually just an empty string for example or a single byte.

python/ola/rpc/StreamRpcChannel.py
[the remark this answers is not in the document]
peternewman: Yeah sounds good, looks like it should only be needed to fix the tests and not even then in most cases.

python/Makefile.mk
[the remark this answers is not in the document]
peternewman: Sorry, missed this. I'd probably prefer to have it per folder in the various Makefile.mk's like our other code. We could have it separately in the Makefile like the lint stuff, but this feels a bit more important than that so should probably run against every test run.

Or put it in Makefile.am so it's one test run at the top level.

Will this need the imports to be present to work? In which case it should probably go in the subfolders so the logic for e.g. numpy for the RDM tests is kept in one place? If not putting it in the root and compiling everything seems like a win.

We could possibly even have some logic to compile **everything** if it's running a Python 2 build, but that might be a lot of faff.

python/ola/OlaClient.py
peternewman: Done.

.travis.yml
```diff
@@ -357,8 +358,8 @@ before_install:
 after_failure:
 # Disabled as otherwise the logfile is too big
 #  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
-  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
-  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
+#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
+#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
```
peternewman, line 362: Possibly put these back if you've got the debug you need as configure is obviously likely to fail first, although I guess test-suites fail more often so perhaps leave it.

I guess if we were being clever we'd set a variable when configure succeeded and use that to decide what diagnostic logs to show!

python/ola/PidStoreTest.py
```diff
@@ -198,6 +198,87 @@ def testCmp(self):
...
+
+  def testPackRanges(self):
+    store = PidStore.PidStore()
+    store.Load([os.path.join(path, "test_pids.proto")])
+
+    pid = store.GetName("REAL_TIME_CLOCK")
+
+    args = ["2020", "6", "20", "20", "20", "20"]
+    blob = pid.Pack(args, PidStore.RDM_SET)
+    self.assertTrue(len(blob) > 1)
```
peternewman, line 229: Any reason not to test this blob is exactly the intended length (and possibly test a data match)?

python/ola/PidStoreTest.py
```diff
@@ -198,6 +198,87 @@ def testCmp(self):
...
+    # invalid negative month
+    with self.assertRaises(PidStore.ArgsValidationError):
+      args = ["2020", "-1", "20", "20", "20", "20"]
+      blob = pid.Pack(args, PidStore.RDM_SET)
+
+    # tests for string with min=max=2
+    pid = store.GetName("LANGUAGE_CAPABILITIES")
+    args = ["en"]
+    blob = pid._responses.get(PidStore.RDM_GET).Pack(args)[0]
+    self.assertTrue(len(blob) > 1)
```
peternewman, line 260: Again make the test more thorough?

python/ola/PidStoreTest.py
```diff
@@ -198,6 +198,87 @@ def testCmp(self):
...
+
+    with self.assertRaises(PidStore.ArgsValidationError):
+      args = ["enx"]
+      blob = pid._responses.get(PidStore.RDM_GET).Pack(args)[0]
+
+    # valid empty string
+    pid = store.GetName("STATUS_ID_DESCRIPTION")
+    args = [""]
+    blob = pid._responses.get(PidStore.RDM_GET).Pack(args)[0]
+    decoded = pid.Unpack(blob, PidStore.RDM_GET)
```
peternewman, line 274: Perhaps a length check of the blob here too? Check it's not adding null termination or anything.

python/ola/RDMTest.py
```diff
@@ -0,0 +1,168 @@
...
+    sends fixed response message."""
+    sockets = socket.socketpair()
+    wrapper = ClientWrapper(sockets[0])
+    pid_store = PidStore.GetStore(pidStorePath)
+    client = wrapper.Client()
+    rdm_api = RDMAPI(client, pid_store)
+
+    class results:
+      gotrequest = False
+      gotresponse = False
```
peternewman, line 51: Perhaps switch to this to match the other code:
```suggestion
      got_request = False
      got_response = False
```

Then equivalent matching changes throughout.

python/ola/RDMTest.py
```diff
@@ -0,0 +1,168 @@
...
+
+    sockets[0].close()
+    sockets[1].close()
+
+    self.assertTrue(results.gotrequest)
+    self.assertTrue(results.gotresponse)
+
+
+if __name__ == '__main__':
+  pidStorePath = (os.environ.get('PIDSTOREDIR', "../data/rdm"))
```
peternewman, line 167: Again perhaps snake case for consistency rather than this being the only camel in the file.

peternewman (commented):
Sorry a few minor nits if you've got time?

.travis.yml
brucelowekamp: Completely forgot I changed those....

Yeah, when someone has time...

python/ola/PidStoreTest.py
brucelowekamp: Doesn't hurt to check.  I was thinking of the range validation in that test and mostly threw that in to check that the valid ranges don't throw, but I'll decode it and confirm it is correct.

python/ola/RDMTest.py
brucelowekamp: I've written to so many coding standards over the years that I tend to be entirely inconsistent in naming variables unless I think about it...

python/ola/Makefile.mk
brucelowekamp: git (and distcheck) seem OK with empty directories

python/ola/PidStoreTest.py
brucelowekamp: added the check to make sure it's not there, though the python code, at least, rstrips nulls from strings

.travis.yml
brucelowekamp: Actually, looking at yet another pair of random mac build failures, it occurred to me that just outputting the test-suite logs first would make sure they appear if the build failed in the test suite.  Seems like 99% of the time if it got that far, that's what you would want to see.

python/Makefile.mk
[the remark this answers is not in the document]
brucelowekamp: I tried to split the difference of leaving the script building in python/Makefile.mk but having a variable that assembled directories that should be compiled and are set in the particular directories that need it.  That seems to me like it captures the spirit of having the opt-in be local while maintaining a single copy of the script and not adding test-generation stuff into the root Makefile.am.

python/ola/PidStoreTest.py
peternewman: Thanks, yeah that was just a random example.

.travis.yml
peternewman: Sometimes the simple fixes are the best!

.travis.yml
```diff
@@ -356,12 +356,12 @@ before_install:
...
 
 after_failure:
-# Disabled as otherwise the logfile is too big
-#  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
-#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/config.log; fi
-#  - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/config.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/test-suite.log; fi
   - if [ -f ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log ]; then cat ${TRAVIS_BUILD_DIR}/ola-*/_build/sub/test-suite.log; fi
+# Disabled as otherwise the logfile is too big
+#  - if [ -f ${TRAVIS_BUILD_DIR}/config.log ]; then cat ${TRAVIS_BUILD_DIR}/config.log; fi
```
peternewman, line 362: I think this one can be left uncommented too now, just in case it exists somewhere.

python/ola/PidStoreTest.py
peternewman: Looks good thanks.

peternewman: Actually, thinking about it, comparing to a byte array might be better. If pack ended up reversing the data or doing incorrect network/host order for the larger values, this potentially wouldn't catch it, whereas confirming the packed value matches the expected would. I think we're already testing unpack elsewhere aren't we?

python/ola/RDMTest.py
peternewman: Done.

python/ola/RDMTest.py
```diff
@@ -159,10 +159,66 @@ def ResponseCallback(self, response, data, unpack_exception):
...
+      response = binascii.unhexlify(
+        "2f0000100802100022290800100218002202000628e001300138004a0908f0f"
+        "4011500ffffff520908f0f40115ac107de05831")
+      sent_bytes = sockets[1].send(response)
+      self.assertEqual(sent_bytes, len(response))
+
+    def ResponseCallback(self, response, data, unpack_exception):
+      results.got_response = True
+      self.assertEqual(response.response_type, client.RDM_NACK_REASON)
+      self.assertEqual(response.pid, 0xe0)
```
peternewman, line 203: You should also be able to check the response.nack_reason matches (should be RDMNack.NR_DATA_OUT_OF_RANGE).

Makefile.am
```diff
@@ -173,6 +176,7 @@ include olad/Makefile.mk
 include protoc/Makefile.mk
 include python/Makefile.mk
 include tools/Makefile.mk
+include scripts/Makefile.mk
```
peternewman, line 179: Can you sort these two alphabetically:
```suggestion
include scripts/Makefile.mk
include tools/Makefile.mk
```

.travis.yml
brucelowekamp: doesn't hurt.  I need to push a change to the pycompiletest anyway b/c I forgot -f breaks the mac build

python/Makefile.mk
```diff
@@ -3,11 +3,11 @@ include python/ola/Makefile.mk
 
 python/PyCompileTest.sh: python/Makefile.mk
 	mkdir -p $(top_builddir)/python
-# 	restore this line when py3 compat is done for whole tree
-#	echo "$(PYTHON) -m compileall -f tools scripts python include data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
-	echo "$(PYTHON) -m compileall python data; exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
+	echo "$(PYTHON) -m compileall -f $(PYTHON_BUILD_DIRS); exit \$$?" > $(top_builddir)/python/PyCompileTest.sh
 	chmod +x $(top_builddir)/python/PyCompileTest.sh
```
peternewman: I still wonder if this should be in Makefile.am, but don't feel too strongly either way, the variable is certainly a great improvement.

peternewman (changes requested):
Just a minor improvement of the tests

python/ola/PidStoreTest.py
brucelowekamp: RDMTest.py has the byte-level encoding validation of a pretty large structure with at least one large value.  So I think that's captured and given the problem updating all these tests if the protobuf is ever changed, IMHO it would be better not to check the wire encoding in too many places.

.travis.yml
peternewman: Ah, how so?

python/Makefile.mk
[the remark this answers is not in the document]
peternewman: Okay, fair enough if you think we should keep Makefile.am cleaner.

python/ola/RDMTest.py
peternewman: I didn't think/explain that very well did I. We could have a GET on DMX_PERSONALITY_DESCRIPTION 0 which NACKs and a SET on DMX_PERSONALITY 1 which ACKs to test set too!

python/ola/RDMTest.py
brucelowekamp: annoyingly that isn't exposed nicely as an enum.  Think that can be fixed another time.

.travis.yml
brucelowekamp: -f forces it to recompile all of the python code regardless of timestamp etc.  I think the problem is that with it trying to run the tests in parallel it winds up trying to delete files while they're being used for other tests and that causes permission denied errors.   For a clean build there really isn't ever a reason to use -f, so I think it's fine without it.  On linux I guess the OS is OK with deleting a file and leaving it open by another process, so it doesn't cause a problem.

python/ola/RDMTest.py
brucelowekamp: I think at this point I'm feeling like code coverage is good enough with what's here.  If I were to do more, I think I would run the test suite with coverage and figure out what else could be tested that way.

Makefile.am
peternewman: Done.

python/ola/RDMTest.py
peternewman: Does this not work?: https://github.com/OpenLightingProject/ola/blob/master/tools/rdm/TestDefinitions.py#L1227

.travis.yml
peternewman: Ah yes the parallel build! We still run the java stuff serially for a different reason but I don't think it's worth slowing down all the other python stuff just to fix this, as you say dropping -f seems good enough.

python/ola/RDMTest.py
```diff
@@ -201,6 +202,9 @@ def ResponseCallback(self, response, data, unpack_exception):
       results.got_response = True
       self.assertEqual(response.response_type, client.RDM_NACK_REASON)
       self.assertEqual(response.pid, 0xe0)
+      self.assertEqual(response.nack_reason,
+                       RDMNack.LookupCode(RDMNack.NACK_SYMBOLS_TO_VALUES
+                                          ['NR_DATA_OUT_OF_RANGE'][0]))
```
peternewman, line 207:
```suggestion
                       RDMNack.NR_DATA_OUT_OF_RANGE)
```

python/ola/PidStoreTest.py
peternewman: Done.

python/ola/PidStoreTest.py
peternewman: pid.Pack should be the RDM wire encoding (which is then put in the Protobuf):
https://github.com/OpenLightingProject/ola/blob/0028d658d6e58ec5158c017ae676f30103d75bbc/python/ola/PidStore.py#L210-L219

So if it changes we're rewriting all of OLA's RDM implementation!

See the C++ equivalent here for example: https://github.com/OpenLightingProject/ola/blob/0028d658d6e58ec5158c017ae676f30103d75bbc/common/rdm/MessageDeserializerTest.cpp

python/ola/RDMTest.py
peternewman: Fair enough.

python/ola/RDMTest.py
brucelowekamp: I entirely missed the setattrs part.  To make myself feel better, I'll add a doc comment about it.

python/ola/PidStoreTest.py
brucelowekamp: My objection is not so much how or whether it might change, but that I think all of the wire/encoding tests of outbound get and set requests with and without parameters as well as inbound responses with data and nacks are fairly thoroughly testing in RDMTest.py, and adding them additionally here where the intention was to just check that the ranges are being enforced correctly isn't necessary and only confuses things.  But it doesn't hurt a whole lot and since I wanted to push the other change, I'll check the blob.

python/ola/RDMTest.py
peternewman: Perfect thanks; brings a new meaning to self-documenting code! :laughing:

python/ola/PidStoreTest.py
peternewman: Ah okay, I assumed you were testing the RDM serialisation in PidStore and the more general Protobuf stuff in RDMTest. Thanks for adding it anyway.

brucelowekamp:
@peternewman anything else pending?  A bit hard to follow but I think all the comments are addressed

peternewman:
> @peternewman anything else pending? A bit hard to follow but I think all the comments are addressed

Yeah I think so too. I'm going to do a final sanity check of the code and comments later and merge. Sorry got distracted earlier so didn't get through it all.

peternewman (approved):
Thanks again for this @brucelowekamp and @sddbs and everyone else who's worked on it!

Issue #1523, opened 2018-12-21
wrong output mode (UART-DMX, but affects virtually elsewhere)


rbalykov:
On startup UART sender has empty DMX frame.
This causes UART sequence BREAK-MAB-STARTCODE repeated without mandatory 24-byte payload.
During normal use, it seems to be full 512-byte frame, set to 0x00.

peternewman:
Hi @rbalykov ,

Well spotted. I think it potentially can continue after start-up depending on how your frame is set, if you use ola_set_dmx, you can populate frame with fewer than 512 slots, which I think will therefore continue to show the problem:
http://docs.openlighting.org/ola/man/man1/ola_set_dmx.1.html

The relevant code is here:
https://github.com/OpenLightingProject/ola/blob/0.10/plugins/uartdmx/UartWidget.cpp#L119-L134

Firstly do you fancy having a go at fixing it? I think it should just be a case of adding a memset of zero if the length returned from the DMX buffer is less than 24, then incrementing the length to write.

If not, are you at least in a position to test it? I've not used the UART plugin, so I'd personally prefer to avoid jumping through all those hoops just to see if it works.

Then we can review which other plugins need fixing too.

rbalykov:
[...]

Issue #280, opened 2013-12-24
Feature list for Patcher v2


ola-importer:
_From [c...@rstenpresser.de](https://code.google.com/u/118222577951750466038/) on November 28, 2013 08:11:17_

This might be usefull since OLA now supports a few Dimmerpacks directly.
When you want to address multiple Dimmers you have to create one Universe for each device.
With a mapper one could copy channels beween Universes. This would be really usefull for that usecase.

We would need to discuss how to set up a map-file. And how to avoid loops in the mapping.

A possible syntax would be 'universe-01.map'
1:u3c3 #reads: fill channel1-data with data from universe3 channel3
2:u3c2
3:u3c1
4:u3c7
5:u3c5
6:u3c4
...

(this would copy data from universe3 to universe1 and reverse the color-order from RGB to BGR).

_Original issue: http://code.google.com/p/open-lighting/issues/detail?id=279_

peternewman:
Swap RGB channels around in a fixture/LED strip:
https://groups.google.com/forum/#!topic/open-lighting/qb-FTXpRYHI

peternewman:
> This might be usefull since OLA now supports a few Dimmerpacks directly. When you want to address multiple Dimmers you have to create one Universe for each device. With a mapper one could copy channels beween Universes. This would be really usefull for that usecase.

I've implemented the main features of this mapper now in here #1895 (as a separate programme for now). I suspect that might be the way to go with most of these for now, they can be written, prototyped and tested, pending some clever integration into the core in the future.

Displaying configuring as a graph/timeline will be a bit challenging. Presenting the config for some of these facilities in an intuitive and user-friendly way will be a lot harder sadly.

PR #1656, opened 2020-08-02
Added '-padding' option to uartdmx plugin.

rbalykov:
Sets minimal slots count in UART frame. Default is 0, will behave as before.

### DIFF
--- a/plugins/uartdmx/UartDmxDevice.h
+++ b/plugins/uartdmx/UartDmxDevice.h
@@ -52,6 +52,7 @@ class UartDmxDevice : public Device {
   // Per device options
   std::string DeviceBreakKey() const;
   std::string DeviceMalfKey() const;
+  std::string DevicePaddingKey() const;
   void SetDefaults();
 
   std::auto_ptr<UartWidget> m_widget;
@@ -60,11 +61,14 @@ class UartDmxDevice : public Device {
   const std::string m_path;
   unsigned int m_breakt;
   unsigned int m_malft;
+  unsigned int m_paddingt;
 
   static const unsigned int DEFAULT_MALF;
   static const char K_MALF[];
   static const unsigned int DEFAULT_BREAK;
   static const char K_BREAK[];
+  static const unsigned int DEFAULT_PADDING;
+  static const char K_PADDING[];
 
   DISALLOW_COPY_AND_ASSIGN(UartDmxDevice);
 };
--- a/plugins/uartdmx/UartWidget.cpp
+++ b/plugins/uartdmx/UartWidget.cpp
@@ -46,6 +46,7 @@
 #include "ola/io/IOUtils.h"
 #include "ola/Logging.h"
 #include "plugins/uartdmx/UartWidget.h"
+#include "plugins/uartdmx/UartDmxDevice.h"
 
 namespace ola {
 namespace plugin {
@@ -54,8 +55,9 @@ namespace uartdmx {
 using std::string;
 using std::vector;
 
-UartWidget::UartWidget(const string& path)
+UartWidget::UartWidget(const std::string &path, unsigned int padding)
     : m_path(path),
+			m_padding (padding),
       m_fd(NOT_OPEN) {
 }
 
@@ -122,6 +124,10 @@ bool UartWidget::Write(const ola::DmxBuffer& data) {
   buffer[0] = DMX512_START_CODE;
 
   data.Get(buffer + 1, &length);
+  if (length < m_padding) {
+  	  memset ((buffer + 1 + length), 0x00, (m_padding - length) );
+  	  length = m_padding;
+    }
 
   if (write(m_fd, buffer, length + 1) <= 0) {
     // TODO(richardash1981): handle errors better as per the test code,
--- a/plugins/uartdmx/UartWidget.h
+++ b/plugins/uartdmx/UartWidget.h
@@ -49,8 +49,7 @@ class UartWidget {
      * Construct a new UartWidget instance for one widget.
      * @param path The device file path of the serial port
      */
-    explicit UartWidget(const std::string &path);
-
+    explicit UartWidget(const std::string &path, unsigned int padding);
     /** Destructor */
     virtual ~UartWidget();
 
@@ -81,6 +80,7 @@ class UartWidget {
 
  private:
   const std::string m_path;
+  unsigned int m_padding;
 
   /**
    * variable to hold the Unix file descriptor used to open and manipulate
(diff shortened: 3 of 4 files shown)


rbalykov:
1) FAIL: common/network/HealthCheckedConnectionTester - do I need to fix it some way?
2) Missing from tarball plugins/uartdmx/README.md - i've checked it out from "master", edited and commited to "0.10". Need smth else?

peternewman:
>     1. FAIL: common/network/HealthCheckedConnectionTester - do I need to fix it some way?

No, some of the tests area bit flaky on Mac, you can ignore this.

>     2. Missing from tarball plugins/uartdmx/README.md - i've checked it out from "master", edited and commited to "0.10". Need smth else?

On 0.10 the plugin description is in the code:
https://github.com/OpenLightingProject/ola/blob/7892723801bc64c62dba0079ffe25d2f7a09c019/plugins/uartdmx/UartDmxPlugin.cpp#L114-L120

However I'd suggest if this code was to go in, that it should go in master, as it's a feature not a bug fix.

Can I ask the broader question, why? The web interface sends a full frame, any API can do so too, what's the reasoning for adding this to this plugin rather than dealing with it at source? I think the Enttec Pro's behave the same way and probably the Open's too. This would also be a perfect candidate for Patcher v2 #280.

rbalykov:
> > 2. Missing from tarball plugins/uartdmx/README.md - i've checked it out from "master", edited and commited to "0.10". Need smth else?
> On 0.10 the plugin description is in the code:

So, keep README or edit in-code descriptption?
Plugin is still "Native UART", nothing has changed.

> However I'd suggest if this code was to go in, that it should go in master, as it's a feature not a bug fix.
> Can I ask the broader question, why?

This code fixes #1523 , I got some "hardware madness" with zero-lenght frames. 
Sure, it could be fixed with ola_set_dmx, but this still gives some "broken" frames on boot.

peternewman:
Sorry this slipped through the net a bit.

> > > 1. Missing from tarball plugins/uartdmx/README.md - i've checked it out from "master", edited and commited to "0.10". Need smth else?
> > >    On 0.10 the plugin description is in the code:
> 
> So, keep README or edit in-code descriptption?
> Plugin is still "Native UART", nothing has changed.

Edit the in-code description if targetting 0.10 or the README if targetting master.

> > However I'd suggest if this code was to go in, that it should go in master, as it's a feature not a bug fix.
> > Can I ask the broader question, why?
> 
> This code fixes #1523 , I got some "hardware madness" with zero-lenght frames.

Ah okay, makes sense.

> Sure, it could be fixed with ola_set_dmx, but this still gives some "broken" frames on boot.

Yeah I think we're agreeing, fixing in the plugin is the most sense.

> Had another PR #1573 with same issue before, but was unfamiliar with github, so mixed two PRs in one commit. This one gives just "-padding" relates code.

Great thanks.

plugins/uartdmx/README.md
```diff
@@ -0,0 +1,34 @@
+Native UART DMX Plugin
```
peternewman, line 1: As discussed, please remove this file and edit PluginDescription if targeting 0.10.

plugins/uartdmx/UartDmxDevice.h
```diff
@@ -60,11 +61,14 @@ class UartDmxDevice : public Device {
   const std::string m_path;
   unsigned int m_breakt;
   unsigned int m_malft;
+  unsigned int m_paddingt;
```
peternewman, line 64: I think the t's were timing, e..g break time, mark after last frame time, so this should probably just be m_padding.

plugins/uartdmx/UartWidget.cpp
```diff
@@ -122,6 +124,10 @@ bool UartWidget::Write(const ola::DmxBuffer& data) {
   buffer[0] = DMX512_START_CODE;
 
   data.Get(buffer + 1, &length);
+  if (length < m_padding) {
+      memset((buffer + 1 + length), 0x00, (m_padding - length) );
+      length = m_padding;
+    }
```
peternewman, line 130: Given as you've said this is probably required in a few other plugins, I wonder if it makes sense to add some extra options to DmxBuffer? e.g. GetWithPadding(uint8_t *data, unsigned int *length, unsigned int min_length). Or alternatively a Pad/Grow function which modifies the buffer in place (to save duplicating lots of functions), although the latter would need to be called on the lower level buffer in UartThread as this one is const so maybe GetWithPadding is simpler.

plugins/uartdmx/UartWidget.h
```diff
@@ -48,9 +48,9 @@ class UartWidget {
     /**
      * Construct a new UartWidget instance for one widget.
      * @param path The device file path of the serial port
+     * @param padding Minimum DMX frame slots count, padded with NULLs
```
peternewman, line 51: Possibly semantics, but I'd say they are zeros here, as they will be actual channel data.

plugins/uartdmx/UartWidget.cpp
```diff
@@ -54,8 +55,9 @@ namespace uartdmx {
 using std::string;
 using std::vector;
 
-UartWidget::UartWidget(const string& path)
+UartWidget::UartWidget(const std::string &path, unsigned int padding)
```
peternewman, line 58: The std:: isn't needed here as we're `using std::string`

plugins/uartdmx/UartWidget.cpp
```diff
@@ -46,6 +46,7 @@
 #include "ola/io/IOUtils.h"
 #include "ola/Logging.h"
 #include "plugins/uartdmx/UartWidget.h"
+#include "plugins/uartdmx/UartDmxDevice.h"
```
peternewman, line 49: I don't think we need this include do we?

plugins/uartdmx/README.md
```diff
@@ -0,0 +1,34 @@
...
+### Per Device Settings (using above device name)
+
+`<device>-break = 100`
+The DMX break time in microseconds for this device (optional).
+
+`<device>-malf = 100`
+The Mark After Last Frame time in microseconds for this device (optional).
+
+`<device>-padding = 0` Minimal slots count in DMX frame (optional). 
+Default is 0, which allows empty frame. Using minimal timing, frame period reaches 140 microseconds, while standard requires minimum 1204us. Using BREAK=88us and MAB=8us, frame has to include 25 slots, that gives frame period 1240us.
```
peternewman, line 34: If you're feeling particularly keen, it would be really nice on initialisation to print an OLA_INFO with the minimum padding value (based on break/MAB/MALF), or even an OLA_WARN if the configured padding is insufficient!

peternewman (changes requested):
Sorry this got missed before @rbalykov , thanks for your contribution, a few comments if you don't mind.

PR #1659, opened 2020-08-14
Add Nortle as a FDTI adaptor manufacturer

arneman44:


arneman44:
Im selling openscource FDTI adaptors under the name NORTLE i would like these to be compatible with OLA

peternewman:
So these are the same as the OpenDMX USB, no microprocessor onboard, effectively just a re-titled FTDI?

plugins/ftdidmx/FtdiWidget.cpp
```diff
@@ -192,7 +192,8 @@ void FtdiWidget::Widgets(vector<FtdiWidgetInfo> *widgets) {
         if (std::string::npos != v.find("FTDI") ||
             std::string::npos != v.find("KMTRONIC") ||
             std::string::npos != v.find("KWMATIK") ||
-            std::string::npos != v.find("WWW.SOH.CZ")) {
+            std::string::npos != v.find("WWW.SOH.CZ") ||
+            std::string::npos != v.find("NORTLE")) {
```
peternewman, line 196: Can you sort the order of the manufacturers please.

arneman44: what do you mean by sort manufacturers

arneman44:
yes these are the same as the opendmxusb but are available with up to a 4 port version 

it can definitely be moved to the 0.10 branch

plugins/ftdidmx/FtdiWidget.cpp
peternewman: As in WWW.SOH.CZ > NORTLE, the string comparisons.

peternewman:
> it can definitely be moved to the 0.10 branch

Great, you'll need to cherry-pick or rebase or something.

peternewman:
Have you had any more time to look at this @arneman44 ?

Issue #779, opened 2015-06-12
Replace gettimeofday() with clock_gettime() where possible


nomis52:
See https://blog.habets.se/2010/09/gettimeofday-should-never-be-used-to-measure-time

and 

http://stackoverflow.com/questions/11680461/monotonic-clock-on-osx

stefansiegfried:
Hi there, found your project through Hacktoberfest. Could I take a look at this issue?

peternewman:
Yes @stefansiegfried , please feel free!

It looks like we use our wrapper function CurrentTime in quite a lot of places, a handful just to see the wall-clock time, but mostly to see how much time has elapsed (i.e. the naughty version).

If you've not found it yet, our Doxygen docs will probably be useful for this:
http://docs.openlighting.org/ola/doc/latest/

I suspect as a start, and realising this actually looks like it could be a simple, but quite involved PR (there's lots of places it's used in the code) adding some stuff to detect when the monotonic clock is available and then probably having a CurrentMonotonicTime or something function which falls back to CurrentTime, then tweaking http://docs.openlighting.org/ola/doc/latest/classola_1_1_time_interval.html to do some offset calculations could make sense, but I'm welcome to other ideas too.

Please get in touch if you've got any more comments.

stefansiegfried:
Thanks, @peternewman. I was able to compile the project's 'make all' target, however I have run into some trouble with 'make check' unit tests and g++ errors (-Werror=deprecated-copy).  Where is the best place troubleshoot that? Should I open a separate ticket, or can we discuss that here?

peternewman:
Ah, sorry to hear that. I'd suggest raising a new issue for that is probably best, as it may affect others too.

When you do so, can you let us know OS, version etc. In the meantime, adding `--disable-fatal-warnings` to configure may allow you to continue.

stefansiegfried:
After doing some reading, I do have some questions on implementation:

1. Is there any concern with CLOCK_MONOTONIC not advancing during suspend?

2. Is the suggestion to replace all "naughty" calls to CurrentTime with CurrentMonotonicTime (which then falls back to CurrentTime)?

3. The clock_gettime function accepts a struct timespec as a parameter versus gettimeofday which takes a struct timeval.  Would you suggest modifying BaseTimeVal/TimeStamp/TimeInterval to also handle struct timespec? 

Or, replace gettimeofday with clock_gettime(CLOCK_REALTIME) and just use timespec _instead_ of timeval?
[...]

Issue #1619, opened 2020-03-12
Makefile flake8 target


peternewman:
Add a flake8 Makefile target since it's available via pypi.

_Originally posted by @brucelowekamp in https://github.com/OpenLightingProject/ola/pull/1605/files_

Sounds good although I suspect the flake8 thing should be part of the lint target rather than check, as per https://github.com/OpenLightingProject/ola/issues/691 because it's nice to encourage people to run make check, but we don't want to force them to have flake8 installed. Although for bonus points configure could handle it all seamlessly. Although I suspect it should still be a separate target, I often write rather messy code while developing, but still running make check, which I wouldn't want to fail on that, before tidying up ready for committing.

_Originally posted by @peternewman in https://github.com/OpenLightingProject/ola/pull/1605/files_

See also #691

peternewman:
This has likely got a lot easier now we've switched to a `.flake8` config file in #1637

stefansiegfried:
Could I take a look at this issue?

peternewman:
Yeah certainly, I've assigned you too.

Issue #1564, opened 2019-06-30
error: explicitly assigning value of variable of type 'ola::DmxBuffer' to itself


peternewman:
```
common/utils/DmxBufferTest.cpp:197:10: error: explicitly assigning value of
      variable of type 'ola::DmxBuffer' to itself
      [-Werror,-Wself-assign-overloaded]
  buffer = buffer;
  ~~~~~~ ^ ~~~~~~
1 error generated.
```

Is the error when I am up to date with origin/master (and for this branch as well)

_Originally posted by @kecramer in https://github.com/OpenLightingProject/ola/pull/1561#issuecomment-502122886_

janosvitok:
https://github.com/OpenLightingProject/ola/blob/master/common/utils/DmxBufferTest.cpp#L197
I guess this is intentional check whether assignment operation survives self-assignment.

However I miss there any explicit checks for consistency after self assignment. Only implicit ones are present (i.e. the code does not crash).

peternewman:
> https://github.com/OpenLightingProject/ola/blob/master/common/utils/DmxBufferTest.cpp#L197
> I guess this is intentional check whether assignment operation survives self-assignment.

I think that's the idea based on the comment @janosvitok .

> However I miss there any explicit checks for consistency after self assignment. Only implicit ones are present (i.e. the code does not crash).

Do you mean checking the contents of buffer is still equal to what it was initialised to? Do you fancy trying to add some further tests around this?

I think to workaround the warning/error we might need to add this flag to the relevant bit of test code (but ideally just that bit, not everywhere), although I think when I looked before, GCC complains if passed this flag (or LLVM if it was GCC throwing it):
-Wno-self-assign-overloaded

peternewman:
> I think to workaround the warning/error we might need to add this flag to the relevant bit of test code (but ideally just that bit, not everywhere), although I think when I looked before, GCC complains if passed this flag (or LLVM if it was GCC throwing it):
> -Wno-self-assign-overloaded

We may be able to workaround this problem using the push/pop idea @yoe is using here:
https://github.com/OpenLightingProject/ola/pull/1583/commits/67234e2cd617b6330c721f003cfc5edd9afa9d28

peternewman:
@kecramer or anyone else, which compiler does this happen on, I think it should be a fairly simple fix but it would be nice to test it.

peternewman:
[...]

Issue #1547, opened 2019-03-19
Make it easier for people to write web apps linked to OLA


peternewman:
As requested in here:
https://groups.google.com/forum/#!topic/open-lighting/sYXZQS9JqwE

Possible options:
- [ ] Implement Cross-Origin Resource Sharing (CORS) 
- [ ] Add the ability for people to load their own files to be served by our webserver (probably don't allow wildcards). Ensure we don't allow path traversal

PR #1673, opened 2020-10-22
Use clock_gettime's CLOCK_MONOTONIC if available

stefansiegfried:
* Fixes OpenLightingProject/ola#779

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,7 +3,8 @@ x/y/2020 ola-0.10.8
  * Support DragonFly BSD
 
  API:
- * 
+ * Deprecated Clock::CurrentTime and added Clock::CurrentMonotonicTime and 
+   Clock::CurrentRealTime
 
  RDM Tests:
  * Fix a bug in the RDM tests to ensure they can cope with 255 personalities
--- a/NEWS
+++ b/NEWS
@@ -4,7 +4,8 @@ x/y/2020 ola-0.10.8
 
  API:
  * Deprecated Clock::CurrentTime and added Clock::CurrentMonotonicTime and 
-   Clock::CurrentRealTime
+   Clock::CurrentRealTime. Occurrences of Clock::CurrentTime have been
+   replaced with Clock::CurrentRealTime where applicable.
 
  RDM Tests:
  * Fix a bug in the RDM tests to ensure they can cope with 255 personalities
--- a/NEWS
+++ b/NEWS
@@ -5,7 +5,7 @@ x/y/2020 ola-0.10.8
  API:
  * Deprecated Clock::CurrentTime and added Clock::CurrentMonotonicTime and 
    Clock::CurrentRealTime. Occurrences of Clock::CurrentTime have been
-   replaced with Clock::CurrentRealTime where applicable.
+   replaced with either variant as applicable.
 
  RDM Tests:
  * Fix a bug in the RDM tests to ensure they can cope with 255 personalities
--- a/common/base/Init.cpp
+++ b/common/base/Init.cpp
@@ -393,16 +393,16 @@ void Daemonise() {
 }
 
 void ClockInit() {
-Clock clock;
-TimeStamp now_monotonic;
-TimeStamp now_realtime;
+  Clock clock;
+  TimeStamp now_monotonic;
+  TimeStamp now_realtime;
 #ifndef CLOCK_MONOTONIC
   OLA_DEBUG << "Monotonic clock unavailable. Falling back to real time clock."
 #endif
-clock.CurrentMonotonicTime(&now_monotonic);
-clock.CurrentRealTime(&now_realtime);
-OLA_DEBUG << "Monotonic clock: " << std::setw(18) << now_monotonic;
-OLA_DEBUG << "Real clock     : " << std::setw(18) << now_realtime;
+  clock.CurrentMonotonicTime(&now_monotonic);
+  clock.CurrentRealTime(&now_realtime);
+  OLA_DEBUG << "Monotonic clock: " << std::setw(18) << now_monotonic;
+  OLA_DEBUG << "Real clock     : " << std::setw(18) << now_realtime;
 }
 /**@}*/
 }  // namespace ola
--- a/common/base/Init.cpp
+++ b/common/base/Init.cpp
@@ -50,6 +50,7 @@
 #endif  // _WIN32
 #include <unistd.h>
 
+#include <ola/Clock.h>
 #include <ola/ExportMap.h>
 #include <ola/Logging.h>
 #include <ola/StringUtils.h>
@@ -60,6 +61,7 @@
 #include <ola/thread/Utils.h>
 #include <ola/system/Limits.h>
 
+#include <iomanip>
 #include <iostream>
 #include <string>
 
@@ -186,6 +188,7 @@ using std::endl;
 using std::string;
 
 bool ServerInit(int argc, char *argv[], ExportMap *export_map) {
+  ClockInit();
   ola::math::InitRandom();
   if (!InstallSEGVHandler()) {
     return false;
@@ -219,6 +222,7 @@ bool AppInit(int *argc,
              char *argv[],
              const string &first_line,
              const string &description) {
+  ClockInit();
   ola::math::InitRandom();
   SetHelpString(first_line, description);
   ParseFlags(argc, argv);
@@ -386,6 +390,19 @@ void Daemonise() {
     exit(EXIT_OSERR);
   }
 #endif  // _WIN32
+}
+
+void ClockInit() {
+Clock clock;
+TimeStamp now_monotonic;
+TimeStamp now_realtime;
+#ifndef CLOCK_MONOTONIC
+  OLA_DEBUG << "Monotonic clock unavailable. Falling back to real time clock."
+#endif
+clock.CurrentMonotonicTime(&now_monotonic);
+clock.CurrentRealTime(&now_realtime);
+OLA_DEBUG << "Monotonic clock: " << std::setw(18) << now_monotonic;
+OLA_DEBUG << "Real clock     : " << std::setw(18) << now_realtime;
 }
 /**@}*/
 }  // namespace ola
--- a/common/math/Random.cpp
+++ b/common/math/Random.cpp
@@ -44,6 +44,8 @@ std::default_random_engine generator_;
 void InitRandom() {
   Clock clock;
   TimeStamp now;
+  // The clock type here should not matter because only the microseconds field
+  // is being used to seed the random number generator.
   clock.CurrentRealTime(&now);
 
   uint64_t seed = (static_cast<uint64_t>(now.MicroSeconds()) << 32) +
--- a/common/math/Random.cpp
+++ b/common/math/Random.cpp
@@ -44,7 +44,7 @@ std::default_random_engine generator_;
 void InitRandom() {
   Clock clock;
   TimeStamp now;
-  clock.CurrentMonotonicTime(&now);
+  clock.CurrentRealTime(&now);
 
   uint64_t seed = (static_cast<uint64_t>(now.MicroSeconds()) << 32) +
                    static_cast<uint64_t>(getpid());
--- a/common/math/Random.cpp
+++ b/common/math/Random.cpp
@@ -44,7 +44,7 @@ std::default_random_engine generator_;
 void InitRandom() {
   Clock clock;
   TimeStamp now;
-  clock.CurrentTime(&now);
+  clock.CurrentMonotonicTime(&now);
 
   uint64_t seed = (static_cast<uint64_t>(now.MicroSeconds()) << 32) +
                    static_cast<uint64_t>(getpid());
--- a/common/thread/PeriodicThread.cpp
+++ b/common/thread/PeriodicThread.cpp
@@ -53,6 +53,8 @@ void *PeriodicThread::Run() {
   Clock clock;
   TimeStamp last_run_at;
 
+  // Real time is used here because it gets passed into pthread_cond_timedwait
+  // which expects an absolute time
   clock.CurrentRealTime(&last_run_at);
   if (!m_callback->Run()) {
     return NULL;
--- a/common/thread/PeriodicThread.cpp
+++ b/common/thread/PeriodicThread.cpp
@@ -53,7 +53,7 @@ void *PeriodicThread::Run() {
   Clock clock;
   TimeStamp last_run_at;
 
-  clock.CurrentTime(&last_run_at);
+  clock.CurrentMonotonicTime(&last_run_at);
   if (!m_callback->Run()) {
     return NULL;
   }
@@ -72,7 +72,7 @@ void *PeriodicThread::Run() {
         continue;
       }
     }
-    clock.CurrentTime(&last_run_at);
+    clock.CurrentMonotonicTime(&last_run_at);
     if (!m_callback->Run()) {
       return NULL;
     }
--- a/common/thread/PeriodicThread.cpp
+++ b/common/thread/PeriodicThread.cpp
@@ -74,7 +74,7 @@ void *PeriodicThread::Run() {
         continue;
       }
     }
-    clock.CurrentMonotonicTime(&last_run_at);
+    clock.CurrentRealTime(&last_run_at);
     if (!m_callback->Run()) {
       return NULL;
     }
--- a/common/thread/PeriodicThread.cpp
+++ b/common/thread/PeriodicThread.cpp
@@ -53,7 +53,7 @@ void *PeriodicThread::Run() {
   Clock clock;
   TimeStamp last_run_at;
 
-  clock.CurrentMonotonicTime(&last_run_at);
+  clock.CurrentRealTime(&last_run_at);
   if (!m_callback->Run()) {
     return NULL;
   }
--- a/common/thread/PeriodicThread.cpp
+++ b/common/thread/PeriodicThread.cpp
@@ -74,6 +74,8 @@ void *PeriodicThread::Run() {
         continue;
       }
     }
+    // Real time is used here again to maintain clock consistency with the
+    // previous call to CurrentRealTime()
     clock.CurrentRealTime(&last_run_at);
     if (!m_callback->Run()) {
       return NULL;
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -62,8 +62,7 @@ BaseTimeVal& BaseTimeVal::operator=(const struct timeval &tv) {
 }
 
 BaseTimeVal& BaseTimeVal::operator=(const struct timespec &ts) {
-  m_tv.tv_sec = ts.tv_sec;
-  m_tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+  Set(ts);
   return *this;
 }
 
@@ -182,6 +181,11 @@ void BaseTimeVal::Set(int64_t interval_useconds) {
 #endif  // HAVE_SUSECONDS_T
 }
 
+void BaseTimeVal::Set(const struct timespec& ts) {
+  m_tv.tv_sec = ts.tv_sec;
+  m_tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+}
+
 TimeInterval& TimeInterval::operator=(const TimeInterval& other) {
   if (this != &other) {
     m_interval = other.m_interval;
@@ -294,12 +298,9 @@ void MockClock::AdvanceTime(int32_t sec, int32_t usec) {
 
 void MockClock::CurrentMonotonicTime(TimeStamp* timestamp) const {
 #ifdef CLOCK_MONOTONIC
-  struct timeval tv;
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
-  tv.tv_sec = ts.tv_sec;
-  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
-  *timestamp = tv;
+  *timestamp = ts;
   *timestamp += m_offset;
 #else
   OLA_DEBUG << "Monotonic clock unavailable. Falling back to CurrentRealTime.";
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -61,6 +61,12 @@ BaseTimeVal& BaseTimeVal::operator=(const struct timeval &tv) {
   return *this;
 }
 
+BaseTimeVal& BaseTimeVal::operator=(const struct timespec &ts) {
+  m_tv.tv_sec = ts.tv_sec;
+  m_tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+  return *this;
+}
+
 bool BaseTimeVal::operator==(const BaseTimeVal &other) const {
   return timercmp(&m_tv, &other.m_tv, ==);
 }
@@ -230,6 +236,11 @@ TimeStamp& TimeStamp::operator=(const struct timeval &tv) {
   return *this;
 }
 
+TimeStamp& TimeStamp::operator=(const struct timespec &ts) {
+  m_tv = ts;
+  return *this;
+}
+
 TimeStamp &TimeStamp::operator+=(const TimeInterval &interval) {
   m_tv += interval.m_interval;
   return *this;
@@ -254,12 +265,9 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
 
 void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
 #ifdef CLOCK_MONOTONIC
-  struct timeval tv;
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
-  tv.tv_sec = ts.tv_sec;
-  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
-  *timestamp = tv;
+  *timestamp = ts;
 #else
   CurrentRealTime(timestamp);
 #endif
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -261,7 +261,6 @@ void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
   tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
   *timestamp = tv;
 #else
-  OLA_DEBUG << "Monotonic clock unavailable. Falling back to CurrentRealTime.";
   CurrentRealTime(timestamp);
 #endif
 }
@@ -285,10 +284,29 @@ void MockClock::AdvanceTime(int32_t sec, int32_t usec) {
   m_offset += interval;
 }
 
-void MockClock::CurrentMonotonicTime(TimeStamp *timestamp) const {
+void MockClock::CurrentMonotonicTime(TimeStamp* timestamp) const {
+#ifdef CLOCK_MONOTONIC
+  struct timeval tv;
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  tv.tv_sec = ts.tv_sec;
+  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+  *timestamp = tv;
+  *timestamp += m_offset;
+#else
+  OLA_DEBUG << "Monotonic clock unavailable. Falling back to CurrentRealTime.";
+  CurrentRealTime(timestamp);
+#endif
+}
+
+void MockClock::CurrentRealTime(TimeStamp* timestamp) const {
   struct timeval tv;
   gettimeofday(&tv, NULL);
   *timestamp = tv;
   *timestamp += m_offset;
+}
+
+void MockClock::CurrentTime(TimeStamp* timestamp) const {
+  CurrentMonotonicTime(timestamp);
 }
 }  // namespace ola
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -27,6 +27,7 @@
  */
 
 #include <ola/Clock.h>
+#include <ola/Logging.h>
 #include <stdint.h>
 #include <sys/time.h>
 
@@ -260,6 +261,7 @@ void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
   tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
   *timestamp = tv;
 #else
+  OLA_DEBUG << "Monotonic clock unavailable. Falling back to CurrentRealTime.";
   CurrentRealTime(timestamp);
 #endif
 }
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -253,6 +253,19 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
 
 void Clock::CurrentTime(TimeStamp *timestamp) const {
   struct timeval tv;
+#ifdef CLOCK_MONOTONIC
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  tv.tv_sec = ts.tv_sec;
+  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+#else
+  gettimeofday(&tv, NULL);
+#endif
+  *timestamp = tv;
+}
+
+void Clock::CurrentRealTime(TimeStamp *timestamp) const {
+  struct timeval tv;
   gettimeofday(&tv, NULL);
   *timestamp = tv;
 }
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -280,7 +280,7 @@ void Clock::CurrentRealTime(TimeStamp *timestamp) const {
 }
 
 void Clock::CurrentTime(TimeStamp *timestamp) const {
-  CurrentMonotonicTime(timestamp);
+  CurrentRealTime(timestamp);
 }
 
 void MockClock::AdvanceTime(const TimeInterval &interval) {
--- a/common/utils/Clock.cpp
+++ b/common/utils/Clock.cpp
@@ -251,7 +251,7 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
   return TimeStamp(m_tv - interval.m_interval);
 }
 
-void Clock::CurrentTime(TimeStamp *timestamp) const {
+void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
   struct timeval tv;
 #ifdef CLOCK_MONOTONIC
   struct timespec ts;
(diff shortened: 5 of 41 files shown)


common/utils/Clock.cpp
```diff
@@ -253,6 +253,19 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
 
 void Clock::CurrentTime(TimeStamp *timestamp) const {
   struct timeval tv;
+#ifdef CLOCK_MONOTONIC
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  tv.tv_sec = ts.tv_sec;
+  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+#else
+  gettimeofday(&tv, NULL);
```
peternewman, line 262: I think it would be good to call the other wrapper function here (i.e. currently CurrentRealTime) rather than calling gettimeofday directly, in case we need to add other wrapping or ifdefs round it in future.

include/ola/Clock.h
```diff
@@ -222,7 +222,11 @@ class Clock {
  public:
   Clock() {}
   virtual ~Clock() {}
+  // Get the current monotonic clock if available,
+  // otherwise falls back to wall clock time.
```
peternewman, line 226: Can you define this as Doxygen, like the class comment.

include/ola/Clock.h
```diff
@@ -222,7 +222,11 @@ class Clock {
...
   virtual void CurrentTime(TimeStamp *timestamp) const;
+  // Get wall clock time.
```
peternewman, line 228: Again Doxygen please.

include/ola/Clock.h
```diff
@@ -222,7 +222,11 @@ class Clock {
...
+  virtual void CurrentRealTime(TimeStamp *timestamp) const;
```
peternewman, line 229: I wonder if CurrentWallClockTime might be a bit less ambiguous, did there seem to be any consistency online when you were looking for stuff?

peternewman (changes requested):
Thanks for this, looks like some great work!

You can ignore the failing Mac test on Travis, some of the Mac tests are a bit intermittent so will work one time and fail another.

Can we have a test of the new function please.

Also I know this is possibly different to what I said before, but thinking about it a bit more I suspect the safest thing to do both in terms of backwards compatibility and future confusion might be:

1. Rename CurrentTime to CurrentMonotonicTime (or similar)
2. Make a new method CurrentTime which just calls CurrentMonotonicTime tag it deprecated in the Doxygen
3. Ensure we have some (basic) tests for all three functions
4. Switch all instances in the codebase to CurrentMonotonicTime/CurrentRealTime as appropriate
5. Profit/get a T-shirt :smile: 

It means we know everything has been pro-actively checked and changed if necessary rather than just accidentally slipping through on the old behaviour.

At some point in the future we can remove CurrentTime entirely.

Let me know if you feel differently, this is an open discussion.

common/utils/Clock.cpp
stefansiegfried: I appreciate all the feedback, @peternewman! I'll start working on the requested changes.

> Rename CurrentTime to CurrentMonotonicTime (or similar)

I did consider this, but felt that if the function falls back to `gettimeofday`, then perhaps CurrentMonotonicTime is a bit of a misnomer.  I don't mind changing it either way, we'll just have to document the behavior.

include/ola/Clock.h
stefansiegfried: I went with CurrentRealTime based on `clock_gettime`'s terminology, although they make it a point to clarify it means wall-clock time. The man page states: 
```
CLOCK_REALTIME
              System-wide  clock that measures real (i.e., wall-clock) time. ...
```
The `time` function also uses the term "real".  My opinion is stick with CurrentRealTime, but provide clarification in the documentation.

peternewman: Okay that makes sense to me, I'm happy to stick with Real.

common/utils/Clock.cpp
peternewman: Agreed on documentation. I'd say the big benefit of renaming it is we can find stuff that hasn't been spotted (e.g. when this gets merged into our main branch there are new plugins, if we keep the same name, or if external people are using it, you can't tell if that code has been checked yet and deemed safe, or just missed).

I take your point about the fall back, do we know the likelyhood of it actually happening roughly (i.e. what doesn't support monotonic)? You do raise one good point, given it's "unexpected" we should log an info message when that path is called (assuming that's not too spammy; perhaps try a test run with it logging on both paths and running at `-l 4` logging). You could also add a warning to configure when it hasn't been found/supported.

include/ola/Clock.h
stefansiegfried: done

peternewman (changes requested):
Looks good, these are mostly style comments and a few things we should check.

Don't forget to add an entry to NEWS and credit yourself in AUTHORS if you'd like.

I'll make sure I either merge this or tag it as hacktoberfest-accepted before the deadline so that's one of your four sorted, hopefully you've got a few others lined up if you want your free T-shirt, or do you want to tackle some other OLA challenges?

common/utils/Clock.cpp
stefansiegfried: > I take your point about the fall back, do we know the likelyhood of it actually happening roughly (i.e. what doesn't support monotonic)?

Unsure of the likelihood, but I suspect "older" operating systems e.g. I believe it will fall back on OS X before 10.12 due to lack of `clock_gettime`.  

> we should log an info message when that path is called (assuming that's not too spammy; perhaps try a test run with it logging on both paths and running at -l 4 logging)

I tried logging these paths, unfortunately it is kind of noisy at about once per second so I removed that debug output.

> You could also add a warning to configure when it hasn't been found/supported.

I was able to add a configure check for clock_gettime, but was unsure of how to check that CLOCK_MONOTONIC is defined. Please let me know if you have any advice on this.

peternewman: > > we should log an info message when that path is called (assuming that's not too spammy; perhaps try a test run with it logging on both paths and running at -l 4 logging)
> 
> I tried logging these paths, unfortunately it is kind of noisy at about once per second so I removed that debug output.

Ah yeah, I guess that doubles the messages we get around the SelectServer stuff.

You could add something in the AppInit and ServerInit functions which calls InitClock to just log a message or something:
https://github.com/OpenLightingProject/ola/blob/d1010976c47de23dc6031749179f8d92596b7e8d/common/base/Init.cpp

Possibly OTT, although it would probably be nice to see the status when the program is run from a debugging point of view, rather than just a warning in configure. It could even print and display both clocks for comparison.

> > You could also add a warning to configure when it hasn't been found/supported.
> 
> I was able to add a configure check for clock_gettime, but was unsure of how to check that CLOCK_MONOTONIC is defined. Please let me know if you have any advice on this.

This would cover the warning:
https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Printing-Messages.html

And this the switching:
https://stackoverflow.com/questions/46704731/check-value-of-c-define-with-autoconf

To be honest I'm still torn, most people will probably ignore a configure warning if it succeeds, the app/server init would at least surface it in a log we might look at when they complain. Equally we've worked fine for the last however many years, so the risk of it causing an issue is hopefully fairly low either way.

common/math/Random.cpp
```diff
@@ -44,7 +44,7 @@ std::default_random_engine generator_;
 void InitRandom() {
   Clock clock;
   TimeStamp now;
-  clock.CurrentTime(&now);
+  clock.CurrentMonotonicTime(&now);
```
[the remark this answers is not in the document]
stefansiegfried: > Don't forget to add an entry to NEWS and credit yourself in AUTHORS if you'd like.

Sounds good, will do.

> I'll make sure I either merge this or tag it as hacktoberfest-accepted before the deadline so that's one of your four sorted, hopefully you've got a few others lined up if you want your free T-shirt, or do you want to tackle some other OLA challenges?

Thank you very much, I appreciate that.  I actually could use two more PRs after this one, so if you have anything in mind that you think I could accomplish by the end of the month, I'd be happy to take a look at those too.


> I assume Monotonic time does have microsecond precision too?

At least on my Linux system, CLOCK_MONOTONIC has nanosecond precision according to clock_getres.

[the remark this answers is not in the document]
peternewman: > > I'll make sure I either merge this or tag it as hacktoberfest-accepted before the deadline so that's one of your four sorted, hopefully you've got a few others lined up if you want your free T-shirt, or do you want to tackle some other OLA challenges?
> 
> Thank you very much, I appreciate that.

No worries, you deserve it, I hope you've enjoyed your open source journey and it would be great if you fancy hanging around and doing the odd future pull request to scratch that itch!

> I actually could use two more PRs after this one, so if you have anything in mind that you think I could accomplish by the end of the month, I'd be happy to take a look at those too.

Here are a selection in my guess at a rough order of complexity:

Fix a compiler warning:
https://github.com/OpenLightingProject/ola/issues/1564

Port some config behaviour from one plugin to another:
https://github.com/OpenLightingProject/ola/issues/508

Or a cleanup task, definitely to target master:
https://github.com/OpenLightingProject/ola/issues/1318

Adding some extra config to the web server to host more files:
https://github.com/OpenLightingProject/ola/issues/1547

How's you're Makefile knowledge? This should be pretty trivial basically just add a phony target to run flake8, then some wrappers around it and in configure to check flake8 is available:
https://github.com/OpenLightingProject/ola/issues/1619

Perhaps part of this:
https://github.com/OpenLightingProject/ola/issues/1529

Or this:
https://github.com/OpenLightingProject/ola/issues/1480

> > I assume Monotonic time does have microsecond precision too?
> 
> At least on my Linux system, CLOCK_MONOTONIC has nanosecond precision according to clock_getres.

Excellent. I guess the quick test would be confirming that now.MicroSeconds() prints a different value each run.

common/thread/PeriodicThread.cpp
```diff
@@ -53,7 +53,7 @@ void *PeriodicThread::Run() {
   Clock clock;
   TimeStamp last_run_at;
 
-  clock.CurrentTime(&last_run_at);
+  clock.CurrentMonotonicTime(&last_run_at);
```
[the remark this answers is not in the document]
stefansiegfried: Good catch, I think this one is actually real time.

common/utils/Clock.cpp
```diff
@@ -251,12 +252,29 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
...
+void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
+#ifdef CLOCK_MONOTONIC
+  struct timeval tv;
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  tv.tv_sec = ts.tv_sec;
+  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
+  *timestamp = tv;
+#else
+  CurrentRealTime(timestamp);
```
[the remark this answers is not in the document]
stefansiegfried: Going to look at logging elsewhere to reduce noise

common/utils/Clock.cpp
```diff
@@ -251,12 +252,29 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
...
 }
 
-void Clock::CurrentTime(TimeStamp *timestamp) const {
+void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
+#ifdef CLOCK_MONOTONIC
+  struct timeval tv;
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  tv.tv_sec = ts.tv_sec;
+  tv.tv_usec = ts.tv_nsec / ONE_THOUSAND;
```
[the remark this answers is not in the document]
stefansiegfried: Great idea, thanks.

common/utils/ClockTest.cpp
```diff
@@ -143,11 +151,49 @@ void ClockTest::testTimeIntervalMutliplication() {
...
+
+  TimeStamp second;
+  clock.CurrentRealTime(&second);
+  OLA_ASSERT_LT(first, second);
+}
 
 /**
- * test the clock
+ * @brief Test the CurrentTime wrapper method
+ *
```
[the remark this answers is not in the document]
stefansiegfried: Done.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   *
+   * The timestamp parameter will be set to the current monotonic time only if
+   * the system has a monotonic clock available.  If a monotonic clock is
+   * unavailable, this method will fallback to CurrentRealTime.
+   *
+   * The system's monotonic clock is not subject to discontinuous jumps due to
+   * administrative action, but may be affected by incremental adjustment due to
+   * time synchronization protocol e.g. NTP.  The monotonic clock does not
+   * advance while the system is suspended.
+   * @param timestamp A TimeStamp pointer
```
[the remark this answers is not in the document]
stefansiegfried: Done.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   * @return void
+   */
+  virtual void CurrentRealTime(TimeStamp* timestamp) const;
+
+  /**
+   * @brief Wrapper around CurrentMonotonicTime.
+   * @param timestamp A TimeStamp pointer
+   * @return void
+   * @deprecated This method is deprecated as of v0.10. Please use either
```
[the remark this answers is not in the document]
stefansiegfried: Done.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   * The timestamp parameter will be set to the current monotonic time only if
+   * the system has a monotonic clock available.  If a monotonic clock is
+   * unavailable, this method will fallback to CurrentRealTime.
+   *
+   * The system's monotonic clock is not subject to discontinuous jumps due to
+   * administrative action, but may be affected by incremental adjustment due to
+   * time synchronization protocol e.g. NTP.  The monotonic clock does not
+   * advance while the system is suspended.
+   * @param timestamp A TimeStamp pointer
+   * @return void
```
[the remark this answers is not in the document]
stefansiegfried: Removed.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   * @brief Sets timestamp to the current real time.
+   *
+   * The timestamp parameter will be set to the current real e.g. wall-clock
+   * time.
+   *
+   * The system's realtime clock is subject to discontinuous and incremental
+   * adjustment due to administrative action or time synchronization protocol
+   * e.g. NTP.
+   * @param timestamp A TimeStamp pointer
+   * @return void
```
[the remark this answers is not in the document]
stefansiegfried: Removed.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   */
+  virtual void CurrentRealTime(TimeStamp* timestamp) const;
+
+  /**
+   * @brief Wrapper around CurrentMonotonicTime.
+   * @param timestamp A TimeStamp pointer
+   * @return void
```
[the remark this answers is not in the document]
stefansiegfried: Removed.

common/thread/PeriodicThread.cpp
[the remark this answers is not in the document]
peternewman: In which case a comment would be good too.

peternewman:
If you want to push up the bits you've fixed already I can start checking them off while you work on the other bits.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   * advance while the system is suspended.
+   * @param timestamp A TimeStamp pointer
+   * @return void
+   */
+  virtual void CurrentMonotonicTime(TimeStamp* timestamp) const;
+
+  /**
+   * @brief Sets timestamp to the current real time.
+   *
+   * The timestamp parameter will be set to the current real e.g. wall-clock
```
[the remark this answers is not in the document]
peternewman: Done

common/utils/ClockTest.cpp
```diff
@@ -162,11 +208,65 @@ void ClockTest::testClock() {
...
+  TimeStamp third;
+  clock.CurrentRealTime(&third);
+  OLA_ASSERT_LT(second, third);
+  OLA_ASSERT_TRUE(ten_point_five_seconds <= (third - second));
+}
 
 /**
- * test the Mock Clock
+ * @brief Test the mock CurrentTime wrapper method
+ *
```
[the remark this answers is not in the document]
peternewman: Done

common/utils/ClockTest.cpp
```diff
@@ -162,11 +208,65 @@ void ClockTest::testClock() {
   OLA_ASSERT_LT(first, second);
 }
 
+/**
+ * @brief Test the mock monotonic clock
+ *
```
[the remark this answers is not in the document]
peternewman: Done

common/utils/ClockTest.cpp
```diff
@@ -143,11 +151,49 @@ void ClockTest::testTimeIntervalMutliplication() {
   OLA_ASSERT_EQ((int64_t) 20000, twenty_seconds.InMilliSeconds());
 }
 
+/**
+ * @brief Test the monotonic clock
+ *
```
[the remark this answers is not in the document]
peternewman: Done

common/utils/ClockTest.cpp
```diff
@@ -143,11 +151,49 @@ void ClockTest::testTimeIntervalMutliplication() {
...
+#endif  // _WIN32
+
+  TimeStamp second;
+  clock.CurrentMonotonicTime(&second);
+  OLA_ASSERT_LT(first, second);
+}
+
+/**
+ * @brief Test the real time clock
+ *
```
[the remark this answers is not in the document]
peternewman: Done

NEWS
```diff
@@ -3,7 +3,8 @@ x/y/2020 ola-0.10.8
  * Support DragonFly BSD
 
  API:
- * 
+ * Deprecated Clock::CurrentTime and added Clock::CurrentMonotonicTime and 
+   Clock::CurrentRealTime
```
peternewman, line 7: Perhaps also mention that stuff has been switched to monotonic where possible.

include/ola/Clock.h
```diff
@@ -214,21 +214,56 @@ class TimeStamp {
...
+   * time.
+   *
+   * The system's realtime clock is subject to discontinuous and incremental
+   * adjustment due to administrative action or time synchronization protocol
+   * e.g. NTP.
```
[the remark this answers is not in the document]
peternewman: Done.

common/utils/Clock.cpp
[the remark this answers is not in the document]
peternewman: Done thanks.

common/utils/Clock.cpp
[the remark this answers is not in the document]
peternewman: Done elsewhere.

common/thread/PeriodicThread.cpp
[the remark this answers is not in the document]
peternewman: Done.

common/math/Random.cpp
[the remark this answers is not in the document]
peternewman: I see you've switched Random back, did it not work as intended? Can we have a comment as such if so.

common/base/Init.cpp
```diff
@@ -387,5 +391,18 @@ void Daemonise() {
...
+Clock clock;
+TimeStamp now_monotonic;
+TimeStamp now_realtime;
+#ifndef CLOCK_MONOTONIC
+  OLA_DEBUG << "Monotonic clock unavailable. Falling back to real time clock."
+#endif
+clock.CurrentMonotonicTime(&now_monotonic);
+clock.CurrentRealTime(&now_realtime);
+OLA_DEBUG << "Monotonic clock: " << std::setw(18) << now_monotonic;
+OLA_DEBUG << "Real clock     : " << std::setw(18) << now_realtime;
```
peternewman, line 405: Can you indent all this by two (aside from the ifndef/endif) please.

common/thread/PeriodicThread.cpp
```diff
@@ -72,7 +74,7 @@ void *PeriodicThread::Run() {
         continue;
       }
     }
-    clock.CurrentTime(&last_run_at);
+    clock.CurrentMonotonicTime(&last_run_at);
```
peternewman, line 77: Doesn't this want to be Real time for the same reason as the above one?

common/utils/Clock.cpp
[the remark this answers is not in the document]
peternewman: Done elsewhere.

common/utils/Clock.cpp
```diff
@@ -251,12 +263,26 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
   return TimeStamp(m_tv - interval.m_interval);
 }
 
-void Clock::CurrentTime(TimeStamp *timestamp) const {
+void Clock::CurrentMonotonicTime(TimeStamp *timestamp) const {
+#ifdef CLOCK_MONOTONIC
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
```
peternewman, line 269: Can you add this to AC_CHECK_FUNCS in configure.ac so we ensure we've actually got clock_gettime available.

include/ola/Clock.h
```diff
@@ -214,21 +216,57 @@ class TimeStamp {
...
+   * the system has a monotonic clock available.  If a monotonic clock is
+   * unavailable, this method will fallback to CurrentRealTime.
+   *
+   * The system's monotonic clock is not subject to discontinuous jumps due to
+   * administrative action, but may be affected by incremental adjustment due to
+   * time synchronization protocol e.g. NTP.  The monotonic clock does not
+   * advance while the system is suspended. An appropriate use of this method is
+   * to measure elapsed time.
+   * @sa Clock::CurrentRealTime
+   * @param timestamp A TimeStamp pointer
```
peternewman, line 239: This flags it as an output parameter in Doxygen land.
```suggestion
   * @param[out] timestamp A TimeStamp pointer
```

include/ola/Clock.h
```diff
@@ -214,21 +216,57 @@ class TimeStamp {
...
+   *
+   * The timestamp parameter will be set to the current real i.e. wall-clock
+   * time.
+   *
+   * The system's realtime clock is subject to discontinuous and incremental
+   * adjustment due to administrative action or time synchronization protocol
+   * e.g. NTP. An appropriate use of this method is to display a user-facing
+   * timestamp.
+   * @sa Clock::CurrentMonotonicTime
+   * @param timestamp A TimeStamp pointer
```
peternewman, line 254: Again

include/ola/Clock.h
```diff
@@ -214,21 +216,57 @@ class TimeStamp {
...
+   */
+  virtual void CurrentRealTime(TimeStamp* timestamp) const;
+
+  /**
+   * @brief Wrapper around CurrentMonotonicTime.
+   * @param timestamp A TimeStamp pointer
```
peternewman, line 260: Again

peternewman (changes requested):
A few more comments

common/utils/Clock.cpp
```diff
@@ -251,12 +263,26 @@ const TimeStamp TimeStamp::operator-(const TimeInterval &interval) const {
...
+}
+
+void Clock::CurrentRealTime(TimeStamp *timestamp) const {
   struct timeval tv;
   gettimeofday(&tv, NULL);
   *timestamp = tv;
 }
 
+void Clock::CurrentTime(TimeStamp *timestamp) const {
+  CurrentMonotonicTime(timestamp);
```
peternewman, line 283: I'm just thinking, this should probably be real time for full backwards compatibility.

NEWS
```diff
@@ -4,7 +4,8 @@ x/y/2020 ola-0.10.8
 
  API:
  * Deprecated Clock::CurrentTime and added Clock::CurrentMonotonicTime and 
-   Clock::CurrentRealTime
+   Clock::CurrentRealTime. Occurrences of Clock::CurrentTime have been
+   replaced with Clock::CurrentRealTime where applicable.
```
peternewman, line 8: Given some became each version.
```suggestion
   Clock::CurrentRealTime. Occurrences of Clock::CurrentTime have been
   replaced with either variant as applicable.
```

common/thread/PeriodicThread.cpp
```diff
@@ -74,7 +74,7 @@ void *PeriodicThread::Run() {
         continue;
       }
     }
-    clock.CurrentMonotonicTime(&last_run_at);
+    clock.CurrentRealTime(&last_run_at);
```
peternewman, line 77: Comment again please.

include/ola/Clock.h
```diff
@@ -251,13 +251,13 @@ class Clock {
...
    * @sa Clock::CurrentMonotonicTime
-   * @param timestamp A TimeStamp pointer
+   * @param[out] timestamp A TimeStamp pointer
    */
   virtual void CurrentRealTime(TimeStamp* timestamp) const;
 
   /**
    * @brief Wrapper around CurrentMonotonicTime.
-   * @param timestamp A TimeStamp pointer
+   * @param timestamp[out] A TimeStamp pointer
```
peternewman, line 260:
```suggestion
   * @param[out] timestamp A TimeStamp pointer
```

peternewman (changes requested):
A few minor tweaks.

include/ola/Clock.h
```diff
@@ -214,21 +216,57 @@ class TimeStamp {
...
+   * adjustment due to administrative action or time synchronization protocol
+   * e.g. NTP. An appropriate use of this method is to display a user-facing
+   * timestamp.
+   * @sa Clock::CurrentMonotonicTime
+   * @param[out] timestamp A TimeStamp pointer
+   */
+  virtual void CurrentRealTime(TimeStamp* timestamp) const;
+
+  /**
+   * @brief Wrapper around CurrentMonotonicTime.
```
peternewman, line 259: Correct this too please, now it's around Real.

common/math/Random.cpp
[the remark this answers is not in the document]
stefansiegfried: I think the change may be inconsequential in this case because random is being seeded with usecs, which repeat every second regardless of clock type.  However, if the seeding was changed in the future and secs were used, then I think the realtime secs _may_ be preferred since they do not repeat, unlike monotonic seconds which usually reset to 0 on system boot.

common/thread/PeriodicThread.cpp
peternewman: Done

include/ola/Clock.h
peternewman: Done.

include/ola/Clock.h
peternewman: Done.

common/utils/Clock.cpp
peternewman: Done.

include/ola/Clock.h
peternewman: Done.

include/ola/Clock.h
peternewman: Done.

include/ola/Clock.h
peternewman: Done.

common/utils/Clock.cpp
peternewman: Done.

common/thread/PeriodicThread.cpp
peternewman: Done.

common/base/Init.cpp
peternewman: Done.

NEWS
peternewman: Done.

common/math/Random.cpp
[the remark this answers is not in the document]
peternewman: Ah I didn't realise monotonic ones were that simple like a system epoch, I assumed they were still based off realtime somehow. As you can probably tell I only skim read the document.

common/utils/Clock.cpp
peternewman: Now done.

include/ola/Clock.h
```diff
@@ -111,6 +112,12 @@ class BaseTimeVal {
                 struct timeval *result) const;
 
   void Set(int64_t interval_useconds);
+
+  /**
+   * @brief Sets m_tv
```
peternewman, line 117: A different comment would probably be better documentation. Perhaps something like this.
```suggestion
   * @brief Sets the value with a timespec.
```

peternewman (changes requested):
You should add yourself to Authors if you want a credit in there!

Just one other minor tweak please then I think we're good.

include/ola/Clock.h
stefansiegfried: I agree :)

common/math/Random.cpp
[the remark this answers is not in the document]
stefansiegfried: I couldn't tell, but it's probably worth taking a second, closer, look over all the instances we replaced with monotonic clock to make sure there are not any more gotchas like PeriodicThread.  I'll have another look at those tomorrow.

common/utils/ClockTest.cpp
```diff
@@ -162,11 +208,65 @@ void ClockTest::testClock() {
...
+  TimeStamp first;
+  clock.CurrentRealTime(&first);
+
+  TimeInterval one_second(1, 0);
+  clock.AdvanceTime(one_second);
+
+  TimeStamp second;
+  clock.CurrentRealTime(&second);
+  OLA_ASSERT_LT(first, second);
+  OLA_ASSERT_TRUE(one_second <= (second - first));
```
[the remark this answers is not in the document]
stefansiegfried: This was a copy-paste of an existing clock unit test and perhaps I'm not seeing what the issue is yet though. Could you elaborate on why this would fail?

common/math/Random.cpp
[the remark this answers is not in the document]
peternewman: Okay great thanks.

common/utils/ClockTest.cpp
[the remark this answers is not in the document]
peternewman: Well the whole reason for switching to monotonic is because real time may move backwards or forwards due to outside influences. Consider the following:

1. `clock.CurrentRealTime(&first);` 00:00:10
1. `clock.AdvanceTime(one_second);` 00:00:11
1. NTP causes wall clock to roll back 2s due to a leap seconds (it got really bad so we added two!)
1. `clock.CurrentRealTime(&second);` 00:00:09
1. `OLA_ASSERT_LT(first, second);` - This is now false
1. `OLA_ASSERT_TRUE(one_second <= (second - first));` 1 <= 9-10 == 1 <= -1 - This is also now false

It's probably worth switching these to the _MSG variants, or adding one here for the non-existent LT_MSG so we can give the user a little message in the unlikely event the test fails:
https://github.com/OpenLightingProject/ola/blob/0.10/include/ola/testing/TestUtils.h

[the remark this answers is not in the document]
peternewman: This is still outstanding @stefansiegfried .

plugins/gpio/GPIODriver.cpp
```diff
@@ -90,7 +90,7 @@ void *GPIODriver::Run() {
     bool update_pins = false;
 
     TimeStamp wake_up;
-    clock.CurrentTime(&wake_up);
+    clock.CurrentMonotonicTime(&wake_up);
```
peternewman, line 93: This is going to pthread_cond_timedwait should it be RealTime too?

plugins/karate/KarateThread.cpp
```diff
@@ -76,7 +76,7 @@ void *KarateThread::Run() {
     if (!k.IsActive()) {
       // try to reopen the device...
       TimeStamp wake_up;
-      clock.CurrentTime(&wake_up);
+      clock.CurrentMonotonicTime(&wake_up);
```
peternewman, line 79: Again going to pthread.

plugins/opendmx/OpenDmxThread.cpp
```diff
@@ -79,7 +79,7 @@ void *OpenDmxThread::Run() {
 
     if (m_fd == INVALID_FD) {
       TimeStamp wake_up;
-      clock.CurrentTime(&wake_up);
+      clock.CurrentMonotonicTime(&wake_up);
```
peternewman, line 82: Pthread

peternewman (changes requested):
Aside from those pthread issues and the outstanding comment I'm happy having re-reviewed all the changes.

I think for now it would make sense to add a comment to https://github.com/OpenLightingProject/ola/blob/d1010976c47de23dc6031749179f8d92596b7e8d/common/thread/Mutex.cpp#L123 that it should use Real time clock values.

I'll have a final read of the other discussion to make sure we've not missed anything but I think this is basically across the line.

common/utils/ClockTest.cpp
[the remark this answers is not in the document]
stefansiegfried: Great explanation, that is a valid point for real clock. My confusion was that the line numbers referenced (253-254) are actually part of the monotonic clock test and in this case I don't believe time can roll back.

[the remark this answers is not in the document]
peternewman: Yeah I don't know what's going on there. I think those numbers are stale. If you look here on the files page, it shows I am within testMockClockRealTime. This caught me out earlier too:
https://github.com/OpenLightingProject/ola/pull/1673/files#diff-b8f53e67619b3b3cb1ef5e4f7011f74d3cfdb1959f4b0e94e6f2d53a1ee82062

plugins/opendmx/OpenDmxThread.cpp
peternewman: Done.

plugins/karate/KarateThread.cpp
peternewman: Done.

plugins/gpio/GPIODriver.cpp
peternewman: Done.

common/utils/ClockTest.cpp
```diff
@@ -35,7 +35,7 @@ class ClockTest: public CppUnit::TestFixture {
   CPPUNIT_TEST_SUITE(ClockTest);
   CPPUNIT_TEST(testTimeStamp);
   CPPUNIT_TEST(testTimeInterval);
-  CPPUNIT_TEST(testTimeIntervalMutliplication);
+  CPPUNIT_TEST(testTimeIntervalMultiplication);
```
peternewman: Good spot!

common/utils/ClockTest.cpp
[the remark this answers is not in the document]
peternewman: Done.

peternewman (approved):
LGTM thanks @stefansiegfried 

I guess a future stretch goal for another PR is switching the pthread_cond_timedwait stuff to monotonic too where possible.

peternewman:
I've just found this, which would have been fun to test it with:
http://manpages.ubuntu.com/manpages/jammy/man1/faketime.1.html

PR #1676, opened 2020-10-28
Add flake8 Makefile target

stefansiegfried:
* Closes OpenLightingProject/ola#1619

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -231,6 +231,6 @@ EXTRA_DIST += \
 # Create a flake8 python linter target
 if FOUND_FLAKE8
 .PHONY : flake8
-flake8:
+flake8: Makefile.am
 	flake8
 endif
--- a/Makefile.am
+++ b/Makefile.am
@@ -225,3 +225,12 @@ EXTRA_DIST += \
   config/stl_hash.m4 \
   mac_build.sh \
   ola.spec
+
+# -----------------------------------------------------------------------------
+
+# Create a flake8 python linter target
+if FOUND_FLAKE8
+.PHONY : flake8
+flake8:
+	flake8
+endif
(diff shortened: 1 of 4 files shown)


Makefile.am
```diff
@@ -225,3 +225,12 @@ EXTRA_DIST += \
   config/stl_hash.m4 \
   mac_build.sh \
   ola.spec
+
+# -----------------------------------------------------------------------------
+
+# Create a flake8 python linter target
+if FOUND_FLAKE8
+.PHONY : flake8
+flake8:
```
peternewman, line 234: I don't know if it's strictly necessary, but we normally list Makefile.am here, like so:
```suggestion
flake8: Makefile.am
```

Which should mean that changes to the config within Makefile.am would cause it to need to be re-run.

Makefile.am
```diff
@@ -225,3 +225,12 @@ EXTRA_DIST += \
...
+	flake8
```
peternewman, line 235: Perhaps consider moving the conditional inside so `make flake8` always exists but throws an error/dies when flake8 is missing. Otherwise if we write some docs saying run `make flake8` people will get a confusing make error if it's not on their system rather than a helpful one saying to install flake8 and re-run configure.

peternewman (changes requested):
It would be good to have an overarching `make lint` command which runs flake8 etc if present on the machine (and can be extended to cover the other tools).

You could then also list this here:
https://github.com/OpenLightingProject/ola/blob/master/configure.ac#L1012-L1015

Its probably good to switch the Travis flake8 to call your Makefile version too:
https://github.com/OpenLightingProject/ola/blob/master/.travis-ci.sh

Makefile.am
peternewman: Done.

Makefile.am
peternewman: Done.

peternewman (approved):
LGTM thanks @stefansiegfried .

Just waiting for a Travis test run to pass then I'll merge.

Feel free to add a NEWS comment too if you want.

PR #1682, opened 2020-11-03
[RFC] Record v2

shenghaoyang:
Adds a V2 show file format that uses absolute timestamps from the start of the recording.

No core changes are required in the recorder's playback engine. The loader automatically converts the absolute timestamps into frame timeouts.

The layout of the show files are unchanged except the header is changed to `OLA Show V2` and the timings are now recorded in `<second> <ms>` format, similar how `struct timeval` stores them as two separate values.

The player automatically detects the file format by inspecting the header and should handle both types automatically.

The recorder gained an option to select the recording file format, defaulting to the v1 relative timestamp format.

This helps make the playback of a recording more accurate as we eliminate accumulated roundoff errors caused by storing relative timestamps.

Depends on #1680

### DIFF
--- a/examples/ShowLoader.cpp
+++ b/examples/ShowLoader.cpp
@@ -25,6 +25,7 @@
 
 #include <errno.h>
 #include <string.h>
+#include <ola/Clock.h>
 #include <ola/DmxBuffer.h>
 #include <ola/Logging.h>
 #include <ola/StringUtils.h>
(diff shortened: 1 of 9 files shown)


shenghaoyang:
Does this new format look sane? If it does I'll go work on some tests and polish the PR a bit more: just throwing ideas out here now.

shenghaoyang:
There's an easier way to do this.

Patch incorporated into #1680

PR #1683, opened 2020-11-13
Extend examples/streaming-client to read universe numbers from STDIN

shenghaoyang:
Makes it a lot more useful for sending data to OLA.

(Also used in the crazy pipeline to use VLC to play converted showfiles - see https://github.com/shenghaoyang/ola_video_convert for the prototype)


examples/ola-streaming-client.cpp
```diff
@@ -78,11 +84,30 @@ int main(int argc, char *argv[]) {
...
+      }
+
+      if (have_universe || !FLAGS_universe_from_source) {
+        SendDataFromString(&ola_client, read_universe, input);
+        have_universe = false;
+      }
     }
   } else {
+    if (FLAGS_universe_from_source) {
+      OLA_FATAL << "Not reading from STDIN. Use -u to sepcify universe.";
```
peternewman, line 108:
```suggestion
      OLA_FATAL << "Not reading from STDIN. Use -u to specify universe.";
```

examples/ola-streaming-client.cpp
```diff
@@ -41,6 +41,12 @@ DEFINE_s_string(dmx, d, "", "Comma separated DMX values to send, e.g. "
 DEFINE_s_uint32(universe, u, 1, "The universe to send data for");
 DEFINE_uint8(priority, ola::dmx::SOURCE_PRIORITY_DEFAULT,
              "The source priority to send data at");
+DEFINE_s_default_bool(universe_from_source, s, false,
+                      "Read universe number from source. Only allowed when "
+                      "processing STDIN. The universe number must be formatted "
+                      "in base 10, and be separated by at least one whitespace"
+                      " character from the start and end of the sequence of "
```
peternewman, line 48: Nit, we generally favour leaving the spaces at the end of a line as it makes it a bit more readable/them less likely to get dropped:
```suggestion
                      "in base 10, and be separated by at least one whitespace "
                      "character from the start and end of the sequence of "
```

examples/ola-streaming-client.cpp
```diff
@@ -41,6 +41,12 @@ DEFINE_s_string(dmx, d, "", "Comma separated DMX values to send, e.g. "
 DEFINE_s_uint32(universe, u, 1, "The universe to send data for");
 DEFINE_uint8(priority, ola::dmx::SOURCE_PRIORITY_DEFAULT,
              "The source priority to send data at");
+DEFINE_s_default_bool(universe_from_source, s, false,
```
peternewman, line 44: I'm wondering if this would be better called --universe-from-stdin or something. Source doesn't jump out as obvious to me.

examples/ola-streaming-client.cpp
```diff
@@ -78,11 +84,30 @@ int main(int argc, char *argv[]) {
...
+
+      if (have_universe || !FLAGS_universe_from_source) {
+        SendDataFromString(&ola_client, read_universe, input);
+        have_universe = false;
+      }
     }
   } else {
+    if (FLAGS_universe_from_source) {
+      OLA_FATAL << "Not reading from STDIN. Use -u to specify universe.";
+      exit(1);
```
peternewman, line 109: It would be better to `exit(ola::EXIT_USAGE);`

examples/ola-streaming-client.cpp
```diff
@@ -78,11 +84,30 @@ int main(int argc, char *argv[]) {
...
+      }
+
+      if (have_universe || !FLAGS_universe_from_source) {
+        SendDataFromString(&ola_client, read_universe, input);
+        have_universe = false;
+      }
     }
   } else {
+    if (FLAGS_universe_from_source) {
+      OLA_FATAL << "Not reading from STDIN. Use -u to specify universe.";
```
peternewman, line 108: This should probably be a cout message in case my logging is going to syslog for example.

examples/ola-streaming-client.cpp
```diff
@@ -78,11 +84,30 @@ int main(int argc, char *argv[]) {
...
+    bool have_universe = false;
+    unsigned int read_universe = FLAGS_universe;
+
     while (!terminate && std::cin >> input) {
-      ola::StringTrim(&input);
-      SendDataFromString(&ola_client, FLAGS_universe, input);
+      if (!have_universe && FLAGS_universe_from_source) {
+        if (!ola::StringToInt(input, &read_universe, true)) {
+          OLA_FATAL << "Could not convert universe number.";
+          exit(1);
```
peternewman, line 94: This should probably be OLA_USAGE too.

examples/ola-streaming-client.cpp
```diff
@@ -78,11 +84,30 @@ int main(int argc, char *argv[]) {
...
     string input;
+    bool have_universe = false;
+    unsigned int read_universe = FLAGS_universe;
+
     while (!terminate && std::cin >> input) {
-      ola::StringTrim(&input);
-      SendDataFromString(&ola_client, FLAGS_universe, input);
+      if (!have_universe && FLAGS_universe_from_source) {
+        if (!ola::StringToInt(input, &read_universe, true)) {
+          OLA_FATAL << "Could not convert universe number.";
```
peternewman, line 93: It would be nice to put the string it's tried to process in the message to aid debugging.

examples/ola-streaming-client.cpp
```diff
@@ -41,6 +41,12 @@ DEFINE_s_string(dmx, d, "", "Comma separated DMX values to send, e.g. "
 DEFINE_s_uint32(universe, u, 1, "The universe to send data for");
 DEFINE_uint8(priority, ola::dmx::SOURCE_PRIORITY_DEFAULT,
              "The source priority to send data at");
+DEFINE_s_default_bool(universe_from_source, s, false,
+                      "Read universe number from source. Only allowed when "
+                      "processing STDIN. The universe number must be formatted "
+                      "in base 10, and be separated by at least one whitespace"
```
peternewman, line 47: What's the actual format on STDIN?
```
Uni chan1,chan2
Uni chan1,chan2
```

Is this also valid?:
```
Uni chan1,chan2 Uni chan1,chan2
```

Just a bit thrown by your end comment.

peternewman (changes requested):
A few more comments

examples/ola-streaming-client.cpp
shenghaoyang: Cool - I'll cleanup the other users of `exit(1)` as well, then.

examples/ola-streaming-client.cpp
shenghaoyang: Yeah, any sort of whitespace serving as a delimiter should be fine.

It was using https://en.cppreference.com/w/cpp/string/basic_string/operator_ltltgtgt to tokenize the input, and that (by default) just extracts chunks of input delimited by whitespace.

examples/ola-streaming-client.cpp
shenghaoyang: Hm, but I'd argue that if you're logging to syslog then you'd expect it to show up there, no?

examples/ola-streaming-client.cpp
shenghaoyang: Makes sense, I'll go swap those out.

examples/ola-streaming-client.cpp
shenghaoyang: Added some examples in the CLI help: 6ff5ae190daa8260e985f3a6996ecaf2f10ddc75

examples/ola-streaming-client.cpp
peternewman: We generally differentiate between messages to the user and logging. It's a bit of grey area, but it's going to bomb out so you probably want to see why immediately. The same as the usage stuff.

examples/ola-streaming-client.cpp
```diff
@@ -41,6 +42,11 @@ DEFINE_s_string(dmx, d, "", "Comma separated DMX values to send, e.g. "
 DEFINE_s_uint32(universe, u, 1, "The universe to send data for");
 DEFINE_uint8(priority, ola::dmx::SOURCE_PRIORITY_DEFAULT,
              "The source priority to send data at");
+DEFINE_s_default_bool(universe_from_stdin, s, false,
+                      "Also read the destination universe number from STDIN "
+                      "when reading DMX data from STDIN. The universe number "
+                      "must precede the channel values, and be delimited "
+                      "by whitespace. E.g. 1 0,255,128 2 0,255,127.");
```
peternewman, line 49: Grammar and clarity.
```suggestion
                      "by whitespace, e.g. 1 0,255,128 2 0,255,127");
```

peternewman: Will this break if I'm just trying to set the first channel of each universe?
`1 255 2 0 5 255`

examples/ola-streaming-client.cpp
peternewman: Either way, there's a cout in SendDataFromString which is inconsistent!

examples/ola-streaming-client.cpp
```diff
@@ -73,18 +79,37 @@ int main(int argc, char *argv[]) {
...
   if (!ola_client.Setup()) {
     OLA_FATAL << "Setup failed";
-    exit(1);
+    exit(ola::EXIT_SOFTWARE);
   }
 
   if (FLAGS_dmx.str().empty()) {
     string input;
+    bool have_universe = false;
+    unsigned int read_universe = FLAGS_universe;
```
peternewman, line 88: I wonder if this should be universe or send_universe given it's both the read and the CLI arg universe?

examples/ola-streaming-client.cpp
peternewman: Sorry, I've just realised this only happens when it's NOT reading from STDIN, so could be running in a script sent to syslog.

examples/ola-streaming-client.cpp
```diff
@@ -73,18 +79,37 @@ int main(int argc, char *argv[]) {
...
+
     while (!terminate && std::cin >> input) {
-      ola::StringTrim(&input);
-      SendDataFromString(&ola_client, FLAGS_universe, input);
+      if (!have_universe && FLAGS_universe_from_stdin) {
+        if (!ola::StringToInt(input, &read_universe, true)) {
+          OLA_FATAL << "Could not convert universe number: read " << input;
```
peternewman, line 93: Minor style tweak:
```suggestion
          OLA_FATAL << "Could not convert universe number, read " << input;
```

examples/ola-streaming-client.cpp
```diff
@@ -73,18 +79,37 @@ int main(int argc, char *argv[]) {
...
+          exit(ola::EXIT_DATAERR);
+        }
+
+        have_universe = true;
+        continue;
+      }
+
+      if (have_universe || !FLAGS_universe_from_stdin) {
+        SendDataFromString(&ola_client, read_universe, input);
```
peternewman, line 102: Does your example actually work, won't it barf trying to parse `128 2 0` as a channel value?

peternewman (changes requested):
A few comments

examples/ola-streaming-client.cpp
shenghaoyang: In that case the interpretation is
- universe 128
  - channel 0 = 0x02
- universe 0
  - *process will block while waiting for channel values*

I'm not sure what you mean by `won't it barf trying to parse 128 2 0 as a channel value?` ... :confused:

The channel value parsing routine comes from the existing implementation. I've only added a change to parse a leading decimal value as a universe number, so existing channel data parsing behavior is still preserved.

A recording of universe 128 & 0 with a sequence of `128 2 0 1` results in:

```
OLA Show
Recording, hit Control-C to end
128 2
4915
0 1
```

examples/ola-streaming-client.cpp
shenghaoyang: Eh.. I think that's fine either way? There's no way to confuse that variable for anything else.

peternewman: It would be nice if it was called something relevant though the flag hasn't really been read from anywhere in that sense.

examples/ola-streaming-client.cpp
peternewman: Sorry, I mean the full example from the man page is:
`1 0,255,128 2 0,255,127`

Does that actually set both universes if it's all on one line (I can't see anything to split the line).

This could just be a case of updating the example to be on two lines.

> I'm not sure what you mean by `won't it barf trying to parse 128 2 0 as a channel value?` ... confused

I mean I think it would fail to parse it as a set of channel values.

> The channel value parsing routine comes from the existing implementation. I've only added a change to parse a leading decimal value as a universe number, so existing channel data parsing behavior is still preserved.

Exactly, and that would expect chan-a,chan-b,chan-c etc. I don't see anything to stop when it hits something invalid and pass it onto the next bit of the system.

> A recording of universe 128 & 0 with a sequence of `128 2 0 1` results in:
> 
> ```
> OLA Show
> Recording, hit Control-C to end
> 128 2
> 4915
> 0 1
> ```

Why is there a bit (4s?) delay between universe 128 frame and universe 0 frame if they're on the same line?

shenghaoyang: The 4s delay was because I wasn't entering the input on one line (separating the two universes works equally well using spaces newlines doesn't really matter, since the tokenization is done using whitespace. See (2) at https://en.cppreference.com/w/cpp/string/basic_string/operator_ltltgtgt - that's what does the "stop when it hits something invalid and pass it onto the next bit of the system." part).

The recording from the input being on one line is this:

```
1 0,255,128
0
2 0,255,127
```

shenghaoyang: When the client reads from STDIN it would get individual strings of "1", "0,255,128", "0", "0,255,127" (see line 90 of `ola-streaming-client.cpp`).

The universe numbers are consumed, and only the comma-separated data values are passed to `SetFromString`: the SetFromString implementation doesn't need to know about the extra universe numbers.

peternewman: Ah apologies @shenghaoyang . I'd always erroneously assumed this would be just newlines and didn't realise any whitespace would work. I see the same thing does indeed work on the existing code too!

examples/ola-streaming-client.cpp
```diff
@@ -41,6 +42,11 @@ DEFINE_s_string(dmx, d, "", "Comma separated DMX values to send, e.g. "
 DEFINE_s_uint32(universe, u, 1, "The universe to send data for");
 DEFINE_uint8(priority, ola::dmx::SOURCE_PRIORITY_DEFAULT,
              "The source priority to send data at");
+DEFINE_s_default_bool(universe_from_stdin, s, false,
+                      "Also read the destination universe number from STDIN "
+                      "when reading DMX data from STDIN. The universe number "
+                      "must precede the channel values, and be delimited "
+                      "by whitespace, e.g. 1 0,255,128 2 0,255,127");
```
peternewman, line 49:
```suggestion
                      "must precede the channel values, and be delimited by "
                      "whitespace, e.g. 1 0,255,128 2 0,255,127");
```

peternewman (approved):
LGTM thanks @shenghaoyang . One minor nit which I'll fix myself then merge it in once Travis passes.

PR #1685, opened 2020-11-20
Release 0.10.8

peternewman:
- Fix some GCC warning changes
- Do some initial python 3 compatibility changes
- Update the PID definitions
- Add some more python pack/unpack blob tests


peternewman:
Force merging as the most recent commit has no real code changes.

PR #1690, opened 2020-12-02
Update to Python 3.9

mikacousin:
Fix AttributeError: 'array.array' object has no attribute 'fromstring'

### DIFF
--- a/python/ola/OlaClient.py
+++ b/python/ola/OlaClient.py
@@ -1173,7 +1173,10 @@ class OlaClient(Ola_pb2.OlaClientService):
 
     if request.universe in self._universe_callbacks:
       data = array.array('B')
-      data.fromstring(request.data)
+      if sys.version >= '3.2':
+        data.frombytes(request.data)
+      else:
+        data.fromstring(request.data)
       self._universe_callbacks[request.universe](data)
     response = Ola_pb2.Ack()
     callback(response)
@@ -1478,7 +1481,10 @@ class OlaClient(Ola_pb2.OlaClientService):
 
     if status.Succeeded():
       data = array.array('B')
-      data.fromstring(response.data)
+      if sys.version >= '3.2':
+        data.frombytes(response.data)
+      else:
+        data.fromstring(response.data)
       universe = response.universe
 
     callback(status, universe, data)
--- a/python/ola/OlaClient.py
+++ b/python/ola/OlaClient.py
@@ -1172,11 +1172,7 @@ class OlaClient(Ola_pb2.OlaClientService):
       return False
 
     if request.universe in self._universe_callbacks:
-      data = array.array('B')
-      if sys.version >= '3.2':
-        data.frombytes(request.data)
-      else:
-        data.fromstring(request.data)
+      data = array.array('B', request.data)
       self._universe_callbacks[request.universe](data)
     response = Ola_pb2.Ack()
     callback(response)
@@ -1480,11 +1476,7 @@ class OlaClient(Ola_pb2.OlaClientService):
     universe = None
 
     if status.Succeeded():
-      data = array.array('B')
-      if sys.version >= '3.2':
-        data.frombytes(response.data)
-      else:
-        data.fromstring(response.data)
+      data = array.array('B', response.data)
       universe = response.universe
 
     callback(status, universe, data)


peternewman (commented):
Thanks @mikacousin . Good spot.

I've updated your PR to target the 0.10 branch. We work in the master/0.x branches and just generate the 0.x.y branches at release for historical reasons.

Would you be interested in adding a test of one or both of these functions to ClientWrapperTest.py in the same way as @brucelowekamp did in there and RDMTest.py as it seems this issue slipped through the net due to having no equivalent test for these functions.

mikacousin:
> Would you be interested in adding a test of one or both of these functions to ClientWrapperTest.py in the same way as @brucelowekamp did in there and RDMTest.py as it seems this issue slipped through the net due to having no equivalent test for these functions.

I'm sorry but I don't master ola and python enough to write those tests. Shame on me, but I still have a lot to learn :-)

brucelowekamp (changes requested):
Good catch.   I hate these if version things, though.  Would this work with just:
data = array.array('B', request.data)
?

mikacousin:
@brucelowekamp : Just tested with python 3.9, it works and it's more elegant.

brucelowekamp:
> > Would you be interested in adding a test of one or both of these functions to ClientWrapperTest.py in the same way as @brucelowekamp did in there and RDMTest.py as it seems this issue slipped through the net due to having no equivalent test for these functions.
> 
> I'm sorry but I don't master ola and python enough to write those tests. Shame on me, but I still have a lot to learn :-)

@peternewman It would definitely be a bit challenging to write the tests, though I think the socket mock could do it.  I don't have time for now.  IMHO it would be better to merge this with a TODO to write tests for it.  I believe the change should work on python2 and python3.

peternewman:
> @peternewman It would definitely be a bit challenging to write the tests, though I think the socket mock could do it. I don't have time for now. IMHO it would be better to merge this with a TODO to write tests for it. I believe the change should work on python2 and python3.

I think you've already done it with the RDM ones that send and receive data @brucelowekamp . I was actually going to try and knock one together and see if @mikacousin or you could test it instead but I keep getting sidetracked today.

peternewman:
Hmm, I thought I'd covered this test in https://github.com/OpenLightingProject/ola/pull/1687/commits/c402fcbcb416edc8d0cc45a0d1d16fec8ea3b34a but it's still passing on Python 3.7 which it looks like it shouldn't from the docs, so I guess I've not tested quite the right bit...

peternewman:
> Hmm, I thought I'd covered this test in [c402fcb](https://github.com/OpenLightingProject/ola/commit/c402fcbcb416edc8d0cc45a0d1d16fec8ea3b34a) but it's still passing on Python 3.7 which it looks like it shouldn't from the docs, so I guess I've not tested quite the right bit...

Okay I'm really confused now, because a local run with some print statements proves it is running the array code in my test. @mikacousin could you give that commit/my branch a test on your machine please?

mikacousin:
No problem, compiling now your clang-latest branch.
I'm sorry but what kind of tests do you need ? Run olad with logs ?

peternewman:
That's perfect thanks @mikacousin . Sorry I was thrown by it working in 3.7 but not in 3.9, but having read more carefully I see it was deprecating in 3.2 and dropped in 3.9.

peternewman (approved):
Thanks again @mikacousin . Works in reality, passes my new tests, what's not to like!

peternewman:
Force merging as we've got some outstanding Travis issues which will be fixed in another PR by me, no point holding this one up.

peternewman:
@brucelowekamp I assume you don't have a nicer way you'd prefer to do this bit which we should ask @mikacousin nicely to test on a machine with 3.9?:
https://github.com/OpenLightingProject/ola/blob/7090ab2f066a9a24a182d62b6049d50b8639adad/python/ola/OlaClient.py#L949-L952

brucelowekamp:
> @brucelowekamp I assume you don't have a nicer way you'd prefer to do this bit which we should ask @mikacousin nicely to test on a machine with 3.9?:
> https://github.com/OpenLightingProject/ola/blob/7090ab2f066a9a24a182d62b6049d50b8639adad/python/ola/OlaClient.py#L949-L952

not that I know of.  problems with renaming methods across releases...

peternewman:
> not that I know of. problems with renaming methods across releases...

Okay thanks. Yeah this Python 3 thing has certainly been painful in terms of migration. Although I guess things like libmicrohttpd didn't do much better either.

PR #1706, opened 2020-12-31
Fix some more Python 3 errors

peternewman:
And a few other minor fixes


peternewman:
@RenZ0 says it's fine now on his machines.

PR #1710, opened 2021-01-22
Fix Python tests for Python 2.6

RenZ0:
Python 2.6 needs the backport called unittest2 (to handle assertIsNone for example).

### DIFF
--- a/python/ola/ClientWrapperTest.py
+++ b/python/ola/ClientWrapperTest.py
@@ -21,7 +21,11 @@ import binascii
 import datetime
 import socket
 # import timeout_decorator
-import unittest
+try:
+  # Python 2.6 needs backport
+  import unittest2 as unittest
+except ImportError:
+  import unittest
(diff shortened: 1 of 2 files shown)


python/ola/ClientWrapperTest.py
```diff
@@ -21,7 +21,11 @@
 import datetime
 import socket
 # import timeout_decorator
-import unittest
+try:
+  # Python 2.6 needs backport
+  import unittest2 as unittest
+except ImportError:
+  import unittest
```
peternewman, line 28: @brucelowekamp may have an opinion between try/execept and sys.version based checks. They may have an opinion about doing this at all too! :smile: 

I also wonder if moving it into ola.TestUtils might work too, to keep the logic once in one place, or if that might upset flake8/pychecker.

peternewman (changes requested):
As mentioned I think you should add the configure change too, so it doesn't just spontaneously break on older versions.

PR #1712, opened 2021-02-02
usbdmx: Move widget deletion to main thread

DanielG:
Currently when a usb device is remove the libusb event thread, though the
hotplug handler, will cause the widget's Device to be deleted in the main
thread and wait for this to complete in the libusb thread.

If any of the destructors triggered by device deletion causes libusb_close
to be called this causes a deadlock as libusb_close will try to take over
handling of events in the main thread while the libusb thread is waiting
for libusb_close, via the destructors, to finish.

Just move both widget and device deletion to the main thread to prevent
this scenario.

### DIFF
--- a/plugins/usbdmx/AsyncPluginImpl.cpp
+++ b/plugins/usbdmx/AsyncPluginImpl.cpp
@@ -264,16 +264,9 @@ void AsyncPluginImpl::DeviceEvent(HotplugAgent::EventType event,
     // Sunlite plugin, if we make the f/w load async we'll need to let the
     // factory cancel the load.
 
-    // Unregister & delete the device in the main thread.
-    if (state->ola_device) {
-      Future<void> f;
-      m_plugin_adaptor->Execute(
-          NewSingleCallback(this, &AsyncPluginImpl::ShutdownDevice,
-                            state->ola_device, &f));
-      f.Get();
-      state->ola_device = NULL;
-    }
-    state->DeleteWidget();
+    // Unregister & delete the device and widget in the main thread.
+    m_plugin_adaptor->Execute(
+        NewSingleCallback(this, &AsyncPluginImpl::ShutdownDeviceState, state));
   }
 }
 
@@ -353,17 +346,26 @@ bool AsyncPluginImpl::StartAndRegisterDevice(Widget *widget, Device *device) {
 
 /*
  * @brief Signal widget removal.
- * @param device_id The device id to remove.
+ * @param state The DeviceState owning the widget to be removed.
  *
- * This is run within the main thread.
+ * This must be run within the main thread. The Device destructor below may
+ * cause libusb_close() to be called, which would deadlock if the hotplug
+ * event thread were to wait for it, say in AsyncPluginImpl::DeviceEvent().
  */
-void AsyncPluginImpl::ShutdownDevice(Device *device, Future<void> *f) {
-  m_plugin_adaptor->UnregisterDevice(device);
-  device->Stop();
-  delete device;
-  if (f) {
-    f->Set();
+void AsyncPluginImpl::ShutdownDeviceState(DeviceState *state) {
+  if (state->ola_device) {
+    Device *device = state->ola_device;
+    m_plugin_adaptor->UnregisterDevice(device);
+    device->Stop();
+    delete device;
+    state->ola_device = NULL;
+  } else {
+    /* This case can be legitimate when the widget setup through the
+     * widget factory is delayed and an unplug event happens before
+     * that is completed. */
+    OLA_DEBUG << "ola_device was NULL at shutdown";
   }
+  state->DeleteWidget();
 }
 }  // namespace usbdmx
 }  // namespace plugin
--- a/plugins/usbdmx/AsyncPluginImpl.h
+++ b/plugins/usbdmx/AsyncPluginImpl.h
@@ -37,7 +37,6 @@
 #include "libs/usb/HotplugAgent.h"
 
 #include "ola/base/Macro.h"
-#include "ola/thread/Future.h"
 #include "olad/Preferences.h"
 #include "plugins/usbdmx/PluginImplInterface.h"
 #include "plugins/usbdmx/SynchronizedWidgetObserver.h"
@@ -111,7 +110,7 @@ class AsyncPluginImpl: public PluginImplInterface, public WidgetObserver {
   template <typename Widget>
   bool StartAndRegisterDevice(Widget *widget, Device *device);
 
-  void ShutdownDevice(Device *device, ola::thread::Future<void> *f);
+  void ShutdownDeviceState(DeviceState *state);
 
   DISALLOW_COPY_AND_ASSIGN(AsyncPluginImpl);
 };


plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -353,17 +346,21 @@ bool AsyncPluginImpl::StartAndRegisterDevice(Widget *widget, Device *device) {
...
-  if (f) {
-    f->Set();
+void AsyncPluginImpl::ShutdownDeviceState(DeviceState *state) {
+  if (state->ola_device) {
+    Device *device = state->ola_device;
+    m_plugin_adaptor->UnregisterDevice(device);
+    device->Stop();
+    delete device;
+    state->ola_device = NULL;
   }
```
peternewman, line 362: It would be potentially nice to log at say WARN or DEBUG level if this happens with a state without ola_device, which presumably means its getting called twice or something? With a log level appropriate to whether it can happen via legitimate means (e.g. if you pull a device OLA doesn't care about).

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -353,17 +346,21 @@ bool AsyncPluginImpl::StartAndRegisterDevice(Widget *widget, Device *device) {
 
 /*
  * @brief Signal widget removal.
- * @param device_id The device id to remove.
+ * @param state The DeviceState owning the widget to be removed.
  *
- * This is run within the main thread.
+ * This is run within the main thread and it must. The Device destructor
+ * below may cause libusb_close() to be called, which will deadlock if the
+ * hotplug event thread waits for it.
```
peternewman, line 353: I wonder if this reads a little better?
```suggestion
 * This is run within the main thread and it must be as the Device destructor
 * below may cause libusb_close() to be called, which will deadlock if the
 * hotplug event thread waits for it.
```

peternewman (commented):
This seems fairly sensible to me, but I don't fully understand the finer detail of why it was done in the first place. Have you seen this issue in your code then?

It was done in https://github.com/OpenLightingProject/ola/pull/930 , specifically https://github.com/OpenLightingProject/ola/commit/70ec24fc1f69895e3126e17c4d5a2b403e422612 .

@nomis52 may be able to comment further if they see this.

peternewman:
@kripton you might be interested in this to avoid merge conflicts with your changes, and you may have a better idea of the code than me!

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: You mean spit out a warning if ola_device is already NULL, right?

IIRC ola_device being NULL here is legitimate because of the way that StartAandRegisterDevice is called by the factory through BaseWidgetFactory::AddWidget and AsyncPluginImpl::NewWidget. The Factory can actually delay calling AddWidget instead of doing it right away in the DeviceAdded function like normal. If the usb device is then unplugged before it does end up calling AddWidget then ola_device will be NULL here.

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: Aww, and I was so proud of my forceful sounding formulation :), but sure I'll incorporate this into the next revision.

DanielG (commented):
Thanks for the review!

> Have you seen this issue in your code then?

Indeed I ran into this deadlock in our out-of-tree code. AFAICS there isn't an in-tree user that is currently triggering it because no one is crazy enough to have the usb tranciever (threads) be part of the (derrived) `Port` class, but there's nothing preventing it from happening in the future.

The approximate backtrace for the main thead (from memory) is something like this:
```
libusb_close()
....
LibUsbAdaptor::Close()
AsyncUsbSender::~AsyncUsbSender()
MyAsyncUsbSender::~MyAsyncUsbSender()
MyPort::~MyPort()
Device::DeleteAllPorts()
~Device::Stop()
AsyncPluginImpl::ShutdownDevice()
...

```

while the libusb thread is stuck waiting for the future `f.Get()` call in `AsyncPluginImpl::DeviceEvent`.

For details on the `libusb_close` behavour see the `libusb_close` definition `libusb/core.c`, it has this:

```
	/* Similarly to libusb_open(), we want to interrupt all event handlers
	 * at this point. More importantly, we want to perform the actual close of
	 * the device while holding the event handling lock (preventing any other
	 * thread from doing event handling) because we will be removing a file
	 * descriptor from the polling loop. If this is being called by the current
	 * event handler, we can bypass the interruption code because we already
	 * hold the event handling lock. */

	if (!handling_events) {
                [...]
		/* take event handling lock */
		libusb_lock_events(ctx);
	}
```

and IIRC libusb_lock_events() is where things deadlock.

If you'd like to see the details I could probably post the relevant parts of our code somewhere, it's just not really ready for inclusion is any way shape or form :)

kripton:
@peternewman, thanks for pinging me here. I can also follow the reasoning of the change but just as you, I wouldn't know if it breaks something. I agree that @nomis52 might be able to comment on it. However since I assume it worked before #930 and nomis52 might just not have been aware of the problem, this could just be merged.
About merge conflicts with the changes I did in my DMX16 branch: true, that would be affected. Currently, I don't plan to create a PR for them since I'm not quite happy with the dongles performance. There are some options pending (like a USB serial based approach instead of USB HID) to be played around with :) Detailed reasoning will follow on the mailing list topic tomorrow.

kripton:
Oh, and I don't have any USB DMX dongle to actually check the code changes ;)
The RP2040-based dongle I made those changes for only works when it is plugged in before olad is started. So the unplugging and hotplug cases are hard to test using that dongle.
Most probable reason (a bit off-topic): The RP2040 dongle has the HID and CDC ACM interface. When libusb sees the device on hotplug, it cannot read the serial number since the kernel is about to configure the ACM interface. libusb tells me that the device is busy ...

Oh, just thinkig about it: I have one old Anyma uDMX around that I can test the code with :D

DanielG:
> When libusb sees the device on hotplug, it cannot read the serial number since the kernel is about to configure the ACM interface. libusb tells me that the device is busy ...

I think you want to be calling adaptor->DetachKernelDriver(usb_handle, <USB iface num>) before trying to access that stuff, see `DMXCProjectsNodleU1.cpp` for an example.

DanielG:
I did test this patch together with our out-of-tree widget support code and it works just fine. I don't have any other supported usbdmx box either though, just some usbpro ones.

DanielG:
@kripton I just found your branch that Peter was talking about (I think): https://github.com/OpenLightingProject/ola/compare/master...kripton:RP2040UsbDmx. I would ask you to consider submitting that as a PR if only so I can do some review. I also have some muti-universe support for usbdmx in my (private) patch queue using a different approach and I think we might want to hammer out a way to do this that works for both of us.

kripton:
@DanielG: I created a Draft-PR as #1713. I'm curious to see what your solution looks like :+1:

plugins/usbdmx/AsyncPluginImpl.cpp
peternewman: There's nothing wrong with it forcefulness wise, just grammatically it doesn't read quite right. We could go with this instead if you want, which is closer to your original:
```suggestion
 * This must be run within the main thread. The Device destructor below may
 * cause libusb_close() to be called, which will deadlock if the hotplug event
 * thread waits for it.
```

You could always chuck in an exclamation mark! :wink:

plugins/usbdmx/AsyncPluginImpl.cpp
peternewman: > You mean spit out a warning if ola_device is already NULL, right?

Yeah

> IIRC ola_device being NULL here is legitimate because of the way that StartAandRegisterDevice is called by the factory through BaseWidgetFactory::AddWidget and AsyncPluginImpl::NewWidget. The Factory can actually delay calling AddWidget instead of doing it right away in the DeviceAdded function like normal. If the usb device is then unplugged before it does end up calling AddWidget then ola_device will be NULL here.

Makes sense. It might be worth adding one as OLA_DEBUG then; it sounds like it should only happen in a few rare circumstances, so being able to see it in a log could be handy if its firing at other times.

peternewman:
> Indeed I ran into this deadlock in our out-of-tree code. AFAICS there isn't an in-tree user that is currently triggering it because no one is crazy enough to have the usb tranciever (threads) be part of the (derrived) `Port` class, but there's nothing preventing it from happening in the future.

> If you'd like to see the details I could probably post the relevant parts of our code somewhere, it's just not really ready for inclusion is any way shape or form :)

No need for that, if you've hit the issue, made the change and its fixed it, that's good enough for me.

peternewman (changes requested):
So from my perspective, LGTM, if we could fix those two minor issues then I'm happy to merge.

Logistics wise, I know #1627 slipped through the net before, is there any benefit to that going in before or afterwards?

DanielG:
> So from my perspective, LGTM, if we could fix those two minor issues then I'm happy to merge.

Awesome, I'll try to get a new revision out by next week.

> Logistics wise, I know #1627 slipped through the net before

No worries, I never had time to fix the CI failures there but I'll get that cleaned up.

FYI @kripton you might want to have a look at that PR since your branch has a vaguely similar multi-transfer mechanism.

> is there any benefit to that going in before or afterwards?

I don't think the order makes any difference.

peternewman:
> > Logistics wise, I know #1627 slipped through the net before
> 
> No worries, I never had time to fix the CI failures there but I'll get that cleaned up.

Looks like it may just be a space around a comment or something trivial!

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: I've added the debug output and a comment explaining the situation when this might legitimately arise.

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: I've applied your suggestion and added a bit more detail around where/how things would deadlock.

plugins/usbdmx/AsyncPluginImpl.cpp
peternewman: Thanks.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -264,16 +264,9 @@ void AsyncPluginImpl::DeviceEvent(HotplugAgent::EventType event,
...
-      m_plugin_adaptor->Execute(
-          NewSingleCallback(this, &AsyncPluginImpl::ShutdownDevice,
-                            state->ola_device, &f));
-      f.Get();
-      state->ola_device = NULL;
-    }
-    state->DeleteWidget();
+    // Unregister & delete the device and widget in the main thread.
+    m_plugin_adaptor->Execute(
+      NewSingleCallback(this, &AsyncPluginImpl::ShutdownDeviceState, state));
```
peternewman, line 269: Nit: Spacing tweak
```suggestion
    m_plugin_adaptor->Execute(
        NewSingleCallback(this, &AsyncPluginImpl::ShutdownDeviceState, state));
```

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -353,17 +346,26 @@ bool AsyncPluginImpl::StartAndRegisterDevice(Widget *widget, Device *device) {
...
+    Device *device = state->ola_device;
+    m_plugin_adaptor->UnregisterDevice(device);
+    device->Stop();
+    delete device;
+    state->ola_device = NULL;
+  } else {
+          /* This case can be legitimate when the widget setup through the
+           * widget factory is delayed and an unplug event happens before
+           * that is completed. */
+          OLA_DEBUG << "device was NULL at shutdown";
```
peternewman, line 366: Nit:
```suggestion
          OLA_DEBUG << "ola_device was NULL at shutdown";
```

peternewman (changes requested):
LGTM within my knowledge of the code @DanielG 

Happy to merge given the tests before and afterwards that you've done. Would you mind resolving the two outstanding nits please.

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: Seems to me this file uses two space indent so I don't really understand why it's four spaces here, but since other parameter list continuations also use four space I suppose it's consistent at least :)

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: Done

DanielG:
@peternewman This is ready for another round of review, also the CI seems to need approval: "First-time contributors need a maintainer to approve running workflows." despite this not being my first rodeo. Weird.

plugins/usbdmx/AsyncPluginImpl.cpp
```diff
@@ -353,17 +346,26 @@ bool AsyncPluginImpl::StartAndRegisterDevice(Widget *widget, Device *device) {
...
+    Device *device = state->ola_device;
+    m_plugin_adaptor->UnregisterDevice(device);
+    device->Stop();
+    delete device;
+    state->ola_device = NULL;
+  } else {
+          /* This case can be legitimate when the widget setup through the
+           * widget factory is delayed and an unplug event happens before
+           * that is completed. */
+          OLA_DEBUG << "ola_device was NULL at shutdown";
```
peternewman, line 366: Because of the push I can't tell if I missed this before or it's newly introduced:
```suggestion
    /* This case can be legitimate when the widget setup through the
     * widget factory is delayed and an unplug event happens before
     * that is completed. */
    OLA_DEBUG << "ola_device was NULL at shutdown";
```

plugins/usbdmx/AsyncPluginImpl.cpp
peternewman: Thanks. It's precisely because of that, it's parameter continuation, not another logical level of nesting.

This is essentially just wrapping to keep the line an acceptable length so we want to differentiate it from control flow.

peternewman (changes requested):
Sorry just one more minor style nit please.

plugins/usbdmx/AsyncPluginImpl.cpp
DanielG: No that was likely alread there. Done.

DanielG:
No worries, thanks for the quick review :)

peternewman (approved):
Thanks for this @DanielG !

I assume you've tested the latest iteration?

It would be nice for backwards compatibility if we could keep the old method too, but I'm guessing we can't do that without risking causing the issue we're trying to fix!

DanielG:
Just to be clear: I have not tested the post-review version. I think we only made cosmetic changes though so I wouldn't expect any breakage having been introduced. Unfortunately my test bed for this is torn down ATM. So if you want it tested again it'll have to sit around until I get around to the project this came from again.

As for your backwards compat point, yeah, the whole point is to prevent a deadlock so I don't see how we can do that while keeping DeleteWidget in the libusb thread.

peternewman:
> Just to be clear: I have not tested the post-review version. I think we only made cosmetic changes though so I wouldn't expect any breakage having been introduced.

Yet another reason to hate force pushing! :)

> Unfortunately my test bed for this is torn down ATM. So if you want it tested again it'll have to sit around until I get around to the project this came from again.

Okay I've merged it anyway, hopefully myself or someone will notice fairly sharpish if this breaks things!

> As for your backwards compat point, yeah, the whole point is to prevent a deadlock so I don't see how we can do that while keeping DeleteWidget in the libusb thread.

Yep, makes sense, just thought I should check.

PR #1716, opened 2021-02-18
Relax timing on ClientWrapperTest

### LINKED ISSUE #1714: Python ClientWrapperTest fails with protobuf 3.14.0
This has apparently been a known issue for a while, but I've narrowed down the circumstances it can happen in while working on OpenLightingProject/website#14.  Warning - it's a bit involved.

When protobuf is installed from pip (i.e. `pip3 install protobuf`), ClientWrapperTest fails.  When protobuf is installed from the system package manager (i.e. `apt-get install python3-protobuf`), the test passes.  Checked with the latest commit in master (a95a6f7 currently) with Ubuntu 20.10 and Ubuntu 20.04.  I haven't properly checked Fedora yet (only a cursory test while writing docs), but I suspect it will have the same issue for reasons detailed below.

Some speculation:

My first thought was an issue between Python2 packages and Python3 packages.  These test systems only have Python3, so that's not it.

Further research reveals that GNOME distros (or, at least, Ubuntu and Fedora) come with `gnome-online-accounts`, which itself depends on `python3-protobuf`.  This means that GNOME users will likely have protobuf for python already installed on their systems.  If they install deps from pip (e.g. with `pip3 install protobuf numpy`), pip won't install another protobuf as it sees the system version already.  This also explains why the issue occurs on headless systems (i.e. Travis), as those systems won't have the preinstalled python protobuf and will install it from pip if you ask.

But what's the difference between the two sources?  My instinct is that something changed in newer Protobuf releases (I know this is a regular problem here).  The latest release on pip is [3.14.0](https://pypi.org/project/protobuf/#history).  Ubuntu 20.10 will get you [3.12.3](https://packages.ubuntu.com/groovy/python3-protobuf).  Ubuntu 20.04 provides [3.6.1](https://packages.ubuntu.com/focal/python3-protobuf).  Fedora 33 has [3.12.4](https://fedora.pkgs.org/33/fedora-x86_64/python3-protobuf-3.12.4-1.fc33.noarch.rpm.html).  Fedora 32 has [3.11.2](https://fedora.pkgs.org/32/fedora-x86_64/python3-protobuf-3.11.2-2.fc32.noarch.rpm.html).

To confirm this theory, I installed protobuf 3.12.4 from pip and ran tests, which all passed.  protobuf 3.13.0 also passes.  This leads me to conclude something about protobuf 3.14 is causing problems.  Caveat - I havn't tested any of this with Python2.

Test log:
```
==================================
   OLA 0.10.8: ./test-suite.log
==================================

# TOTAL: 95
# PASS:  94
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: python/ola/ClientWrapperTest.sh
=====================================

...F/usr/lib/python3.8/unittest/case.py:704: ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
  outcome.errors.clear()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
.
======================================================================
FAIL: testEventLoop (__main__.ClientWrapperTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./python/ola/ClientWrapperTest.py", line 195, in testEventLoop
    self.assertTrue(results.d_called - self.start >=
AssertionError: False is not true

----------------------------------------------------------------------
Ran 5 tests in 0.039s

FAILED (failures=1)
FAIL python/ola/ClientWrapperTest.sh (exit status: 1)
```

danielskeenan:
Fixes #1714.

I allowed up to 500 microseconds of drift in the checks, this way we still catch anything that's wildly off but allow for some minor variance in Python's execution.

### DIFF
--- a/python/ola/ClientWrapperTest.py
+++ b/python/ola/ClientWrapperTest.py
@@ -191,7 +191,7 @@ class ClientWrapperTest(unittest.TestCase):
     # Check when the callbacks were called. Allow 500 microseconds of drift.
     # Called immediately
     a_diff = results.a_called - self.start
-    self.assertAlmostEqual(a_diff, datetime.timedelta(microseconds=0),
+    self.assertAlmostEqual(a_diff, datetime.timedelta(milliseconds=0),
                            delta=datetime.timedelta(microseconds=500))
 
     # Called in 5 milliseconds
--- a/python/ola/ClientWrapperTest.py
+++ b/python/ola/ClientWrapperTest.py
@@ -173,6 +173,7 @@ class ClientWrapperTest(unittest.TestCase):
     wrapper.AddEvent(datetime.timedelta(milliseconds=5), b)
     wrapper.AddEvent(10, c)
 
+    # Nothing has been called yet
     self.assertIsNone(results.a_called)
     self.assertIsNone(results.b_called)
     self.assertIsNone(results.c_called)
@@ -181,19 +182,32 @@ class ClientWrapperTest(unittest.TestCase):
     self.start = datetime.datetime.now()
     wrapper.Run()
 
+    # Everything has been called
     self.assertIsNotNone(results.a_called)
     self.assertIsNotNone(results.b_called)
     self.assertIsNotNone(results.c_called)
     self.assertIsNotNone(results.d_called)
 
-    self.assertTrue(results.a_called - self.start <
-                    datetime.timedelta(milliseconds=5))
-    self.assertTrue(results.b_called - self.start >=
-                    datetime.timedelta(milliseconds=5))
-    self.assertTrue(results.c_called - self.start >=
-                    datetime.timedelta(milliseconds=10))
-    self.assertTrue(results.d_called - self.start >=
-                    datetime.timedelta(milliseconds=15))
+    # Check when the callbacks were called. Allow 500 microseconds of drift.
+    # Called immediately
+    a_diff = results.a_called - self.start
+    self.assertAlmostEqual(a_diff, datetime.timedelta(microseconds=0),
+                           delta=datetime.timedelta(microseconds=500))
+
+    # Called in 5 milliseconds
+    b_diff = results.b_called - self.start
+    self.assertAlmostEqual(b_diff, datetime.timedelta(milliseconds=5),
+                           delta=datetime.timedelta(microseconds=500))
+
+    # Called in 10 milliseconds
+    c_diff = results.c_called - self.start
+    self.assertAlmostEqual(c_diff, datetime.timedelta(milliseconds=10),
+                           delta=datetime.timedelta(microseconds=500))
+
+    # Called in 15 milliseconds
+    d_diff = results.d_called - self.start
+    self.assertAlmostEqual(d_diff, datetime.timedelta(milliseconds=15),
+                           delta=datetime.timedelta(microseconds=500))
 
     sockets[0].close()
     sockets[1].close()
@@ -229,6 +243,9 @@ class ClientWrapperTest(unittest.TestCase):
     wrapper.Run()
 
     self.assertTrue(results.gotdata)
+
+    sockets[0].close()
+    sockets[1].close()
 
 
 if __name__ == '__main__':


python/ola/ClientWrapperTest.py
```diff
@@ -173,6 +173,7 @@ def d():
     wrapper.AddEvent(datetime.timedelta(milliseconds=5), b)
     wrapper.AddEvent(10, c)
 
+    # Nothing is called yet
```
peternewman, line 176: Probably a little clearer:
```suggestion
    # Nothing has been called yet
```

python/ola/ClientWrapperTest.py
```diff
@@ -230,6 +244,9 @@ def DataCallback(self):
 
     self.assertTrue(results.gotdata)
 
+    sockets[0].close()
+    sockets[1].close()
+
```
peternewman, line 249: Thanks for fixing this too!

python/ola/ClientWrapperTest.py
```diff
@@ -181,19 +182,32 @@ def d():
...
+    # Called in 5 milliseconds
+    b_diff = results.b_called - self.start
+    self.assertLessEqual(abs(b_diff - datetime.timedelta(milliseconds=5)),
+                         datetime.timedelta(microseconds=500))
+
+    # Called in 10 milliseconds
+    c_diff = results.c_called - self.start
+    self.assertLessEqual(abs(c_diff - datetime.timedelta(milliseconds=10)),
+                            datetime.timedelta(microseconds=500))
```
peternewman, line 205: Spacing, align with Equal('s bracket.

python/ola/ClientWrapperTest.py
```diff
@@ -181,19 +182,32 @@ def d():
...
+
+    # Called in 15 milliseconds
+    d_diff = results.d_called - self.start
+    self.assertLessEqual(abs(d_diff - datetime.timedelta(milliseconds=15)),
```
peternewman, line 209: Again

python/ola/ClientWrapperTest.py
```diff
@@ -181,19 +182,32 @@ def d():
...
-                    datetime.timedelta(milliseconds=5))
-    self.assertTrue(results.c_called - self.start >=
-                    datetime.timedelta(milliseconds=10))
-    self.assertTrue(results.d_called - self.start >=
-                    datetime.timedelta(milliseconds=15))
+    # Check when the callbacks were called. Allow 500 microseconds of drift.
+    # Called immediately
+    a_diff = results.a_called - self.start
+    self.assertLessEqual(a_diff,
+                         datetime.timedelta(microseconds=500))
```
peternewman, line 195: It looks like https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertAlmostEqual and feeding in a delta should work and will likely give some far more friendly diagnostics, given if it fails you actually want to know the two numeric values being compared, not the delta.

Untested, but basically this:
```suggestion
    a_diff = results.a_called - self.start
    self.assertAlmostEqual(results.a_called - self.start, 0, None, None, timedelta(microseconds=500))
```

And equivalent for the others.

peternewman (changes requested):
As a bug-fix, please can this target 0.10 branch rather than master.

Some other comments too.

danielskeenan:
Yikes, git-splosion.  Let me try that again...

danielskeenan:
Should be better now... every time I think I know what `rebase` will do I'm surprised...

peternewman:
> Should be better now... every time I think I know what `rebase` will do I'm surprised...

I've always given up and just cherry-picked in the past!

python/ola/ClientWrapperTest.py
```diff
@@ -181,19 +182,32 @@ def d():
...
-                    datetime.timedelta(milliseconds=5))
-    self.assertTrue(results.c_called - self.start >=
-                    datetime.timedelta(milliseconds=10))
-    self.assertTrue(results.d_called - self.start >=
-                    datetime.timedelta(milliseconds=15))
+    # Check when the callbacks were called. Allow 500 microseconds of drift.
+    # Called immediately
+    a_diff = results.a_called - self.start
+    self.assertAlmostEqual(a_diff, datetime.timedelta(microseconds=0),
+                           delta=datetime.timedelta(microseconds=500))
```
peternewman, line 195: Personally I'd go for something like this, but I don't mind leaving a_diff if you'd rather.

```suggestion
    self.assertAlmostEqual((results.a_called - self.start), datetime.timedelta(microseconds=0),
                           delta=datetime.timedelta(microseconds=500))
```

Also have you done some dummy testing of this just to check assertAlmostEqual does actually work, e.g. change the offset of the B side a bit to confirm it passes (and breaks as appropriate)?

python/ola/ClientWrapperTest.py
```diff
@@ -181,19 +182,32 @@ def d():
...
-    self.assertTrue(results.b_called - self.start >=
-                    datetime.timedelta(milliseconds=5))
-    self.assertTrue(results.c_called - self.start >=
-                    datetime.timedelta(milliseconds=10))
-    self.assertTrue(results.d_called - self.start >=
-                    datetime.timedelta(milliseconds=15))
+    # Check when the callbacks were called. Allow 500 microseconds of drift.
+    # Called immediately
+    a_diff = results.a_called - self.start
+    self.assertAlmostEqual(a_diff, datetime.timedelta(microseconds=0),
```
peternewman, line 194: While it's no technical change to the output, this will avoid any typos or mistakes in future!
```suggestion
    self.assertAlmostEqual(a_diff, datetime.timedelta(microseconds=0),
```

peternewman (changes requested):
A minor style change, a style suggestion and a query. Other than that, pending CI this looks good to merge thanks!

python/ola/ClientWrapperTest.py
danielskeenan: Can confirm the test does what it says on the tin (the output when it fails is a bit verbose because of the `datetime.timedelta` everywhere).

I went with the local var mostly because I think it makes the test output more clear at a glance when it fails, like:

```
Traceback (most recent call last):
  File "./python/ola/ClientWrapperTest.py", line 194, in testEventLoop
    self.assertAlmostEqual(a_diff, datetime.timedelta(microseconds=0),
AssertionError: datetime.timedelta(microseconds=4) != datetime.timedelta(0) within datetime.timedelta(0) delta (datetime.timedelta(microseconds=4) difference)
```

python/ola/ClientWrapperTest.py
danielskeenan: Not sure what you're asking?  Both lines appear the same.

peternewman: :facepalm: Edited!

python/ola/ClientWrapperTest.py
peternewman: Fair enough, sounds good!

peternewman (approved):
LGTM thanks

peternewman:
Looks like it's perhaps still not relaxed enough on Arm on a VM (sometimes at least) https://travis-ci.org/github/OpenLightingProject/ola/jobs/759708456 :
FAIL: testEventLoop (__main__.ClientWrapperTest)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "../../python/ola/ClientWrapperTest.py", line 201, in testEventLoop

    delta=datetime.timedelta(microseconds=500))

AssertionError: datetime.timedelta(0, 0, 5709) != datetime.timedelta(0, 0, 5000) within datetime.timedelta(0, 0, 500) delta

PR #1717, opened 2021-03-02
Set a default value for the enabling Eurolite Mk2 preference

peternewman:
CC @ridicolos and @aroffringa

This sets a default value (of false) for enable_eurolite_mk2 and writes it into the relevant conf file making things easier for the user.

Sorry @aroffringa I'm not sure why I didn't suggest this before.

### DIFF
--- a/plugins/usbdmx/EuroliteProFactory.h
+++ b/plugins/usbdmx/EuroliteProFactory.h
@@ -45,6 +45,8 @@ class EuroliteProFactory : public BaseWidgetFactory<class EurolitePro> {
 
   static bool IsEuroliteMk2Enabled(Preferences *preferences);
 
+  static const char ENABLE_EUROLITE_MK2_KEY[];
+
  private:
   ola::usb::LibUsbAdaptor *m_adaptor;
   bool m_enable_eurolite_mk2;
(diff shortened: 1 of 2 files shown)


ridicolos:
Hey @peternewman, I will test this feature in the course of the week. I will give you an update then.

peternewman:
> Hey @peternewman, I will test this feature in the course of the week. I will give you an update then.

Thanks @ridicolos . There's really no reason it should fail, and I've confirmed it generates the default false option correctly, but not having the hardware it would be nice to know I haven't somehow broken that.

ridicolos:
@peternewman so on friday the date is set. There I can test the whole thing. The last few days I have unfortunately not got my hands on the hardware.

peternewman:
> @peternewman so on friday the date is set. There I can test the whole thing. The last few days I have unfortunately not got my hands on the hardware.

Thanks @ridicolos , no huge rush, Friday is fine!

ridicolos:
So I tried it now. I installed OLA from scratch with this branch.

After I ran `olad` I checked the ola-usbdmx.conf. It was set to `enable_eurolite_mk2 = false`. I deleted that line, started olad again and it immediately added the line with `false` again. 

`olad -l 4` gave the following output:

```
plugins/usbdmx/EuroliteProFactory.cpp:107: Connected FTDI device could be a Eurolite USB-DMX512-PRO MK2 but was ignored, because enable_eurolite_mk2 was false.
```

If I set it to `true` manually in the config it works.

The rest of the output you can check here: <https://pastebin.com/wPDSXgpf>

peternewman:
> If I set it to `true` manually in the config it works.

Perfect thanks. I'd done the other basic tests myself, but it's always reassuring to know it hasn't somehow broken the support for the device in some weird way, I'll get this merged now.

PR #1718, opened 2021-03-09
Sigrok based RDM sniffer

peternewman:
This supports a far wider range of devices than just Saleae devices.

- [ ] Extract common LogicReader code into another class so sigrok and Saleae just deal with getting data in
- [ ] Connected check for sigrok
- [ ] Readme
- [ ] Separate folder/packaging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -226,6 +226,8 @@ tools/e133/slp_sa_test
 tools/e133/slp_sa_test.exe
 tools/logic/logic_rdm_sniffer
 tools/logic/logic_rdm_sniffer.exe
+tools/logic/sigrok_rdm_sniffer
+tools/logic/sigrok_rdm_sniffer.exe
 tools/ola_trigger/config.tab.cpp
 tools/ola_trigger/config.tab.h
 tools/ola_trigger/lex.yy.cpp
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ addons:
      - libavahi-glib1
      - libncurses5-dev
      - libmicrohttpd-dev
+     - libsigrok-dev
     packages: &core_build_gpp_latest
      - *core_build
      - g++-10
@@ -68,6 +69,7 @@ addons:
      # Looks like this is Python 3 only, so install via pip
      # - numpy
      - libusb
+     # Sigrok
     packages: &mac_cppunit_113
      # Installing via the addon isn't currently working so do it the old fashioned way below
      # - https://raw.githubusercontent.com/Homebrew/homebrew-core/e6e43cf6a3^/Formula/cppunit.rb # install a slightly older cppunit, as latest needs C++11 support
--- a/configure.ac
+++ b/configure.ac
@@ -567,13 +567,21 @@ PKG_CHECK_MODULES([libsigrok], [libsigrok], [have_libsigrok="yes"], [have_libsig
 AM_CONDITIONAL([HAVE_LIBSIGROK],
                [test "x$have_libsigrok" = xyes])
 
-have_libsigrok_session="no"
+have_libsigrok_context="no"
 AS_IF([test "x$have_libsigrok" != xno],
-      [PKG_CHECK_MODULES([libsigrok_session], [libsigrok >= 0.4.0],
-                         [have_libsigrok_session="yes"], [true])])
-AS_IF([test "x$have_libsigrok_session" = xyes],
-      [AC_DEFINE([HAVE_LIBSIGROK_SESSION], [1],
-                 [define if libsigrok is installed and has sessions])])
+      [PKG_CHECK_MODULES([libsigrok_context], [libsigrok >= 0.4.0],
+                         [have_libsigrok_context="yes"], [true])])
+AS_IF([test "x$have_libsigrok_context" = xyes],
+      [AC_DEFINE([HAVE_LIBSIGROK_CONTEXT], [1],
+                 [define if libsigrok is installed and uses it's context])])
+
+have_libsigrok_dev_inst_opaque="no"
+AS_IF([test "x$have_libsigrok" != xno],
+      [PKG_CHECK_MODULES([libsigrok_context], [libsigrok >= 0.4.0],
+                         [have_libsigrok_dev_inst_opaque="yes"], [true])])
+AS_IF([test "x$have_libsigrok_dev_inst_opaque" = xyes],
+      [AC_DEFINE([HAVE_LIBSIGROK_DEV_INST_OPAQUE], [1],
+                 [define if libsigrok is installed and the sr_dev_inst struct is opaque])])
 
 # Features
 #####################################################
--- a/configure.ac
+++ b/configure.ac
@@ -567,6 +567,14 @@ PKG_CHECK_MODULES([libsigrok], [libsigrok], [have_libsigrok="yes"], [have_libsig
 AM_CONDITIONAL([HAVE_LIBSIGROK],
                [test "x$have_libsigrok" = xyes])
 
+have_libsigrok_session="no"
+AS_IF([test "x$have_libsigrok" != xno],
+      [PKG_CHECK_MODULES([libsigrok_session], [libsigrok >= 0.4.0],
+                         [have_libsigrok_session="yes"], [true])])
+AS_IF([test "x$have_libsigrok_session" = xyes],
+      [AC_DEFINE([HAVE_LIBSIGROK_SESSION], [1],
+                 [define if libsigrok is installed and has sessions])])
+
 # Features
 #####################################################
 
--- a/configure.ac
+++ b/configure.ac
@@ -559,9 +559,13 @@ AS_IF([test "x$have_avahi" = xyes],
       [AC_DEFINE([HAVE_AVAHI], [1], [Defined to use Avahi])])
 AM_CONDITIONAL([HAVE_AVAHI], [test "x$have_avahi" = xyes])
 
-# Look for -lSaleaeDevice, if we have it we build the logic sniffer.
+# Look for -lSaleaeDevice, if we have it we build the Saleae logic sniffer.
 SALEAE_DEVICE
 
+# Look for -lsigrok, if we have it we build the sigrok logic sniffer.
+PKG_CHECK_MODULES([libsigrok], [libsigrok], [have_libsigrok="yes"], [have_libsigrok="no"])
+AM_CONDITIONAL([HAVE_LIBSIGROK],
+               [test "x$have_libsigrok" = xyes])
 
 # Features
 #####################################################
--- a/tools/logic/DMXSignalProcessor.cpp
+++ b/tools/logic/DMXSignalProcessor.cpp
@@ -69,6 +69,7 @@ DMXSignalProcessor::DMXSignalProcessor(DataCallback *callback,
       m_ticks(0),
       m_may_be_in_break(false),
       m_ticks_in_break(0) {
+  OLA_DEBUG << "Microseconds per tick " << m_microseconds_per_tick;
   if (m_sample_rate % DMX_BITRATE) {
     OLA_WARN << "Sample rate is not a multiple of " << DMX_BITRATE;
   }
--- a/tools/logic/DMXSignalProcessor.cpp
+++ b/tools/logic/DMXSignalProcessor.cpp
@@ -118,7 +118,7 @@ void DMXSignalProcessor::ProcessSample(bool bit) {
           SetState(MAB);
         } else {
           OLA_WARN << "Break too short, was " << TicksAsMicroSeconds()
-                   << " us";
+                   << " us (" << m_ticks << " ticks)";
           SetState(IDLE);
         }
       } else {
@@ -136,7 +136,8 @@ void DMXSignalProcessor::ProcessSample(bool bit) {
           // OLA_INFO << "In start bit!";
           SetState(START_BIT);
         } else {
-          OLA_WARN << "Mark too short, was " << TicksAsMicroSeconds() << "us";
+          OLA_WARN << "Mark too short, was " << TicksAsMicroSeconds() << "us ("
+                   << m_ticks << " ticks)";
           SetState(UNDEFINED);
         }
       }
@@ -222,7 +223,7 @@ void DMXSignalProcessor::ProcessBit(bool bit) {
       SetState(static_cast<State>(m_state + 1));
     } else {
       OLA_WARN << "Bit " << m_state << " was too short, was "
-               << TicksAsMicroSeconds() << "us";
+               << TicksAsMicroSeconds() << "us (" << m_ticks << " ticks)";
       SetState(UNDEFINED);
     }
   }
@@ -283,7 +284,7 @@ void DMXSignalProcessor::HandleFrame() {
  */
 void DMXSignalProcessor::SetState(State state, unsigned int ticks) {
   OLA_INFO << "Transition to " << state << ", prev duration was "
-           << TicksAsMicroSeconds();
+           << TicksAsMicroSeconds() << " us (" << m_ticks << " ticks)";
   m_state = state;
   m_ticks = ticks;
   if (state == UNDEFINED) {
--- a/tools/logic/Makefile.mk
+++ b/tools/logic/Makefile.mk
@@ -2,11 +2,24 @@ if HAVE_SALEAE_LOGIC
 bin_PROGRAMS += tools/logic/logic_rdm_sniffer
 endif
 
+if HAVE_LIBSIGROK
+bin_PROGRAMS += tools/logic/sigrok_rdm_sniffer
+endif
+
 tools_logic_logic_rdm_sniffer_SOURCES = \
     tools/logic/DMXSignalProcessor.cpp \
     tools/logic/DMXSignalProcessor.h \
     tools/logic/logic-rdm-sniffer.cpp
 tools_logic_logic_rdm_sniffer_LDADD = common/libolacommon.la \
                                       $(libSaleaeDevice_LIBS)
+
+tools_logic_sigrok_rdm_sniffer_SOURCES = \
+    tools/logic/DMXSignalProcessor.cpp \
+    tools/logic/DMXSignalProcessor.h \
+    tools/logic/sigrok-rdm-sniffer.cpp
+tools_logic_sigrok_rdm_sniffer_CXXFLAGS = $(COMMON_CXXFLAGS) \
+                                          $(libsigrok_CFLAGS)
+tools_logic_sigrok_rdm_sniffer_LDADD = common/libolacommon.la \
+                                       $(libsigrok_LIBS)
 
 EXTRA_DIST += tools/logic/README.md
--- a/tools/logic/logic-rdm-sniffer.cpp
+++ b/tools/logic/logic-rdm-sniffer.cpp
@@ -90,6 +90,10 @@ DEFINE_s_default_bool(display_dmx, d, false,
 DEFINE_uint16(dmx_slot_limit, ola::DMX_UNIVERSE_SIZE,
               "Only display the first N slots of DMX data.");
 DEFINE_uint32(sample_rate, 4000000, "Sample rate in HZ.");
+// Set the sample rate.  Must be a supported value, i.e.:  24000000, 16000000,
+// 12000000, 8000000, 4000000, 2000000, 1000000, 500000, 250000, 200000,
+// 100000, 50000, 25000
+// TODO(Peter): Trap and report this, don't just crash
 DEFINE_string(pid_location, "",
               "The directory containing the PID definitions.");
 
--- a/tools/logic/sigrok-rdm-sniffer.cpp
+++ b/tools/logic/sigrok-rdm-sniffer.cpp
@@ -95,8 +95,8 @@ DEFINE_string(pid_location, "",
 //DEFINE_uint32(sigrok_log_level, SR_LOG_NONE, "Sigrok log level, from "
 //                + SR_LOG_NONE + " to " + SR_LOG_SPEW + ".");
 DEFINE_uint32(sigrok_log_level, SR_LOG_NONE, "Set the Sigrok logging level from 0 .. 5.");
-DEFINE_uint32(sigrok_samples, 200, "Set the Sigrok sample count.");
-DEFINE_uint32(sigrok_time, 2000, "Set the Sigrok sample time in ms.");
+DEFINE_uint32(sigrok_samples, 2000, "Limit capture to this many samples.");
+DEFINE_uint32(sigrok_time, 2000, "Limit capture to this many ms.");
 DEFINE_string(sigrok_device, "demo", "Set the Sigrok device to use.");
 
 #ifdef HAVE_LIBSIGROK_DEV_INST_OPAQUE
@@ -399,6 +399,8 @@ void *SigrokThread::Run() {
                 << sr_strerror(ret);
       return NULL;
     }
+  } else {
+    OLA_INFO << "Sigrok will capture continuously";
   }
 
 #ifdef HAVE_LIBSIGROK_CONTEXT
--- a/tools/logic/sigrok-rdm-sniffer.cpp
+++ b/tools/logic/sigrok-rdm-sniffer.cpp
@@ -109,7 +109,7 @@ DEFINE_string(sigrok_device, "demo", "Set the Sigrok device to use.");
 //                void *user_data);
 //void OnError(U64 device_id, void *user_data);
 //void ProcessData(uint8_t *data, uint64_t data_length);
-void ProcessData(const sr_datafeed_logic *logic);
+//void ProcessData(const sr_datafeed_logic *logic);
 
 class LogicReader {
  public:
@@ -157,7 +157,6 @@ class LogicReader {
 
 
     void ProcessData(uint8_t *data, uint64_t data_length);
-    //void ProcessData(uint64_t data_length);
     //void ProcessData(const sr_datafeed_logic *logic);
     void DisplayDMXFrame(const uint8_t *data, unsigned int length);
     void DisplayRDMFrame(const uint8_t *data, unsigned int length);
@@ -171,6 +170,13 @@ static void sigrok_feed_callback(const struct sr_dev_inst *sdi,
   sr_dev_driver *driver = SIGROK_DRIVER_FROM_INSTANCE(sdi);
 
   OLA_DEBUG << "Got feed callback for " << driver->name;
+
+  if (!cb_data) {
+    OLA_WARN << "Callback missing our data";
+    // Free data
+    //DevicesManagerInterface::DeleteU8ArrayPtr(data);
+    return;
+  }
 
   if (packet->type == SR_DF_HEADER) {
     OLA_INFO << "Sigrok header for " << driver->name << " got header";
@@ -354,12 +360,13 @@ void *SigrokThread::Run() {
   OLA_INFO << "New sample rate is " << g_variant_get_uint64(gvar) << "Hz";
   g_variant_unref(gvar);
 
-//  if ((ret = sr_config_set(sdi, NULL, SR_CONF_LIMIT_SAMPLES, g_variant_new_uint64(FLAGS_sigrok_samples))) != SR_OK) {
-//    OLA_FATAL << "Error setting config limit samples via libsigrok driver "
-//              << driver->name << " (" << sr_strerror_name(ret) << "): "
-//              << sr_strerror(ret);
-//    return NULL;
-//  }
+/*  // Just samples based
+  if ((ret = sr_config_set(sdi, NULL, SR_CONF_LIMIT_SAMPLES, g_variant_new_uint64(FLAGS_sigrok_samples))) != SR_OK) {
+    OLA_FATAL << "Error setting config limit samples via libsigrok driver "
+              << driver->name << " (" << sr_strerror_name(ret) << "): "
+              << sr_strerror(ret);
+    return NULL;
+  } */
 
   if (sr_dev_has_option(sdi, SR_CONF_LIMIT_MSEC)) {
     gvar = g_variant_new_uint64(FLAGS_sigrok_time);
@@ -384,13 +391,6 @@ void *SigrokThread::Run() {
       return NULL;
     }
   }
-
-/*    // Just samples based
-    gvar = g_variant_new_uint64(FLAGS_sigrok_samples);
-    if (sr_config_set(sdi, NULL, SR_CONF_LIMIT_SAMPLES, gvar) != SR_OK) {
-      OLA_FATAL << "Failed to configure time-based sample limit.";
-      return NULL;
-    }*/
 
 #ifdef HAVE_LIBSIGROK_CONTEXT
   if ((ret = sr_session_datafeed_callback_add(sr_sess, sigrok_feed_callback,
@@ -521,7 +521,6 @@ void LogicReader::DataReceived(uint8_t *data, uint64_t data_length) {
   OLA_DEBUG << "Got " << data_length << " samples";
   m_ss->Execute(
       NewSingleCallback(this, &LogicReader::ProcessData, data, data_length));
-//      NewSingleCallback(this, &LogicReader::ProcessData, data_length));
 //      NewSingleCallback(this, &LogicReader::ProcessData, logic));
 //  for (unsigned int i = 0 ; i < data_length; i++) {
 //    OLA_DEBUG << "Got sample (after) " << ToHex(data[i]);
@@ -572,7 +571,6 @@ void LogicReader::Stop() {
  * @param data_length the size of the data
  */
 void LogicReader::ProcessData(uint8_t *data, uint64_t data_length) {
-//void LogicReader::ProcessData(uint64_t data_length) {
 //void LogicReader::ProcessData(const sr_datafeed_logic *logic) {
   //OLA_DEBUG << "Got " << logic->length << " samples";
   OLA_DEBUG << "Got " << data_length << " samples";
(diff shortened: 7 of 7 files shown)


peternewman:
@kripton I noticed a sigrok related thing in your repos, I don't know if you fancy giving this a little test with whichever logic analyser you have? You might find it really useful for https://github.com/OpenLightingProject/rp2040-dongle/ too!

kripton:
Sure, I will give this a try the coming days. I have multiple `fx2la` devices here, one Open Bench Logic Sniffer and a scope that can be theoretically hooked up to sigrok as well :)

Since I already saw that yesterday:
* Do you think this could make it as a multi-port input-only OLA-plugin? ;)
* Are you aware of the "official" DMX512 decoder? https://github.com/sigrokproject/libsigrokdecode/blob/master/decoders/dmx512/pd.py Maybe some RDM decoding capabilities could be included there as well?

peternewman:
> Sure, I will give this a try the coming days. I have multiple `fx2la` devices here, one Open Bench Logic Sniffer and a scope that can be theoretically hooked up to sigrok as well :)

Cool, the logic sniffer should just work (it needs to be on pin 1 currently). In theory the scope should be possible although I haven't tried analogue signals yet, as I can't generate them unless the dummy device does. Happy to have a go at it though and it should be fairly simple too. If you try connecting the scope up can you share some -l 4 and --sigrok-log-level 5 captures of the setup.

>     * Do you think this could make it as a multi-port input-only OLA-plugin? ;)

I'd never even considered that. Yeah it should be possible and in theory each pin would be an input port with some minimal code changes. Obviously DMX only as it couldn't reply. Perhaps I'll try and mock it up as a client first as I'm not sure what the delay might be like from a signal arriving on the pin to getting into OLA, depending on the blocks of samples being transferred. Do you think people want multi-universe DMX only capture? I guess for a recording setup maybe.

>     * Are you aware of the "official" DMX512 decoder? https://github.com/sigrokproject/libsigrokdecode/blob/master/decoders/dmx512/pd.py Maybe some RDM decoding capabilities could be included there as well?

I am, I think I submitted some DMX and RDM samples in the past to try and help get it extended.

The main challenge with RDM is it's a lot more complicated. To get something high level you could just decode the command type and list the PID name, but depending on what exactly you're trying to do, you probably want to know the contents of the different PIDs, particularly the discovery stuff which is where things often go wrong, but possibly some other basic initial traffic, at which point you need to implement large chunks of:
https://github.com/OpenLightingProject/ola/blob/master/python/ola/PidStore.py
As well as the UID stuff:
https://github.com/OpenLightingProject/ola/blob/master/python/ola/UID.py

Plus our protobuf stuff to support that. At which point you've reimplemented large chunks of OLA. I guess we could just include OLA's library but I'm not sure if they're really interested in that sort of approach. Finally from my brief playing with PulseView/sigrok this time around, I wasn't really sure how suitable it actually was for that purpose. When you're doing discovery of a few fixtures you're quickly into tens of messages which just seemed to result in a lot of scrolling from my brief experience. Compared to something like Wireshark which seems better suited to that type of messaging.

All I've done in comparison in this PR is connected a different source of samples into the existing DMXSignalProcessor code, I've written very little new or novel code here.

I do mean to get in touch with the sigrok team about it, perhaps they'll be able to signpost to this or maybe they'll want to write a full decoder themselves.

kripton:
Okay, first results. I tried the scope in pulseview first. I was able to see my DMX frames but I was unable to do continuously capturing. And reading from the scope's memory (over LAN, didn't yet try USB) was about 64kByte/s. Don't know why but I stopped there.

Then I hooked up my FX2LAFW device. With pulseview no problems, continuous capture at 1MHz, 1 channel and DMX data was clearly visible. The internal decoder did lots of mis-interpretations but attached fixtures took the signal all well.

Finally, I tried the `sigrok_rdm_sniffer` but it didn't take the signal all that well:
```
kripton@momo / $ sigrok-cli --scan
The following devices were found:
demo - Demo device with 13 channels: D0 D1 D2 D3 D4 D5 D6 D7 A0 A1 A2 A3 A4
fx2lafw:conn=1.8 - sigrok FX2 LA (16ch) with 16 channels: D0 D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 D11 D12 D13 D14 D15
```

Results:
```
kripton@momo ~ $ sigrok_rdm_sniffer --sigrok-log-level 2 --sigrok-device "fx2lafw" -d -l 3
tools/logic/sigrok-rdm-sniffer.cpp:722: Running...
common/thread/Thread.cpp:200: Thread , policy SCHED_OTHER, priority 0
tools/logic/sigrok-rdm-sniffer.cpp:285: Got driver: fx2lafw - fx2lafw (generic driver for FX2 based LAs)
tools/logic/sigrok-rdm-sniffer.cpp:307: Found 1 devices
tools/logic/sigrok-rdm-sniffer.cpp:314: Found device (using first):
tools/logic/sigrok-rdm-sniffer.cpp:316:         Vendor: sigrok
tools/logic/sigrok-rdm-sniffer.cpp:318:         Model: FX2 LA (16ch)
tools/logic/sigrok-rdm-sniffer.cpp:320:         Version: 
tools/logic/sigrok-rdm-sniffer.cpp:347: Initial sample rate is 20000Hz
tools/logic/sigrok-rdm-sniffer.cpp:360: New sample rate is 4000000Hz
tools/logic/sigrok-rdm-sniffer.cpp:403: Sigrok will capture continuously
tools/logic/sigrok-rdm-sniffer.cpp:182: Sigrok header for fx2lafw got header
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 0 us (0 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 0.25 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 0.25 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 0.25 us (1 ticks)
(continues with lots of those messages, no valid DMX data seen)
```

I then changed the sample rate to 1MHz but results were similar:
```
kripton@momo ~ $ sigrok_rdm_sniffer --sigrok-log-level 2 --sigrok-device "fx2lafw" -d -l 3 --sample-rate 1000000
tools/logic/sigrok-rdm-sniffer.cpp:722: Running...
common/thread/Thread.cpp:200: Thread , policy SCHED_OTHER, priority 0
tools/logic/sigrok-rdm-sniffer.cpp:285: Got driver: fx2lafw - fx2lafw (generic driver for FX2 based LAs)
tools/logic/sigrok-rdm-sniffer.cpp:307: Found 1 devices
tools/logic/sigrok-rdm-sniffer.cpp:314: Found device (using first):
tools/logic/sigrok-rdm-sniffer.cpp:316:         Vendor: sigrok
tools/logic/sigrok-rdm-sniffer.cpp:318:         Model: FX2 LA (16ch)
tools/logic/sigrok-rdm-sniffer.cpp:320:         Version: 
tools/logic/sigrok-rdm-sniffer.cpp:347: Initial sample rate is 20000Hz
tools/logic/sigrok-rdm-sniffer.cpp:360: New sample rate is 1000000Hz
tools/logic/sigrok-rdm-sniffer.cpp:403: Sigrok will capture continuously
tools/logic/sigrok-rdm-sniffer.cpp:182: Sigrok header for fx2lafw got header
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 0 us (0 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 17 us (17 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
```

The DMX line is directly (= TTL level, no RS-485 bus drivers and receivers involved) hooked up `D0` (= `PB0` on my board). And pulseview is happy as well (picture uploading fails due to a CORS error?)

kripton:
And something about the cli parameters seems wrong. The FX2LA-device is shown in `sigrok-cli --scan` as `fx2lafw:conn=1.8`. However, when I specify this as value for `--sigrok-device` at the `sigrok_rdm_sniffer`, it tells me: `tools/logic/sigrok-rdm-sniffer.cpp:291: Failed to find driver fx2lafw:conn=1.8`. When I only specify the driver without giving a specific device, it will pick the first device offered by that driver.

With the scope (driver `rigol-ds`) it's different. The driver supports USB or LAN. USB might work with autodetect, LAN needs the IP address. `sigrok-cli -d rigol-ds:conn=tcp-raw/192.168.3.16/5555 -l 5 --scan` gives:
```
sr: [00:00.000001] log: libsigrok loglevel set to 5.
sr: [00:00.000060] backend: libsigrok 0.5.2/5:1:1.
sr: [00:00.000139] backend: Libs: glib 2.66.4 (rt: 2.66.7/6607:7), libzip 1.7.3, libserialport 0.1.1/1:0:1 (rt: 0.1.1/1:0:1), libusb-1.0 1.0.24.11586 API 0x01000108, hidapi 0.8.0-rc1, bluez 5.55, libftdi 1.5.
sr: [00:00.000160] backend: Host: x86_64-pc-linux-gnu, little-endian.
sr: [00:00.000172] backend: SCPI backends: TCP, serial, USBTMC.
sr: [00:00.000180] backend: Firmware search paths:
sr: [00:00.000252] backend:  - /home/kripton/.local/share/sigrok-firmware
sr: [00:00.000265] backend:  - /usr/share/sigrok-firmware
sr: [00:00.000273] backend:  - /usr/local/share/sigrok-firmware
sr: [00:00.000286] backend:  - /usr/share/sigrok-firmware
sr: [00:00.000325] backend: Sanity-checking all drivers.
sr: [00:00.000341] backend: Sanity-checking all input modules.
sr: [00:00.000355] backend: Sanity-checking all output modules.
sr: [00:00.000375] backend: Sanity-checking all transform modules.
srd: libsigrokdecode loglevel set to 5.
sr: [00:00.003782] hwdriver: sr_config_list(): key 2147418112 (NULL) sdi (nil) cg NULL -> [uint32 20000, 20001]
sr: [00:00.003817] scpi: Opening RAW TCP device tcp-raw/192.168.2.82/5555.
sr: [00:00.022901] scpi_tcp: Successfully sent SCPI command: '*IDN?'.
sr: [00:00.026305] scpi: Got response: 'RIGOL TECHNOLOGIES,DS1104Z,DS1ZA194820861,00.04.04.SP3', length 54.
sr: [00:00.026443] hwdriver: Scan found 1 devices (rigol-ds).
The following devices were found:
rigol-ds - Rigol DS1104Z 00.04.04.SP3 with 4 channels: CH1 CH2 CH3 CH4
sr: [00:00.026537] hwdriver: Cleaning up all drivers.
```

So the correct "sigrok device string" including driver and device would be `rigol-ds:conn=tcp-raw/192.168.3.16/5555`. This value fails with the rdm sniffer:
```
kripton@momo ~ $ sigrok_rdm_sniffer --sigrok-device "rigol-ds:conn=tcp-raw/192.168.2.82/5555" --sigrok-log-level 5
sr: [00:00.000000] log: libsigrok loglevel set to 5.
sr: [00:00.000089] backend: libsigrok 0.5.2/5:1:1.
sr: [00:00.000152] backend: Libs: glib 2.66.4 (rt: 2.66.7/6607:7), libzip 1.7.3, libserialport 0.1.1/1:0:1 (rt: 0.1.1/1:0:1), libusb-1.0 1.0.24.11586 API 0x01000108, hidapi 0.8.0-rc1, bluez 5.55, libftdi 1.5.
sr: [00:00.000169] backend: Host: x86_64-pc-linux-gnu, little-endian.
sr: [00:00.000179] backend: SCPI backends: TCP, serial, USBTMC.
sr: [00:00.000187] backend: Firmware search paths:
sr: [00:00.000221] backend:  - /home/kripton/.local/share/sigrok-firmware
sr: [00:00.000229] backend:  - /usr/share/sigrok-firmware
sr: [00:00.000235] backend:  - /usr/local/share/sigrok-firmware
sr: [00:00.000240] backend:  - /usr/share/sigrok-firmware
sr: [00:00.000265] backend: Sanity-checking all drivers.
sr: [00:00.000275] backend: Sanity-checking all input modules.
sr: [00:00.000282] backend: Sanity-checking all output modules.
sr: [00:00.000291] backend: Sanity-checking all transform modules.
tools/logic/sigrok-rdm-sniffer.cpp:291: Failed to find driver rigol-ds:conn=tcp-raw/192.168.2.82/5555
```

When only naming the driver "rigol-ds", without specifying a device, the driver is poked for a device but doesn't find one.

kripton:
Okay, the thing with the scope in pulseview is a bug in the current sigfox release. Fixed in git but no release since then:
https://github.com/sigrokproject/libsigrok/commit/704910e32ca97421e59009f1e384b9fc94a5a75b and https://github.com/sigrokproject/libsigrok/commit/1cbb3b1cfbd0334001bde052413cc7aa869d5994

peternewman:
> Okay, first results. I tried the scope in pulseview first. I was able to see my DMX frames but I was unable to do continuously capturing. And reading from the scope's memory (over LAN, didn't yet try USB) was about 64kByte/s. Don't know why but I stopped there.

I had issues with continuous capturing with a FX2LAFW in older releases. I assume your scope channels are listed as analogue, in which case they won't work until I write more code.

> Then I hooked up my FX2LAFW device. With pulseview no problems, continuous capture at 1MHz, 1 channel and DMX data was clearly visible. The internal decoder did lots of mis-interpretations but attached fixtures took the signal all well.

It probably doesn't bode too well if the built in decoder didn't like it. Was it just bit length issues?

> Finally, I tried the `sigrok_rdm_sniffer` but it didn't take the signal all that well:
> 
> ```
> kripton@momo / $ sigrok-cli --scan
> The following devices were found:
> demo - Demo device with 13 channels: D0 D1 D2 D3 D4 D5 D6 D7 A0 A1 A2 A3 A4
> fx2lafw:conn=1.8 - sigrok FX2 LA (16ch) with 16 channels: D0 D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 D11 D12 D13 D14 D15
> ```

Ah, I think I know what's happening, my naive implementation is ignore the channel and capture width so effectively is alternating between D0 and D9 values. I think you could prove this by either adding some printing to `DMXSignalProcessor::ProcessSample`, perhaps tying one high and one low, or electrically coupling D0 and D9 to the same input. I think I removed all the useful debug from those bits of code.

This would tie up with it changing nearly every tick regardless of the frequency.

> The DMX line is directly (= TTL level, no RS-485 bus drivers and receivers involved) hooked up `D0` (= `PB0` on my board). And pulseview is happy as well

Yeah that makes sense, and should work I think.

> picture uploading fails due to a CORS error?

GitHub does seem to have got a little flaky recently.

peternewman:
> And something about the cli parameters seems wrong. The FX2LA-device is shown in `sigrok-cli --scan` as `fx2lafw:conn=1.8`. However, when I specify this as value for `--sigrok-device` at the `sigrok_rdm_sniffer`, it tells me: `tools/logic/sigrok-rdm-sniffer.cpp:291: Failed to find driver fx2lafw:conn=1.8`. When I only specify the driver without giving a specific device, it will pick the first device offered by that driver.

The code for that bit was inspired by the collectd plugin:
https://collectd.org/wiki/index.php/Plugin:sigrok

I might try and offer both options, as the existing means you can juse use fx2lafw and not worry about which port it's connected to, but as you say you can't pass more options or potentially control which LA if you've got two.

> With the scope (driver `rigol-ds`) it's different. The driver supports USB or LAN. USB might work with autodetect, LAN needs the IP address. `sigrok-cli -d rigol-ds:conn=tcp-raw/192.168.3.16/5555 -l 5 --scan` gives:

Yeah I hadn't even considered those sort of settings.

kripton:
> I had issues with continuous capturing with a FX2LAFW in older releases. I assume your scope channels are listed as analogue, in which case they won't work until I write more code.

True, the DMX decoder only works on binary data. Pulseview can convert the analog channel to a digital one, that's not the issue. I've built libsigrok from git master by now and the capture is much more stable now. The DMX decoder is still flaky: The two STOP bits of the START CODE are detected properly. The, the START bit of channel 1 follows but the decoder interprets it as "INTERFRAME". Thus, every byte shifts the decoder by one bit.... Trying to attach a picture: (Failed again ...)

> It probably doesn't bode too well if the built in decoder didn't like it. Was it just bit length issues?

Nope, actually same as above. The built-in DMX decoder behaves the same, no matter if the data is captured via scope of FX2 device.
I'll open and decode the DMX samples from the sigrok homepage tomorrow.

> Ah, I think I know what's happening, my naive implementation is ignore the channel and capture width so effectively is alternating between D0 and D9 values. I think you could prove this by either adding some printing to `DMXSignalProcessor::ProcessSample`, perhaps tying one high and one low, or electrically coupling D0 and D9 to the same input. I think I removed all the useful debug from those bits of code.

Ah, you're saying that your LA has 8 channels, mine has 16 and your code is hard-coded for 8 channels?

Bridging PB0 (D0) to PD0 (D8) or PD1 (D9) doesn't change much, still lots of errors and no useful data.

However, the output is different:

DB0 bridged to PD0:
```
tools/logic/sigrok-rdm-sniffer.cpp:360: New sample rate is 1000000Hz
tools/logic/sigrok-rdm-sniffer.cpp:403: Sigrok will capture continuously
tools/logic/sigrok-rdm-sniffer.cpp:182: Sigrok header for fx2lafw got header
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 0 us (0 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 58 us (58 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 58 us (58 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 16 us (16 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 72 us (72 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 72 us (72 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 16 us (16 ticks)
``` 


DB0 bridged to PD1:
```
tools/logic/sigrok-rdm-sniffer.cpp:360: New sample rate is 1000000Hz
tools/logic/sigrok-rdm-sniffer.cpp:403: Sigrok will capture continuously
tools/logic/sigrok-rdm-sniffer.cpp:182: Sigrok header for fx2lafw got header
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 0 us (0 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 1 us (1 ticks)
tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 1 us (1 ticks
``` 

> GitHub does seem to have got a little flaky recently.

Yep, also experienced other issues the last two week :(

kripton:
Regarding the incorrect decoding of obviously valid DMX in pulseview: As with my scope: "Fixed in git, no release since then":
https://sigrok.org/bugzilla/show_bug.cgi?id=1442 and https://github.com/sigrokproject/libsigrokdecode/commit/dca19fbfdf0650dba693c0ac6213f6cdb748a8c3
Exactly what I have been experiencing. Looks like I also need to build libsigrokdecode from git HEAD ;)

kripton:
> I might try and offer both options, as the existing means you can juse use fx2lafw and not worry about which port it's connected to, but as you say you can't pass more options or potentially control which LA if you've got two.
> 
> > With the scope (driver `rigol-ds`) it's different. The driver supports USB or LAN. USB might work with autodetect, LAN needs the IP address. `sigrok-cli -d rigol-ds:conn=tcp-raw/192.168.3.16/5555 -l 5 --scan` gives:
> 
> Yeah I hadn't even considered those sort of settings.

So what about this: When only a driver is specified (as with `--sigrok-device fx2la`), it chooses the first device, just as now. If parameters (separated from the driver name by a colon) are present, they are passed to the driver.

I had to use the `--sigrok-device` parameter in any case. If it was missing, it used the `demo` device: `demo - Demo device with 13 channels: D0 D1 D2 D3 D4 D5 D6 D7 A0 A1 A2 A3 A4`. That of course might never produce valid DMX data.

peternewman:
> True, the DMX decoder only works on binary data. Pulseview can convert the analog channel to a digital one, that's not the issue.

No, but it will be an issue with this code, which rejects any non-logic packets.

> I've built libsigrok from git master by now and the capture is much more stable now.

That's good to hear.

> The DMX decoder is still flaky: The two STOP bits of the START CODE are detected properly. The, the START bit of channel 1 follows but the decoder interprets it as "INTERFRAME". Thus, every byte shifts the decoder by one bit.... Trying to attach a picture: (Failed again ...)

I wonder if that's our bug too?

> > Ah, I think I know what's happening, my naive implementation is ignore the channel and capture width so effectively is alternating between D0 and D9 values. I think you could prove this by either adding some printing to `DMXSignalProcessor::ProcessSample`, perhaps tying one high and one low, or electrically coupling D0 and D9 to the same input. I think I removed all the useful debug from those bits of code.
> 
> Ah, you're saying that your LA has 8 channels, mine has 16 and your code is hard-coded for 8 channels?

Yeah, it tries to process the first bit of every byte of sampled logic data, there's a flag present which I'm currently ignoring which would deal with that.

> Bridging PB0 (D0) to PD0 (D8) or PD1 (D9) doesn't change much, still lots of errors and no useful data.
> 
> However, the output is different:
> 
> DB0 bridged to PD0:
> 
> ```
> tools/logic/sigrok-rdm-sniffer.cpp:360: New sample rate is 1000000Hz
> tools/logic/sigrok-rdm-sniffer.cpp:403: Sigrok will capture continuously
> tools/logic/sigrok-rdm-sniffer.cpp:182: Sigrok header for fx2lafw got header
> tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 0 us (0 ticks)
> tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 58 us (58 ticks)
> tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 58 us (58 ticks)
> tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 16 us (16 ticks)
> tools/logic/DMXSignalProcessor.cpp:121: Break too short, was 72 us (72 ticks)
> tools/logic/DMXSignalProcessor.cpp:287: Transition to 1, prev duration was 72 us (72 ticks)
> tools/logic/DMXSignalProcessor.cpp:287: Transition to 2, prev duration was 16 us (16 ticks)
> ```

That does look more promising. It might be worth trying at 4Mhz (which is all we've really tested at). Did you wait a while for data to run through, and what channel data are you using (all full, all off, or real random data)? Come to think of it, I think my tests are still only sending 20 channels which probably isn't great either, but I think should work. In your case worst case it may have to process an entire frame before it finds the break and locks on. If you run it with `-d` and possibly lower the log levels, it should just output DMX lines every good frame.

peternewman:
> So what about this: When only a driver is specified (as with `--sigrok-device fx2la`), it chooses the first device, just as now. If parameters (separated from the driver name by a colon) are present, they are passed to the driver.

I'll have a look at their code and see how guaranteed that is, or just offer two flags.
 
> I had to use the `--sigrok-device` parameter in any case. If it was missing, it used the `demo` device: `demo - Demo device with 13 channels: D0 D1 D2 D3 D4 D5 D6 D7 A0 A1 A2 A3 A4`. That of course might never produce valid DMX data.

Oops, yeah I should remove that test code! I didn't get it to decode but I didn't try for too long (I think it has a restricted pattern of data).

peternewman:
> That does look more promising. It might be worth trying at 4Mhz (which is all we've really tested at). Did you wait a while for data to run through, and what channel data are you using (all full, all off, or real random data)? Come to think of it, I think my tests are still only sending 20 channels which probably isn't great either, but I think should work. In your case worst case it may have to process an entire frame before it finds the break and locks on. If you run it with `-d` and possibly lower the log levels, it should just output DMX lines every good frame.

I realised by paralleling workaround was unlikely to work properly @kripton as the frequency of the signal would be halved because it would get two samples where it should only be getting one. Anyway I know honour the unit width field. If it doesn't just burst into life with the logic analyser, can you share up to the first couple of transition/break to short with at least `-l 4` and ideally `--sigrok-log-level 5` so I can try and work out what's going on. But maybe it will just work. :crossed_fingers: 

I've not done anything on proper driver config strings yet, or the analogue reading of your scope (which given they are linked won't have much impact as far as I'm aware so I might push that into a future PR.

kripton:
> > I had to use the `--sigrok-device` parameter in any case. If it was missing, it used the `demo` device: `demo - Demo device with 13 channels: D0 D1 D2 D3 D4 D5 D6 D7 A0 A1 A2 A3 A4`. That of course might never produce valid DMX data.
> 
> Oops, yeah I should remove that test code! I didn't get it to decode but I didn't try for too long (I think it has a restricted pattern of data).

Are you sure that this is in your code? I can also see that demo device when I run `sigrok-cli --scan`. So it seems to be provided by libsigrok.

kripton:
> I realised by paralleling workaround was unlikely to work properly @kripton as the frequency of the signal would be halved because it would get two samples where it should only be getting one. Anyway I know honour the unit width field. If it doesn't just burst into life with the logic analyser, can you share up to the first couple of transition/break to short with at least `-l 4` and ideally `--sigrok-log-level 5` so I can try and work out what's going on. But maybe it will just work. crossed_fingers

Oh, great. You're right, the frequency would be halved. However, with 1MHz sample rate configured, I'd expect to sample with 500kHz, which should be okay to sample DMX data.
I'll re-test your code the following days, today if I can make it.
Besides the FX2-based devices, I also have one [HydraBus](https://hydrabus.com/hydrabus-1-0-specifications/) here that I would like to test sometime. It should work fine with sigrok (https://github.com/hydrabus/hydrafw/issues/82) but I didn't yet try that. 

> 
> I've not done anything on proper driver config strings yet, or the analogue reading of your scope (which given they are linked won't have much impact as far as I'm aware so I might push that into a future PR.

Don't worry so much about the scope. The configuration is already so difficult in pulseview and/or the data rate for the analog values is too low. It helps a lot to see the actual signal in pulseview but I would be really surprised if it can be managed to get this working reliably and easy from the command line.

peternewman:
> > Oops, yeah I should remove that test code! I didn't get it to decode but I didn't try for too long (I think it has a restricted pattern of data).
> 
> Are you sure that this is in your code? I can also see that demo device when I run `sigrok-cli --scan`. So it seems to be provided by libsigrok.

Yes, well sort of, my code no longer defaults to demo and instead complains if you've omitted the `--sigrok-device` argument:
https://github.com/OpenLightingProject/ola/pull/1718/commits/7ac19ebb1f8d4b29d530c2d645679cdbd625eae2

> Oh, great. You're right, the frequency would be halved. However, with 1MHz sample rate configured, I'd expect to sample with 500kHz, which should be okay to sample DMX data.

Yes, but the code doesn't actually do timing in the traditional sense (i.e. end-start), it just counts samples, so if you give it a 1MHz sample rate, it works out each sample will last 1 microsecond. If you consider something lasting for four sampling periods on an eight channel LA, i.e. 4 microseconds, because you've paralleled them, it counts it twice and gets a duration of 8 microseconds. Whether this is acceptable or not I'd need to check the docs and would depend on your source DMX refresh rate too I'd imagine.

> I'll re-test your code the following days, today if I can make it.

Excellent thanks!

> Besides the FX2-based devices, I also have one [HydraBus](https://hydrabus.com/hydrabus-1-0-specifications/) here that I would like to test sometime. It should work fine with sigrok ([hydrabus/hydrafw#82](https://github.com/hydrabus/hydrafw/issues/82)) but I didn't yet try that.

Cool, in theory it shouldn't care at all, as long as it (currently has logic outputs) and a sufficient sample rate, it should just work regardless.

> Don't worry so much about the scope. The configuration is already so difficult in pulseview and/or the data rate for the analog values is too low. It helps a lot to see the actual signal in pulseview but I would be really surprised if it can be managed to get this working reliably and easy from the command line.

Ah that's interesting, I assumed it would sample okay, but I guess the data rate has to be much higher to get the analogue level of detail. Perhaps I will park that for now. I guess I'd be curious if you can get the scope to decode DMX in PulseView in which case it might be worthwhile investigating in future.

kripton:
> Yes, well sort of, my code no longer defaults to demo and instead complains if you've omitted the `--sigrok-device` argument:
> [7ac19eb](https://github.com/OpenLightingProject/ola/commit/7ac19ebb1f8d4b29d530c2d645679cdbd625eae2)

Yes, I can confirm that and I agree that it makes sense.

> Excellent thanks!

Re-tested, working :)
`DMX 512:17 36 64 9f e0 00 ff 00 00 00 ...`. Just the values I would've expected. Nice :+1: 
Unfortunately, no RDM capable Interface here, yet. And right now, the Arduino is configured as RDM responder, not controller.

> Cool, in theory it shouldn't care at all, as long as it (currently has logic outputs) and a sufficient sample rate, it should just work regardless.

Yes. The big advantage of the FX2-based boards is that they can do continuous sampling. Most other devices I have here, including the HydraBus and the "original" OpenLogicSniffer allocate some memory and can hold a limited amount of samples. Depending on channel count and sample rate, this ranges from a few µs to several ms.

I didn't try the HydraBus in PulseView since the number of samples at the sample rate we need is way too low. I did try the original OpenLogicSniffer and it displays the correct bits in PulseView. However the number of samples at 500kHz is limited to about 2 DMX frames. For it to work with the `sigrok_rdm_sniffer`, I need to pass the path of the serial port device to the `ols` driver (`--sigrok-device ols:conn=/dev/ttyACM1`) and that doesn't yet seem to work. FX2LA is all good :)

> Ah that's interesting, I assumed it would sample okay, but I guess the data rate has to be much higher to get the analogue level of detail. Perhaps I will park that for now. I guess I'd be curious if you can get the scope to decode DMX in PulseView in which case it might be worthwhile investigating in future.

Yes, I managed to decode DMX512 with pulseview. But not in realtime. Capture (triggered on the TXEN line), tranfser to pulseview, decode. Interesting to see rise and fall times but not actually interesting to capture any live DMX or RDM data with it.

man/sigrok_rdm_sniffer.1
```diff
@@ -0,0 +1,48 @@
+.TH sigrok_rdm_sniffer 1 "March 2021"
+.SH NAME
+sigrok_rdm_sniffer \- Decode DMX/RDM data from devices supported by sigrok
```
kripton, line 3: I'd replace `devices` with `logic sniffers` here

man/sigrok_rdm_sniffer.1
```diff
@@ -0,0 +1,48 @@
...
+.SH SYNOPSIS
+.B sigrok_rdm_sniffer
+[ options ]
+.SH DESCRIPTION
+.B sigrok_rdm_sniffer
+Decode DMX/RDM data from devices supported by sigrok
```
kripton, line 9: same as line 3

man/sigrok_rdm_sniffer.1
```diff
@@ -0,0 +1,48 @@
...
+.SH OPTIONS
+.IP "-d, --display-dmx"
+Display DMX Frames. Defaults to false.
+.IP "-h, --help"
+Display the help message
+.IP "-l, --log-level <int8_t>"
+Set the logging level 0 .. 4.
+.IP "-r, --full-rdm"
+Unpack RDM parameter data.
+.IP "-t, --timestamp"
```
kripton, line 19: Is that implemented? Doesn't seem to work with just `-d -t`. Maybe it's just for RDM frames?

kripton (commented):
Minor inline comments

man/sigrok_rdm_sniffer.1
peternewman: I went for logic analyzer (with the US spelling) to match what sigrok call them in their docs.

man/sigrok_rdm_sniffer.1
peternewman: Yeah they all come from the same place via magic like this:
`./tools/logic/sigrok_rdm_sniffer --gen-manpage > man/sigrok_rdm_sniffer.1`

I've fixed the source and updated it all.

man/sigrok_rdm_sniffer.1
peternewman: No, I found out some time back it wasn't implemented on the original Logic but left is present just in case it broke people's CLIs:
https://github.com/peternewman/ola/blob/sigrok/tools/logic/logic-rdm-sniffer.cpp#L86-L87

Obviously that appeared when I duplicated this code. I've commented it out for now entirely in the sigrok based sniffer. When it all gets pulled out to some common code I'll try and finally implement it too, then they'll both get it!

peternewman:
> Re-tested, working :)
> `DMX 512:17 36 64 9f e0 00 ff 00 00 00 ...`. Just the values I would've expected. Nice +1

:partying_face: :tada: 

I've also still not tested recent code with a full size frame, so glad to see that works too!

> Unfortunately, no RDM capable Interface here, yet. And right now, the Arduino is configured as RDM responder, not controller.

Yeah no worries, I accidentally tested it when I started the sniffer before my olad generator and it did a universe discovery. It should all just work though given it's just different start codes and that worked fine in the original Logic sniffer. I think the only outstanding thing here would be to analyse and decode the DUB response.

> Yes. The big advantage of the FX2-based boards is that they can do continuous sampling. Most other devices I have here, including the HydraBus and the "original" OpenLogicSniffer allocate some memory and can hold a limited amount of samples. Depending on channel count and sample rate, this ranges from a few µs to several ms.

Ah I didn't realise the FX2 ones were somewhat unique in that regard. Yeah not being able to do continuous would be a real pain with all my past experience of using these.

> I did try the original OpenLogicSniffer and it displays the correct bits in PulseView. However the number of samples at 500kHz is limited to about 2 DMX frames.

Personally I think that would just be hugely frustrating debugging RDM stuff, even if you got a couple more frames of RDM out, due to smaller packet sizes.

> Yes, I managed to decode DMX512 with pulseview. But not in realtime. Capture (triggered on the TXEN line), tranfser to pulseview, decode. Interesting to see rise and fall times but not actually interesting to capture any live DMX or RDM data with it.

Okay, perhaps I'll park that for now then.

kripton:
Hey Peter, great to see code updates in this PR! I'd love to test them right away but unfortunately, we are relocating and most stuff (including the logic analyzers) is already packed. Just wanted to inform you that I'm monitoring this here but can't give solid feedback right now :)

peternewman:
> Hey Peter, great to see code updates in this PR!

I'm impressed to see you're watching it at the commit level!

> I'd love to test them right away but unfortunately, we are relocating and most stuff (including the logic analyzers) is already packed.

Damn, you were potentially on my list at some point to ping, although realistically since you last tried it I've only been doing some refactoring so shouldn't have broken anything! :crossed_fingers: 

I suspect I might get the main bits done then merge this PR, then possibly park the enhancements for the future. I was also toying with implementing some other decoders for some of the other standardised ASCs, probably mostly just for the fun of it, I'm not sure how much they exist in the wild.

I assume that's also why there's not been much activity on the RP2040 is it?

kripton:
Well, I subscribed to this PR, so I got an email ;)
Oh yes, rp2040-dongle is doing very well: https://github.com/kripton/rp2040-dongle/tree/rewrite. Don't want to cross-post much, but it has a webserver running, it can detect an nrf24 module and use it for channel scanning, it detects the IO modules and could read a config from them or the internal flash as fallback. Individual channels can be set via HTTP (not realtime capable) and dmx buffers can be fetched via HTTP in a compressed form (to prove compression and base64 are working). Lots of things still missing, though. And not much progress on hardware things.

PR #1723, opened 2021-06-01
Remaining Python 3 Compatibility

peternewman:
I think this fixes the remaining Python 3 compatibility.

I've not really given most of it any serious testing, so just getting some people to try using it would be a good start!

I'd appreciate any input or comments, whether general or from testing from anyone please.

CC @sdbbs @brucelowekamp @RenZ0 @danielskeenan @mikacousin

### DIFF
--- a/tools/ola_mon/ola_mon.py
+++ b/tools/ola_mon/ola_mon.py
@@ -18,7 +18,6 @@
 
 from __future__ import print_function
 import getopt
-import httplib
 import rrdtool
 import time
 import os.path
@@ -27,6 +26,11 @@ import socket
 import sys
 import textwrap
 import threading
+
+try:
+    import http.client as httplib
+except ImportError:
+    import httplib
 
 DEFAULT_CONFIG = 'ola_mon.conf'
 DEFAULT_PORT = 9090
@@ -59,7 +63,10 @@ class OlaFetcher(object):
     try:
       response = connection.getresponse()
       if response.status == 200:
-        return response.read()
+        if sys.version >= '3.2':
+          return response.read().decode('utf-8')
+        else:
+          return response.read()
     except httplib.BadStatusLine:
       return None
     return None
@@ -197,7 +204,8 @@ def LoadConfig(config_file):
     A dict with the config parameters.
   """
   locals = {}
-  execfile(config_file, {}, locals)
+  # Python 2 and 3 compatible version of execfile
+  exec(open(config_file).read(), {}, locals)
 
   keys = set(['OLAD_SERVERS', 'DATA_DIRECTORY', 'VARIABLES', 'WWW_DIRECTORY',
               'CDEFS'])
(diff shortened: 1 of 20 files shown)


tools/ola_mon/ola_mon.py
```diff
@@ -58,7 +67,10 @@ def _FetchDebug(self):
     try:
       response = connection.getresponse()
       if response.status == 200:
-        return response.read()
+        if sys.version >= '3.2':
```
peternewman, line 70: As per #1755
```suggestion
        if sys.version_info >= (3, 2):
```

peternewman (commented):
Todo to fix more of #1755 !

tools/ola_mon/ola_mon.py
mikacousin: Good catch !

mikacousin:
With adding PR #1755 , `make check` success with python 3.10. 

Maybe, we can use `threading.current_thread()` instead of `threading.currentThread()` as is marked as deprecated (line 270 in ClientWrapper.py).
As I understand, it should work with python 2.6+. What is minimal python version ola support ?

peternewman:
> With adding PR #1757 , `make check` success with python 3.10.

Excellent thanks, merged, as you've probably seen.

> Maybe, we can use `threading.current_thread()` instead of `threading.currentThread()` as is marked as deprecated (line 270 in ClientWrapper.py). As I understand, it should work with python 2.6+. What is minimal python version ola support ?

Yeah that sounds fine. My vague theory is to continue to target Python 2.6+ for 0.10 and master, and then master+1 can drop all of that (given the current codebase should support it, and at least one person was still using some ancient OS on an embedded system, and it seems a shame to needlessly throw away that work). My laptop isn't much newer than that, although it really should be...

peternewman:
> > Maybe, we can use `threading.current_thread()` instead of `threading.currentThread()` as is marked as deprecated (line 270 in ClientWrapper.py). As I understand, it should work with python 2.6+. What is minimal python version ola support ?
> 
> Yeah that sounds fine.

Done by @mikacousin in #1758 .

tools/rdm/ResponderTestTest.py
```diff
@@ -54,5 +54,62 @@ def testCmp(self):
...
+    self.assertEqual('%s' % rtf._EscapeData(['a']), "['a']")
+    self.assertEqual('%s' % rtf._EscapeData(["foo", 'a']), "['foo', 'a']")
+    self.assertEqual('%s' % rtf._EscapeData(['caf\xe9']), "['caf\\\\xe9']")
+    self.assertEqual('%s' % rtf._EscapeData(["foo", u'foo\u2014bar']),
+                     "['foo', 'foo\\\\u2014bar']")
+
+    self.assertEqual('%s' % rtf._EscapeData({"a": 0}), "{'a': 0}")
+    # The tests below that are commented out are non-deterministic on Python 3
+    # Might be able to get round it with a sort or something? We don't
+    # actually care about the order...
```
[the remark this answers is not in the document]
peternewman: Done in https://github.com/OpenLightingProject/ola/pull/1687/commits/d26d041132c33416c19194c2146b51153a0230f6 which will get merged in due course.

PR #1731, opened 2021-06-29
Switch to Travis-ci.com

peternewman:
- Only run Travis builds against the zero credit partner queue solution for now
- Temporarily disable the flake8 build until #1723 is in.
- This will be blocked on https://github.com/codespell-project/codespell/pull/1997 to fix the codespell false positive.


peternewman:
Force merged because having working Travis is good, even if we've got some outstanding typos.

PR #1729, opened 2021-06-19
Support for FTDI device "JMS USB2DMX PRO V2.1"

cmouttet:
#1728

Issue #1280, opened 2017-07-08
The beta UI fader window, set all channels to zero doesn't work


joshdurbin:
When in the fader window, in the beta UI, the set all channels to zero doesn't work as expected. The set all channels to zero operation should POST to set_dmx with two params; one for the universe `u`, and one with the value assignments `d` for all 512 channels. The beta UI executes the post but with a static value for `u`, `0` yielding no change. POSTed params:

u:0
d:0

The same operation in the old, classic, UI DMX Console works as expected... POSTed params:

u:0
[...]

PR #874, opened 2015-08-04
a bit of cleanup of the new-ui

daveol:

PR #1739, opened 2021-08-01
store highest channel number used

cmouttet:
With this fix the Fader UI and Keypad UI are able to send zeros to channels with the highest number.

Due to an optimisation the tailing zeros are cut off. With this stored highest channel number, we can fill up the DMX array with zeros until the maximum channel used.


javascript/new-src/src/factories/ola.js
```diff
@@ -63,6 +76,9 @@ ola.factory('$ola', ['$http', '$window', 'OLA',
           strip = false;
         }
       }
+      while (integers.length < highestChannelNumberUsed) {
+        integers.push(OLA.MIN_CHANNEL_VALUE);
+      }
```
peternewman, line 81: I think you could drop this and tweak the strip stuff, which was trying to do the same thing (introduced in #874) but had this bug.

javascript/new-src/src/factories/ola.js
```diff
@@ -118,6 +134,7 @@ ola.factory('$ola', ['$http', '$window', 'OLA',
             }
           })
             .then(function(response) {
+              updatehighestChannelNumberUsed(response.data.dmx);
```
peternewman, line 137: I forget the exact workflow here, but all updates to the faders come via this function do they? So both local and remote stuff is covered? Ideally this would just be called to deal with the local fader/keypad stuff to let OLAD deal with LTP merges.

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,19 @@
 ola.factory('$ola', ['$http', '$window', 'OLA',
   function($http, $window, OLA) {
     'use strict';
+    var highestChannelNumberUsed = 0;
```
peternewman, line 25: Does this need resetting when you switch from one universe to another?

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,19 @@
...
+
+    var updatehighestChannelNumberUsed = function(dmx) {
```
peternewman, line 27: Can you just check if OLAd has two universes patched this value is correctly kept local to each one?

peternewman (changes requested):
Great, thanks very much for this @cmouttet .

Given https://groups.google.com/g/open-lighting/c/IvFInOVAbTE do you want to maybe add a README.developer in olad/www maybe, signposting people to the other folders?

olad/www/new/css/style.min.css
```diff
@@ -1 +1 @@
...
...ebkit-transition:border-color ease-in-out .15s,-webkit-box-shadow ease-in-out .15s}.table-condensed>tbody>tr>td,.table-condensed>tbody>tr>th,.table-condensed>tfoot>tr>td,.table-condensed>tfoot>tr>th,.table-condensed>thead>tr>td,.table-condensed>thead>tr>th{padding:3px}.number{cursor:text!important}input.form-control[type=checkbox]{display:none}input.form-control[type=checkbox]+label span.glyphicon:before{width:20px;height:20px;border-radius:4px;border:1px solid #ccc;content:"\e014";color:#d9534f;background-color:#fff;cursor:pointer}input.form-control[type=checkbox]:checked+label span.glyphicon:before{content:"\e013";color:#5cb85c}input.form-control[type=checkbox][disabled]+label span.glyphicon:before{color:#a9a9a9}label.fake-check{margin-bottom:0}.nav-tabs{overflow-x:auto;overflow-y:hidden;display:-webkit-box;display:-moz-box}.nav-tabs>li{float:none}.nav-tabs::-webkit-scrollbar{-webkit-appearance:none}.nav-tabs::-webkit-scrollbar:vertical{width:12px}.nav-tabs::-webkit-scrollbar:horizontal{height:12px}.nav-tabs::-webkit-scrollbar-thumb{background-color:rgba(0,0,0,.5);border-radius:10px;border:2px solid #fff}.nav-tabs::-webkit-scrollbar-track{border-radius:10px;background-color:#fff}
```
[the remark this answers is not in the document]
cmouttet: It's the generated output from `grunt build`

javascript/new-src/src/factories/ola.js
cmouttet: Good point. I will simplify the logic as soon as possible.

javascript/new-src/src/factories/ola.js
cmouttet: Oh, I didn't check this. But I guess some kind of reset is needed.

javascript/new-src/src/factories/ola.js
cmouttet: I saw that the faders' data is updated by polling get_dmx. Remote updates are handled by the backend code. I think get_dmx will return the correctly updated/merged content.

javascript/new-src/src/factories/ola.js
cmouttet: done

javascript/new-src/src/factories/ola.js
cmouttet: changing the universe resets `highestChannelNumberUsed` now.

javascript/new-src/src/factories/ola.js
cmouttet: same here: changing the universe resets highestChannelNumberUsed

README.developer
```diff
@@ -200,6 +200,9 @@ Javascript
 Javascript is used for the olad web UI. Instructions for building the
 javascript can be found in javascript/README.
 
+The sources for the new UI is located in javascript/new-src. The built artifact
```
peternewman, line 203: Given the CSS is minified from the same place...
```suggestion
The sources for the new UI is located in javascript/new-src. The built artifacts
```

README.developer
```diff
@@ -200,6 +200,9 @@ Javascript
 Javascript is used for the olad web UI. Instructions for building the
 javascript can be found in javascript/README.
```
peternewman, line 201: Deal with the fact there are two UIs
```suggestion
Javascript is used for the olad web UIs. Instructions for building the
original UI's Javascript can be found in javascript/README.
```

AUTHORS
```diff
@@ -29,3 +29,4 @@ Contributors:
   Stefan Krüger, added APA102 support to the SPI Plugin
   Tobi Schäfer, for the MacPort files
   Stefan S, improved timing with monotonic clock
+  Christian Mouttet, fix in new web-ui
```
peternewman, line 32: Feel free to add something about #1729 or various fixes or whatever.

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,19 @@
...
+    var highestChannelNumberUsed = 0;
+
+    var updateHighestChannelNumberUsed = function(dmx) {
+      for (var channel = dmx.length; channel > highestChannelNumberUsed;
+           channel--) {
+
+        if ((dmx[channel - 1] > OLA.MIN_CHANNEL_VALUE) &&
+            (highestChannelNumberUsed < channel)) {
+          highestChannelNumberUsed = channel;
+        }
```
peternewman, line 34: Would something like this be easier?
```suggestion
        if (dmx[channel - 1] > OLA.MIN_CHANNEL_VALUE) {
          highestChannelNumberUsed = $window.Math.max(highestChannelNumberUsed, channel);
        }
```

Otherwise if you want to be efficient, can't we do some sort of early return, if we're counting down, as we are here, then the first used (non-zero I suspect), channel is what we need, we don't need to check all the following channels.

Edit: Actually I see you already have in the for loop, which I think makes the second part of the `if` redundant doesn't it? If we're looking at this value it's already `channel > highestChannelNumberUsed`.

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,19 @@
...
   function($http, $window, OLA) {
     'use strict';
+    var highestChannelNumberUsed = 0;
+
+    var updateHighestChannelNumberUsed = function(dmx) {
+      for (var channel = dmx.length; channel > highestChannelNumberUsed;
+           channel--) {
+
+        if ((dmx[channel - 1] > OLA.MIN_CHANNEL_VALUE) &&
+            (highestChannelNumberUsed < channel)) {
```
peternewman, line 32: If it did say (which as below I don't think it needs to, swapping the two variables around so it matches the for loop would be less confusing).

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,19 @@
 ola.factory('$ola', ['$http', '$window', 'OLA',
   function($http, $window, OLA) {
     'use strict';
+    var highestChannelNumberUsed = 0;
+
+    var updateHighestChannelNumberUsed = function(dmx) {
+      for (var channel = dmx.length; channel > highestChannelNumberUsed;
```
peternewman, line 28: A comment might be nice here, just so we don't automatically assume it's iterating from dmx.length to zero as I did. Something like "we only care about checking channels higher than our existing highest" or similar maybe.

javascript/new-src/src/controllers/universe.js
```diff
@@ -24,6 +24,7 @@ ola.controller('universeCtrl',
       'use strict';
       $scope.dmx = [];
       $scope.Universe = $routeParams.id;
+      $ola.resetHighestChannelNumberUsed();
```
peternewman, line 27: We're better off calling a function here than just initialising a scope variable then are we (like the DMX array)?

javascript/new-src/src/factories/ola.js
peternewman: Great thanks.

javascript/new-src/src/factories/ola.js
```diff
@@ -52,20 +65,28 @@ ola.factory('$ola', ['$http', '$window', 'OLA',
...
-          i === OLA.MIN_CHANNEL_NUMBER) {
-          integers[i - 1] = value;
-          strip = false;
+            channel <= highestChannelNumberUsed) {
+
+          integers[channel - 1] = value;
+
+          if (highestChannelNumberUsed < channel) {
+            highestChannelNumberUsed = channel;
+          }
```
peternewman, line 81: Again would $window.Math.max be easier?

peternewman (changes requested):
A few minor tweaks and suggestions.

Have you checked if it fixes this:

> Keypad on the beta UI supports setting a concrete channel to a value. If you set the last non-zero channel to zero, the current value doesn't change.

_Originally posted by @cmouttet in https://github.com/OpenLightingProject/ola/issues/1280#issuecomment-864527789_

Also does blackout work as intended?

olad/www/new/css/style.min.css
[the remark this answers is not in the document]
peternewman: Ah fair enough, at least it's not changed second time around.

javascript/new-src/src/factories/ola.js
peternewman: Thanks.

javascript/new-src/src/factories/ola.js
peternewman: Ah yeah you should be fine then, I just wasn't sure if there were a few different paths on the JS side of things. The backend will indeed deal with merging a remote update.

Although that might mean that bringing up a high numbered channel in e.g. the ola_dmxconsole will ripple into the JS faders too, where it perhaps didn't before, but I can't think of an easy solution to that.

javascript/new-src/src/factories/ola.js
cmouttet: When using `max` only the `> OLA.MIN_CHANNEL_VALUE` part is needed. `highestChannelNumberUsed < channel` will be removed.

javascript/new-src/src/factories/ola.js
cmouttet: Yes, it works. Thanks.

javascript/new-src/src/controllers/universe.js
cmouttet: I'm not very familiar with JavaScript. My home is Java on the backend. :)
I tried to to work with something like this `$scope.highestChannelNumberUsed = 0` but I failed.

AUTHORS
peternewman: Done thanks,

README.developer
```diff
@@ -197,8 +197,11 @@ Please follow the Sun Java style guide.
 Javascript
 ===============================================================================
 
-Javascript is used for the olad web UI. Instructions for building the
-javascript can be found in javascript/README.
+Javascript is used for the olad web UIs. Instructions for building the
+original UI's Javascript can be found in javascript/README.
+
+The sources for the new UI is located in javascript/new-src. The built artifacts
+is copied to olad/www/new/js/app.min.js.
```
peternewman, line 204: Fix some grammar:
```suggestion
The sources for the new UI are located in javascript/new-src. The built artifacts
are copied to olad/www/new/js/app.min.js.
```

AUTHORS
```diff
@@ -29,3 +29,4 @@ Contributors:
   Stefan Krüger, added APA102 support to the SPI Plugin
   Tobi Schäfer, for the MacPort files
   Stefan S, improved timing with monotonic clock
```
peternewman, line 31: Sort the file:
```suggestion
  Stefan Krüger, added APA102 support to the SPI Plugin
  Stefan S, improved timing with monotonic clock
  Tobi Schäfer, for the MacPort files
```

AUTHORS
```diff
@@ -29,3 +29,4 @@ Contributors:
...
+  Christian Mouttet, various fixes
```
peternewman, line 32: Can you sort your name entry too please (which avoids any issues with who did the most).

Feel free to change the top line to be "Contributors (in alphabetical order):" or similar.

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,29 @@
 ola.factory('$ola', ['$http', '$window', 'OLA',
   function($http, $window, OLA) {
     'use strict';
+    // holds the highest channel that was used by faders or the keypad
+    var highestChannelNumberUsed = 0;
+
+    // Search for the highest channel in the array `dmx`
+    // that having a value greater than MIN_CHANNEL_VALUE
```
peternewman, line 29: Grammar wise we want either:
```suggestion
    // that has a value greater than MIN_CHANNEL_VALUE
```
Or:
```suggestion
    // having a value greater than MIN_CHANNEL_VALUE
```

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,29 @@
 ola.factory('$ola', ['$http', '$window', 'OLA',
   function($http, $window, OLA) {
     'use strict';
+    // holds the highest channel that was used by faders or the keypad
+    var highestChannelNumberUsed = 0;
+
+    // Search for the highest channel in the array `dmx`
```
peternewman, line 28: Would this help clarify your last sentence?
```suggestion
    // Search for the new highest channel in the array `dmx`
```

javascript/new-src/src/factories/ola.js
```diff
@@ -22,6 +22,29 @@
...
+    // that having a value greater than MIN_CHANNEL_VALUE
+    // and update `highestChannelNumberUsed` if needed.
+    //
+    // Only channels higher than the current `highestChannelNumberUsed`
+    // will be checked.
+    var updateHighestChannelNumberUsed = function(dmx) {
+      for (var channel = dmx.length; channel > highestChannelNumberUsed;
+           channel--) {
+
+        var value = parseInt(dmx[channel - 1], 10);
```
peternewman, line 38: This should probably call `channelValueCheck` to sanity check the value the field was set to. But then `updateHighestChannelNumberUsed` is just `dmxConvert` without the return, so perhaps we should just have one function?

javascript/new-src/src/controllers/universe.js
peternewman: So it seems like $scope should work if you change it everywhere, but only within one controller, or there's $rootScope, or our $ola might be able to take a variable.

Otherwise @daveol or @FloEdelmann might be able to help, all this whizzy JavaScript library stuff isn't really my area of expertise either unfortunately.

peternewman (changes requested):
Sorry for the very delayed review @cmouttet . A few more minor comments if you've got some time? It would definitely be good to get this bugfix into 0.10.9.

Issue #1744, opened 2021-09-09
Sending DMX data to USB only if changed?


paradajz:
Disclaimer: this is not an issue or bug, but a question.

I'm developing a dual-mode device which registers on PC as both MIDI and CDC interface (virtual COM port). I'm using CDC interface for OLA via usbserial plugin. I have QLC+ connected to OLA, OLA sends data to CDC (aka my device) and then my device sends DMX signals to actual fixtures. On my device, I have internal DMX buffer and I am taking care of all the timings, meaning I basically need only raw data. My device implements Enttec DMX USB Pro Widget API so that the communication with OLA is possible. All fine and dandy - at least on devices with STM32 MCUs.

This entire setup also works with some AVR (8-bit, 16MHz) MCUs that I'm using, however, there is a considerable lag between changing values in lightning software and the actual change in lightning on fixtures. The reason for this is quite obvious: OLA sends values for all DMX channels *constantly* to device. Now, this is not fault of OLA, it's just the way Widget API works, I guess. The result of this that my AVR-based device needs to poll USB constantly for data that isn't changing all that often, so most of that processing time is spent for nothing. Normally, due to the high amount of time spent in polling and processing data from USB, the output will lag as well.

With all this, my question is: does OLA support "diff mode" protocol for DMX via USB, where only the changed data would get sent to device? Does any of the supported devices work like this? I've checked but haven't found exactly what I'm looking for, but then again, I might've missed something. The idea is basically this:

* OLA receives DMX data from some source (in my case, QLC+)
* OLA keeps internal buffer for all DMX channels as the source-of-truth
* Upon receiving new DMX data, OLA compares data for DMX channels with the values in source-of-truth buffer
* If the data differs, internal buffer is updated, and OLA sends a packet something like this:
  * `<start> <channel> <new_value> <end>` 

Would something like this be feasible?

peternewman:
> Disclaimer: this is not an issue or bug, but a question.

Thanks for the up front note.

> This entire setup also works with some AVR (8-bit, 16MHz) MCUs that I'm using,
> The result of this that my AVR-based device needs to poll USB constantly for data that isn't changing all that often, so most of that processing time is spent for nothing. Normally, due to the high amount of time spent in polling and processing data from USB, the output will lag as well.

I'm a bit surprised by that, I didn't think our Enttec implementation on an Arduino had those sort of issues but I may be wrong.
https://github.com/OpenLightingProject/rgbmixer

[...]

PR #1746, opened 2021-09-30
Add support for OpenDeck boards

paradajz:
This PR adds support for OpenDeck boards.

What is OpenDeck?

OpenDeck is open-source platform for building MIDI controllers which supports many different boards (some AVR, some STM32). More info on the [website](https://shanteacontrols.com/). Since firmware version 6.0.0, I have added support for DMX - more specifically, through CDC interface, Enttec Widget API was implemented so that the boards are compatible with OLA. A drawback for my boards was that in default USB Pro implementation, DMX data was sent constantly to the boards, even if data hasn't changed. Since my boards do many other things other than DMX, performance issues arose (at least on AVR, STM32 can handle it). I've started discussion about it [here](https://github.com/OpenLightingProject/ola/issues/1744).

What this PR does is:

* Copies UltraDMX Pro device/widget files and renames them to OpenDeck device/widget files, respectively (UltraDMX Pro seemed to be the simplest one on which I could base my boards)
* Implements only one output port (no input)
* Changes a bit handling of serial numbers so that in OLA, they show up the same as USB serial number which I already use on my boards
* Changes the handling of firmware version which gets shown in OLA Web interface (displayed in major.minor.revision format)
* Sends DMX data on board *only* if something has changed, and then:
  * If the amount of changed channels is > 128, full DMX frame is sent
  * Otherwise, send only the changed channels (two bytes for channel, one byte for value)


plugins/usbpro/OpenDeckDevice.h
```diff
@@ -0,0 +1,134 @@
...
+                    OpenDeckWidget *widget,
+                    uint16_t esta_id,
+                    uint16_t device_id,
+                    uint32_t serial,
+                    uint16_t firmware_version,
+                    unsigned int fps_limit);
+
+  std::string DeviceId() const { return m_serial; }
+  // both output ports can be bound to the same universe
+  bool AllowMultiPortPatching() const { return true; }
```
paradajz, line 56: Not sure on what this is supposed to return when there is only 1 output port.

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,243 @@
...
+ */
+OpenDeckDevice::OpenDeckDevice(ola::PluginAdaptor *plugin_adaptor,
+                                     ola::AbstractPlugin *owner,
+                                     const string &name,
+                                     OpenDeckWidget *widget,
+                                     OLA_UNUSED uint16_t esta_id,
+                                     OLA_UNUSED uint16_t device_id,
+                                     uint32_t serial,
+                                     uint16_t firmware_version,
+                                     unsigned int fps_limit):
```
peternewman, line 58: Can you align all these arguments like so:
```suggestion
OpenDeckDevice::OpenDeckDevice(ola::PluginAdaptor *plugin_adaptor,
                               ola::AbstractPlugin *owner,
                               const string &name,
                               OpenDeckWidget *widget,
                               OLA_UNUSED uint16_t esta_id,
                               OLA_UNUSED uint16_t device_id,
                               uint32_t serial,
                               uint16_t firmware_version,
                               unsigned int fps_limit):
```

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,243 @@
...
+ * Handle device config messages
+ * @param controller An RpcController
+ * @param request the request data
+ * @param response the response to return
+ * @param done the closure to call once the request is complete
+ */
+void OpenDeckDevice::Configure(RpcController *controller,
+                                  const string &request,
+                                  string *response,
+                                  ConfigureCallback *done) {
```
peternewman, line 113: Again

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,243 @@
...
+      done->Run();
+  }
+}
+
+
+/**
+ * Update the cached param values
+ */
+void OpenDeckDevice::UpdateParams(bool status,
+                                     const usb_pro_parameters &params) {
```
peternewman, line 139: Again

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,243 @@
...
+ * Handle a parameter request. This may set some parameters in the widget.
+ * If no parameters are set we simply fetch the parameters and return them to
+ * the client. If we are setting parameters, we send a SetParam() request and
+ * then another GetParam() request in order to return the latest values to the
+ * client.
+ */
+void OpenDeckDevice::HandleParametersRequest(RpcController *controller,
+                                                const Request *request,
+                                                string *response,
+                                                ConfigureCallback *done) {
```
peternewman, line 159: Again

plugins/usbpro/OpenDeckDevice.h
```diff
@@ -0,0 +1,134 @@
...
+ public:
+  OpenDeckDevice(ola::PluginAdaptor *plugin_adaptor,
+                    ola::AbstractPlugin *owner,
+                    const std::string &name,
+                    OpenDeckWidget *widget,
+                    uint16_t esta_id,
+                    uint16_t device_id,
+                    uint32_t serial,
+                    uint16_t firmware_version,
+                    unsigned int fps_limit);
```
peternewman, line 52: Again

plugins/usbpro/OpenDeckDevice.h
```diff
@@ -0,0 +1,134 @@
...
+ */
+class OpenDeckOutputPort: public BasicOutputPort {
+ public:
+  OpenDeckOutputPort(OpenDeckDevice *parent,
+                        OpenDeckWidget *widget,
+                        unsigned int id,
+                        const std::string &description,
+                        const TimeStamp *wake_time,
+                        unsigned int max_burst,
+                        unsigned int rate)
```
peternewman, line 105: Again

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,75 @@
...
+ * OpenDeckWidget Constructor
+ */
+OpenDeckWidget::OpenDeckWidget(
+  ola::io::ConnectedDescriptor *descriptor)
+    : GenericUsbProWidget(descriptor) {
+}
+
+
+bool OpenDeckWidget::SendDMX(const DmxBuffer &data) {
+    if (data != internal_buffer) {
```
peternewman, line 38: This wants de-denting by 2

peternewman (changes requested):
Some initial linting comments to start off with

plugins/usbpro/OpenDeckDevice.cpp
paradajz: done

plugins/usbpro/OpenDeckDevice.cpp
paradajz: done

plugins/usbpro/OpenDeckDevice.cpp
paradajz: done

plugins/usbpro/OpenDeckDevice.cpp
paradajz: done

plugins/usbpro/OpenDeckDevice.h
paradajz: done

plugins/usbpro/OpenDeckDevice.h
paradajz: done

plugins/usbpro/OpenDeckWidget.cpp
paradajz: done

paradajz:
Can I get some update on this?

peternewman:
Sorry for not replying sooner @paradajz .

> Had another look at this - not sure what to do next.

There are still a few outstanding comments (e.g. your own rate limiting key if you still want to do it). I'll see if the tests pass cleanly on a second run and try and re-review my outstanding comments and the code soon.

> Also don't know why the CI is failing, on my PC everything passes normally.

I think that test has occasionally been a bit intermittent so it might clear with a re-run now you've fixed the broken test. I've (accidentally) updated the online version of your branch to match master, so you may need to pull on your development machine.

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,251 @@
...
+  str.str("");
+
+  // Firmware version is stored in major.minor format.
+  // This scheme uses the following encoding:
+  // MMMMMMMM mmmmmmmm
+  // M: major
+  // m: minor
+  str << "Serial #: " << m_serial << ", firmware "
+      << (firmware_version >> 8)
+      << "." << (firmware_version & 0xff);
```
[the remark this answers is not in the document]
paradajz: Done!

paradajz:
Thanks for the reply - I've rebased on master to remove the merge commit.

plugins/usbpro/UsbSerialPlugin.cpp
```diff
@@ -201,6 +202,22 @@ void UsbSerialPlugin::NewWidget(UltraDMXProWidget *widget,
...
+  AddDevice(new OpenDeckDevice(
+      m_plugin_adaptor,
+      this,
+      GetDeviceName(information),
+      widget,
+      information.esta_id,
+      information.device_id,
+      information.serial,
+      information.firmware_version,
+      GetUltraDMXProFrameLimit()));
```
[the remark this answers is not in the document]
paradajz: Added new key.

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,256 @@
...
+    m_break_time(0),
+    m_mab_time(0),
+    m_rate(0) {
+  std::ostringstream str;
+  str << std::setfill('0');
+  uint8_t *ptr = reinterpret_cast<uint8_t*>(&serial);
+  for (int i = 0; i < UsbProWidgetInformation::SERIAL_LENGTH; i++) {
+    str << std::setfill('0') << std::setw(2) << std::hex
+        << static_cast<int>(ptr[i]);
+  }
```
peternewman, line 73: A comment here would be good to confirm the data format, or maybe even pull it out to it's own function so you can run a few tests against it. It looks like it's stored as hex digits and then turned to a string.

You may just be able to use our ToHex function:
http://docs.openlighting.org/ola/doc/latest/namespaceola_1_1strings.html#aedd1827ba98aeee25dbb9fdc8673d225

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * client.
+ */
+void OpenDeckDevice::HandleParametersRequest(RpcController *controller,
+                                             const Request *request,
+                                             string *response,
+                                             ConfigureCallback *done) {
+  if (request->has_parameters() &&
+      (request->parameters().has_break_time() ||
+       request->parameters().has_mab_time() ||
+       request->parameters().has_rate())) {
```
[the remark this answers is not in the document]
peternewman: Maybe you should just do this bit then if it's read only, which I think should allow get still:
```
      controller->SetFailed("SetParameters failed");
      done->Run();
      return;
```

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,256 @@
...
+    bool status,
+    const usb_pro_parameters &params) {
+  if (!status) {
+    controller->SetFailed("GetParameters failed");
+  } else {
+    UpdateParams(true, params);
+    Reply reply;
+    reply.set_type(ola::plugin::usbpro::Reply::USBPRO_PARAMETER_REPLY);
+    ola::plugin::usbpro::ParameterReply *parameters_reply =
+      reply.mutable_parameters();
```
peternewman, line 223: Fix the indent:
```suggestion
        reply.mutable_parameters();
```

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,256 @@
...
+ */
+void OpenDeckDevice::HandleSerialRequest(
+    RpcController *controller,
+    const Request *request,
+    string *response,
+    ConfigureCallback *done) {
+  Reply reply;
+  reply.set_type(ola::plugin::usbpro::Reply::USBPRO_SERIAL_REPLY);
+  ola::plugin::usbpro::SerialNumberReply *serial_reply =
+    reply.mutable_serial_number();
```
peternewman, line 247: Fix the indent:
```suggestion
      reply.mutable_serial_number();
```

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+
+bool OpenDeckWidget::SendDMX(const DmxBuffer &data) {
+  if (data != internal_buffer) {
+    std::vector<uint8_t> send_buffer = {};
+
+    for (size_t index = 0; index < data.Size(); index++) {
+      if (data.Get(index) != internal_buffer.Get(index)) {
+        send_buffer.push_back((index+1) & 0xFF);
+        send_buffer.push_back((index+1) >> 8);
```
peternewman, line 46: You could probably do SplitUInt16 here.

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,79 @@
...
+        send_buffer.push_back(data.Get(index));
+      }
+    }
+
+    internal_buffer = data;
+
+    if (send_buffer.size() >= MAX_DIFF_CHANNELS) {
```
peternewman, line 53: Perhaps put this check inside the for loop too, otherwise for a blackout it will generate the full send_buffer first, then realise it doesn't need to.

Also either this calculation is wrong, or the variable name is vague, if the MAX_DIFF_CHANNELS is 3, then it will only be one actual channel, comprising of it's high byte, low byte and value.

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,79 @@
...
+        uint8_t dmx[DMX_UNIVERSE_SIZE];
+      } full_frame;
+
+      full_frame.start_code = DMX512_START_CODE;
+      unsigned int length = DMX_UNIVERSE_SIZE;
+      data.Get(full_frame.dmx, &length);
+
+      return SendMessage(DMX_LABEL,
+                         reinterpret_cast<uint8_t*>(&full_frame),
+                         length + 1);
```
peternewman, line 66: This may be nicer (and should get optimised to the same):
```suggestion
                         length + sizeof(full_frame.start_code));
```

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,79 @@
...
+    } else {
+      return SendMessage(DMX_SLOT_VALUE_DIFF_LABEL,
+                         &send_buffer[0],
+                         send_buffer.size());
+    }
+  } else {
+    // Nothing new to send
```
peternewman, line 73: Please add an OLA_DEBUG here so we can see what's going on.

Will your system keep sending the old packet indefinitely?

What happens if the board is unplugged and replugged without OLA being restarted? Does it reconnect, does the internal buffer get cleared on disconnect?

plugins/usbpro/UsbSerialPlugin.cpp
```diff
@@ -278,6 +296,11 @@ bool UsbSerialPlugin::SetDefaultPreferences() {
   save |= m_preferences->SetDefaultValue(DEVICE_DIR_KEY, StringValidator(),
                                          DEFAULT_DEVICE_DIR);
 
+  save |= m_preferences->SetDefaultValue(OPENDECK_FPS_LIMIT_KEY,
+                                         UIntValidator(0,
+                                         MAX_OPENDECK_FPS_LIMIT),
```
peternewman, line 301: Assuming this doesn't fit:
```suggestion
                                         UIntValidator(0, MAX_OPENDECK_FPS_LIMIT),
```

Please do something like this:
```suggestion
                                         UIntValidator(0,
                                             MAX_OPENDECK_FPS_LIMIT),
```

plugins/usbpro/WidgetDetectorThreadTest.cpp
```diff
@@ -207,6 +212,18 @@ class WidgetDetectorThreadTest: public CppUnit::TestFixture,
       m_received_widget_type = ULTRA_DMX;
       m_ss.Terminate();
     }
+    void NewWidget(OpenDeckWidget *widget,
+                   const UsbProWidgetInformation &information) {
+      OLA_ASSERT_EQ(static_cast<uint16_t>(0x6555), information.esta_id);
+      OLA_ASSERT_EQ(string("Shantea Controls"), information.manufacturer);
+      OLA_ASSERT_EQ(static_cast<uint16_t>(0x0), information.device_id);
```
peternewman, line 219:
```suggestion
      OLA_ASSERT_EQ(static_cast<uint16_t>(0x0000), information.device_id);
```

plugins/usbpro/WidgetDetectorThreadTest.cpp
```diff
@@ -207,6 +212,18 @@ class WidgetDetectorThreadTest: public CppUnit::TestFixture,
...
+      OLA_ASSERT_EQ(string("OpenDeck"), information.device);
+      OLA_ASSERT_EQ(static_cast<uint32_t>(0xABCDEF99), information.serial);
+      OLA_ASSERT_EQ(static_cast<uint16_t>(0x600), information.firmware_version);
```
peternewman, line 222:
```suggestion
      OLA_ASSERT_EQ(static_cast<uint16_t>(0x0600), information.firmware_version);
```

peternewman (changes requested):
Some more comments and suggested improvements.

plugins/usbpro/OpenDeckDevice.cpp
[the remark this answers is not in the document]
paradajz: Done

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,256 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * OpenDeckDevice.cpp
+ * An OpenDeck Device
+ * Copyright (C) 2021 Simon Newton, Igor Petrovic
```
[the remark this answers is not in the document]
paradajz: Done

plugins/usbpro/OpenDeckWidget.cpp
paradajz: I'm not sure where should I handle the "disconnect" event in OLA for my device.

plugins/usbpro/OpenDeckWidget.cpp
paradajz: Done

plugins/usbpro/OpenDeckWidget.cpp
paradajz: Done

plugins/usbpro/OpenDeckWidget.cpp
paradajz: You're right. I've reworked this entire function so please take another look.

plugins/usbpro/OpenDeckWidget.cpp
paradajz: But OLA_DEBUG is added

plugins/usbpro/UsbSerialPlugin.cpp
paradajz: Done

plugins/usbpro/OpenDeckDevice.cpp
paradajz: Done

plugins/usbpro/UsbSerialPlugin.cpp
```diff
@@ -278,6 +296,11 @@ bool UsbSerialPlugin::SetDefaultPreferences() {
   save |= m_preferences->SetDefaultValue(DEVICE_DIR_KEY, StringValidator(),
                                          DEFAULT_DEVICE_DIR);
 
+  save |= m_preferences->SetDefaultValue(OPENDECK_FPS_LIMIT_KEY,
+                                         UIntValidator(0,
+                                             MAX_OPENDECK_FPS_LIMIT),
+                                             DEFAULT_OPENDECK_FPS_LIMIT);
```
peternewman, line 302: De-dent this line, as it's just the line above in the UIntValidator function
```suggestion
                                         DEFAULT_OPENDECK_FPS_LIMIT);
```

plugins/usbpro/OpenDeckDevice.cpp
peternewman: Its not applied?

plugins/usbpro/OpenDeckDevice.cpp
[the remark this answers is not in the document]
peternewman: Cool, does this actually work as expected? Assuming the message gets propagated, perhaps make it more useful like "Parameter are read only" or "Setting of parameters isn't supported".

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+
+/*
+ * Handle a parameter request. This may set some parameters in the widget.
+ * If no parameters are set we simply fetch the parameters and return them to
+ * the client. If we are setting parameters, we send a SetParam() request and
+ * then another GetParam() request in order to return the latest values to the
+ * client.
+ */
```
peternewman, line 167: This comment is probably stale now isn't it. Maybe replace it with one briefly explaining the technical issues why you can't change your timing?

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -0,0 +1,239 @@
...
+    UsbSerialDevice(owner, name, widget),
+    m_opendeck_widget(widget),
+    m_serial(),
+    m_got_parameters(false),
+    m_break_time(0),
+    m_mab_time(0),
+    m_rate(0) {
+  // Serial number is stored as series of 4 bytes.
+  // Display the serial in hex format so that it
+  // matches (partially) with USB serial number.
```
peternewman, line 71: Even better, I'd not even checked how they behave! Does this also match representations you use within your system; that's probably more important knowing you're talking about the same device both ends!

peternewman (commented):
A few more comments

plugins/usbpro/OpenDeckDevice.cpp
paradajz: Weird, was sure that I did... Fixed now.

plugins/usbpro/OpenDeckDevice.cpp
paradajz: Done

plugins/usbpro/OpenDeckDevice.cpp
paradajz: Yeah it does now, although I had the wrong byte order on the device which is fixed now.

OLA:
`olad/plugin_api/DeviceManager.cpp:105: Installed device: Shantea Controls - discovery:5-0x002e001d`

dmesg:
```
[363290.808545] usb 1-2.1: Product: OpenDeck | discovery
[363290.808546] usb 1-2.1: Manufacturer: Shantea Controls
[363290.808547] usb 1-2.1: SerialNumber: 002E001D574850012032364B
[363290.822563] cdc_acm 1-2.1:1.0: ttyACM1: USB ACM device
```

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,93 @@
...
+          } full_frame;
+
+          full_frame.start_code = DMX512_START_CODE;
+          unsigned int length = DMX_UNIVERSE_SIZE;
+          data.Get(full_frame.dmx, &length);
+          internal_buffer = data;
+
+          return SendMessage(DMX_LABEL,
+                            reinterpret_cast<uint8_t*>(&full_frame),
+                            length + sizeof(full_frame.start_code));
```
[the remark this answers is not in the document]
paradajz: Done

plugins/usbpro/OpenDeckDevice.cpp
paradajz: Additional change to truly match (no 0x prefix and uppercase now)

`olad/plugin_api/DeviceManager.cpp:105: Installed device: Shantea Controls - discovery:5-002E001D`

paradajz:
Any news on this?

plugins/usbpro/OpenDeckDevice.cpp
peternewman: I assume there's no benefit to taking one end of the serial or the other from your device? e.g. is the lower end more likely to be different between two devices produced around the same time?

plugins/usbpro/OpenDeckDevice.cpp
peternewman: That's unrelated. The rate limiting is internal to the plugin, whereas MAB and Break (in Enttec land) are set on the device and define the actual output from the on-board micro.

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,93 @@
...
+          full_frame.start_code = DMX512_START_CODE;
+          unsigned int length = DMX_UNIVERSE_SIZE;
+          data.Get(full_frame.dmx, &length);
+          internal_buffer = data;
+
+          return SendMessage(DMX_LABEL,
+                             reinterpret_cast<uint8_t*>(&full_frame),
+                             length + sizeof(full_frame.start_code));
+        }
+      }
```
peternewman, line 79: You should be able to just call the existing parent code rather than rewriting it:
```suggestion
          // Just send the full frame in this case
          internal_buffer = data;
          GenericUsbProWidget::SendDMX(data);          
```

peternewman (changes requested):
A few more fairly minor improvements if you wouldn't mind.

plugins/usbpro/OpenDeckDevice.cpp
paradajz: They're both equally random.

plugins/usbpro/OpenDeckDevice.cpp
paradajz: Ah I see, thanks!

plugins/usbpro/OpenDeckDevice.cpp
```diff
@@ -70,8 +70,9 @@ OpenDeckDevice::OpenDeckDevice(ola::PluginAdaptor *plugin_adaptor,
   // Display the serial in hex format so that it
   // matches (partially) with USB serial number.
   std::ostringstream str;
-  str << ToHex(serial);
+  str << ToHex(serial, false);
   m_serial = str.str();
+  ToUpper(&m_serial);
```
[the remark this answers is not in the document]
paradajz: I did, although I feel like this is unrelated change. It's in dedicated commit and I've also had to touch `StringUtils.cpp`,  `IntToHexString` function (this seems to be used only in tests, though).

paradajz:
Hopefully no more changes are required. The CI is still failing but from what I see it's something completely unrelated (spell checker?).

peternewman:
> Hopefully no more changes are required.

:crossed_fingers: 

> The CI is still failing but from what I see it's something completely unrelated (spell checker?).

If you want to cherry-pick this commit it will shut it up:
https://github.com/OpenLightingProject/ola/pull/1770/commits/05564726d6aebed391463206ebd0100533421fc7

plugins/usbpro/OpenDeckWidget.cpp
peternewman: This looks much better now thanks!

plugins/usbpro/OpenDeckDevice.cpp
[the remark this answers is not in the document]
peternewman: Yeah sorry and thanks, I was going to try and do a commit but you beat me to it. I'd agree it's unrelated, although if it doesn't go in now it's just never-ending tech debt.

I've just chucked some tests in.

plugins/usbpro/OpenDeckDevice.h
```diff
@@ -0,0 +1,131 @@
...
+        m_wake_time(wake_time)
+        {}
+
+  bool WriteDMX(const DmxBuffer &buffer, OLA_UNUSED uint8_t priority) {
+    if (m_bucket.GetToken(*m_wake_time)) {
+      return m_widget->SendDMX(buffer);
+    } else {
+      OLA_INFO << "Port rated limited, dropping frame";
+    }
+    return true;
```
peternewman, line 117: While it will make it a little more complicated, you probably want to move the rate limiting check to within OpenDeckWidget.cpp, otherwise you risk wasting your tokens on unchanged packets which don't actually get sent, rather than saving them for the first diff or full packet that arrives!

Depending on your processing times, you might also choose to take another token (I think you can) if processing one (e.g. unpacking the diff) takes longer than the other one on your device.

plugins/usbpro/OpenDeckWidget.h
```diff
@@ -0,0 +1,54 @@
...
+ public:
+    explicit OpenDeckWidget(ola::io::ConnectedDescriptor *descriptor);
+    ~OpenDeckWidget() {}
+    void Stop() { GenericStop(); }
+
+    bool SendDMX(const DmxBuffer &buffer);
+
+ private:
+    static const size_t MAX_DIFF_CHANNELS = 128;
+    static const uint8_t DMX_SLOT_VALUE_DIFF_LABEL = 80;
```
peternewman, line 48: Label 80 (aka 0x50) doesn't seem to appear in our codebase, so looks good to go! :partying_face:

peternewman (commented):
Really sorry, just the one niggle with the token bucket that I think will be worth testing in the long term.

plugins/usbpro/OpenDeckWidget.cpp
peternewman: Please can you test this workflow, I think the following should work:

1. Ideally have something/a few somethings connected to DMX output (e.g. LEDs)
2. Turn on channel 1 and confirm output
3. Wait until it's logging the nothing changed messages
4. Unplug OpenDeck
5. Turn on channel 2
6. Wait until it's logging the nothing changed messages
7. Replug device
8. Confirm channel 2 lights up (i.e. it doesn't continue to use the internal buffer where nothing has changed and hence doesn't send new data)
9. Turn on channel 3 (this should fix the issue in either case if it's reconnected okay)

If the above doesn't work as expected, then some reading of the -l 4 logs or testing my side should hopefully figure out where we need to fix it...

plugins/usbpro/OpenDeckDevice.h
paradajz: Good point. I've added initial, maybe naive approach. Let me know what you think.

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -56,20 +59,30 @@ bool OpenDeckWidget::SendDMX(const DmxBuffer &data) {
...
-    return SendMessage(DMX_SLOT_VALUE_DIFF_LABEL,
+    internal_buffer = buffer;
+
+    if (bucket.GetToken(*wake_time)) {
+      if (send_full_frame) {
+        return GenericUsbProWidget::SendDMX(buffer);
+      } else {
+        return SendMessage(DMX_SLOT_VALUE_DIFF_LABEL,
                        &send_buffer[0],
                        send_buffer.size());
```
peternewman, line 80: Fix the whitespace:
```suggestion
        return SendMessage(DMX_SLOT_VALUE_DIFF_LABEL,
                           &send_buffer[0],
                           send_buffer.size());
```

common/utils/StringUtilsTest.cpp
```diff
@@ -393,7 +393,7 @@ void StringUtilsTest::testIntToHexString() {
   str << ToHex((uint16_t)0xABCD, false);
   OLA_ASSERT_EQ(string("abcd"), str.str());
   str.str("");
-  
+
```
peternewman, line 396: Thanks, pesky GitHub editor!

plugins/usbpro/OpenDeckDevice.h
peternewman: Any particular reason you haven't just moved all the TokenBucket code to OpenDeckWidget, rather than keeping it outside and passing it in?

peternewman (commented):
Minor comments

paradajz:
Can't seem to reply to this comment, so I'll do it here:

> Any particular reason you haven't just moved all the TokenBucket code to OpenDeckWidget, rather than keeping it outside and passing it in?

I wanted to do that, but to create TokenBucket, I need some parameters first, namely:

    max_burst, rate, max_burst, *wake_time

These aren't accessible in OpenDeckWidget. I might be missing something here.

plugins/usbpro/OpenDeckWidget.cpp
paradajz: done

plugins/usbpro/OpenDeckDevice.h
paradajz: I wanted to do that, but to create TokenBucket, I need some parameters first, namely:

    max_burst, rate, max_burst, *wake_time
These aren't accessible in OpenDeckWidget. I might be missing something here.

paradajz: This seems to be a bit beyond my understanding of OLA... Don't know what to do here.

paradajz: Added simpler, alternative approach.

plugins/usbpro/OpenDeckWidget.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * OpenDeckWidget Constructor
+ */
+OpenDeckWidget::OpenDeckWidget(
+  ola::io::ConnectedDescriptor *descriptor)
+    : GenericUsbProWidget(descriptor) {
+}
+
+bool OpenDeckWidget::IsBufferDiff(const DmxBuffer &buffer) {
+  return buffer != internal_buffer;
+}
```
peternewman, line 44: Great minds! I literally just came here to write this after reviewing #1766 who's doing a very similar thing...

plugins/usbpro/OpenDeckWidget.cpp
paradajz: I've just tried this. After the reconnection to USB it takes few seconds for OLA to find the device again, after which it immediately resends all values, and everything is back as it was before the disconnection. Was this the thing you were interested in?

paradajz: Also, changing the values of some channels once the device is unplugged results in those channels being correctly applied once the device is connected and detected by OLA again.

paradajz:
So is this ready for merging?

PR #1750, opened 2021-11-13
Attempt to enable native UART on modern systems

jahudka:
Both current Raspberry Pi OS and Ubuntu for the Pi are missing the `stropts.h` header file, which causes the OLA ExtendedSerial helper to be compiled to always return `false` from `LinuxHelper::SetDmxBaud()`. This in turn makes it impossible to use the native UART DMX output plugin. This PR is an attempt to fix that by including an alternative bundled header file which defines the `ioctl()` function, which is the only thing needed out of `stropts.h`.

I'd like to say up front that I'm not a C/C++ developer - in fact this is only the second time in my life that I'm attempting to mess around with C code - so please don't be too harsh with me if I've made newbie mistakes! As it is, the code compiles on Ubuntu 21.10 for RPi without errors and from the output of `olad` it is apparent that the issue is fixed by this workaround - that is, where previously the UART DMX plugin logged a warning about being unable to set the correct baud rate for the UART device, now the plugin outputs the expected line about the input and output rates being the expected `250000`. I've tried to provide a "minimal" fix which would only be applied when `stropts.h` isn't present, so it should compile on other and / or older systems exactly as before, unless I'm mistaken.

Of course, the usual UART configuration still applies - you still need to modify `/boot/firmware/config.txt` to disable bluetooth via a `dtoverlay`, enable UART and set the appropriate UART clock, and disable the `hciuart` service - but even with all those things in place, neither the distribution `ola` package nor a fresh compile of OLA produced the expected result.

Thanks everyone on the OLA team for creating and maintaining this project!

### DIFF
--- a/common/io/ExtendedSerial.cpp
+++ b/common/io/ExtendedSerial.cpp
@@ -85,7 +85,7 @@ bool LinuxHelper::SetDmxBaud(int fd) {
   }
   return true;
 #else
-  OLA_INFO << "Failed to set baud rate, due to missing stropts.h or termios2";
+  OLA_INFO << "Failed to set baud rate, missing termios2";
   return false;
   (void) fd;
 #endif  // defined(HAVE_TERMIOS2)
--- a/common/io/ExtendedSerial.cpp
+++ b/common/io/ExtendedSerial.cpp
@@ -32,6 +32,10 @@
 #ifdef HAVE_STROPTS_H
 // this provides ioctl() definition without conflicting with asm/termios.h
 #include <stropts.h>
+#else
+// this provides a minimal ioctl() definition for modern systems
+// which don't have <stropts.h> anymore
+#include <ola/io/IOCtl.h>
 #endif  // HAVE_STROPTS_H
 
 #ifdef HAVE_ASM_TERMIOS_H
@@ -52,7 +56,7 @@ namespace ola {
 namespace io {
 
 bool LinuxHelper::SetDmxBaud(int fd) {
-#if defined(HAVE_STROPTS_H) && defined(HAVE_TERMIOS2)
+#if defined(HAVE_TERMIOS2)
   static const int rate = 250000;
 
   struct termios2 tio;  // linux-specific terminal stuff
@@ -84,7 +88,7 @@ bool LinuxHelper::SetDmxBaud(int fd) {
   OLA_INFO << "Failed to set baud rate, due to missing stropts.h or termios2";
   return false;
   (void) fd;
-#endif  // defined(HAVE_STROPTS_H) && defined(HAVE_TERMIOS2)
+#endif  // defined(HAVE_TERMIOS2)
 }
 }  // namespace io
 }  // namespace ola
--- a/include/ola/io/IOCtl.h
+++ b/include/ola/io/IOCtl.h
@@ -23,7 +23,7 @@
 #define INCLUDE_OLA_IO_IOCTL_H_
 
 extern "C" {
-  extern int ioctl(int __fd, unsigned long int __request, ...) __THROW;  // NOLINT(runtime/int)
+  extern int ioctl(int f, unsigned long r, ...) __THROW;  // NOLINT(runtime/int)
 }
 
 #endif  // INCLUDE_OLA_IO_IOCTL_H_
--- a/include/ola/io/IOCtl.h
+++ b/include/ola/io/IOCtl.h
@@ -1,0 +1,29 @@
+/*
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ *
+ * IOCtl.h
+ * This is a minimal definition of ioctl() to allow the UART DMX plugin
+ * to compile with custom baud rate support on newer systems which don't
+ * have <stropts.h>.
+ */
+
+#ifndef INCLUDE_OLA_IO_IOCTL_H_
+#define INCLUDE_OLA_IO_IOCTL_H_
+
+extern "C" {
+  extern int ioctl(int __fd, unsigned long int __request, ...) __THROW;  // NOLINT(runtime/int)
+}
+
+#endif  // INCLUDE_OLA_IO_IOCTL_H_
--- a/include/ola/io/Makefile.mk
+++ b/include/ola/io/Makefile.mk
@@ -4,6 +4,7 @@ olaioinclude_HEADERS = \
     include/ola/io/ByteString.h \
     include/ola/io/Descriptor.h \
     include/ola/io/ExtendedSerial.h \
+    include/ola/io/IOCtl.h \
     include/ola/io/IOQueue.h \
     include/ola/io/IOStack.h \
     include/ola/io/IOUtils.h \


peternewman (changes requested):
Hi @jahudka ,

Thanks very much for this, #1749 beat you to the bug report, but not the fix, so it's certainly appreciated!

Please can you target our 0.10 branch with this fix, as it should go into those patch releases too. You'll need to recommit against those or cherry-pick or similar as your branch has all the 0.10 to master changes in it already.

More generally this doesn't feel quite right, I'd assume it still exists in a file somewhere and we're just not loading it, we might need to do some more digging. My strpopts.h (on an older system) comes from libc6-dev.

I might generate another PR for you to test if you don't mind, to start off by us logging if this fails in future, rather than silently stopping.

common/io/ExtendedSerial.cpp
```diff
@@ -32,6 +32,10 @@
 #ifdef HAVE_STROPTS_H
 // this provides ioctl() definition without conflicting with asm/termios.h
 #include <stropts.h>
+#else
+// this provides a minimal ioctl() definition for modern systems
+// which don't have <stropts.h> anymore
+#include <ola/io/IOCtl.h>
```
peternewman, line 38: I think from a bit of reading Linux shouldn't necessarily have stropts.h (as it's a POSIX thing).

common/io/ExtendedSerial.cpp
```diff
@@ -32,6 +32,10 @@
...
 #endif  // HAVE_STROPTS_H
 
 #ifdef HAVE_ASM_TERMIOS_H
```
peternewman, line 41: Which would suggest it's actually this bit that's breaking

include/ola/io/IOCtl.h
```diff
@@ -0,0 +1,29 @@
...
+ * This is a minimal definition of ioctl() to allow the UART DMX plugin
+ * to compile with custom baud rate support on newer systems which don't
+ * have <stropts.h>.
+ */
+
+#ifndef INCLUDE_OLA_IO_IOCTL_H_
+#define INCLUDE_OLA_IO_IOCTL_H_
+
+extern "C" {
+	extern int ioctl (int __fd, unsigned long int __request, ...) __THROW;
```
peternewman, line 26: Is that definition in an existing header on your machine, if so which one?

jahudka:
Hi, wow, fast and thorough! I'll go ahead and remake this PR against the 0.10 branch (I'm assuming I can't simply update this PR, I'll have to create a new one, right?).

Regarding `stropts.h`, yes, it does come from `libc6-dev`, but it looks like it has been removed sometime before the release of Ubuntu 20.04, since that's where most people started reporting it missing, according to my research. Which would explain why it works on older systems.

common/io/ExtendedSerial.cpp
jahudka: Yes, true, as I understand it, the Streams interface probably wasn't ever even really implemented, the header files were mostly present for decades before it was finally purged, I don't recall where but I read something along those lines in the past couple of days..

common/io/ExtendedSerial.cpp
jahudka: I don't think it is - `./configure` reports that both `termios2.h` and `asm/termios.h` are present and usable, whereas `stropts.h` is not; and in any case, if it were this, then my "fix" wouldn't actually fix anything.. so unless something really non-obvious is happening I don't think the issue is here.

include/ola/io/IOCtl.h
jahudka: No, I found the source of `stropts.h` [here](https://code.woboq.org/qt5/include/stropts.h.html) and extrapolated from there.

peternewman:
> Hi, wow, fast and thorough!

The perils of notifications! And I'm actually free today... :laughing: 

> I'll go ahead and remake this PR against the 0.10 branch (I'm assuming I can't simply update this PR, I'll have to create a new one, right?).

Yeah sadly, a cherry-pick is the best I've managed, or you might be able to rebase or something (I don't think anything has actually changed in this file).

> Regarding `stropts.h`, yes, it does come from `libc6-dev`, but it looks like it has been removed sometime before the release of Ubuntu 20.04, since that's where most people started reporting it missing, according to my research. Which would explain why it works on older systems.

Yeah, it still feels like that ioctl should be defined *somewhere* even now, if we can use it.

Would you also mind testing that this logs correctly on failure before you're patch (you'll need -l 3 or -l 4) #1751. Then we can either merge it as is, or you can pull it into your branch and adapt if we end up changing the includes requirements.

peternewman:
Also please heed and fix the cpplint errors on Travis:
```
./include/ola/io/IOCtl.h:29:  #endif line should be "#endif  // INCLUDE_OLA_IO_IOCTL_H_"  [build/header_guard] [5]
./include/ola/io/IOCtl.h:26:  Tab found; better to use spaces  [whitespace/tab] [1]
./include/ola/io/IOCtl.h:26:  Extra space before ( in function call  [whitespace/parens] [4]
./include/ola/io/IOCtl.h:26:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
```

jahudka:
`ioctl()` is actually defined in `sys/ioctl.h`, but including it in place of `stropts.h` causes conflicts with the `winsize` and `termio` structs which I've been unable to resolve with my limited skills..

Regarding the logging: I've actually tried a _lot_ of that when I was trying to get to the bottom of this - mostly exactly what you have in your PR, except in a somewhat.. _sharper_... language (I was getting slightly frustrated at that point lol). And the fun part is, I never actually got the log messages to display - I would update the code, run `make` and `make install` again, wait for an hour.. and then get _nothing_! I pulled a lot of my hair out last night. Anyway, I'll try a recompile from your PR later tonight to check if I get the expected `missing stropts.h or termios2` message which should be evidence my theory is correct, and then I'll post an updated PR based on 0.10, taking care of the `cpplint` errors as well (sorry, I didn't really run it before pushing this PR, I'm doing all the compilation etc on a Pi 3 and everything takes forever, so I figured I'd have the CI do some of the work for me..).

jahudka:
quick question re: `./include/ola/io/IOCtl.h:26:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]` from cpplint: which type would be the correct one to use here? since even `sys/ioctl.h` defines the parameter as `unsigned long`, according to several sources..

shenghaoyang:
> quick question re: `./include/ola/io/IOCtl.h:26: Use int16/int64/etc, rather than the C type long [runtime/int] [4]` from cpplint: which type would be the correct one to use here? since even `sys/ioctl.h` defines the parameter as `unsigned long`, according to several sources.. sorry, I know this is probably one of the noobest C++ questions ever :-D

You should stick with ulong. The syscall wrapper in libc is expecting a ulong when it's called, so using any other fixed-width type will cause issues on systems where ulong != the type used.

In any case using `asm/termbits.h` & `asm/ioctls.h` allows use of `sys/ioctl.h` without any conflicts (at lest on my system):

```
#include <asm/ioctls.h>
#include <asm/termbits.h>
#include <sys/ioctl.h>

int set_baud(int fd) {
    const speed_t rate = 250000u;
    termios2 tio;

    if (ioctl(fd, TCGETS2, &tio) == -1)
        return -1;

    tio.c_cflag &= ~CBAUD;
    tio.c_cflag |= BOTHER;
    tio.c_ispeed = rate;
    tio.c_ospeed = rate;

    if (ioctl(fd, TCSETS2, &tio) == -1)
        return -1;

    return 0;
}
``` 

Other ideas are discussed at https://stackoverflow.com/questions/37710525/including-termios-h-and-asm-termios-h-in-the-same-project - if all else fails `#undef` hacks could be used.

jahudka:
@peternewman I've finally successfully completed a build of your PR with the debug code - the pertinent section of the output looks like this:

```
olad/PluginManager.cpp:195: Trying to start UART native DMX
plugins/uartdmx/UartDmxPlugin.cpp:67: Trying to open UART device /dev/ttyAMA0
plugins/uartdmx/UartWidget.cpp:70: Opening serial port /dev/ttyAMA0
plugins/uartdmx/UartWidget.cpp:76: Opened serial port /dev/ttyAMA0
common/io/ExtendedSerial.cpp:84: Failed to set baud rate, due to missing stropts.h or termios2
plugins/uartdmx/UartWidget.cpp:180: Failed to set baud rate to 250k
plugins/uartdmx/UartDmxPlugin.cpp:82: Unable to setup device for output, device ignored /dev/ttyAMA0
olad/PluginManager.cpp:200: Started UART native DMX
```

@shenghaoyang I've seen the discussion about `#undef` hacks you linked and I've tried using said hacks - it looked a little like this:

```
#ifdef HAVE_STROPTS_H
// this provides ioctl() definition without conflicting with asm/termios.h
#include <stropts.h>
#else
#define winsize sys_ioctl_h_winsize
#define termio sys_ioctl_h_termio
#include <sys/ioctl.h>
#undef winsize
#undef termio
#endif
```

This just caused my build to hang without apparent reason (to the point that I couldn't even SSH to the machine anymore).

Whether we should or shouldn't be directly including things like `asm/<anything>` or `<any>bits.h` is another question for more experienced C/C++ devs than I am - I've read somewhere that we shouldn't, but what do I know..

jahudka:
Hi, I've rebased the PR onto the `0.10` branch and fixed the CppLint errors, but now two of the CI jobs have failed on seemingly unrelated issues - in one there was a segfault, and both failed the `ClientWrapperTest.py` test. I'm not sure exactly what the test is supposed to do, or if it might be one of those tests which sometimes _just fail_ (happens to some workflows in our company CI all the time) - help please?

Also, @peternewman, I'm not sure how to interpret your confused smiley reactions in the two discussions above - would you please clarify? I know that the `IOCtl.h` file I'm proposing with this PR is missing a Copyright notice - I'm not sure how to exactly proceed with that - on the one hand it obviously wouldn't be right to credit myself, since I didn't exactly _invent_ that bit of code out of thin air, but on the other hand it isn't a _verbatim_ copy of the original code, either - so I'm not sure the author(s) of the original would even _want_ to be credited for something they didn't sign off on.. I don't want to step on anyone's toes, I'd very much like to do it right, I'm just not sure what the proper etiquette is, here.

Sorry to put so much work on you guys, I know I should do my homework and I'm trying, it's just that this is all pretty new to me - I come from a web development background, C/C++ really isn't my strong suit, and I don't contribute to OSS projects that often, either.. so please bear with me and know that I have the best of intentions :-)

peternewman:
> In any case using `asm/termbits.h` & `asm/ioctls.h` allows use of `sys/ioctl.h` without any conflicts (at least on my system):

What's your system for reference @shenghaoyang ?

> @peternewman I've finally successfully completed a build of your PR with the debug code - the pertinent section of the output looks like this:

Great, thanks for doing that, merged as you've probably seen.

> This just caused my build to hang without apparent reason (to the point that I couldn't even SSH to the machine anymore).

I'd be surprised if it was something around the build specifically (unless it ran out of memory or something)!

> Whether we should or shouldn't be directly including things like `asm/<anything>` or `<any>bits.h` is another question for more experienced C/C++ devs than I am - I've read somewhere that we shouldn't, but what do I know..

I'd need to read up, but using a shipped header file generally feels much nicer to me than duplicating the definition in our own file (and risking it becoming out of date). It could also have issues on other OSes where the declaration differs slightly.

> Hi, I've rebased the PR onto the `0.10` branch and fixed the CppLint errors, but now two of the CI jobs have failed on seemingly unrelated issues - in one there was a segfault, and both failed the `ClientWrapperTest.py` test. I'm not sure exactly what the test is supposed to do, or if it might be one of those tests which sometimes _just fail_ (happens to some workflows in our company CI all the time) - help please?

It's probably just flaky tests, although maybe not the segfault, I'll re-run them or have a look.

> Also, @peternewman, I'm not sure how to interpret your confused smiley reactions in the two discussions above - would you please clarify?

Hopefully the above covered it, I'd be far happier if we can just include an existing header file somehow for the declaration.

> Sorry to put so much work on you guys, I know I should do my homework and I'm trying, it's just that this is all pretty new to me - I come from a web development background, C/C++ really isn't my strong suit, and I don't contribute to OSS projects that often, either.. so please bear with me and know that I have the best of intentions :-)

No worries at all, thanks for your efforts and contribution already!

shenghaoyang:
> What's your system for reference @shenghaoyang ?

@peternewman I ran that on Debian "Bullseye".

But it does bring in the kernel definition of `struct termios` (which may cause conflicts if the version that libc provides needs to be used in the same source file) - I don't think that's used in `ExtendedSerial.cpp`, so we should be fine.

Alternatively,

See https://github.com/npat-efault/picocom/blob/master/termbits2.h - that header explains the duplicate definition problem in more detail and picocom's approach.

Also, https://github.com/sigrokproject/libserialport/blob/master/linux_termios.c is interesting as well - libserialport isolates the linux specific declarations to a separate compilation unit.

peternewman:
Sorry for taking a while to get back. I think I'd prefer @shenghaoyang 's solution, or just a rather hacky #undef like we've used elsewhere for Windows and like https://stackoverflow.com/a/48521433 rather than redefining the call ourselves.

jahudka:
@peternewman well, in that case I'm afraid I'm not the right man for the job - it's all a little over my head - so we can probably close this and you or somebody else who knows what they're doing can open a new PR?

peternewman:
> @peternewman well, in that case I'm afraid I'm not the right man for the job - it's all a little over my head - so we can probably close this and you or somebody else who knows what they're doing can open a new PR?

No worries at all and thanks for all your attempts on this. I left it open because it did indeed work if people needed a quick fix. Now I've managed to have a look and come up with #1760 , I'll close this (as hopefully that works in the physical domain too).

> I come from a web development background

We're always looking for people to work on the web front-end (especially the shiny new version) for OLA (and indeed some of our related projects) if you ever want to do some bits of that... :smiley:

PR #1751, opened 2021-11-13
Add more debugging so we can tell why setting the baud rate failed

peternewman:


peternewman:
This solves it for @jahudka:
https://github.com/OpenLightingProject/ola/pull/1750#issuecomment-968783831

And @MagnusHerrlin:
https://github.com/OpenLightingProject/ola/issues/1749#issuecomment-968174636

So will get this in as it improves the logging.

Issue #1726, opened 2021-06-09
OLA is crashing on adding Artnet Universe


freaknils:
HI,

I have just build the current master of OLA, but when I add a new Artnet Universe via the web page the _olad_ crashes:

```
common/io/IOUtils.cpp:39: open(/dev/dmx0): No such file or directory
plugins/opendmx/OpenDmxPlugin.cpp:81: Could not open /dev/dmx0 No such file or directory
common/io/IOUtils.cpp:39: open(/dev/kldmx0): No such file or directory
plugins/karate/KaratePlugin.cpp:80: Could not open /dev/kldmx0 No such file or directory
/usr/include/c++/11.1.0/bits/stl_queue.h:627: std::priority_queue<_Tp, _Sequence, _Compare>::const_reference std::priority_queue<_Tp, _Sequence, _Compare>::top() const [with _Tp = ola::io::TimeoutManager::Event*; _Sequence = std::vector<ola::io::TimeoutManager::Event*>; _Compare = ola::io::TimeoutManager::ltevent; std::priority_queue<_Tp, _Sequence, _Compare>::const_reference = ola::io::TimeoutManager::Event* const&]: Assertion '!this->empty()' failed.
Received Segmentation fault
/usr/lib/libolacommon.so.0(+0x990dc)[0x7f12105290dc]
/usr/lib/libc.so.6(+0x3cda0)[0x7f12100cfda0]
/usr/lib/libc.so.6(abort+0x1d5)[0x7f12100b9921]
/usr/lib/libolacommon.so.0(+0xaad3a)[0x7f121053ad3a]
/usr/lib/libolacommon.so.0(_ZN3ola2io14TimeoutManager15ExecuteTimeoutsEPNS_9TimeStampE+0x252)[0x7f1210542082]
/usr/lib/libolacommon.so.0(_ZN3ola2io12SelectPoller4PollEPNS0_14TimeoutManagerERKNS_12TimeIntervalE+0x229)[0x7f12105442b9]
/usr/lib/libolacommon.so.0(_ZN3ola2io12SelectServer14CheckForEventsERKNS_12TimeIntervalE+0xc6)[0x7f121053eeb6]
/usr/lib/libolacommon.so.0(_ZN3ola2io12SelectServer3RunEv+0x7c)[0x7f121053efcc]
/usr/lib/libolaserver.so.0(_ZN3ola4http10HTTPServer3RunEv+0x88)[0x7f1210716188]
[...]

PR #1753, opened 2021-12-01
Fix UB in TimeoutManager::ExecuteTimeouts

baltitenger:
The outer for loop would call `m_events.top()` before checking if it's empty, which raises an exception when compiled with `-D_GLIBCXX_ASSERTIONS`.

### DIFF
--- a/common/io/TimeoutManager.cpp
+++ b/common/io/TimeoutManager.cpp
@@ -97,8 +97,8 @@ TimeInterval TimeoutManager::ExecuteTimeouts(TimeStamp *now) {
   if (m_events.empty())
     return TimeInterval();
 
-  for (e = m_events.top(); !m_events.empty() && (e->NextTime() <= *now);
-       e = m_events.top()) {
+  // make sure we only try to access m_events.top() if m_events isn't empty
+  while (!m_events.empty() && ((e = m_events.top())->NextTime() <= *now)) {
     m_events.pop();
 
     // if this was removed, skip it
--- a/.codespellignorewords
+++ b/.codespellignorewords
@@ -1,3 +1,4 @@
-thead
 acn
 ACN
+inflight
+thead
--- a/.travis.yml
+++ b/.travis.yml
@@ -349,6 +349,13 @@ jobs:
            - lintian
            - moreutils
   allow_failures:
+    - os: linux
+      dist: bionic
+      arch: arm64
+      env: TASK='jshint'
+      addons:
+        apt:
+          packages:
 #    - os: osx
 #      osx_image: xcode9.4
 #      compiler: clang
--- a/include/ola/base/FlagsPrivate.h
+++ b/include/ola/base/FlagsPrivate.h
@@ -307,7 +307,7 @@ bool Flag<T>::SetValue(const std::string &input) {
 
 
 /**
- * @brief This class holds all the flags, and is responsbile for parsing the
+ * @brief This class holds all the flags, and is responsible for parsing the
  * command line.
  */
 class FlagRegistry {
--- a/javascript/ola/full/dmx_console_tab.js
+++ b/javascript/ola/full/dmx_console_tab.js
@@ -66,7 +66,7 @@ ola.DmxConsoleTab.prototype.setUniverse = function(universe_id) {
 
 
 /**
- * Called when the tab changes visibiliy.
+ * Called when the tab changes visibility.
  */
 ola.DmxConsoleTab.prototype.setActive = function(state) {
   ola.DmxConsoleTab.superClass_.setActive.call(this, state);
--- a/javascript/ola/full/dmx_monitor_tab.js
+++ b/javascript/ola/full/dmx_monitor_tab.js
@@ -38,7 +38,7 @@ goog.inherits(ola.DmxMonitorTab, ola.common.BaseUniverseTab);
 
 
 /**
- * Called when the tab changes visibiliy.
+ * Called when the tab changes visibility.
  */
 ola.DmxMonitorTab.prototype.setActive = function(state) {
   ola.DmxMonitorTab.superClass_.setActive.call(this, state);


common/io/TimeoutManager.cpp
```diff
@@ -94,11 +94,7 @@ void TimeoutManager::CancelTimeout(timeout_id id) {
 
 TimeInterval TimeoutManager::ExecuteTimeouts(TimeStamp *now) {
   Event *e;
-  if (m_events.empty())
-    return TimeInterval();
```
peternewman, line 98: Sorry I don't follow @baltitenger , surely it's done the empty check here?

Or are you talking about some race condition where it gets emptied by another external process after this check?

baltitenger: I only removed those lines because they were redundant. The actual problem is in the for loop's iteration expression: `e = m_events.top()` would run before the loop's test (`!m_events.empty()`), but this can call `.top()` on an empty queue. Since you didn't access `e` in that case, you didn't actually get a segfault, but it's still undefined behavior, so there is a check for it in glibc when compiled with `_GLIBCXX_ASSERTIONS`.

peternewman: > I only removed those lines because they were redundant.

But they weren't.

> The actual problem is in the for loop's iteration expression: `e = m_events.top()` would run before the loop's test (`!m_events.empty()`), but this can call `.top()` on an empty queue.

But it can't because it's covered by the fail early (which you've removed) which had already checked whether m_events was empty.

> Since you didn't access `e` in that case, you didn't actually get a segfault, but it's still undefined behavior, so there is a check for it in glibc when compiled with `_GLIBCXX_ASSERTIONS`.

Would you mind sharing the compiler error output please.

baltitenger: There's no compile error, only a runtime crash (saying some assertion failed).
I'm not talking about the initial entry to the function, that was indeed protected by that check. However, on the next iteration of the loop,  the queue can be empty again.

peternewman: > There's no compile error, only a runtime crash on line 101 (saying some assertion failed).

Ah gotcha, in which case the next question is does the test catch it, if so it's probably worth adding to our CI so we can catch such issues automagically in future.

> I'm not talking about the initial entry to the function, that was indeed protected by that check. However, on the next iteration of the loop, the queue can be empty again.

Yes, I see now, if we had one event, it would pop it, then call m_events.top() which is undefined, although it wouldn't actually use that value afterwards.

common/io/TimeoutManager.cpp
```diff
@@ -94,11 +94,7 @@ void TimeoutManager::CancelTimeout(timeout_id id) {
...
-
```
peternewman, line 99: Code review wise, please could we put back the early return, yes it's technically redundant as the return after the while loop will catch it, but it's a coding style we use throughout so I think we'd prefer to leave it please.

common/io/TimeoutManager.cpp
```diff
@@ -94,11 +94,7 @@ void TimeoutManager::CancelTimeout(timeout_id id) {
...
-  for (e = m_events.top(); !m_events.empty() && (e->NextTime() <= *now);
-       e = m_events.top()) {
+  while (!m_events.empty() && (e = m_events.top())->NextTime() <= *now) {
```
peternewman, line 97: A few more brackets for clarity:
```suggestion
  while (!m_events.empty() && ((e = m_events.top())->NextTime() <= *now)) {
```

peternewman (changes requested):
Code change wise, LGTM, just some minor style stuff.

As per the other discussion, if the tests can catch this, I'd like to add it to the CI before we merge this please (in a TDD style).

peternewman:
Also would you mind targeting the 0.10 branch please?

baltitenger:
Well, I don't think you could add a test for this. You could just define `_GLIBCXX_ASSERTIONS`, assuming your testing environment uses glibc. It will happen consistently when creating a universe iirc. (However, it's possible this would introduce many other crashes)

peternewman:
> Well, I don't think you could add a test for this. You could just define `_GLIBCXX_ASSERTIONS`, assuming your testing environment uses glibc. It will happen consistently when creating a universe iirc. (However, it's possible this would introduce many other crashes)

So if you run `make test` with that defined, does it throw this error (or fail on an earlier on)?

I'm sort of curious how you came across it too...

baltitenger:
I haven't tried running the test suite yet. I just created an archlinux package for ola, but arch's default compile flags include `-D_GLIBCXX_ASSERTIONS`, so it just occured naturally (when creating a new universe)

baltitenger:
BTW, I just tried `make && make check` with CFLAGS containing `-D_GLIBCXX_ASSERTIONS`, and 5 tests fail without my fix. On my branch all of them complete successfully. Just in case you need it, the failing tests are `common/io/SelectServerTester`, `common/io/TimeoutManagerTester`, `plugins/artnet/ArtNetTester`, `plugins/usbpro/DmxterWidgetTester`, `plugins/usbpro/WidgetDetectorThreadTester`, and the flags I used were arch's defaults:
```sh
CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions \
        -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security \
        -fstack-clash-protection -fcf-protection"
CXXFLAGS="$CFLAGS -Wp,-D_GLIBCXX_ASSERTIONS"
LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"
```

cbix (approved):
Would be great to see this merged and released, this fixes #1726 for me on Arch Linux as well!

peternewman:
> BTW, I just tried `make && make check` with CFLAGS containing `-D_GLIBCXX_ASSERTIONS`, and 5 tests fail without my fix. On my branch all of them complete successfully.

Great, that's what we want!

common/io/TimeoutManager.cpp
```diff
@@ -97,8 +97,7 @@ TimeInterval TimeoutManager::ExecuteTimeouts(TimeStamp *now) {
   if (m_events.empty())
     return TimeInterval();
 
-  for (e = m_events.top(); !m_events.empty() && (e->NextTime() <= *now);
-       e = m_events.top()) {
+  while (!m_events.empty() && ((e = m_events.top())->NextTime() <= *now)) {
```
peternewman, line 100: Can we add a comment for our future selves please, something like this:

```suggestion
  // Use a while loop rather than a for loop, so the iteration expression
  // assigning m_events.top() isn't executed after the loop on a potentially
  // empty m_events queue.
  while (!m_events.empty() && ((e = m_events.top())->NextTime() <= *now)) {
```

peternewman (approved):
Sorry for the huge delay (as often unfortunately).

Can you just add that comment, or something similar, and resync and then this can be merged!

baltitenger:
@peternewman what should I resync with, `0.10` or `master`?

peternewman:
> @peternewman what should I resync with, `0.10` or `master`?

0.10 please @baltitenger

baltitenger:
> > @peternewman what should I resync with, `0.10` or `master`?
> 
> 0.10 please @baltitenger

All right, that should be done. Added a comment as well, hope that's enough.

baltitenger:
um, did you mean to push those here?

peternewman (approved):
Great, thanks for the final changes

peternewman:
> um, did you mean to push those here?

Yeah sorry, just trying to get the build clean before merging so updating some typos and similar and sneaking it into your PR! 😁

Issue #1755, opened 2021-12-23
Problem with python version 3.10 test


mikacousin:
Hi,
I just found a bug with the python version check.
It does not work with python 3.10.

Currently, we have:
`if sys.version >= '3.2':`
which does not match with 3.10.

It looks like this is working:
`if sys.version_info >= (3, 2):`

Regards

peternewman:
Hi @mikacousin ,

Thanks for reporting.

I was initially a bit confused, but its obviously because the broken version is doing a naïve string comparison and 1 < 2!

I assume you mean here (and elsewhere in that file):
https://github.com/OpenLightingProject/ola/blob/c9add79a8d1c83ab209146006dae87906803b64f/python/ola/PidStore.py#L661

And likewise (and elsewhere):
https://github.com/OpenLightingProject/ola/blob/c9add79a8d1c83ab209146006dae87906803b64f/python/ola/OlaClient.py#L955

We already do that in https://github.com/OpenLightingProject/ola/blob/c9add79a8d1c83ab209146006dae87906803b64f/include/ola/gen_callbacks.py#L25

So we should be good.

Was that all you needed to fix Python 3.10 compatibility? Would you mind opening a PR with those changes please?

Would you mind testing #1723 please while you're there, so we can get that merged in too, and then I think we'll hopefully be properly Python 3 compatible!

mikacousin:
Hi @peternewman,

That's the only problem I found with python 3.10 for now :-)
PR is done, but I'll test #1723 in a few days if I found some time ...
Thanks a lot for all your work on ola !

peternewman:
> That's the only problem I found with python 3.10 for now :-)

:crossed_fingers: Roughly what have you done of these via Python:
Send DMX
Receive DMX
Patch universes
List plugins
Send RDM

> PR is done, but I'll test #1723 in a few days if I found some time ...

Great thanks.

> Thanks a lot for all your work on ola !

No worries

peternewman:
Fixed in #1757

PR #1756, opened 2021-12-24
Better way to test python version

mikacousin:
Fix issue  #1755


peternewman (changes requested):
Please can you do this against 0.10, but otherwise LGTM thanks!

PR #1757, opened 2021-12-24
Better way to test Python version

### LINKED ISSUE #1755: Problem with python version 3.10 test

mikacousin:
Fix #1755


peternewman (approved):
LGTM thanks!

PR #1758, opened 2021-12-29
currentThread is deprecated, so let's use current_thread (should work…

mikacousin:
… with Python 2.6+)


peternewman (approved):
LGTM Thanks

PR #1761, opened 2021-12-31
Fix vast majority of outstanding Python 3 issues

peternewman:
Found when trying to actually run the tools.

~Untested on Python < 3.~

- [x] Fix `--list-tests` and the RDM responder test class selection stuff in `GetTestClasses` due to them not being loaded
- [x] Fix `DownloadModelData` returning blank info
- [ ] Fix SetNonAsciiDeviceLabel causing a UTF-8 decode error on the validate
- [ ] Decide if this UTF-8 supporting behaviour is the correct way to go, or if we should wait for the new E1.20 to be released with it in


kripton (approved):
As far as I'm able to understand Python, it all makes sense to me :)

Issue #1674, opened 2020-10-26
Replace UUCP locks with flock


M-Reimer:
I'm no user of OLA. I just did some research about device locking as I had trouble with a closed source application which expects to have write access to /var/lock (which seems to be not the case on a increasing number of distributions).

UUCP locking is considered obsolete and for legacy applications only. Some more information can be found on this Debian bug report where they deprecate their own library for simplifying UUCP locks:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=734086

The (way more simple) alternative is to just use flock() on the opened TTY device:

https://linux.die.net/man/2/flock

Replacing the old UUCP lock code with just flock would also fix #1189

peternewman:
Thanks for this @M-Reimer .

I suspect as they mention in the Debian bug we probably need to add an option to fallback to UUCP or possibly even do both (because all your tools need to use the same locking method), but default to flock if available on your OS.

M-Reimer:
If you target Linux only you should be able to remove the whole UUCP stuff completely. Every "modern" kernel should have flock support.

If you also target BSD or similar Unix derivatives, then you would have to check for those. I don't have any experience with anything besides Linux.

Please don't do both. If /var/lock is non-writable, you would still fail and as only a fragment of applications ever implemented the complex UUCP locking, it is very likely that someone will just use the port without any kind of lock, anyway.

PR #1770, opened 2022-02-07
Use flock() instead of UUCP-style locking for serial devices

### LINKED ISSUE #1189: UUCP locking fail in archlinuxARM
Hi,
I've got some problems running ola on my raspberry pi B2, with ArchlinuxARM.
I have the latest version of everything.
Basically ola is telling that it does not have the permission to access / modify /var/lock/LCK..ttyUSB0

I have explained everything on the [archlinux ARM forum](https://archlinuxarm.org/forum/viewtopic.php?f=60&t=11184), and posted a message in ola's [google mailing list])https://groups.google.com/d/msg/open-lighting/GYt4Ne3qQzM/1j9-OGfICQAJ) too

In #960, i see that you added uucp locking. I got the message that the permission required is not checked, but in fact there is no LCK..ttyUSB0 file in /var/lock/ ... So I don't know who to tell this to, you the ola dev team or the archlinux ARM team for driver oriented solution maybe ?

Please see my archlinux ARM post for full logs and infos.
Thank you for reading.

taw10:
This replaces AcquireUUCPLockAndOpen() with AcquireLockAndOpen().  Depending on a compile-time option (--enable-uucp-locking), either flock() or UUCP-style lockfiles will be used as the underlying lock mechanism.  The default is to use flock().  The configure script will fall back on UUCP locking if flock() is not found.

It would also be possible to preserve the API by naming the top-level open/lock routine AcquireUUCPLockAndOpen, but then the routine name would not be an accurate description.  I've updated the plugins using these routines (only two of them, StageProfi and UsbPro).

ReleaseUUCPLock() is still present, becoming a no-op if flock() is being used.  With flock(), the lock is automatically released when the file descriptor is closed.

In practice, the TIOCEXCL lock (when available) is much more helpful, because it prevents any other process from opening the device and doesn't rely on all processes using the same locking scheme.  This lock is still used when flock() is available.

Fixes #1189, #1674 and many related problems on various platforms (e.g. running in a Toolbox container on Fedora Silverlight).

### DIFF
--- a/common/io/Serial.cpp
+++ b/common/io/Serial.cpp
@@ -27,6 +27,7 @@
 #include <signal.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/file.h>
 #ifdef _WIN32
 #define VC_EXTRALEAN
 #define WIN32_LEAN_AND_MEAN
@@ -57,6 +58,7 @@ using std::string;
 
 namespace {
 
+#ifdef UUCP_LOCKING
 string GetLockFile(const string &path) {
   const string base_name = ola::file::FilenameFromPath(path);
   return ola::file::JoinPaths(UUCP_LOCK_DIR, "LCK.." + base_name);
@@ -113,6 +115,7 @@ bool RemoveLockFile(const string &lock_file) {
   }
   return true;
 }
+#endif  // UUCP_LOCKING
 
 }  // namespace
 
@@ -140,7 +143,41 @@ bool UIntToSpeedT(uint32_t value, speed_t *output) {
   return false;
 }
 
+#ifdef HAVE_FLOCK
+bool OpenAndFlock(const std::string &path, int oflag, int *fd) {
+  // First, check if the path exists, there's no point trying to open it if not
+  if (!FileExists(path)) {
+    OLA_INFO << "Device " << path << " doesn't exist.";
+    return false;
+  }
 
+  // Now try to open the serial device.
+  if (!TryOpen(path, oflag, fd)) {
+    OLA_DEBUG << "Failed to open device " << path;
+    return false;
+  }
+
+  if (flock(*fd, LOCK_EX | LOCK_NB) == -1) {
+    OLA_INFO << "Failed to flock() device " << path;
+    close(*fd);
+    return false;
+  }
+
+#if HAVE_SYS_IOCTL_H
+  // As a final safety mechanism, use ioctl(TIOCEXCL) if available to prevent
+  // further opens.
+  if (ioctl(*fd, TIOCEXCL) == -1) {
+    OLA_WARN << "TIOCEXCL " << path << " failed: " << strerror(errno);
+    close(*fd);
+    return false;
+  }
+#endif  // HAVE_SYS_IOCTL_H
+  return true;
+}
+#endif // HAVE_FLOCK
+
+
+#ifdef UUCP_LOCKING
 bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
   // This is rather tricky since there is no real convention for LCK files.
   // If it was only a single process doing the locking we could use fnctl as
@@ -231,9 +268,11 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
 #endif  // HAVE_SYS_IOCTL_H
   return true;
 }
+#endif // UUCP_LOCKING
 
 
 void ReleaseUUCPLock(const std::string &path) {
+#ifdef UUCP_LOCKING
   const string lock_file = GetLockFile(path);
 
   pid_t locked_pid;
@@ -247,6 +286,16 @@ void ReleaseUUCPLock(const std::string &path) {
       OLA_INFO << "Released " << lock_file;
     }
   }
+#endif  /* else no action needed */
 }
+
+bool AcquireLockAndOpen(const std::string &path, int oflag, int *fd) {
+#ifdef UUCP_LOCKING
+	return AcquireUUCPLockAndOpen(path, oflag, fd);
+#else
+	return OpenAndFlock(path, oflag, fd);
+#endif
+}
+
 }  // namespace io
 }  // namespace ola
--- a/include/ola/io/Serial.h
+++ b/include/ola/io/Serial.h
@@ -60,21 +60,27 @@ typedef enum {
 bool UIntToSpeedT(uint32_t value, speed_t *output);
 
 /**
- * @brief Try to open the path, respecting UUCP locking.
+ * @brief Try to open the path and obtain a lock to control access
  * @param path the path to open
  * @param oflag flags passed to open
  * @param[out] fd a pointer to the fd which is returned.
  * @returns true if the open succeeded, false otherwise.
  *
+ * Depending on the compile-time configuration, this will use either flock()
+ * (the default) or UUCP locking.  See: ./configure --enable-uucp-locking.
+ *
  * This fails-fast, it we can't get the lock immediately, we'll return false.
  */
-bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd);
+bool AcquireLockAndOpen(const std::string &path, int oflag, int *fd);
 
 /**
  * @brief Remove a UUCP lock file for the device.
  * @param path The path to unlock.
  *
  * The lock is only removed if the PID matches.
+ *
+ * Does nothing if UUCP locking is not in use
+ * (see ./configure --enable-uucp-locking).
  */
 void ReleaseUUCPLock(const std::string &path);
 }  // namespace io
--- a/plugins/usbpro/BaseUsbProWidget.cpp
+++ b/plugins/usbpro/BaseUsbProWidget.cpp
@@ -129,8 +129,7 @@ ola::io::ConnectedDescriptor *BaseUsbProWidget::OpenDevice(
     const string &path) {
   struct termios newtio;
   int fd;
-  if (!ola::io::AcquireUUCPLockAndOpen(path, O_RDWR | O_NONBLOCK | O_NOCTTY,
-                                       &fd)) {
+  if (!ola::io::AcquireLockAndOpen(path, O_RDWR | O_NONBLOCK | O_NOCTTY, &fd)) {
     return NULL;
   }
 
--- a/configure.ac
+++ b/configure.ac
@@ -787,6 +787,22 @@ AC_ARG_ENABLE(
   [AS_HELP_STRING([--disable-doxygen-version],
   [Substitute the Doxygen version with latest, for the website])])
 
+# Use UUCP locking?
+AC_ARG_ENABLE(
+  [uucp-locking],
+  [AS_HELP_STRING([--enable-uucp-locking],
+  [Use UUCP locking instead of flock()])])
+
+AC_CHECK_FUNC([flock], [AC_DEFINE([HAVE_FLOCK], [1],
+                                  [Define if flock() is available])
+                        have_flock=yes], [])
+
+AS_IF([test "x$enable_uucp_locking" = xyes || test "x$have_flock" != xyes],
+      [AC_DEFINE([UUCP_LOCKING], [1],
+                 [Define if UUCP locking should be used instead of flock()])
+       SERIAL_LOCK_METHOD=UUCP],
+      [SERIAL_LOCK_METHOD=flock])
+
 # UUCP Lock directory
 AC_ARG_WITH([uucp-lock],
             [AS_HELP_STRING([--with-uucp-lock],
@@ -1022,6 +1038,7 @@ Enable HTTP Server: ${have_microhttpd}
 RDM Responder Tests: ${enable_rdm_tests}
 Ja Rule: ${BUILDING_JA_RULE}
 Enabled Plugins:${PLUGINS}
+Serial port locking: $SERIAL_LOCK_METHOD
 UUCP Lock Directory: $UUCPLOCK
 
 Now type 'make @<:@<target>@:>@'
--- a/plugins/stageprofi/StageProfiDetector.cpp
+++ b/plugins/stageprofi/StageProfiDetector.cpp
@@ -164,9 +164,9 @@ ConnectedDescriptor* StageProfiDetector::ConnectToUSB(
   struct termios newtio;
 
   int fd;
-  if (!ola::io::AcquireUUCPLockAndOpen(widget_path,
-                                       O_RDWR | O_NONBLOCK | O_NOCTTY,
-                                       &fd)) {
+  if (!ola::io::AcquireLockAndOpen(widget_path,
+                                   O_RDWR | O_NONBLOCK | O_NOCTTY,
+                                   &fd)) {
     return NULL;
   }
 


common/io/Serial.cpp
```diff
@@ -231,9 +268,11 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
 #endif  // HAVE_SYS_IOCTL_H
   return true;
 }
+#endif // UUCP_LOCKING
 
 
 void ReleaseUUCPLock(const std::string &path) {
+#ifdef UUCP_LOCKING
```
peternewman, line 275: I think it would be good to be consistent here too and create a ReleaseLock function, where the non UUCP path does nothing.

common/io/Serial.cpp
```diff
@@ -27,6 +27,7 @@
 #include <signal.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/file.h>
```
peternewman, line 30: Has sys/file.h existed forever, so we're not breaking UUCP builds which don't have it?

common/io/Serial.cpp
```diff
@@ -57,6 +58,7 @@ using std::string;
 
 namespace {
 
+#ifdef UUCP_LOCKING
```
peternewman, line 61: I'm not sure of the benefit of not defining these if we're not in UUCP mode, it looks like they should work in all cases.

common/io/Serial.cpp
```diff
@@ -57,6 +58,7 @@ using std::string;
...
 string GetLockFile(const string &path) {
```
peternewman, line 62: This should probably be renamed though.

common/io/Serial.cpp
```diff
@@ -140,7 +143,41 @@ bool UIntToSpeedT(uint32_t value, speed_t *output) {
   return false;
 }
 
+#ifdef HAVE_FLOCK
+bool OpenAndFlock(const std::string &path, int oflag, int *fd) {
```
peternewman, line 147: I think we've generally gone for having this the other way round and returning false, rather than this way where the build would break when it only impacts a small fraction of the functionality.

common/io/Serial.cpp
```diff
@@ -247,6 +286,16 @@ void ReleaseUUCPLock(const std::string &path) {
       OLA_INFO << "Released " << lock_file;
     }
   }
+#endif  /* else no action needed */
```
peternewman, line 289: Please add an else and at least an OLA_DEBUG if not OLA_WARN, so the logs will show the function incorrectly getting called.

include/ola/io/Serial.h
```diff
@@ -60,21 +60,27 @@ typedef enum {
...
  * @param[out] fd a pointer to the fd which is returned.
  * @returns true if the open succeeded, false otherwise.
  *
+ * Depending on the compile-time configuration, this will use either flock()
+ * (the default) or UUCP locking.  See: ./configure --enable-uucp-locking.
+ *
  * This fails-fast, it we can't get the lock immediately, we'll return false.
  */
-bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd);
+bool AcquireLockAndOpen(const std::string &path, int oflag, int *fd);
```
peternewman, line 74: Please leave the other one documented, but deprecate it and @see the new function.

common/io/Serial.cpp
```diff
@@ -113,6 +115,7 @@ bool RemoveLockFile(const string &lock_file) {
   }
```
peternewman, line 115: And this function ^ wants renaming too please.

peternewman (changes requested):
Hi @taw10 ,

Thanks for this!

There are some lint issues if you could fix them please:
https://app.travis-ci.com/github/OpenLightingProject/ola/jobs/558598505#L1483-L1487

If possible a test for this would be awesome! I.e. lock a file and try and open it again and confirm it fails etc. Which will then let people know during make test if their chosen locking hasn't worked.

common/io/Serial.cpp
taw10: Done: added ReleaseSerialPortLock

common/io/Serial.cpp
taw10: I'm not sure, I'll investigate.  However the header is checked by the configure script, and used without any #ifdefs in one other place (KarateLight).

common/io/Serial.cpp
taw10: Ok, I removed the ifdefs, which reduces the noise a bit.

common/io/Serial.cpp
taw10: Done - renamed to GetUUCPLockFIle.

common/io/Serial.cpp
taw10: Ok, done.  Now it will try to open the device, but return false instead of locking it.

common/io/Serial.cpp
taw10: No longer relevant, I think.  ReleaseUUCPLock now unconditionally tries a UUCP-style unlock, so the error should be obvious if it gets called instead of the new ReleaseSerialPortLock.

include/ola/io/Serial.h
taw10: Ok, done.  The two pairs of routines (lock/unlock) now "@see" each other, and the two UUCP versions "@deprecated" to the corresponding new routines.

common/io/Serial.cpp
taw10: Done - now it's RemoveUUCPLockFile.

taw10:
Two more things still to do (add a test and check the situation with sys/file.h), but this should cover most of the comments.  The lint errors should be fixed (I may have introduced new errors, let's see).  I've also renamed the top-level lock/unlock routines, because "ReleaseLock" seemed too vague.  Now it's ReleaseSerialPortLock.

Note that in the flock() case, the unit test will really only test the TIOCEXCL locking, since that will prevent the open() call from succeeding before we try to flock().  Working round that would introduce some nasty complexity.  But, as long as one of the locking mechanisms works, it's OK.  With UUCP lockfiles, the check is done before we try to open the port itself.

common/io/Serial.cpp
```diff
@@ -157,11 +154,16 @@ bool OpenAndFlock(const std::string &path, int oflag, int *fd) {
...
 
+#ifdef HAVE_FLOCK
   if (flock(*fd, LOCK_EX | LOCK_NB) == -1) {
     OLA_INFO << "Failed to flock() device " << path;
     close(*fd);
     return false;
   }
+#else  // HAVE_FLOCK
+  close(*fd);
```
peternewman, line 164: Probably chuck a log line in here saying it's tried to flock without flock, so someone doesn't get confused why it never locks if this branch is ever triggered.

common/io/Serial.cpp
```diff
@@ -282,20 +282,26 @@ void ReleaseUUCPLock(const std::string &path) {
...
-    if (RemoveLockFile(lock_file)) {
+    if (RemoveUUCPLockFile(lock_file)) {
       OLA_INFO << "Released " << lock_file;
     }
   }
-#endif  /* else no action needed */
 }
 
-bool AcquireLockAndOpen(const std::string &path, int oflag, int *fd) {
+bool AcquireLockAndOpenSerialPort(const std::string &path, int oflag, int *fd) {
```
peternewman, line 291: How come this only works for a serial port? Aside from the fact we've got it in Serial.cpp? Won't it work in any file?

include/ola/io/Serial.h
```diff
@@ -59,6 +59,31 @@ typedef enum {
...
+/**
+ * @brief Try to open the path, respecting UUCP locking.
+ * @param path the path to open
+ * @param oflag flags passed to open
+ * @param[out] fd a pointer to the fd which is returned.
+ * @returns true if the open succeeded, false otherwise.
+ *
+ * This fails-fast, it we can't get the lock immediately, we'll return false.
+ *
+ * @deprecated Use AcquireLockAndOpenSerialPort() instead.
```
peternewman, line 71: Please add a date to the deprecation.

include/ola/io/Serial.h
```diff
@@ -59,6 +59,31 @@ typedef enum {
...
+ */
+bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd);
+
+/**
+ * @brief Remove a UUCP lock file for the device.
+ * @param path The path to unlock.
+ *
+ * The lock is only removed if the PID matches.
+ *
+ * @deprecated Use ReleaseSerialPortLock() instead.
```
peternewman, line 82: Again

include/ola/io/Serial.h
```diff
@@ -70,19 +95,24 @@ bool UIntToSpeedT(uint32_t value, speed_t *output);
...
-bool AcquireLockAndOpen(const std::string &path, int oflag, int *fd);
+bool AcquireLockAndOpenSerialPort(const std::string &path, int oflag, int *fd);
 
 /**
- * @brief Remove a UUCP lock file for the device.
+ * @brief Release a lock for the serial port
  * @param path The path to unlock.
  *
- * The lock is only removed if the PID matches.
+ * If UUCP locking was used (see AcquireLockAndOpenSerialPort), the lockfile
```
peternewman, line 107: I can't remember if this will auto-link, or whether you need to give the full namespace path for that.

common/io/Serial.cpp
peternewman: Thanks, although please can we move the equivalent debug call to the new ReleaseSerialPortLock function instead please.

common/io/Serial.cpp
peternewman: Ah okay, obviously that plugin breaking wouldn't have quite so much impact, but the plugin has been around for a while and no-one has complained so we're probably good...

plugins/usbpro/BaseUsbProWidget.cpp
```diff
@@ -129,7 +129,7 @@ ola::io::ConnectedDescriptor *BaseUsbProWidget::OpenDevice(
     const string &path) {
   struct termios newtio;
   int fd;
-  if (!ola::io::AcquireLockAndOpen(path, O_RDWR | O_NONBLOCK | O_NOCTTY, &fd)) {
+  if (!ola::io::AcquireLockAndOpenSerialPort(path, O_RDWR | O_NONBLOCK | O_NOCTTY, &fd)) {
```
peternewman, line 132: To fix the lint error:
```suggestion
  if (!ola::io::AcquireLockAndOpenSerialPort(path,
                                             O_RDWR | O_NONBLOCK | O_NOCTTY,
                                             &fd)) {
```

peternewman (changes requested):
A few more tweaks

peternewman:
> Two more things still to do (add a test and check the situation with sys/file.h), but this should cover most of the comments.

Great thanks.

> The lint errors should be fixed (I may have introduced new errors, let's see).

Only the one! :laughing: 

> I've also renamed the top-level lock/unlock routines, because "ReleaseLock" seemed too vague. Now it's ReleaseSerialPortLock.

As mentioned in the comment, won't this lock work on any file descriptor? Maybe the UUCP stuff is only implemented for devices?

> Note that in the flock() case, the unit test will really only test the TIOCEXCL locking, since that will prevent the open() call from succeeding before we try to flock(). Working round that would introduce some nasty complexity. But, as long as one of the locking mechanisms works, it's OK.

Exactly, I'm more interested in catching an issue during CI/build and around our logic than I am bothered about thoroughly testing this particular system feature. If we prove we can't open a file twice due to the lock then that's a good test as far as I'm concerned.

> With UUCP lockfiles, the check is done before we try to open the port itself.

Yep, understood.

common/io/Serial.cpp
taw10: he flock call itself conforms to 4.4BSD, i.e. it's been around since 1997.  So I think we're pretty safe here.

I could add guards around the #include (ifdef HAVE_SYS_FILE_H), but if it's worked so far then I don't see any reason to add the extra complexity.   On inspection, I'm not actually sure why KarateLight includes this header at all.

common/io/Serial.cpp
taw10: I've added an OLA_INFO message for the flock() path in ReleaseSerialPortLock.  Is that what you meant?  Or, did you mean to add a return value to ReleaseUUCPLock and check it in ReleaseSerialPortLock?

common/io/Serial.cpp
taw10: Ok, done.  Now it will OLA_WARN "Tried to lock XXX, but flock() is unavailable.".

common/io/Serial.cpp
taw10: Technically, flock() will work for any file of course.  UUCP wouldn't, though, because the lockfile name is made from the base name.  Semantically, the UUCP method is not really defined for anything other than a serial port.  So I don't think it makes sense to expand these routines to other things.

include/ola/io/Serial.h
taw10: Done!

include/ola/io/Serial.h
taw10: Done!

include/ola/io/Serial.h
taw10: I've checked, and added parentheses but not the whole path.  This makes it link correctly.

taw10:
Unit test is still to come (I have to get myself acquainted with the test system first).  One more question: would you prefer to have these commits squashed down to at least three (Codespell fix, unit test, everything else)?  The overall diff is very small and understandable.  Or do you prefer to have the complete history?

taw10:
I've added a test, but it's not very satisfactory for a couple of reasons.  Let me know what you think!

The problem is that we don't have a real serial port to test with, and the TIOCEXCL lock won't work with a normal file.  To avoid creating a new file, I used the ChangeLog file for the test since it's always present at the top level of the project.  This feels bad to me, but the alternatives all seem less portable.  Maybe you or someone else has a suggestion.

To prevent the normal file from failing the test, I changed the two locking routines to not fail if `ioctl(fd, TIOCEXCL)` fails.  Perhaps this is an improvement, because we only need one of the three locking mechanisms to work (flock, UUCP or TIOCEXCL), but it also feels bad.

taw10:
> To avoid creating a new file, I used the ChangeLog file for the test since it's always present at the top level of the project. This feels bad to me, but the alternatives all seem less portable. Maybe you or someone else has a suggestion.

This doesn't work with out-of-tree builds (nice that these work with OLA!), so now the test will create the file itself.  The file gets deleted afterwards.  If the file exists already for some reason (`serialLockTestFile` in the working directory), the test will fail.

common/io/Serial.cpp
peternewman: Yeah very good point about the fact it's worked!

I'm not sure about KarateLight either; @cpresser ?

common/io/Serial.cpp
peternewman: No that was what I meant, moving where the log message is generated.

common/io/Serial.cpp
```diff
@@ -299,6 +295,8 @@ bool AcquireLockAndOpenSerialPort(const std::string &path, int oflag, int *fd) {
 void ReleaseSerialPortLock(const std::string &path) {
 #ifdef UUCP_LOCKING
   ReleaseUUCPLock(path);
+#else   // UUCP_LOCKING
+  OLA_INFO << "No unlock necessary for " << path;
```
peternewman, line 299: Can we change to one of these:
```suggestion
  OLA_INFO << "No unlock necessary with flock() for " << path;
```
or
```suggestion
  OLA_INFO << "No unlock necessary with for " << path << " as we're using flock()";
```

common/io/Serial.cpp
```diff
@@ -161,6 +161,7 @@ bool OpenAndFlock(const std::string &path, int oflag, int *fd) {
     return false;
   }
 #else  // HAVE_FLOCK
```
peternewman, line 163: We don't normally bother commenting the else (and I personally think it makes it a bit more confusing as it's !HAVE_FLOCK).
```suggestion
#else
```

common/io/Serial.cpp
```diff
@@ -170,8 +171,6 @@ bool OpenAndFlock(const std::string &path, int oflag, int *fd) {
   // further opens.
   if (ioctl(*fd, TIOCEXCL) == -1) {
     OLA_WARN << "TIOCEXCL " << path << " failed: " << strerror(errno);
-    close(*fd);
```
peternewman: Shouldn't we still close to tidy up?

common/io/Serial.cpp
```diff
@@ -170,8 +171,6 @@ bool OpenAndFlock(const std::string &path, int oflag, int *fd) {
...
-    return false;
```
peternewman: Shouldn't it still return false if it's failed to TIOCEXCL?

common/io/Serial.cpp
```diff
@@ -263,9 +262,6 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
   // further opens.
   if (ioctl(*fd, TIOCEXCL) == -1) {
     OLA_WARN << "TIOCEXCL " << path << " failed: " << strerror(errno);
-    close(*fd);
```
peternewman, line 227: Again we should close.

common/io/Serial.cpp
```diff
@@ -263,9 +262,6 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
...
-    RemoveUUCPLockFile(lock_file);
```
peternewman: We should definitely try and tidy up shouldn't we so we don't leave a stale file?

common/io/Serial.cpp
```diff
@@ -263,9 +262,6 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
...
-    return false;
```
peternewman, line 229: Again we should return false.

common/io/SerialLockTester.cpp
```diff
@@ -0,0 +1,73 @@
...
+  void testLock();
+};
+
+
+CPPUNIT_TEST_SUITE_REGISTRATION(SerialLockTest);
+
+void SerialLockTest::testLock() {
+  bool r1, r2;
+  int fd1, fd2, fd3;
+  const std::string path = "serialLockTestFile";
```
peternewman, line 50: Given this could appear in system UUCP directories couldn't it?
```suggestion
  const std::string path = "olaSerialLockTestFile";
```

common/io/SerialLockTester.cpp
```diff
@@ -0,0 +1,73 @@
...
+#ifndef _WIN32
+             | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH
+#endif  // !_WIN32
+             );  // NOLINT(whitespace/parens)
+  OLA_ASSERT_FALSE(fd3 < 0);
+  close(fd3);
+
+  r1 = ola::io::AcquireLockAndOpenSerialPort(path, O_RDWR, &fd1);
+  OLA_ASSERT_TRUE(r1);
+
```
peternewman, line 65: Perhaps check fd1 is valid/can be written to?

include/ola/io/Serial.h
```diff
@@ -68,7 +68,7 @@ bool UIntToSpeedT(uint32_t value, speed_t *output);
  *
  * This fails-fast, it we can't get the lock immediately, we'll return false.
  *
- * @deprecated Use AcquireLockAndOpenSerialPort() instead.
+ * @deprecated Use AcquireLockAndOpenSerialPort() instead (19 Feb 2022).
```
peternewman, line 71:
```suggestion
 * @deprecated Use the more general AcquireLockAndOpenSerialPort() instead (19 Feb 2022).
```

peternewman (changes requested):
A few minor tweaks

peternewman:
> One more question: would you prefer to have these commits squashed down to at least three (Codespell fix, unit test, everything else)? The overall diff is very small and understandable.

If you see some of the commit history, you'll see I'm not too hot on that!

>  Or do you prefer to have the complete history?

Certainly while doing code reviews, I hate it when people force push as it breaks a lot of comment features within GitHub! I'm not really bothered either way and I think it occasionally helps when looking at blames etc.

common/io/Makefile.mk
```diff
@@ -47,7 +47,8 @@ test_programs += \
     common/io/MemoryBlockTester \
     common/io/SelectServerTester \
     common/io/StreamTester \
-    common/io/TimeoutManagerTester
+    common/io/TimeoutManagerTester \
+    common/io/SerialLockTester
```
peternewman, line 51: Please add this new test to .gitignore.

peternewman:
> I've added a test, but it's not very satisfactory for a couple of reasons. Let me know what you think!

Great thanks.

> The problem is that we don't have a real serial port to test with, and the TIOCEXCL lock won't work with a normal file.

Is there any reason not to try locking a real serial port too if one exists? See the messaging ability at the bottom of this comment it could just fail saying maybe the port is in use. Especially if it tested ALL ports on a machine there's a good chance one would be available...

> To prevent the normal file from failing the test, I changed the two locking routines to not fail if `ioctl(fd, TIOCEXCL)` fails. Perhaps this is an improvement, because we only need one of the three locking mechanisms to work (flock, UUCP or TIOCEXCL), but it also feels bad.

Yeah it doesn't feel great to me.

Would it make sense to split each of the individual lock mechanism functions out, then you could have it like so (or whatever order it's currently in):
```
CheckFileExists
#ifdef flock
LockFlock
#else
LockUUCP
#endif
LockTIOCEXCL
```

So you could then test the constituent parts individually which would work around this issue.

> > To avoid creating a new file, I used the ChangeLog file for the test since it's always present at the top level of the project. This feels bad to me, but the alternatives all seem less portable. Maybe you or someone else has a suggestion.
> 
> This doesn't work with out-of-tree builds (nice that these work with OLA!), so now the test will create the file itself. The file gets deleted afterwards.

You can thanks Simon for the out-of-tree builds working. Although I've been scratching my head a bit getting a Python test to work with them...

> If the file exists already for some reason (`serialLockTestFile` in the working directory), the test will fail.

Presumably the test failing/bombing out once before deletion would cause all subsequent tests to fail afterwards?

See the best way to tell users that's the case: https://github.com/OpenLightingProject/ola/blob/c9add79a8d1c83ab209146006dae87906803b64f/ola/OlaClientWrapperTest.cpp#L52-L53

We could also possibly generate a random filename/suffix to test against to avoid some of that (although clutter up their drive if the test ever fails).

common/io/Serial.cpp
cpresser: Unfortunately, I am not sure what to answer here. My C++ is quite rusty, I have not used it in years and sure have no idea which include contains which functionality.

I recall that I did include file locking of the serial-port-device in the karatelight code. A quick search turned up this piece of code here:
https://github.com/OpenLightingProject/ola/blob/master/plugins/karate/KarateLight.cpp#L121

Does that help?

common/io/Makefile.mk
taw10: It's actually already covered by these lines in .gitignore:
```
*Tester
*Tester.exe
```

common/io/Serial.cpp
taw10: The `#include <sys/file.h>` has been there since the first version of KarateDevice.cpp (f69a26b065adc853c3f7273d298c1589b1a5df68), but the `flock` call is in a different file (KarateLight.cpp).  I suspect the `flock` calls were originally in KarateDevice.cpp and got moved.

It does mean, however, that there's not only the #include but also a real use of flock() in OLA already without any tests!  Therefore we're probably pretty safe using it for all serial devices.  Nevertheless, we'll keep the tests and the conditionals :)

taw10:
> > The problem is that we don't have a real serial port to test with, and the TIOCEXCL lock won't work with a normal file.
> 
> Is there any reason not to try locking a real serial port too if one exists? See the messaging ability at the bottom of this comment it could just fail saying maybe the port is in use. Especially if it tested ALL ports on a machine there's a good chance one would be available...

The idea of poking real hardware as part of `make check` worries me - we don't know what's on the other end of the serial port. Plus I'm not sure I want to write a load of serial port enumeration stuff (probably Linux-specific as well) just for this purpose.

> Would it make sense to split each of the individual lock mechanism functions out, then you could have it like so (or whatever order it's currently in):
> 
> ```
> CheckFileExists
> #ifdef flock
> LockFlock
> #else
> LockUUCP
> #endif
> LockTIOCEXCL
> ```
> 
> So you could then test the constituent parts individually which would work around this issue.

I'll implement this approach.  This way, we have a new locking mechanism (flock) and a test that it works, the UUCP stuff is still there as a fallback, and the TIOCEXCL is on top of either without a test.  This is already a big improvement over the old situation (legacy UUCP locking without a test).  The KarateLight experience (see above) suggests that the UUCP locking will _never_ be used, so the idea of testing UUCP might be purely theoretical anyway.

> Presumably the test failing/bombing out once before deletion would cause all subsequent tests to fail afterwards?

Unfortunately yes.  I'll add some cleanup so that the file gets removed in any case, and a message to make it clear that the file already exists.  However I want to avoid deleting a random user file in the event that my choice of filename wasn't "obscure" enough.

Update pending...

taw10:
Ok, AcquireLockAndOpenSerialPort() now does this:
```
check port exists
if UUCP: acquire UUCP lock
open port
if FLOCK: apply flock()
if HAVE_IOCTL: apply TIOCEXCL
```
If any stage fails, it will bail out, closing the port and releasing UUCP lock if applicable.  So, all of the available locking mechanisms (potentially three of them, see below) have to succeed.  The bailout logic is a little more complicated than I'd like, because flock() happens after opening, whereas UUCP happens before.

AcquireUUCPLockAndOpen() is still available, although no-one should be using it.  It does this:
```
check port exists
acquire UUCP lock
open port
if HAVE_IOCTL: apply TIOCEXCL
```

I've split off `acquire UUCP lock` and `apply TIOCEXCL` as separate routines.

Each step prints at least an OLA_INFO on success, so you can see which locks are used.  If configured with `--enable-uucp-locking` and flock() is available, it will actually do both - there's no reason not to use flock() if the routine is available.  The information about KarateLight suggests that flock() is _always_ available, which would lead me to suggest completely removing the UUCP locking.  But perhaps there are systems still relying on it.

The test program (SerialLockTester) tests that the flock() call succeeds, but not that it prevents a second lock.  That would be a lot of trouble to test, because the second flock would have to be from a different process.  I consider it more a simple test that the routine really exists - it doesn't seem unreasonable to rely on it doing what the specification says it should!  It now uses a PID-based filename, to prevent one failure from causing subsequent ones.  UUCP locking and TIOCEXCL are not tested at all because of the difficulties noted above, but it's still overall an improvement on the previous situation.

common/io/Serial.cpp
taw10: It now closes (and unlocks, if necessary) on all exit paths.

common/io/Serial.cpp
taw10: It now closes (and unlocks, if necessary) on all exit paths.

common/io/Serial.cpp
taw10: It now closes (and unlocks, if necessary) on all exit paths.

common/io/Serial.cpp
```diff
@@ -249,21 +205,51 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
...
+#if HAVE_SYS_IOCTL_H
+  // This is a final safety mechanism, on top of UUCP locking or flock()
+  if (ioctl(fd, TIOCEXCL) == -1) {
+    OLA_WARN << "TIOCEXCL " << path << " failed: " << strerror(errno);
+    return false;
+  } else {
+    OLA_INFO << "Set TIOCEXCL on " << path;
+  }
+#else
+    OLA_INFO << "Not setting " << path << " - ioctl.h unavailable";
```
peternewman, line 221:
```suggestion
    OLA_INFO << "Not setting TIOCEXCL on " << path << " - ioctl.h unavailable";
```

common/io/Serial.cpp
```diff
@@ -249,21 +205,51 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd) {
...
+    close(*fd);
     return false;
   }
 
-#if HAVE_SYS_IOCTL_H
-  // As a final safety mechanism, use ioctl(TIOCEXCL) if available to prevent
-  // further opens.
-  if (ioctl(*fd, TIOCEXCL) == -1) {
-    OLA_WARN << "TIOCEXCL " << path << " failed: " << strerror(errno);
+  if (!LockTIOCEXCL(*fd, path)) {
```
peternewman, line 247: We may as well keep the comment:
```suggestion
  // As a final safety mechanism, use ioctl(TIOCEXCL) if available to prevent
  // further opens.
  if (!LockTIOCEXCL(*fd, path)) {
```

common/io/Serial.cpp
```diff
@@ -285,19 +271,55 @@ void ReleaseUUCPLock(const std::string &path) {
...
 
 void ReleaseSerialPortLock(const std::string &path) {
 #ifdef UUCP_LOCKING
   ReleaseUUCPLock(path);
-#else   // UUCP_LOCKING
-  OLA_INFO << "No unlock necessary for " << path;
-#endif  // UUCP_LOCKING
+#else
+  OLA_INFO << "No unlock necessary for " << path << " (UUCP locking not used)";
+#endif
```
peternewman, line 322: I think we do want the comment here to keep cpplint happy, just not on the else:
```suggestion
#endif  // UUCP_LOCKING
```

common/io/SerialLockTester.cpp
```diff
@@ -45,29 +46,38 @@ class SerialLockTest: public CppUnit::TestFixture {
...
 void SerialLockTest::testLock() {
-  bool r1, r2;
-  int fd1, fd2, fd3;
-  const std::string path = "serialLockTestFile";
+  int fd;
+  pid_t our_pid = getpid();
 
-  OLA_ASSERT_FALSE(ola::io::FileExists(path));
+  std::stringstream str;
+  str << "serialLockTestFile." << our_pid;
```
peternewman, line 53: Probably pedantry, but it would be nice if the file doesn't look like it's some sort of special file by ending in our PID...
```suggestion
  str << "serialLockTestFile." << our_pid << ".pid";
```

peternewman (commented):
Some minor tweaking

peternewman:
> The idea of poking real hardware as part of `make check` worries me - we don't know what's on the other end of the serial port.

I think we do already as part of our autostart tests TBH (but I can see your point)...

> Plus I'm not sure I want to write a load of serial port enumeration stuff (probably Linux-specific as well) just for this purpose.

I think that code should already exist as we already do essentially that in some of the plugins that will use this new locking!

> I'll implement this approach. This way, we have a new locking mechanism (flock) and a test that it works, the UUCP stuff is still there as a fallback, and the TIOCEXCL is on top of either without a test. This is already a big improvement over the old situation (legacy UUCP locking without a test). The KarateLight experience (see above) suggests that the UUCP locking will _never_ be used, so the idea of testing UUCP might be purely theoretical anyway.

Yeah fair point.

> > Presumably the test failing/bombing out once before deletion would cause all subsequent tests to fail afterwards?
> 
> Unfortunately yes. I'll add some cleanup so that the file gets removed in any case, and a message to make it clear that the file already exists. However I want to avoid deleting a random user file in the event that my choice of filename wasn't "obscure" enough.

Lovely thanks. That should be good enough, as long as the test fails with a non-cryptic message it shouldn't generate too many issues (judging by the auto-start stuff when it's already running as a service which obviously breaks more often).

include/ola/io/Serial.h
```diff
@@ -75,8 +78,41 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd);
...
+
+/**
+ * @brief Try to open the path and obtain a lock to control access
+ * @param path the path to open
+ * @param oflag flags passed to open
+ * @param[out] fd a pointer to the fd which is returned.
+ * @returns true if the open succeeded, false otherwise.
+ *
+ * Depending on the compile-time configuration, this will use either flock()
+ * (the default) or UUCP locking.  See: ./configure --enable-uucp-locking.
```
peternewman, line 95: Isn't this now doing both/all?

include/ola/io/Serial.h
```diff
@@ -75,8 +78,41 @@ bool AcquireUUCPLockAndOpen(const std::string &path, int oflag, int *fd);
...
+bool AcquireLockAndOpenSerialPort(const std::string &path, int oflag, int *fd);
+
+/**
+ * @brief Release a lock for the serial port
+ * @param path The path to unlock.
+ *
+ * If UUCP locking was used (see AcquireLockAndOpenSerialPort()), the lockfile
+ * will be removed (but only if the PID matches).
+ *
+ * Does nothing if flock() was used (see ./configure --enable-uucp-locking).
```
peternewman, line 110: Again can't this do all/more now?

peternewman:
> Ok, AcquireLockAndOpenSerialPort() now does this:
> If any stage fails, it will bail out, closing the port and releasing UUCP lock if applicable. So, all of the available locking mechanisms (potentially three of them, see below) have to succeed. The bailout logic is a little more complicated than I'd like, because flock() happens after opening, whereas UUCP happens before.

Amazing!

> AcquireUUCPLockAndOpen() is still available, although no-one should be using it. It does this:

Cool.

> I've split off `acquire UUCP lock` and `apply TIOCEXCL` as separate routines.
> 
> Each step prints at least an OLA_INFO on success, so you can see which locks are used. If configured with `--enable-uucp-locking` and flock() is available, it will actually do both - there's no reason not to use flock() if the routine is available.

👍 

> The information about KarateLight suggests that flock() is _always_ available, which would lead me to suggest completely removing the UUCP locking. But perhaps there are systems still relying on it.

Yeah given some of the random architectures we've compiled on, they might just not have the KarateLight plugin included/compiled...

> The test program (SerialLockTester) tests that the flock() call succeeds, but not that it prevents a second lock. That would be a lot of trouble to test, because the second flock would have to be from a different process.

Ah yeah that would be quite a hassle! Although we could have a shell script wrapper which runs two copies if that's easy enough (like we do for some Python stuff), but as you say may not be worth the hassle.

> I consider it more a simple test that the routine really exists - it doesn't seem unreasonable to rely on it doing what the specification says it should! It now uses a PID-based filename, to prevent one failure from causing subsequent ones. UUCP locking and TIOCEXCL are not tested at all because of the difficulties noted above, but it's still overall an improvement on the previous situation.

Yeah sounds great!

taw10:
Ok, some updates in addition to your suggested changes (which were all applied):
- Made sure that all the `#endif`s have comments
- Fixed the comments about the behaviour of `AcquireLockAndOpenSerialPort` and `ReleaseSerialPortLock`
- Made sure that `RemoveUUCPLockFile()` happens _after_ `close()` (as it was before)
- Added missing `ReleaseUUCPLock()` calls on the error paths of `AcquireLockAndOpenSerialPort`

taw10:
The CI is failing, but the error doesn't make any sense to me (`ERROR: No matching distribution found for setuptools>=64`). Any suggestions?

DaAwesomeP:
@peternewman should this go into `0.10` or `master`?

PR #1773, opened 2022-04-13
ola: fix compilation with musl 1.2.3

neheb:
musl 1.2.3 defines NULL as nullptr. Cannot use reinterpret_cast with nullptr.

No idea why this is even here. NULL in C == 0.


peternewman (changes requested):
So this particular ifdef went in as part of https://github.com/OpenLightingProject/ola/pull/1460/files#diff-8876d81b49c35412d4a86a252895a940559f9806452ff00a0648559004e15fc6

And the original cast went in under #1226 .

I assume you're referring to https://git.musl-libc.org/cgit/musl/commit/?id=c8a9c22173f485c8c053709e1dfa0a617cb6be1a or similar?

I'm just a bit nervous this will break one of the other many and varied systems we compile on or trigger e.g. a warning in a version of clang.

However given musl refuses to define a __MUSL__ I'm not sure we've got much other option...

Please can we have a comment summarising this and why we're going against #1226 so we don't just put it back in future...

Is this the only fix required for musl? I'm wondering if we should add it to Travis too?

neheb:
Similar to musl I could match is against the cplusplus version. But still. NULL is 0. I don't understand the point of the ifdef.

neheb:
The musl commit that breaks ola compilation is http://git.musl-libc.org/cgit/musl/commit/?id=98e688a9da5e7b2925dda17a2d6820dddf1fb287

neheb:
Made the change.

The longterm solution is to migrate this project to C++11. Newer protobuf requires it.

neheb:
travis failures seem unrelated:

```
/usr/bin/ld: cannot find -lpthreads
collect2: error: ld returned 1 exit status
```

kripton:
@peternewman, could you give this another look, please? Compilation failure against musl would also be a problem for Gentoo Linux (where we do have ola packaged as well). Thanks!
I could investigate if I can add a github action that runs against musl ...

neheb:
@kripton absolutely. There’s a setup-alpine GitHub actions workflow than can be used.

btw, why does this project support less than C++11?

neheb:
Answered my own question:

```
C++ 11
------
As much as we'd like to start using C++11, Ubuntu 12.04 still comes with g++
4.6 which has partial support. Once 12.04 is end-of-lifed, we'll consider
using C++ 11 constructs.
```

12.04 is already EOL. Only 18.04 remains non EOL at LTS.

PR #1784, opened 2022-07-24
Switch node to the latest LTS version we can, switch to an older codespell so it works

peternewman:
And expect jshint to pass again once we've fixed it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -349,13 +349,6 @@ jobs:
            - lintian
            - moreutils
   allow_failures:
-    - os: linux
-      dist: bionic
-      arch: arm64
-      env: TASK='jshint'
-      addons:
-        apt:
-          packages:
 #    - os: osx
 #      osx_image: xcode9.4
 #      compiler: clang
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -472,6 +465,8 @@ before_install:
  - if [ "$PYTHON" == "python3" -a "$TRAVIS_CPU_ARCH" != "arm64" ]; then pyenv global 3.8.1; fi
 #On arm64 (at least) pip is now pip3, so symlink it to pip2 unless we're using python3.
  - if [ "$PYTHON" != "python3" ]; then sudo rm /usr/local/bin/pip && sudo ln -s /usr/local/bin/pip2 /usr/local/bin/pip; fi
+#Switch node to version 17, as newer ones don't work with Ubuntu 18.04 and some older ones are too old
+ - if [ "$TASK" = "jshint" ]; then nvm install 17 && nvm use 17; fi
```
FloEdelmann, line 469: You should install either v16 or v18, as those are LTS releases. See https://github.com/nodejs/release#release-schedule:

> Odd-numbered release lines are not promoted to LTS - they will not go through the 'Active LTS' or 'Maintenance' phases.

peternewman: Thanks @FloEdelmann . I didn't realise they'd still be doing updates for it. We can't use 18 unfortunately as it doesn't work on 18.04.

PR #1800, opened 2022-12-02
UART plugin recovers granularity if timing improves - closes #1798

### LINKED ISSUE #1798: UART plugin does not send correct DMX data when granularity check happens during high load
Hi,

I run OLA on a RPI 4 within a docker container and use UART to send out DMX data. Most of the time it works just fine, however, from time to time the DMX protocol appears to break completely. After some testing and investigating, I found that the granularity is set to BAD when the DMX protocol appears to break. Apparently this can happen quite easily, when the RPI has a lot of other stuff to do during the granularity check. My guess would be, that the ola server doesn't get the CPU in time. The code can be found in `plugins/uartdmx/UartDmxThread.cpp`.
As a result, the DMX break signal is skipped which causes the DMX recipient to ignore the DMX data following since it cannot interpret it correctly. The only way I've found to resolve this problem is to restart the entire OLA server.
  
Recreating this issue should be fairly easy, just force all your CPUs to 100% / stress test your hardware and restart your OLA server with the UART DMX plugin enabled. Maybe you'll have to restart the server a few times until it happens since this issue is not deterministic. It occurred with the versions 0.10.3 and 0.10.8, I haven't tested the others, but I don't expect different results.
  
I think it would be cleaner to fix this issue by fixing the granularity check, or by changing the impact the granularity has on the data sent. However, I'm not sure how to fix this since I have no experience in coding so close to the hardware. But for the time being, could you perhaps provide a config parameter that allows to force the granularity to GOOD?

I know that this is basically undermining it's entire purpose, but I also know that the hardware can handle the timings and it isn't an option for me to basically depend on luck that this plugin is sending out DMX data correctly.
  
Thanks for your help!

markus983:
Hi,
with these changes the UART plugin can recover the granularity if timing improves, similar to how the FTDI plugin solves this issue.

### DIFF
--- a/plugins/uartdmx/UartDmxThread.cpp
+++ b/plugins/uartdmx/UartDmxThread.cpp
@@ -72,10 +72,13 @@ bool UartDmxThread::WriteDMX(const DmxBuffer &buffer) {
  * The method called by the thread
  */
 void *UartDmxThread::Run() {
-  TimeStamp ts1, ts2;
+  TimeStamp ts1, ts2, ts3;
   Clock clock;
   CheckTimeGranularity();
   DmxBuffer buffer;
+
+  int frameTime = static_cast<int>(floor(
+    (m_malft / static_cast<double>(1000)) + static_cast<double>(0.5)));
 
   // Setup the widget
   if (!m_widget->IsOpen())
@@ -92,6 +95,8 @@ void *UartDmxThread::Run() {
       ola::thread::MutexLocker locker(&m_buffer_mutex);
       buffer.Set(m_buffer);
     }
+
+    clock.CurrentMonotonicTime(&ts1);
 
     if (!m_widget->SetBreak(true))
       goto framesleep;
@@ -110,7 +115,31 @@ void *UartDmxThread::Run() {
 
   framesleep:
     // Sleep for the remainder of the DMX frame time
-    usleep(m_malft);
+    clock.CurrentMonotonicTime(&ts2);
+    TimeInterval elapsed = ts2 - ts1;
+
+    if (m_granularity == GOOD) {
+      while (elapsed.InMilliSeconds() < frameTime) {
+        usleep(1000);
+        clock.CurrentMonotonicTime(&ts2);
+        elapsed = ts2 - ts1;
+      }
+    } else {
+      // See if we can drop out of bad mode.
+      usleep(1000);
+      clock.CurrentMonotonicTime(&ts3);
+      TimeInterval interval = ts3 - ts2;
+      if (interval.InMilliSeconds() <= BAD_GRANULARITY_LIMIT) {
+        m_granularity = GOOD;
+        OLA_INFO << "Switching from BAD to GOOD granularity for UART thread";
+      }
+      
+      elapsed = ts3 - ts1;
+      while (elapsed.InMilliSeconds() < frameTime) {
+        clock.CurrentMonotonicTime(&ts2);
+        elapsed = ts2 - ts1;
+      }
+    }
   }
   return NULL;
 }
@@ -122,17 +151,13 @@ void *UartDmxThread::Run() {
 void UartDmxThread::CheckTimeGranularity() {
   TimeStamp ts1, ts2;
   Clock clock;
-  /** If sleeping for 1ms takes longer than this, don't trust
-   * usleep for this session
-   */
-  const int threshold = 3;
 
   clock.CurrentMonotonicTime(&ts1);
   usleep(1000);
   clock.CurrentMonotonicTime(&ts2);
 
   TimeInterval interval = ts2 - ts1;
-  m_granularity = interval.InMilliSeconds() > threshold ? BAD : GOOD;
+  m_granularity = interval.InMilliSeconds() > BAD_GRANULARITY_LIMIT ? BAD : GOOD;
   OLA_INFO << "Granularity for UART thread is "
            << (m_granularity == GOOD ? "GOOD" : "BAD");
 }
--- a/plugins/uartdmx/UartDmxThread.h
+++ b/plugins/uartdmx/UartDmxThread.h
@@ -53,6 +53,7 @@ class UartDmxThread : public ola::thread::Thread {
   void CheckTimeGranularity();
 
   static const uint32_t DMX_MAB = 16;
+  static const uint32_t BAD_GRANULARITY_LIMIT = 3;
 
   DISALLOW_COPY_AND_ASSIGN(UartDmxThread);
 };


plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -122,17 +151,13 @@ void *UartDmxThread::Run() {
...
-   */
-  const int threshold = 3;
 
   clock.CurrentMonotonicTime(&ts1);
   usleep(1000);
   clock.CurrentMonotonicTime(&ts2);
 
   TimeInterval interval = ts2 - ts1;
-  m_granularity = interval.InMilliSeconds() > threshold ? BAD : GOOD;
+  m_granularity = interval.InMilliSeconds() > BAD_GRANULARITY_LIMIT ? BAD : GOOD;
```
peternewman, line 160: ./plugins/uartdmx/UartDmxThread.cpp:160:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
```suggestion
  m_granularity = (interval.InMilliSeconds() > BAD_GRANULARITY_LIMIT) ?
      BAD : GOOD;
```

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -110,7 +115,31 @@ void *UartDmxThread::Run() {
...
+      // See if we can drop out of bad mode.
+      usleep(1000);
+      clock.CurrentMonotonicTime(&ts3);
+      TimeInterval interval = ts3 - ts2;
+      if (interval.InMilliSeconds() <= BAD_GRANULARITY_LIMIT) {
+        m_granularity = GOOD;
+        OLA_INFO << "Switching from BAD to GOOD granularity for UART thread";
+      }
+      
+      elapsed = ts3 - ts1;
```
peternewman, line 137: /plugins/uartdmx/UartDmxThread.cpp:136:  Line ends in whitespace.  Consider deleting these extra spaces. [whitespace/end_of_line] [4]
```suggestion
      }

      elapsed = ts3 - ts1;
```

peternewman (changes requested):
There are a few minor lint issues I've highlighted and suggested fixes from by copying from the existing FTDI plugin, please could you fix those.

Could you also target the 0.10 branch please and we can get this into our upcoming 0.10.9 release which will go into Debian etc. You'll probably need to cherry-pick or just manually re-apply the changes to a new branch unfortunately.

markus983:
I took care of the lint issues and the PR is now targeting the branch 0.10

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -122,17 +151,14 @@ void *UartDmxThread::Run() {
...
-  const int threshold = 3;
 
   clock.CurrentMonotonicTime(&ts1);
   usleep(1000);
   clock.CurrentMonotonicTime(&ts2);
 
   TimeInterval interval = ts2 - ts1;
-  m_granularity = interval.InMilliSeconds() > threshold ? BAD : GOOD;
+  m_granularity = (interval.InMilliSeconds() > BAD_GRANULARITY_LIMIT) ?
+	  BAD : GOOD;
```
peternewman, line 161: /plugins/uartdmx/UartDmxThread.cpp:161:  Tab found; better to use spaces  [whitespace/tab] [1]
```suggestion
      BAD : GOOD;
```

peternewman (changes requested):
I'm not sure if this was my error or yours, but there's one outstanding lint issue remaining if you wouldn't mind.

Have you re-tested on the 0.10 branch?

markus983:
> I'm not sure if this was my error or yours, but there's one outstanding lint issue remaining if you wouldn't mind.

It should be fixed now.

> Have you re-tested on the 0.10 branch?

Yes I did some testing on the 0.10 branch, and with these fixes, OLA behaved as expected.

plugins/uartdmx/UartDmxThread.cpp
peternewman: Done.

peternewman:
> > I'm not sure if this was my error or yours, but there's one outstanding lint issue remaining if you wouldn't mind.
> 
> It should be fixed now.

Great thanks. For future reference you can also merge them directly from the web UI:

> > Have you re-tested on the 0.10 branch?
> 
> Yes I did some testing on the 0.10 branch, and with these fixes, OLA behaved as expected.

That's brilliant news!

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -110,7 +115,31 @@ void *UartDmxThread::Run() {
 
   framesleep:
     // Sleep for the remainder of the DMX frame time
-    usleep(m_malft);
+    clock.CurrentMonotonicTime(&ts2);
+    TimeInterval elapsed = ts2 - ts1;
+
+    if (m_granularity == GOOD) {
+      while (elapsed.InMilliSeconds() < frameTime) {
```
peternewman, line 122: As above this value, or the calculation of this value, needs changing...

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -110,7 +115,31 @@ void *UartDmxThread::Run() {
...
+      usleep(1000);
+      clock.CurrentMonotonicTime(&ts3);
+      TimeInterval interval = ts3 - ts2;
+      if (interval.InMilliSeconds() <= BAD_GRANULARITY_LIMIT) {
+        m_granularity = GOOD;
+        OLA_INFO << "Switching from BAD to GOOD granularity for UART thread";
+      }
+
+      elapsed = ts3 - ts1;
+      while (elapsed.InMilliSeconds() < frameTime) {
```
peternewman, line 138: And again.

peternewman (approved):
Thanks @markus983 . Unfortunately I've just spotted an issue with how this calculation is mapped across to this code from the FTDI plugin...

You probably also want to add yourself to Authors for your contribution:
https://github.com/OpenLightingProject/ola/blob/0.10/AUTHORS

rewolff:
Just to be sure: The "write" returns immediately. So a working solution should probably get a timestamp at the moment the BREAK condition is released (A scheduling delay during the break then won't affect us), Then issue the write and calculate how long that should take, add in the mark-after-last-frame and wait until that moment before starting the next cycle. 

When I was looking at the code, the person writing the code assumed that write would return when all bytes had been written. My suggestion above does not depend on whether the write returns immediately or not.

markus983:
@peternewman Thanks for the info, I fixed the calculation of the frame time and it should honor, the boundaries now. I tested it as well, and the UART plugin still behaves as expected. Personally I don't like goto statements so I just split the code into more methods, but it still behaves the same (and it helped me with debugging).

@rewolff Thanks for your input, I incorporated this in my newest changes.

rewolff:
Glad to have been of help. I haven't been "intimate" with the code in a while, so good to hear that my "research" from a while back wasn't wasted and helped the actual code. :-)

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -72,11 +74,21 @@ bool UartDmxThread::WriteDMX(const DmxBuffer &buffer) {
...
+  // the +1 at the slots_per_universe is for the dmx start code
+  // the basic_frame_time is in microseconds
+  int basic_frame_time = DMX_MAB + (4 * 11 * (DMX_UNIVERSE_SIZE + 1)) + m_malft;
+
+  // the maximum time between breaks should be 1 second, and the minimum time 1204
+  // the capping of max and min values is done to honor the standard
+  int basic_frame_time_capped = std::max(std::min(basic_frame_time, 1000000), 1204);
+
+  // converts the frame time into milliseconds for later use
+  m_frameTime = static_cast<int>(floor(basic_frame_time_capped / static_cast<double>(1000)));
```
peternewman, line 90: I'd like to get rid of these magic numbers please and turn them into constants in the core of OLA as if we e.g. rewrote the FTDI plugin to match, it wants to use the same numbers, and things like the sniffer might want these values, indeed it already has their ranges:
https://github.com/OpenLightingProject/ola/blob/master/tools/logic/DMXSignalProcessor.cpp

There's some possible inspiration around naming here in another of our bits of software:
https://github.com/OpenLightingProject/ja-rule/blob/master/firmware/src/transceiver_timing.h

We've also got some generic clock related values too:
https://github.com/OpenLightingProject/ola/blob/master/include/ola/Clock.h

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -93,46 +105,94 @@ void *UartDmxThread::Run() {
...
+  }
+  return NULL;
+}
 
-    if (m_granularity == GOOD)
-      usleep(m_breakt);
+/**
+ * Write the DMX data to the actual UART interface.
+ */
+void UartDmxThread::writeDmxData(const DmxBuffer &buffer) {
```
peternewman, line 116: We capitalise the first character too:
```suggestion
void UartDmxThread::WriteDmxData(const DmxBuffer &buffer) {
```

Given we've got the interface method WriteDMX perhaps we can come up with something less ambiguous here, WriteDMXToUART or something? The other equivalent classes don't seem to split it off so no real inspiration.

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -93,46 +105,94 @@ void *UartDmxThread::Run() {
...
+  TimeStamp ts2, ts3;
+  Clock clock;
+
+  // Sleep for the remainder of the DMX frame time
+  clock.CurrentMonotonicTime(&ts2);
+  TimeInterval elapsed = ts2 - ts1;
+
+  if (m_granularity == GOOD) {
+    while (elapsed.InMilliSeconds() < m_frameTime) {
+      usleep(1000);
```
peternewman, line 160: Similar comment, and I suspect this is previous code, but is it a magic value or just a random second, I think it's the max break to break to ensure we start afresh, but...

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -93,46 +105,94 @@ void *UartDmxThread::Run() {
...
+      clock.CurrentMonotonicTime(&ts2);
+      elapsed = ts2 - ts1;
+    }
+  } else {
+    // See if we can drop out of bad mode.
+    usleep(1000);
```
peternewman, line 166: Again.

plugins/uartdmx/UartDmxThread.cpp
```diff
@@ -93,46 +105,94 @@ void *UartDmxThread::Run() {
...
 
 /**
  * Check the granularity of usleep.
  */
 void UartDmxThread::CheckTimeGranularity() {
   TimeStamp ts1, ts2;
   Clock clock;
-  /** If sleeping for 1ms takes longer than this, don't trust
-   * usleep for this session
-   */
```
peternewman, line 127: This comment is really useful to know that 3 isn't the number of times before its bad or something, could you re-attach it to BAD_GRANULARITY_LIMIT please.

plugins/uartdmx/UartDmxThread.h
```diff
@@ -49,10 +49,15 @@ class UartDmxThread : public ola::thread::Thread {
   DmxBuffer m_buffer;
   ola::thread::Mutex m_term_mutex;
   ola::thread::Mutex m_buffer_mutex;
+  unsigned int m_frameTime;
```
peternewman, line 52: Could we stick to snake_case here to match the other variables:
```suggestion
  unsigned int m_frame_time;
```

peternewman (changes requested):
Sorry for ignoring this for a bit, all sorts of stuff has been going on and I've been fighting the good fight with Python.

I'm hoping we can still sneak this into 0.10.9 if you've got a bit of time to just tweak it a bit?

AUTHORS
```diff
@@ -29,3 +29,4 @@ Contributors:
   Stefan Krüger, added APA102 support to the SPI Plugin
   Tobi Schäfer, for the MacPort files
   Stefan S, improved timing with monotonic clock
+  Markus, improved uartdmx plugin
```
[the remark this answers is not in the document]
markus983: Yeah, that's more accurate

plugins/uartdmx/UartDmxThread.cpp
markus983: I decided to stick with WriteDMXToUART

plugins/uartdmx/UartDmxThread.h
markus983: Done

plugins/uartdmx/UartDmxThread.cpp
markus983: I replaced that with a constant and added a comment for better understanding

plugins/uartdmx/UartDmxThread.cpp
markus983: That's a pretty good idea. I've replaced these magic numbers with constants.

peternewman:
> I tweaked the code as you've suggested - thanks for the feedback! Hopefully, this PR can be included in 0.10.9.

🤞 

> Unfortunately, it looks like I've somehow messed up the git history of my forked repo. I was able to push my code to the correct branch in the end, I just hope that this didn't mess up anything else.

Yeah that's about what normally happens with me too!

Do you want to open a PR against this which I think is what you actually intend to merge now?:
https://github.com/OpenLightingProject/ola/compare/0.10...markus983:ola:0.10

plugins/uartdmx/UartDmxThread.h
```diff
@@ -49,10 +50,20 @@ class UartDmxThread : public ola::thread::Thread {
...
+  void frameSleep(const TimeStamp &ts1);
+  void WriteDMXToUART(const DmxBuffer &buffer);
+
   static const uint32_t DMX_MAB = 16;
+  static const uint32_t BAD_GRANULARITY_LIMIT = 3;
+  static const int DMX_BREAK_TIME_MIN = 1204; // in microseconds
+  static const int DMX_BREAK_TIME_MAX = USEC_IN_SECONDS; 
+  static const uint32_t DMX_TIME_PER_BIT = 4; // in microseconds
+  static const uint32_t DMX_BITS_PER_SLOT = 11;
+  static const uint32_t DMX_SLOT_START_CODE = 1;
```
peternewman, line 66: This can move too I think, I'd be tempted to effectively make it 513 (set as `DMX_UNIVERSE_SIZE + 1`, or maybe both the number of slots that the start code takes (1) and the summed total size of a frame plus start code (513).

peternewman (changes requested):
As mentioned can we have the PR from your 0.10 branch to our 0.10 one and some other fixes to add to it too.

peternewman:
FWIW, although we've just cut 0.10.9, I'm planning to roll out 0.10.10 within the next few days ideally, or few weeks if a few key outstanding PRs like this one get resolved and merged.

markus983:
> Do you want to open a PR against this which I think is what you actually intend to merge now?:
> https://github.com/OpenLightingProject/ola/compare/0.10...markus983:ola:0.10
Yeah, that would make it easier. The "new" PR is now here: https://github.com/OpenLightingProject/ola/pull/1825

PR #1807, opened 2022-12-06
Ensure the dummy data we provide is binary to be compatible with Python 3

peternewman:


kripton (approved):
Looks good to me, thanks!

PR #1812, opened 2023-01-29
GitHub Actions Debian Builds

DaAwesomeP:
This pull adds preliminary automatic builds for Debian on GitHub Actions. This is most useful as a way of producing easily installed artifacts for testing new features/branches without needing every person wishing to test to build it themselves.

The build workflow was modeled off of the [Debian GitLab CI workflow](https://salsa.debian.org/wouter/ola/-/blob/a38a396f6994b2b1af8efec9e208aee4e67f77aa/.gitlab-ci.yml). I have not yet added the `autopkgtest` workflow. Building for Debian in this repo may alleviate the need for as many specific patches in that repository to fix Debian build issues (should that be desired).

I had some difficulties getting it to build surrounding outdated/missing Python 2 packages no longer available in different Debian releases. I simply substituted the Python 3 equivalents, but I did so without checking for proper compatibility. Please let me know if the code base is actually running on Python 2 or if these changes are correct.


DaAwesomeP:
It seems that `autopkgtest` fails, but I don't think that is a result of this pull. The [most recent pipeline on the Debian CI](https://salsa.debian.org/wouter/ola/-/pipelines/447952) also successfully builds but fails to test.

DaAwesomeP:
Requesting review from @peternewman

DaAwesomeP:
Possibly related (Python 3) to #1506 and #1761.

.github/workflows/debian.yml
```diff
@@ -0,0 +1,72 @@
+# This action originally modeled off of the Debian build action:
+# https://salsa.debian.org/wouter/ola/-/blob/a38a396f6994b2b1af8efec9e208aee4e67f77aa/.gitlab-ci.yml
+
+name: debian
+on: [push, pull_request]
+jobs:
+  debian-build:
+    name: 'Debian Build ${{ matrix.image_tag }} amd64'
```
kripton, line 8: Interesting that one can use the `matrix` variable before having defined the actual matrix. Nice!

.github/workflows/debian.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
```
kripton, line 9: Any reason not to use a Debian based container/runner?

kripton (commented):
Looks good to me, thanks!

.github/workflows/debian.yml
DaAwesomeP: This is a GH Actions syntax weirdness. GitHub only provides runners natively running Ubuntu, but then see [line 14](https://github.com/OpenLightingProject/ola/pull/1812/files#diff-a57b041fa9cb45a2f493ff99a1b48b1dbd581f0f90dbc905cebb941a314fcf06R14). It runs a Docker container with the official Debian image on top of the Ubuntu runner.

kripton: Understood, thanks for the explanation!

DaAwesomeP:
Possibly @yoe or @peternewman may have some ideas about the failed test phases? I imagine these issue would possibly affect the official Debian builds as well.

peternewman:
> It seems that `autopkgtest` fails, but I don't think that is a result of this pull.

Agreed, I can't see any changes you've made that would cause this.

> The [most recent pipeline on the Debian CI](https://salsa.debian.org/wouter/ola/-/pipelines/447952) also successfully builds but fails to test.

Note that the Debian CI fails to find some packages so doesn't test:
https://salsa.debian.org/wouter/ola/-/jobs/3450438

Whereas we try and test and get some symbol type errors.

.github/workflows/debian.yml
peternewman: It's containers all the way down!

peternewman:
> This pull adds preliminary automatic builds for Debian on GitHub Actions. This is most useful as a way of producing easily installed artifacts for testing new features/branches without needing every person wishing to test to build it themselves.

If we're going down that route, we should do this eventually too:
https://github.com/mjemmeson/Date-Holidays-GB/pull/14

> I had some difficulties getting it to build surrounding outdated/missing Python 2 packages no longer available in different Debian releases. I simply substituted the Python 3 equivalents, but I did so without checking for proper compatibility.

That looks about right thanks.

> Please let me know if the code base is actually running on Python 2 or if these changes are correct.

I've been testing the Python 3 changes with Python 3 libraries so I think they'll be about right.

DaAwesomeP:
> Agreed, I can't see any changes you've made that would cause this.

I think there's an "allow failure" flag that I can add to that job for now so that this can be merged without marking everyone's pulls as broken.

> If we're going down that route, we should do this eventually too:
> [mjemmeson/Date-Holidays-GB#14](https://github.com/mjemmeson/Date-Holidays-GB/pull/14)

Yes! Nightly builds in an automatic repo would be fantastic for continually testing new features/stability! I've tried to do this [manually once before](https://github.com/DaAwesomeP/debian-nginx-module-shibboleth/blob/de9fa388ff8374c21ada72185997d96e496a67b7/.github/workflows/build.yaml#L127-L186) and failed utterly, but that GitHub Action looks like a one-and-done solution!

DaAwesomeP:
OK, wow a proper allow failure feature that allows a job to fail, mark it as failed, but not prevent merge is contentious: https://github.com/actions/runner/issues/2347

For now it will soft-fail but won't raise any alarms. Once the packaging issue is fixed it can be re-enabled.

Should this target 0.10?

DaAwesomeP:
> More generally this could target 0.10 too, like the other ones.

Will do!

> If you're feeling really keen, fancy trying a PR against 0.10 which pulls in most of @yoe 's stuff from https://salsa.debian.org/wouter/ola/-/tree/master (as he's not keeping his GitHub mirror up to date anymore sadly. I think we want to keep our postinst stuff in Debian, and the local JS libs, but most of the rest of it could be ported back from downstream which would avoid gotchas like this compilation failure!

I'll take a stab at it!

peternewman:
> I'll take a stab at it!

Amazing, as mentioned, I think cherry-picking that one commit into this branch should be enough to fix the autopkgtest stuff which is hopefully enough to get this into the stage where it can be merged in too!

DaAwesomeP:
@peternewman See this build: https://github.com/OpenLightingProject/ola/actions/runs/4235745316/jobs/7359746616

Is there a flag to configure to not require cpplint and flake8?

DaAwesomeP:
OK, that patch got us further for Debian stable (bullseye):
https://github.com/DaAwesomeP/ola/actions/runs/4235912149/jobs/7360306928#step:8:1225
```
autopkgtest [19:06:16]: test command6: set -e; g++ -o linktest debian/tests/hw.cc $(pkg-config --cflags --libs libola); ./linktest
autopkgtest [19:06:16]: test command6: [-----------------------
Hello World!
success: t
autopkgtest [19:06:17]: test command6: -----------------------]
autopkgtest [19:06:17]: test command6:  - - - - - - - - - - results - - - - - - - - - -
command6             PASS
autopkgtest [19:06:17]: @@@@@@@@@@@@@@@@@@@@ summary
command1             PASS
command2             PASS
command3             FAIL non-zero exit status 1
command4             FAIL non-zero exit status 1
command5             PASS
command6             PASS
```

And Debian testing (bookworm):
https://github.com/DaAwesomeP/ola/actions/runs/4235912149/jobs/7360307162#step:8:1132
```
autopkgtest [19:05:47]: test command6: set -e; g++ -o linktest debian/tests/hw.cc $(pkg-config --cflags --libs libola); ./linktest
autopkgtest [19:05:47]: test command6: [-----------------------
Hello World!
success: t
autopkgtest [19:05:48]: test command6: -----------------------]
autopkgtest [19:05:49]: test command6:  - - - - - - - - - - results - - - - - - - - - -
command6             PASS
autopkgtest [19:05:49]: @@@@@@@@@@@@@@@@@@@@ summary
command1             PASS
command2             PASS
command3             FAIL badpkg
blame: arg:built/libola-dev_0.10.9-1_amd64.deb deb:libola-dev arg:built/libola1-dbgsym_0.10.9-1_amd64.deb deb:libola1-dbgsym arg:built/libola1_0.10.9-1_amd64.deb deb:libola1 arg:built/ola-dbgsym_0.10.9-1_amd64.deb deb:ola-dbgsym arg:built/ola-python_0.10.9-1_all.deb deb:ola-python arg:built/ola-rdm-tests_0.10.9-1_all.deb deb:ola-rdm-tests arg:built/ola_0.10.9-1_amd64.deb deb:ola .
badpkg: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
command4             FAIL badpkg
blame: arg:built/libola-dev_0.10.9-1_amd64.deb deb:libola-dev arg:built/libola1-dbgsym_0.10.9-1_amd64.deb deb:libola1-dbgsym arg:built/libola1_0.10.9-1_amd64.deb deb:libola1 arg:built/ola-dbgsym_0.10.9-1_amd64.deb deb:ola-dbgsym arg:built/ola-python_0.10.9-1_all.deb deb:ola-python arg:built/ola-rdm-tests_0.10.9-1_all.deb deb:ola-rdm-tests arg:built/ola_0.10.9-1_amd64.deb deb:ola .
badpkg: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
command5             PASS
command6             PASS
```

DaAwesomeP:
Hmmm....I think Python 2 may strike again.

See [here](https://github.com/DaAwesomeP/ola/actions/runs/4235912149/jobs/7360306928#step:8:689) and [here](https://github.com/DaAwesomeP/ola/actions/runs/4235912149/jobs/7360306928#step:8:814). It's trying to run `rdm_responder_test.py`, first with an automatically chosen interpreter (that file has `#!/usr/bin/env python`) and then explicitly with Python 2.7 and can't find OLA.

I wonder if Python 2 doesn't see package in `site-packages`. AFAIK, this is where they should live though.

DaAwesomeP:
On bookworm there appears to be some dependency issues: https://github.com/DaAwesomeP/ola/actions/runs/4235933356/jobs/7360413159#step:8:556

DaAwesomeP:
I think possible `/usr/bin/install` during build is (correctly) placing them in `site-packages`, as simply running the `make install` on the build is not a distribution package. However, they should be installed to `dist-package` as a part of installing the `*.deb` package, not `site-packages`.

peternewman:
> @peternewman See this build: https://github.com/OpenLightingProject/ola/actions/runs/4235745316/jobs/7359746616
> 
> Is there a flag to configure to not require cpplint and flake8?

I think this might be down to your changes in #1817 @DaAwesomeP , assuming it wasn't doing it before? I think the order of the if and the cpplint definition means it now evaluates and throws the error. I suspect it's some magic Makefile.am->Makefile thing.

DaAwesomeP:
> I think this might be down to your changes in https://github.com/OpenLightingProject/ola/pull/1817 @DaAwesomeP , assuming it wasn't doing it before? I think the order of the if and the cpplint definition means it now evaluates and throws the error. I suspect it's some magic Makefile.am->Makefile thing.

It always has bothered me for flake8, even before that pull. I think it may come down to that the else statement calls `$(error ...)`. Maybe there is an info or neutral logging equivalent?

DaAwesomeP:
> I think if you rework the cpplint one to be more like it was, and do the same to flake8, the issue will go away.

I think I'm misunderstanding what you mean. Cpplint used to avoid this error by downloading the missing files for you instead of erroring. Flake8 never did that, it would just error during configure as well.

> But it is an error, we need to flag to the user the command failed to complete if they try to run it.

Yes, but this error should not happen during configure, right? A user shouldn't be required to have all of the testing and linting tools to build and run?

Maybe I should rework both to do this instead of `$(error`:
```
echo "flake8 not found. Install flake8 and re-run configure."
exit 1
```
That way I think the error won't be executed during configure.

peternewman:
> > I think if you rework the cpplint one to be more like it was, and do the same to flake8, the issue will go away.
> 
> I think I'm misunderstanding what you mean. Cpplint used to avoid this error by downloading the missing files for you instead of erroring. Flake8 never did that, it would just error during configure as well.

It only downloaded them if you explicitly ran `make cpplint.py`.

As I mentioned before, my machine returns:
> configure:27383: checking for flake8
> configure:27411: result: yes
> configure:27429: checking for cpplint
> configure:27457: result: no

It also says:

> Makefile.am:246: warning: error flake8 not found. Install flake8 and re-run configure.: non-POSIX variable name
> Makefile.am:246: (probably a GNU make extension)
> Makefile.am:265: warning: error cpplint not found. Install forked cpplint (e.g. via pip for the latest version: non-POSIX variable name
> Makefile.am:265: (probably a GNU make extension)

It now fails with:

> Makefile:18391: *** cpplint not found. Install forked cpplint (e.g. via pip for the latest version) and re-run configure.. Stop.

I've only got cpplint.py in some of my working directories, I've previously been able to build and now can't, so that PR has definitely upset it for machines without cpplint on. Maybe I've always had flake8 as I probably got it flake8 clean before adding it to the Makefile and not noticed that issue, but the PR has definitely introduced an issue.

> Yes, but this error should not happen during configure, right? A user shouldn't be required to have all of the testing and linting tools to build and run?

Agreed entirely, it didn't, and doesn't block my configure, only when I subsequently run make.

> Maybe I should rework both to do this instead of `$(error`:
> 
> ```
> echo "flake8 not found. Install flake8 and re-run configure."
> exit 1
> ```

Given it's worked fine on my machine for years, I think we should still be able to fix it somehow.

> That way I think the error won't be executed during configure.

As mentioned, for me it happens during make not configure.

I'll try and have a look in the next few days unless you beat me to it.

DaAwesomeP:
There may be more commits we want to pull in. I'm going to try this one:

https://salsa.debian.org/wouter/ola/-/commit/944a0b2ae990aac4a8edd328a7fd8c5694a226ac

DaAwesomeP:
> I'll try and have a look in the next few days unless you beat me to it.

I'm really not experienced with autotools (I'm more familiar to CMake and vanilla Makefile), so I may leave that to you.

DaAwesomeP:
Hmm, for some reason `ola-rdm-tests` is requiring python2, which is causing it to fail to in install at all on Bookworm and Sid.

DaAwesomeP:
Huzzah! Thanks to #1835 the test part is working on bullseye! Now I just need to figure out why it insists on installing Python 2 for the RDM tests!

DaAwesomeP:
@peternewman OK, I figured out the issue. Debian is inferring that it needs Python 2 by looking at the shebangs in files that it installs. You can see the commit history; I tried everything I could think of to force it to use Python 3. Python 2 works on bullseye because python2 is still available, but it fails on anything later than that.

The only workaround I can think of besides changing the shebang is to somehow patch these files during the build process to use a Python 3 shebang.

@yoe simply disabled the RDM tests in Salsa repo, but I'm not sure that we want to do that here: https://salsa.debian.org/wouter/ola/-/blob/175a46c5e0f8bdadc26a35597711159ebbfeff6d/debian/control#L42-62

Thoughts?

DaAwesomeP:
Also a note that `python-is-python3` as a dependency is not a workaround for packaging, and our CI that currently makes use of it in Bullseye will break when Bookworm becomes Debian stable:
> For the Debian 11 release (bullseye), the `/usr/bin/python` command is provided in the `python-is-python2` package (pointing to `/usr/bin/python2`)...These packages will not be part of the Debian 12 release (bookworm).

> The packages `python-is-python3`, `python-dev-is-python3`, `python-is-python2` and `python-dev-is-python2` must not be used as build dependencies, dependencies, recommendations or suggestions.

https://www.debian.org/doc/packaging-manuals/python-policy/

A messy workaround would maybe be to provide our own environment `olapython` that shims the intended Python executable, but this make installing or running outside of the package a mess.

I'm wondering if there is a way we can patch all instances of `python` with `python3` specifically during the Debian build process without changing it in the source tree.

DaAwesomeP:
Hey, I solved it! Turns out you can tell `dh_python3` to replace all Python shebangs during build.

This does mean that we need to be sure that every single file with a `/usr/bin/env python` shebang is actually cross-compatible with Python 3, but it seems like we are already there!

This also means that we're on-track to once again offer the `ola-rdm-tests` package! At least it will be available in the form of build artifacts in this repo for now (until Debian pulls to their downstream).

@peternewman This is ready for review!

DaAwesomeP:
@peternewman polite ping!

debian/changelog
```diff
@@ -1,3 +1,13 @@
+ola (0.10.10-1) UNRELEASED; urgency=medium
```
peternewman, line 1: This should be something like this, as it's not a new OLA release, just a new Debian revision:
```suggestion
ola (0.10.9-2) UNRELEASED; urgency=medium
```

debian/rules
```diff
@@ -11,14 +11,21 @@
...
 
 override_dh_installinit:
 	dh_installinit -p ola --name=olad
 	dh_installinit -p ola-rdm-tests --name=rdm_test_server
 
 override_dh_makeshlibs:
 	dh_makeshlibs -V
+
+override_dh_clean:
+	find . -type d -name '__pycache__' -print0 | xargs -0 rm -rf
```
peternewman, line 27: I thought there might be a better clean tool and it seems it's this:
https://manpages.ubuntu.com/manpages/bionic/man1/py3clean.1.html

debian/control
```diff
@@ -42,7 +42,7 @@ Description: Open Lighting Architecture - Python Classes
 Package: ola-rdm-tests
 Section: libs
 Architecture: all
-Depends: ola (>= ${source:Version}), ola (<< ${source:Version}.~), ola-python (= ${binary:Version}), ${python:Depends}, ${misc:Depends}, python-numpy, libjs-jquery, libjs-jquery-ui, lsb-base
+Depends: ola (>= ${source:Version}), ola (<< ${source:Version}.~), ola-python (= ${source:Version}), ${python3:Depends}, ${misc:Depends}, python3-numpy, libjs-jquery, libjs-jquery-ui, lsb-base
```
peternewman, line 45: I'm no expert, but I'm surprised this would need changing from binary to source.

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        image_tag: [bullseye, bookworm, sid]
```
peternewman, line 13: I'm no Debian naming expert, I guess we want to use static names so people know what they want, although if we used dynamic ones we'd always be testing the bleeding edge in testing.

peternewman (changes requested):
Looks good, just some minor potential improvements

debian/rules
peternewman: Maybe also check what other packages use?

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
...
+        working-directory: built
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ola-built-debian-${{ matrix.image_tag }}-amd64
+          path: ./built
+  debian-test:
+    name: 'Debian Test ${{ matrix.image_tag }} amd64'
+    needs: debian-build
+    if: "always()" # Run if some builds fail but ensure they all complete first
```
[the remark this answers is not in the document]
DaAwesomeP: > We can't easily make them run as three two step actions, rather than three actions linked to another three actions can we?

No, the real limitation is that GitHub Actions has no way to properly define matrix dependencies, only whole-job dependencies. What we can do is probably a better check than `always()` so that is properly shows those matrix iterations as skipped instead of failed. I will look into it for a future PR.

debian/control
DaAwesomeP: I don't exactly remember why I did this so I can try without.

debian/rules
DaAwesomeP: I got this from downstream: https://salsa.debian.org/wouter/ola/-/blob/4a0ec01039b06f8ff2d350520a1d031b86e7920b/debian/rules#L27-29

Not many results (although this is not a definitive search) of using `py3clean` with an override: https://www.google.com/search?q=%22override_dh_clean%22+py3clean

It seems that `dh_python3` [may run `py3clean` at some point (`prerm-py3clean` step)](https://salsa.debian.org/python-team/tools/dh-python/-/blob/master/autoscripts/prerm-py3clean) but given that this was deemed necessary downstream I sort of trust that it is deleting something that isn't getting removed otherwise? I suppose this should be easy to test out.

debian/control
DaAwesomeP: I guess is it also strange that we require a source version of `ola` here?

DaAwesomeP: Pushed a change, if the test steps run fine then it works fine!

DaAwesomeP: Seems fine!

debian/rules
DaAwesomeP: OK, I checked all of the `*.deb` files in the bullseye build and there are not any `__pycache__` folders, `*.pyc` files or `*.pyo` files!

debian/control
peternewman: The source version for the other one went in here:
https://salsa.debian.org/wouter/ola/-/commit/dc0081bce482e87f1413ab0b2ad52876788a0360

I'd trust @yoe 's Debian Fu over my random guessing! But it feels like it could make sense. The ola package is a source package where as ola-python is binary only I believe...

DaAwesomeP: Yeah assuming the ola-python is binary then it makes sense for ola-rdm-tests to depend on ola-python binary. I think my reverted change makes sense.

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
+# This action originally modeled off of the Debian build action:
+# https://salsa.debian.org/wouter/ola/-/blob/a38a396f6994b2b1af8efec9e208aee4e67f77aa/.gitlab-ci.yml
+
+name: debian
+on: [push, pull_request]
+jobs:
+  debian-build:
+    name: 'Debian Build ${{ matrix.image_tag }} amd64'
```
peternewman, line 8: Given it seems to be trivially easy:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

Can we just hard-code a matrix of one (amd64) now, so that value can be used in the workflow name, image name etc.

```suggestion
    name: 'Debian Build ${{ matrix.image_tag }} ${{ matrix.architecture }}'
```

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
...
+        working-directory: built
+      - name: SHA256 built files
+        if: always()
+        shell: bash
+        run: find . -type f -exec sha256sum {} \;
+        working-directory: built
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ola-built-debian-${{ matrix.image_tag }}-amd64
```
peternewman, line 58:
```suggestion
          name: ola-built-debian-${{ matrix.image_tag }}-${{ matrix.architecture }}
```

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
...
+      fail-fast: false
+      matrix:
+        image_tag: [bullseye, bookworm, sid]
+    container: debian:${{ matrix.image_tag }}
+    steps:
+      - uses: actions/checkout@master
+      - name: Download build artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: ola-built-debian-${{ matrix.image_tag }}-amd64
```
peternewman, line 75:
```suggestion
          name: ola-built-debian-${{ matrix.image_tag }}-${{ matrix.architecture }}
```

peternewman (approved):
Just the one minor improvement before we merge I think.

Feel free to merge once the arch stuff is working.

Thanks again for all of this!

.github/workflows/debian.yml
DaAwesomeP: I'm going to push all of these as one commit instead of clicking accept on all of these individually.

DaAwesomeP: Done.

.github/workflows/debian.yml
DaAwesomeP: Done.

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
...
+        shell: bash
+        run: find . -type f -exec sha256sum {} \;
+        working-directory: built
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ola-built-debian-${{ matrix.image_tag }}-amd64
+          path: ./built
+  debian-test:
+    name: 'Debian Test ${{ matrix.image_tag }} amd64'
```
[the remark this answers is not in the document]
DaAwesomeP: Done.

.github/workflows/debian.yml
DaAwesomeP: Done.

.github/workflows/debian.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Update package database
+        run: apt-get update -y
+      - name: Install test tools
+        run: apt-get -y install autopkgtest
+      - name: Test
+        run: autopkgtest --output-dir=test-output built/*ges -- null
+      - uses: actions/upload-artifact@v3
+        if: always() # Always upload the test output, even on failed tests
+        with:
+          name: ola-test-output-debian-${{ matrix.image_tag }}-amd64
```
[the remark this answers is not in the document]
DaAwesomeP: Done.

.github/workflows/debian.yml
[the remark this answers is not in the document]
DaAwesomeP: Done.

.github/workflows/debian.yml
peternewman: > I'm going to push all of these as one commit instead of clicking accept on all of these individually.

You should also be able to queue them up and do them as one commit from GitHub's web interface if you want!

peternewman (approved):
Thanks for those final tweaks.

debian/rules
yoe: > @yoe is there some technical reason we're missing why this needs to be included? Maybe because we build and compile files for ola-rdm-tests, but you don't currently clean them correctly as you don't currently install/package ola-rdm-tests?

tbh, I'm not sure why I added that line and not the py3clean one. For sure the cache and pyc and pyo files are required to not be shipped but be generated on Debian systems, so the fact that we remove them before packaging is not something we can get rid of, but how we do it is very much open for discussion...

PR #1816, opened 2023-02-14
NEWS updates

peternewman:


kripton (approved):
Looks good to me

NEWS
```diff
@@ -1,22 +1,45 @@
-x/y/2022 ola-0.10.9
+x/y/2023 ola-0.10.9
  Features:
- *
+ * Python 3 compatibility across the board (including the RDM Responder Tests)!
```
kripton, line 3: It's not yet solved but since 0.10.9 won't be released before this is fixed, I assume we can already mention it in the NEWS

NEWS
```diff
@@ -1,22 +1,45 @@
...
+   #1506
+ * Support for the JMS USB2DMX PRO V2.1 device via the FTDI plugin #1728
 
  API:
  * Python: Add a fetch current DMX example.
 
  RDM Tests:
- * 
+ * Python 3 compatibility of the RDM Tests #1599
+ * Fix a longstanding bug in the GetMaxPacketSize RDM test around timeouts
```
kripton, line 12: any reference? PR# or commit?

NEWS
```diff
@@ -1,22 +1,45 @@
...
+ * Python 3 compatibility across the board (including the RDM Responder Tests)!
+   #1506
+ * Support for the JMS USB2DMX PRO V2.1 device via the FTDI plugin #1728
 
  API:
  * Python: Add a fetch current DMX example.
 
  RDM Tests:
- * 
+ * Python 3 compatibility of the RDM Tests #1599
```
kripton, line 11: Also not merged yet but same as above: will be merged before 0.10.9 is released

kripton (approved):
Re-review done :)

NEWS
peternewman: It was #1736 but looking at NEWS we've historically not bothered putting PR numbers in, just issue numbers when people reported it. I think I probably just caught this by inspection or when doing something else.

NEWS
peternewman: Yeah I must admit I didn't really think about this, I guess the responsible thing will be to take both of these lines out that you've flagged and put them in #1761 in case we did end up with a more ad-hoc Debian release...

NEWS
peternewman: Again.

NEWS
peternewman: I've taken them out for now and I'll put them in the other PR so they just appear at the correct time...

peternewman:
Merging as I've addressed the majority of @kripton 's comments.

NEWS
peternewman: Reverted in https://github.com/OpenLightingProject/ola/pull/1761/commits/c41589de0b67015f5d94eafb061eba511801ef69

NEWS
peternewman: Again reverted in https://github.com/OpenLightingProject/ola/pull/1761/commits/c41589de0b67015f5d94eafb061eba511801ef69

Issue #1815, opened 2023-02-09
Replace Travis CI with GitHub Actions


DaAwesomeP:
This is a progress-tracking issue for replacing Travis CI with Github Actions as mentioned in https://github.com/OpenLightingProject/ola/pull/1813#issuecomment-1423351075. I will begin working on one or more pulls shortly! Please assign me to this issue.

Tasks:
- [ ] Lint
- [ ] Check licenses
- [ ] Spellintian
- [ ] Spellintian Duplicates
- [ ] Codespell (Python 3)
- [ ] Doxygen
- [ ] Coverage
- [ ] Coverity
- [ ] Weblint
- [ ] Flake8 (Python 3)
- [ ] Flake8 (Python 2.7)
- [ ] Pychecker (Python 2.7)
- [ ] Pychecker WIP (Python 2.7)
- [ ] Compile Linux, matrix compilers: gcc, clang
- [ ] Compile Linux, matrix arch: arm64, amd64, s390x (may require slow QEMU emulation)
- [ ] Compile Linux, matrix Python: 3, 2.7
- [ ] Compile Mac, matrix compilers: gcc, clang

Current pulls:

- [ ] #1809
- [ ] #1812 

Related issues/pulls:
- #1813

A few questions:

1. At this point can we abandon the Python 2.7 lint tasks and only move forward with the Python 3 tasks?
2. Most of the Travis tasks appear to be running on `arm64` instead of `amd64`. Is there a particular reason for this?
3. I definitely understand the need for some `arm64`/`aarch64` builds, but are we still interested in `s390x` builds?

cc @peternewman

DaAwesomeP:
The version of cpplint from Google seems to require Python 2. I'm going to look into switching to this: https://github.com/cpplint/cpplint

kripton:
I would say let's bury Python 2 better sooner than later. It's EOL and most distros don't ship it anymore. Or at least make it hard to install it ;)

peternewman:
See also #1794 which I think will be unlocked by this too...

> * [ ]  Flake8 (Python 3)
> * [ ]  Flake8 (Python 2.7)

We've got a version of these already (although ideally they should `make builtfiles` first:
https://github.com/OpenLightingProject/ola/blob/master/.github/workflows/annotation.yml

> 1. At this point can we abandon the Python 2.7 lint tasks and only move forward with the Python 3 tasks?

Yeah I guess, especially if it's a nightmare.

[...]

PR #1817, opened 2023-02-16
GitHub Actions Lint: flake8, cpplint, enforce_licence, generic NOLINTs

DaAwesomeP:
Working toward #1815. This is still a draft. I am force-pushing to avoid massive commit spamming; I'll update the commit message to be very descriptive once I'm done.

I am using Debian stable as packaging on Debian seems to be very actively attended to by OLA contributors and not just Debian packagers. There also exist build instructions for Debian within the OLA repo.

So far:
- Build once, lint multiple times from that one build
- Don't require building first for lint tasks that don't need it
- Run flake8 the exact same way it would be run on a dev machine (from `Makefile`), but with annotations
- Move cpplint to latest `pip` version
- Add GH Actions annotations to `enforce_licence.py`
- Move the generic `NOLINT` check from the Travis-only script to a script that can be run anywhere
- Add GH Actions annotations to the generic `NOLINT` check

Still lots to do, but progress!


DaAwesomeP:
There seems to be a weird bug or some other issue with the new cpplint. It shows header guard name errors on a seemingly arbitrary set of files (a subset of files only within the `tools` directory) like:

```
./tools/ola_trigger/VariableInterpolator.h:21:  #ifndef header guard has wrong style, please use: ___W_OLA_OLA_TOOLS_OLA_TRIGGER_VARIABLEINTERPOLATOR_H_  [build/header_guard] [5]
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,119 @@
+name: lint
+on: [push, pull_request]
+env:
+  GH_ACTION: "TRUE"
```
FloEdelmann, line 4: `GITHUB_ACTIONS` is implicitly always set to `true` by default, so you could use that instead, see https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables.

peternewman: Exactly what I'd come to say!

Although I'd personally probably favour problem matchers over making our scripts GitHub Actions specific, they're pretty easy to do, even if you might need to tweak our script output to be easier to parse:
https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

peternewman:
> There seems to be a weird bug or some other issue with the new cpplint. It shows header guard name errors on a seemingly arbitrary set of files (a subset of files only within the `tools` directory) like:
> 
> ```
> ./tools/ola_trigger/VariableInterpolator.h:21:  #ifndef header guard has wrong style, please use: ___W_OLA_OLA_TOOLS_OLA_TRIGGER_VARIABLEINTERPOLATOR_H_  [build/header_guard] [5]
> ```

It seems to be more than just tools to me, it looked like everything. I think it's because your source archive is extracted into w/ola/ola. I don't know if that's technically a bug with our Makefile (it might need one of the relative or absolute variables added like some of the Python stuff, you should be able to experiment and prove it locally I think), or you need to `cd` before calling it, or both.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,119 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Count the number of generic NOLINTs
+        shell: bash
+        run: ./scripts/count_generic_nolints.sh
+  build:
+    name: Build for Lint Tasks
+    runs-on: ubuntu-latest
+    container: debian:stable
```
peternewman, line 28: Given we should build on Ubuntu too, is this really worth the overhead for linting? I don't know how much inefficiency it actually adds, but still.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Autoreconf
+        run: autoreconf -i
+      - name: Configure
+        run: ./configure --enable-rdm-tests --enable-ja-rule --enable-e133
+      - name: Build
+        run: make builtfiles VERBOSE=1
+      - uses: actions/upload-artifact@v3
+        with:
+          name: ola-debian-stable-built-source-tree
+          path: .
```
peternewman, line 57: :heart: That's a really neat idea!

peternewman: For bonus steps I wonder if you can need a certain step in the task, so it could fork here and do a normal make from now on? Although `make distcheck` is probably still better for compilation.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,119 @@
...
+            debhelper devscripts wget python3-pip
+      - name: Install Python lint tools
+        run: python3 -m pip install --no-input cpplint flake8
+      - name: Install build dependencies
+        shell: bash
+        run: |
+          apt-get -y install libcppunit-dev uuid-dev libncurses5-dev \
+            libmicrohttpd-dev protobuf-compiler python3-protobuf \
+            libprotobuf-dev libprotoc-dev zlib1g-dev libftdi-dev \
+            libusb-1.0-0-dev liblo-dev libavahi-client-dev python3-numpy
```
peternewman, line 85: I'm guessing this is all because we're calling configure during the make?

Given we don't (currently) do any awareness of configure about packages during cpplint, you could use the bare minimum requirements here and ./configure --disable-all-plugins etc to need fewer dependencies. Or possibly just ignore them regardless unless it re-runs configure?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Install build tools
+        shell: bash
+        run: |
+          apt-get -y install pkg-config libtool autoconf \
+            automake g++ bison flex make bash-completion dh-autoreconf \
+            debhelper devscripts wget python3-pip
+      - name: Install Python lint tools
+        run: python3 -m pip install --no-input cpplint flake8
+      - name: Setup flake8 annotations
+        uses: rbialon/flake8-annotations@v1
```
peternewman, line 110: I didn't realise there was another one around that just did annotation!

Makefile.am
```diff
@@ -259,16 +258,9 @@ CPP_LINT_FILES = $(shell find . \( -name "*.h" -or -name "*.cpp" \) -and ! \( \
...
 else
-cpplint: Makefile.am cpplint.py
-	./cpplint.py --filter=$(CPP_LINT_FILTER) $(CPP_LINT_FILES)
-endif
-
-if !FOUND_CPPLINT
-cpplint.py:
-	wget -O cpplint.py $(CPP_LINT_URL)
-	chmod u+x cpplint.py
+  $(error cpplint not found. Install cpplint and re-run configure.)
```
peternewman, line 265: Can we somehow clarify that it's no longer the Google one? Or does that not matter?

scripts/count_generic_nolints.sh
```diff
@@ -0,0 +1,32 @@
...
+#  GNU Library General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# count_generic_nolints.sh
+# Copyright (C) 2023 Perry Naseck, Peter Newman
+
+# This test is based on a Travis CI test by Peter Newman
```
peternewman, line 19:
```suggestion
# This script is based on a Travis CI test by Peter Newman
```

scripts/count_generic_nolints.sh
```diff
@@ -0,0 +1,32 @@
...
+nolints="$(grep -n --exclude "$(basename $0)" -IR NOLINT * | grep -v "NOLINT(")"
+if [[ -v GH_ACTION ]]; then
+  nolints="$(printf "%s\n" "$nolints" | sed 's/^\(\(.*\):\([0-9]\+\):\(.*\)\)$/\1\n::error file=\2,line=\3,title=generic-nolints::Generic NOLINT not permitted/g')"
```
peternewman, line 22: Same discussion as before.

scripts/enforce_licence.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
```
peternewman, line 1: They really don't set python to be any Python now? :cry:

scripts/enforce_licence.py
```diff
@@ -281,13 +281,16 @@ def CheckLicenceForFile(file_name, licence, lang, diff, fix):
   f = open(file_name)
   # + 1 to include the newline to have a complete line
   header_size = len(licence) + 1
+  file_name_line_count = licence.count('\n') + 2
```
peternewman, line 284: Is this not just:
```suggestion
  file_name_line_count = header_size + 1
```

peternewman (changes requested):
Some comments and potential improvements

peternewman:
> I am force-pushing to avoid massive commit spamming

I think force pushes often confuse the review comments, but we can see how we go...

> I am using Debian stable as packaging on Debian seems to be very actively attended to by OLA contributors and not just Debian packagers. There also exist build instructions for Debian within the OLA repo.

As mentioned I think Ubuntu should be fine and ought to just work too with probably less overhead...
 
> So far:
> 
>     * Build once, lint multiple times from that one build

:heart: 

>     * Don't require building first for lint tasks that don't need it

I think they probably all want `make builtfiles` to check we've got licences etc on them.

> Still lots to do, but progress!

:tada:

peternewman:
> Still lots to do, but progress!

Also I think we can definitely break this down a bit and iterate on it. Once we've got at least once working action, we can merge that as it's better than the none we've got thanks to Travis. E.g. we could merge something without cpplint say, because the other bits will benefit.

I also forgot to say we should probably target 0.10 with this so we can check that branch works as intended too.

.github/workflows/lint.yaml
DaAwesomeP: The "initializing containers" task says it takes 4 seconds, so I don't think it's changing much.

Even though it shouldn't matter for this particular task, I think for consistency it would better to have them all the same unless they explicitly should be different (i.e. have a separate task for building on Ubuntu).

scripts/enforce_licence.py
DaAwesomeP: It's assumed that simply `python` is Python 2, which usually doesn't exist anymore.

scripts/enforce_licence.py
DaAwesomeP: `header_size` is the total character count of the license, while `licence.count('\n')` finds specifically how many lines it consumes. This is then used to find the line number where the filename line should be and create an annotation on that line.

scripts/count_generic_nolints.sh
DaAwesomeP: Fixed.

.github/workflows/lint.yaml
DaAwesomeP: Changed the env var! See comment below about moving to problem matchers later (something is better than nothing for now).

DaAwesomeP:
> As mentioned I think Ubuntu should be fine and ought to just work too with probably less overhead...

See also note above, not much overhead. I am partial to Debian because of the docs that exist in this repo specific to Debian and the long release cycles for every release (as opposed to Ubuntu where not every release is LTS, and then a larger jump). If you would like to switch to Ubuntu, please let me know which version to run against.

> I think they probably all want `make builtfiles` to check we've got licences etc on them.

I think I'm not clear on the myriad of tasks that `make builtfiles`  does. Does it also prepare Python files, CPP source files, and the license files? 

> Also I think we can definitely break this down a bit and iterate on it. Once we've got at least once working action, we can merge that as it's better than the none we've got thanks to Travis. E.g. we could merge something without cpplint say, because the other bits will benefit.

Sounds good! I think I fixed the cpplint issue! I'm going to leave my current GH annotations implementation in for now (better than nothing, and I promise I will migrate!) and I will move them to the parser method later.

> I also forgot to say we should probably target 0.10 with this so we can check that branch works as intended too.

Will do, may involve a fresh pull.

.github/workflows/lint.yaml
DaAwesomeP: Yeah, running `make cpplint` doesn't seem to just run the cpplint command (which honestly surprised me).

What does the second `./configure` do for cpplint that isn't performed the first time? Is there a way we can disable having it call `./configure` again if it isn't necessary? Same for `make flake8`.

DaAwesomeP:
OK, I think I managed to change the target branch and re-branch and cherry-pick my source branch without causing mass destruction.

Makefile.am
DaAwesomeP: Fixed, added note to install from pip.

scripts/enforce_licence.py
peternewman: Ah yeah, makes sense

DaAwesomeP:
OK, by archiving and unarchiving the artifacts I just shaved 6+ minutes off of the total actions time. It now takes 4ish minutes total instead of over 10 minutes.

See https://github.com/actions/upload-artifact/issues/199, GitHub is very slow at dealing with large volumes of files.

.github/workflows/lint.yaml
DaAwesomeP: Yeah, you can definitely upload artifacts multiple times under different names! What should the order of things be?

DaAwesomeP:
OK, I have converted the license and `NOLINT` checks to use problem matchers!

scripts/count_generic_nolints.sh
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: Implemented the problem matchers after all!

DaAwesomeP:
OK, I went down a [rabbit hole](https://github.com/actions/runner/issues/763) with why file matching for problem matching annotations were not working, but I fixed it! There are now fully-functioning annotations via problem matchers for cpplint, flake8, check-licences, and generic-nolints.

Note that one [limitation with the problem matchers](https://github.com/actions/toolkit/blob/d2b7d85e7cb61467395604aae1b84439c320cb84/docs/problem-matchers.md#limitations) is that it will never show more than 10 annotations of one type, but it also doesn't mention anywhere that there may be more hidden annotations when more than 10 exist.

I think barring the above unresolved issues and comments with possible additional necessary `make` and `./configure` compilation tasks or prerequisites for other tasks, this should be good to merge!

.github/workflows/lint.yaml
peternewman: > The "initializing containers" task says it takes 4 seconds, so I don't think it's changing much.

I was more meaning what's the overhead on the emulation throughout the run. Essentially what's the time difference with and without that line?

> Even though it shouldn't matter for this particular task, I think for consistency it would better to have them all the same unless they explicitly should be different (i.e. have a separate task for building on Ubuntu).

Yeah fair point. I don't think we need to test Ubuntu AND Debian, and I can see the benefit of testing the deb package on the primary place we release to, just wondering if that should be the edge case. E.g. we test on clang and g++, we could do one of those on Debian and one on Ubuntu for broader testing and minimal overhead/risk of false positives with it being due to the OS not the compiler.

.github/workflows/lint.yaml
peternewman: Well the `make distcheck` which we ought to run at least once should probably test from a clean build, but if we e.g. test on different architectures they could start from this point. Probably one for a different PR when we get to the compilation tests though...

.github/workflows/lint.yaml
DaAwesomeP: >  we could do one of those on Debian and one on Ubuntu for broader testing

Yeah, I was even thinking of including an RPM distro in the build/test workflow if that's something easily supported.

.github/workflows/lint.yaml
DaAwesomeP: I'll get to compilation eventually! But yes that would go in a separate workflow.

.github/workflows/lint.yaml
peternewman: I had a quick look at the Makefile and can't spot an obvious issue. Does it do the same thing for you too? I suspect a file has a different timestamp or something so it thinks it's been touched. Perhaps something around the artefact transfer?

Do you want to just make a note of this somewhere for now, maybe in the main issue as a list of things to improve in future?

DaAwesomeP: > Do you want to just make a note of this somewhere for now, maybe in the main issue as a list of things to improve in future?

Sure, I'll add it as a TODO in the overall tracking issue to circle back to once we have all the workflows minimally working.

scripts/enforce_licence.py
peternewman: Not as per https://peps.python.org/pep-0394/ which suggests python (probably ideally via env) is the only way to do a script which runs on 2 and 3 (which this should). At least in this release and 0.11 I'm rather loath to break compatibility unnecessarily. People doing releases can always either tweak the files or depend on python-is-python3 type packages (although I guess the latter would be naught to force system-wide).

peternewman:
> See also note above, not much overhead. I am partial to Debian because of the docs that exist in this repo specific to Debian and the long release cycles for every release (as opposed to Ubuntu where not every release is LTS, and then a larger jump). If you would like to switch to Ubuntu, please let me know which version to run against.

Although the docs are aimed at Debian they should work fine on any Debian based OS (Ubuntu/Raspbian etc). I used to just target and switch based on LTSes (which I think is all GitHub Actions offers Ubuntu wise) and tried to run on the bleeding edge unless there seemed to be other issues e.g. with Coverity.
 
> I think I'm not clear on the myriad of tasks that `make builtfiles` does. Does it also prepare Python files, CPP source files, and the license files?

So it was initially targeted at Doxygen:
https://github.com/OpenLightingProject/ola/blob/1151e8d001e52b0bf1da7de0cbd42bea7d1b6a17/Makefile.am#L197-L200

More here:
https://www.gnu.org/software/automake/manual/html_node/Built-Sources-Example.html

But basically it's every source file that the build itself generates somehow, regardless of the programming language it might be in or how it might get generated (e.g. Makefile, protoc or shell script).

So while it might mean we've generated some C++ files unnecessarily when we're running a Python based flake8, it's probably easier than trying to work out which ones we need and keep track of it separately.

> OK, I think I managed to change the target branch and re-branch and cherry-pick my source branch without causing mass destruction.

Well done, that's better than I ever manage! :laughing: 

> OK, by archiving and unarchiving the artifacts I just shaved 6+ minutes off of the total actions time. It now takes 4ish minutes total instead of over 10 minutes.

Nice work.

> See [actions/upload-artifact#199](https://github.com/actions/upload-artifact/issues/199), GitHub is very slow at dealing with large volumes of files.

Did you see the comments about how cheating and using the caching functions may be faster still, maybe something to play with in future.

> Note that one [limitation with the problem matchers](https://github.com/actions/toolkit/blob/d2b7d85e7cb61467395604aae1b84439c320cb84/docs/problem-matchers.md#limitations) is that it will never show more than 10 annotations of one type, but it also doesn't mention anywhere that there may be more hidden annotations when more than 10 exist.

I didn't know that, hopefully they'll fix it in future, but the user will spot anyway when they fix the first batch of annotations!

> I think barring the above unresolved issues and comments with possible additional necessary `make` and `./configure` compilation tasks or prerequisites for other tasks, this should be good to merge!

So as mentioned, aside from a few minor bits like which Python shebang which I think we should fix before merge, I think the other outstanding bits can just be listed as a TODO tick list in the main issue and then we can merge this so we start to benefit, and then we can individually pick off the outstanding bits to e.g. make the flake8 build run configure less times etc.

.github/problem-matcher-lint-cpplint.json
```diff
@@ -0,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "lint-cpplint",
+            "severity": "error",
```
peternewman, line 5: Any reason for your severity inconsistencies with the matchers? I can't remember if it just defaults to error or something?

Edit, you've matched on the others! :facepalm:

.github/problem-matcher-lint-check-licences.json
```diff
@@ -0,0 +1,16 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "lint-check-licences",
+            "pattern": [
+                {
+                    "regexp": "^(notice|error)(:(file|dir):([^:]+)(:lines? (\\d+)(-(\\d+))?)?)?: (.+)$",
+                    "severity": 1,
```
peternewman, line 8: I don't think GitHub recognises notice as a severity, just error and warning.
https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md#single-line-matchers

.github/problem-matcher-lint-generic-nolints.json
```diff
@@ -0,0 +1,16 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "lint-generic-nolints",
+            "pattern": [
+                {
+                    "regexp": "^(notice|error)(:(file|dir):([^:]+)(:lines? (\\d+)(-(\\d+))?)?)?: (.+)$",
+                    "severity": 1,
```
peternewman, line 8: Again regarding notice.

.github/workflows/annotation.yml
```diff
@@ -1,24 +0,0 @@
...
-  flake8-annotation-python2:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@v2
-        with:
-          python-version: '2.7'
-      - name: Flake8 with annotations - Python 2
-        # Using the flake8 options in the .flake8 file
-        uses: TrueBrain/actions-flake8@main
```
peternewman, line 14: I'm sort of inclined to leave this one running until we give up on Python 2 completely (some of us don't want to kill it).

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
+name: lint
+on: [push, pull_request]
+jobs:
+  check-licences:
```
peternewman, line 4: This should ideally use/depend on the build step.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Enable Problem Matcher for GitHub annotations
+        run: echo "::add-matcher::.github/problem-matcher-lint-check-licences.json"
+      - name: Check licenses
+        shell: bash
+        run: ./scripts/enforce_licence.py
+  generic-nolints:
```
peternewman, line 17: And this too.

scripts/enforce_licence.py
DaAwesomeP: When trying either `/usr/bin/python` or `/usr/bin/env python` on a fresh Debian 11, I get:
```
-bash: ./scripts/enforce_licence.py: /usr/bin/python: bad interpreter: No such file or directory
```
or
```
/usr/bin/env: ‘python’: No such file or directory
```
That method is only guaranteed to work in a virtual environment (i.e. venv or virtualenv). The PEP says to use `#!/usr/bin/env python` specifically for virtual environments. The paragraph below says that a compromise must be made outside of a virtual environment.

What is the oldest OS release supported by packagers and/or OLA?

I'll change it back to `#!/usr/bin/env python`, but I think this should really be considered for a breaking change in 0.11. Python 2 has been EOL since January 1, 2020 and isn't included in any active OS distros anymore. I can open a separate issue for that.

.github/workflows/lint.yaml
peternewman: Well we've got a probably horribly out of date spec file!

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+    runs-on: ubuntu-latest
+    container: debian:stable
+    steps:
+      - name: Update package database
+        run: apt-get update -y
+      # See comments beginning at https://github.com/actions/runner/issues/763#issuecomment-1435474884
+      # Without Git, actions/checkout@v3 will resort to REST and will not
+      # create a .git folder or .git.config. The Problem Matcher looks for
+      # .git/config to find where the root of the repo is, so it must be
+      # present.
```
peternewman, line 38: If we're going for something like 80 characters we ought to try and hit it more I'd think!
```suggestion
      # See comments beginning at
      # https://github.com/actions/runner/issues/763#issuecomment-1435474884
      # Without Git, actions/checkout@v3 will resort to REST and will not
      # create a .git folder or .git.config. The Problem Matcher looks for
      # .git/config to find where the root of the repo is, so it must be
      # present.
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Autoreconf
+        run: autoreconf -i
+      - name: Configure
+        run: ./configure --enable-rdm-tests --enable-ja-rule --enable-e133
+      - name: Build
+        run: make builtfiles VERBOSE=1
+      - name: Archive artifacts to speed up slow GH Actions upload/download
+        shell: bash
+        run: |
+          touch ola-debian-stable-built-source-tree.tar.gz
```
peternewman, line 66: How come we need to touch it if it doesn't exist? Does the exclude throw a warning or something?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+          tar --exclude=ola-debian-stable-built-source-tree.tar.gz -cvzf ola-debian-stable-built-source-tree.tar.gz .
+      - name: SHA256 artifact archive
+        run: sha256sum ola-debian-stable-built-source-tree.tar.gz
```
peternewman, line 69: Is this just for reference/testing, does it actually need to be run in future?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - uses: actions/upload-artifact@v3
+        with:
+          name: ola-debian-stable-built-source-tree
```
peternewman, line 72: At some point we might need to add at least the branch/PR and possibly some other bits to this name for info. And I guess ideally the source filename.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+    container: debian:stable
+    needs: build
+    steps:
+      - name: Download built source tree archive
+        uses: actions/download-artifact@v3
+        with:
+          name: ola-debian-stable-built-source-tree
+          path: .
+      - name: SHA256 artifact archive
+        run: sha256sum ola-debian-stable-built-source-tree.tar.gz
```
peternewman, line 86: Presumably again for troubleshooting?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Unarchive artifacts and delete archive
+        shell: bash
+        run: |
+          tar -xvzf ola-debian-stable-built-source-tree.tar.gz .
+          rm ola-debian-stable-built-source-tree.tar.gz
+      - name: Display structure of extracted files
+        run: ls -alR
```
peternewman, line 93: Comment this out? Unless we can leverage the debug option of action builds?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+    container: debian:stable
+    needs: build
+    steps:
+      - name: Download built source tree archive
+        uses: actions/download-artifact@v3
+        with:
+          name: ola-debian-stable-built-source-tree
+          path: .
+      - name: SHA256 artifact archive
+        run: sha256sum ola-debian-stable-built-source-tree.tar.gz
```
peternewman, line 127: Again

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Unarchive artifacts and delete archive
+        shell: bash
+        run: |
+          tar -xvzf ola-debian-stable-built-source-tree.tar.gz .
+          rm ola-debian-stable-built-source-tree.tar.gz
+      - name: Display structure of extracted files
+        run: ls -alR
```
peternewman, line 134: Again

Makefile.am
```diff
@@ -240,13 +240,12 @@ lint: flake8 cpplint
 .PHONY : flake8
 flake8: Makefile.am
 if FOUND_FLAKE8
-	flake8
+	flake8 --statistics --count
```
peternewman, line 243: I assume this doesn't hide the individual faults?

Makefile.am
```diff
@@ -258,16 +257,9 @@ CPP_LINT_FILES = $(shell find . \( -name "*.h" -or -name "*.cpp" \) -and ! \( \
         -wholename "./tools/ola_trigger/config.tab.*" -or \
         -wholename "./tools/ola_trigger/lex.yy.cpp" \) | xargs)
 .PHONY : cpplint
-if FOUND_CPPLINT
 cpplint: Makefile.am
-	cpplint.py --filter=$(CPP_LINT_FILTER) $(CPP_LINT_FILES)
+if FOUND_CPPLINT
+	cpplint --repository=. --filter=$(CPP_LINT_FILTER) $(CPP_LINT_FILES)
```
peternewman, line 262: Would --root work if repository is still needed, which would be backwards compatible with the old tool too?

Makefile.am
```diff
@@ -258,16 +257,9 @@ CPP_LINT_FILES = $(shell find . \( -name "*.h" -or -name "*.cpp" \) -and ! \( \
...
 else
-cpplint: Makefile.am cpplint.py
-	./cpplint.py --filter=$(CPP_LINT_FILTER) $(CPP_LINT_FILES)
-endif
-
-if !FOUND_CPPLINT
-cpplint.py:
-	wget -O cpplint.py $(CPP_LINT_URL)
-	chmod u+x cpplint.py
+  $(error cpplint not found. Install cpplint via pip and re-run configure.)
```
peternewman, line 264: Could we clarify this a bit as to which one it is, maybe something like:
```suggestion
  $(error cpplint not found. Install forked cpplint (e.g. via pip) and re-run configure.)
```

We also can be more packaging agnostic:
```suggestion
  $(error cpplint not found. Install forked cpplint (e.g. via pip for the latest version) and re-run configure.)
```

As it's also in some nice friendly debs:
https://packages.ubuntu.com/kinetic/cpplint

scripts/enforce_licence.py
DaAwesomeP: @peternewman fixed!

configure.ac
```diff
@@ -959,7 +959,7 @@ fi
 # Linters
 AC_CHECK_PROG([flake8],[flake8],[yes],[no])
 AM_CONDITIONAL([FOUND_FLAKE8], [test "x$flake8" = xyes])
-AC_CHECK_PROG([cpplint],[cpplint.py],[yes],[no])
+AC_CHECK_PROG([cpplint],[cpplint],[yes],[no])
```
peternewman, line 962: Ah this might make my backwards compatible comment a bit less relevant.

protoc/StrUtil.cpp
```diff
@@ -360,39 +360,39 @@ char* FastUInt32ToBufferLeft(uint32_t u, char* buffer) {
     buffer[0] = ASCII_digits[0];
     buffer[1] = ASCII_digits[1];
     buffer += 2;
- sublt100_000_000:
+  sublt100_000_000:
```
peternewman, line 363: Ironic that this is an (old) Google file, or maybe it's cpplint not handling partially indented labels?

scripts/enforce_licence.py
```diff
@@ -232,11 +232,15 @@ def GetDirectoryLicences(root_dir):
...
     if LICENCE_FILE in files:
-      f = open(os.path.join(dir_name, LICENCE_FILE))
+      licence_path = os.path.join(dir_name, LICENCE_FILE)
+      f = open(licence_path)
       lines = f.readlines()
       f.close()
       licences[dir_name] = TransformLicence(lines)
-      print('Found LICENCE for directory %s' % dir_name)
+      rel_licence_path = "./" + os.path.relpath(licence_path)
+      rel_dir_path = "./" + os.path.relpath(dir_name)
```
peternewman, line 241: This smells a bit fishy, surely we can do something more pythonic and stable to get the relative path than just prepending ./?

scripts/enforce_licence.py
```diff
@@ -280,38 +284,44 @@ def CheckLicenceForFile(file_name, licence, lang, diff, fix):
     return 0
 
   f = open(file_name)
+  rel_path = "./" + os.path.relpath(file_name)
```
peternewman, line 287: Again

scripts/enforce_licence.py
```diff
@@ -280,38 +284,44 @@ def CheckLicenceForFile(file_name, licence, lang, diff, fix):
...
     if lang == PYTHON and first_line is not None:
       licence = first_line + licence
     ReplaceHeader(file_name, licence, lang)
     return 1
   else:
-    print("File %s does not start with \"%s...\"" % (
-        file_name,
-        licence.split('\n')[(0 if (lang == PYTHON) else 1)]))
+    print("error:file:%s:lines 1-%s: File does not start with or not exact "
+          "match of \"%s...\"" %
```
peternewman, line 322: I think this is slightly better English:
```suggestion
    print("error:file:%s:lines 1-%s: File does not start with, or not exact "
          "match of, \"%s...\"" %
```

peternewman (changes requested):
Some comments (before I've seen your very latest changes...)

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Enable Problem Matcher for GitHub annotations
+        run: echo "::add-matcher::.github/problem-matcher-lint-check-licences.json"
+      - name: Check licenses
+        shell: bash
+        run: python3 ./scripts/enforce_licence.py
```
peternewman, line 16: I think installing python-is-python3 should simplify this back to:
```suggestion
        run: ./scripts/enforce_licence.py
```

.github/problem-matcher-lint-cpplint.json
DaAwesomeP: Error is the default, but I wanted to be explicit because my experience with the problem matchers has been that the documentation is...garbage unfortunately.

.github/problem-matcher-lint-check-licences.json
DaAwesomeP: It does! I found it in the [source](https://github.com/actions/runner/blob/aaf02ab34cf874e67c43fc418338db21679e5d52/src/Runner.Worker/Handlers/OutputManager.cs#L218-L221). It causes nice messages like these: https://github.com/DaAwesomeP/ola/actions/runs/4218995033

.github/workflows/annotation.yml
DaAwesomeP: OK, I'll add it back.

.github/workflows/lint.yaml
DaAwesomeP: Will do.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,155 @@
...
+        shell: bash
+        run: |
+          apt-get -y install libcppunit-dev uuid-dev libncurses5-dev \
+            libmicrohttpd-dev protobuf-compiler python3-protobuf \
+            libprotobuf-dev libprotoc-dev zlib1g-dev libftdi-dev \
+            libusb-1.0-0-dev liblo-dev libavahi-client-dev python3-numpy
+      - name: Enable Problem Matcher for GitHub annotations
+        run: echo "::add-matcher::.github/problem-matcher-lint-cpplint.json"
+      - name: cpplint
+        run: make cpplint VERBOSE=1
```
peternewman, line 114: I'm wondering for all these commands we run via make, whether we should go a step deeper and leave them wrapped in our (probably renamed) ci.sh. So `make foo` is how to run the command for the normal user, and TASK=foo ci.sh is how to run it for our CI, which internally in the shell script would run this command with the verbose option (maybe making the configure steps optional for uses like this.

I'm just thinking it might simplify our usage a bit, or at least tie us less into a particular CI tool, and would help if we can get Travis going again for e.g. s390x compilation. Probably one for a future PR though...

.github/workflows/lint.yaml
DaAwesomeP: Yes, tar throws an error because it doesn't actually exclude the file if it does not exist when starting the archive.

.github/workflows/lint.yaml
DaAwesomeP: I can take it out if you'd like; I usually leave it in as a sanity check that something out of our control has not failed. If the checksum doesn't match then it can same some hours of troubleshooting.

DaAwesomeP: I always leave it in for actions creating downloadable files (i.e. end of the build task of this workflow), as then you can verify where the download came from.

.github/workflows/lint.yaml
DaAwesomeP: I can take it out if you'd like; I usually leave it in as a sanity check that something out of our control has not failed. If the checksum doesn't match then it can same some hours of troubleshooting.

.github/workflows/lint.yaml
DaAwesomeP: Yes, again for debugging.

Makefile.am
DaAwesomeP: No, does not hide any information.

.github/workflows/annotation.yml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: I can take it out if you'd like; I usually leave it in as a sanity check that something out of our control has not failed. If the checksum doesn't match then it can same some hours of troubleshooting.

.github/workflows/lint.yaml
DaAwesomeP: I'll add to the TODO!

Makefile.am
DaAwesomeP:
```
root=subdir
      The root directory used for deriving header guard CPP variable.
      This directory is relative to the top level directory of the repository
      which by default is determined by searching for a directory that contains
      .git, .hg, or .svn but can also be controlled with the --repository flag.
      If the specified directory does not exist, this flag is ignored.
```
It seems that because I resolved the issue with the missing `.git` folder I can just take out this argument entirely? I'll try that.

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: I'll add it to the TODO!

protoc/StrUtil.cpp
DaAwesomeP: That's interesting. Should I undo these indent fixes and add NOLINTs or leave them fixed like this?

configure.ac
DaAwesomeP: I think we could also completely take this and flake8 out of dependency checking, if you're OK with that. I don't think that someone building OLA should necessarily be required to have cpplint and flake8 (or maybe there is already a `./configure` flag for that).

Makefile.am
DaAwesomeP: Yup, that did the trick! Fixed!

Makefile.am
DaAwesomeP: Fixed!

scripts/enforce_licence.py
DaAwesomeP: I don't think there is a more Pythonic way; I wanted to match the relative dirs starting with `./` for consistency with the other tasks. I'll take them out.

DaAwesomeP: Fixed!

scripts/enforce_licence.py
DaAwesomeP: Fixed!

scripts/enforce_licence.py
DaAwesomeP: Fixed!

protoc/StrUtil.cpp
peternewman: No I think leaving them fixed is fine, I'm sort of curious if the CPP styleguide has moved on, or it just isn't handling some edge cases around labels.

configure.ac
peternewman: I don't believe it should block the build if they're not present.

Indeed my config.log has:
```
configure:27245: checking for flake8
configure:27273: result: yes
configure:27291: checking for cpplint.py
configure:27319: result: no
```

It just means we can present a more user-friendly message if they try to run the tools.

.github/workflows/lint.yaml
peternewman: Makes sense.

.github/workflows/lint.yaml
peternewman: :facepalm: Maybe add a comment to that effect?

.github/workflows/annotation.yml
```diff
@@ -0,0 +1,24 @@
...
+  flake8-annotation-python3:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+      - name: Flake8 with annotations - Python 3
+        # Using the flake8 options in the .flake8 file
+        uses: TrueBrain/actions-flake8@main
```
peternewman, line 24: I'm quite happy to drop the Python 3 one as that's covered in your new workflow, I was just thinking about keeping the Python 2 one:
```suggestion
```

.github/problem-matcher-lint-check-licences.json
peternewman: :heart: :facepalm: 
Perhaps you'd care to open a PR to improve their docs?

.github/problem-matcher-lint-cpplint.json
peternewman: Yeah makes perfect sense, I'd missed the regex matching elsewhere initially.

peternewman:
One other minor procedural comment, which I'm sure is down to personal preference, I've normally reviewed stuff where the person commenting makes a comment, the contributor then fixes it and replies to the comment with a "Done" or similar, or indeed suggests why my comment is daft and they shouldn't, then I can go back and resolve the discussion when I've reviewed the fix/agreed with my stupidity. It just makes it a bit easier to track what comments I've re-reviewed and resolved and what still needs looking at.

.github/workflows/lint.yaml
peternewman: Looks like we could make it configurable via here:
https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging

I'm just wondering if normally it's going to spam the log versus how often we're actually likely to need it?

scripts/enforce_licence.py
peternewman: I think doing `os.path.join(".", os.path.relpath(licence_path))` would do what you intended in a slightly less fishy way (ideally with a comment explaining why), but if it's not needed that's even better!

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,187 @@
...
+        run: |
+          apt-get -y install libcppunit-dev uuid-dev libncurses5-dev \
+            libmicrohttpd-dev protobuf-compiler python3-protobuf \
+            libprotobuf-dev libprotoc-dev zlib1g-dev libftdi-dev \
+            libusb-1.0-0-dev liblo-dev libavahi-client-dev python3-numpy
+      - name: Autoreconf
+        run: autoreconf -i
+      - name: Configure
+        run: ./configure --enable-rdm-tests --enable-ja-rule --enable-e133
+      - name: Build
```
peternewman, line 39: To clarify it's not doing the whole build:
```suggestion
      - name: Build builtfiles
```

peternewman (approved):
LGTM thanks.

I think the only outstanding bits that could be done are:

- [x] Dropping flake8-annotation-python3 again
- [ ] Minor workflow comments
- [ ] Possibly make some workflow debug logging optional, I know this will inevitably bite me either way, we'll leave it in and I'll spend my life scrolling through it, or we'll take it out and I'll immediately need it, so I'll let you pick what to do (then I can curse someone else :rofl: ).

.github/problem-matcher-lint-check-licences.json
DaAwesomeP: Will do!

.github/workflows/annotation.yml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: I left this one always on because it is so small and would possibly be a hint to temporarily enable the directory listing one.

.github/workflows/lint.yaml
DaAwesomeP: OK, I made the directory listing enabled only with that env var. FWIW, it never bothered me because normally the GitHub UI collapses them. When you download the log archive there is also some segmentation, but I can't remember if it is by job or step.

.github/workflows/lint.yaml
DaAwesomeP: See above, only enabled with the env var now.

.github/workflows/lint.yaml
DaAwesomeP: See below, left this one in because it is so small but made the directory listing only appear when debug logging is enabled

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: One thing to note to save some headache in the future: It doesn't seem possible to set secrets/env vars in GitHub Actions for branches with open pull requests. Obviously it won't use the target repo secrets/vars because that's a security risk, but it also won't use the source repo vars.

In order to use this debugging feature to fix another issue earlier I had to branch from this branch and enable this variable there.

DaAwesomeP:
> One other minor procedural comment, which I'm sure is down to personal preference, I've normally reviewed stuff where the person commenting makes a comment, the contributor then fixes it and replies to the comment with a "Done" or similar, or indeed suggests why my comment is daft and they shouldn't, then I can go back and resolve the discussion when I've reviewed the fix/agreed with my stupidity. It just makes it a bit easier to track what comments I've re-reviewed and resolved and what still needs looking at.

I'm very sorry! I will stop marking anything as resolved.

Anyway, I think I fixed the remaining bits!

.github/workflows/lint.yaml
peternewman: FYI you can also merge my suggestions directly in:

.github/workflows/lint.yaml
peternewman: > OK, I made the directory listing enabled only with that env var. FWIW, it never bothered me because normally the GitHub UI collapses them.

Yeah, at least with Travis I often found myself looking at the raw log, I think maybe because it couldn't handle all the lines of the main build or something.

> When you download the log archive there is also some segmentation, but I can't remember if it is by job or step.

That's only by job, but actually the way the build is split, that will help a bit.

> One thing to note to save some headache in the future: It doesn't seem possible to set secrets/env vars in GitHub Actions for branches with open pull requests. Obviously it won't use the target repo secrets/vars because that's a security risk, but it also won't use the source repo vars.

Ah, I must admit I'd just assumed if you made some changes and then looked on your local repo's action run (as a push) that it would give you the debug info.

> In order to use this debugging feature to fix another issue earlier I had to branch from this branch and enable this variable there.

Hmm that's quite a faff.

We can certainly see how we go I think, if we start to find this being really annoying we can certainly revert it and output it all the time.

DaAwesomeP:
@peternewman I was very confused (I'm sure this was not intentional), but apparently GitHub lets you edit others' comments?
image

peternewman:
> @peternewman I was very confused (I'm sure this was not intentional)

Indeed it wasn't, that was supposed to be a quote reply. I did wonder at the time why it hadn't commented some of the lines! :facepalm: 

> but apparently GitHub lets you edit others' comments?

I assume that's a privelege of being an organisation admin. I'm guessing you can't edit mine?

peternewman:
What I intended to write:

> I'm very sorry! I will stop marking anything as resolved.

No worries, I can completely see the logic in what you did too.
 
> Anyway, I think I fixed the remaining bits!
 
Amazing thanks! Onto the next one... laughing

peternewman:
I've now added branch protection for 0.10 of these new checks thanks!

PR #1819, opened 2023-02-20
Master resync

peternewman:
@kripton please could you work your magic and rebuild the JS stuff here so it's got the equivalent fixes from #1813 . I think I've merged it all correctly, but we've got local changes in this branch, and that stuff so can't just take one or the other of the JS stuff...


kripton:
Hi @peternewman, sure thing, I'll add it to my Todo-list. Pretty amazed how much traction OLA got in the last weeks! I might not have the time on Tuesday but I'll do it on Wednesday

kripton:
@peternewman: Done and pushed to https://github.com/kripton/ola/tree/master-resync. Indeed, it seems that you merged it all together correctly.

There's one thing that confuses me: In `olad/www/new/js', there are three files in ´master` (https://github.com/OpenLightingProject/ola/tree/master/olad/www/new/js) but only two in `0.10` (https://github.com/OpenLightingProject/ola/tree/0.10/olad/www/new/js). The file called `app.js` might not be required at all, at least from my understanding. The "source" that this file is generated from lives in `javascript/new-src/src` and for the actual package and a running instance of olad, the `app.min.js` is all that is required. The `.map`-file is to aid debugging (to make the minified code a lot more readable, like debugging symbols in C).

I didn't want to dig deep into the history right now to find out why the file is there now and not in `0.10` and since it's only 31kByte it should be negligible. Nevertheless, if we don't actually need it, we could simply drop it at some point.

kripton:
Found the commit that introduced the new file: https://github.com/OpenLightingProject/ola/commit/556a9a9521de8ded20741980c8a6abe32b1d4e99

peternewman:
> Pretty amazed how much traction OLA got in the last weeks!

Yeah it's really sped along, everyone must be able to tell there's a release coming!

> There's one thing that confuses me: In `olad/www/new/js', there are three files in ´master` (https://github.com/OpenLightingProject/ola/tree/master/olad/www/new/js) but only two in `0.10` (https://github.com/OpenLightingProject/ola/tree/0.10/olad/www/new/js). The file called `app.js` might not be required at all, at least from my understanding. The "source" that this file is generated from lives in `javascript/new-src/src` and for the actual package and a running instance of olad, the `app.min.js` is all that is required. The `.map`-file is to aid debugging (to make the minified code a lot more readable, like debugging symbols in C).
> 
> I didn't want to dig deep into the history right now to find out why the file is there now and not in `0.10` and since it's only 31kByte it should be negligible. Nevertheless, if we don't actually need it, we could simply drop it at some point.

Did you see the discussion in #1035 although it doesn't look like it got added to olad/OladHTTPServer.cpp so it could be served up. I can certainly see the benefit of the plain file, it's far easier to see what's going on during a merge looking at that compared to just the min file.

PR #1822, opened 2023-02-25
Master resync

peternewman:


.codespellignorewords
```diff
@@ -2,3 +2,4 @@ acn
 ACN
 inflight
 thead
+wronly
```
kripton, line 5: where does this occur that we need to ignore it instead of fixing it?

kripton (commented):
All changes look reasonable to me

.codespellignorewords
DaAwesomeP: It's a part of the file descriptor mode `O_WRONLY`, as in "write only." Example: https://github.com/OpenLightingProject/ola/blob/c9add79a8d1c83ab209146006dae87906803b64f/plugins/opendmx/OpenDmxThread.cpp#L71

Previous versions of codespell just didn't pick it up it seems.

kripton: aah, O_WRONLY absolutely makes sense. I thought it to be a misspelling of "wrongly"

DaAwesomeP: > I thought it to be a misspelling of "wrongly"

Indeed, codespell seemed to think exactly that!

peternewman:
> @peternewman : Do you want me to remove the "retrospective-review-required" label when I did a review or will you remove it after you had a look at my comments?

Yeah that would be amazing, as soon as it's had someone else's eyeballs it can be removed. Although I guess PRs like this, there's only a bit of new content, most of it is what's already been reviewed.

.codespellignorewords
peternewman: Yeah the default codespell word regex doesn't do snake case so doesn't need to ignore O_WRONLY type cases as it just thinks they're one not two words.

We could do the line ignore, but given we only use the word wrongly once, it's probably a bit irrelevant for now.

PR #1826, opened 2023-03-03
Something else we support apparently  

peternewman:
We're even listed in the manual!


plugins/usbdmx/README.md
```diff
@@ -10,6 +10,7 @@ This plugin supports various USB DMX devices including:
 * DMXCreator 512 Basic
 * Eurolite USB-DMX512 PRO
 * Eurolite USB-DMX512 PRO MK2 (when `enable_eurolite_mk2 = true`)
+* Eurolite freeDMX Wi-Fi
```
kripton, line 13: Since the "Eurolite freeDMX Wi-Fi" is actually out-of-production and seems to be replaced by the "eurolite freeDMX Wi-Fi AP" (which is a completely different product and doesn't reference OLA in its manual), I would make this (and probably the other product references as well) a link to its product page. Such as:
* [Eurolite freeDMX Wi-Fi](https://www.steinigke.de/en/mpn51860125-eurolite-freedmx-wi-fi-interface.html)

kripton (commented):
See inline comment. I'd prefer it to be a link to the product page

plugins/usbdmx/README.md
peternewman: Seem very reasonable, but I'm tempted to push all of that to a later change/convert to an issue for now (in case someone else fancies picking it up), partly as this was mostly just tidying up a long-standing commit and also because we'll definitely need some dead link checking at that point as Enttec for one have been terrible over the years at not understanding the idea of permalinks and constantly throwing away old ones as we find out on the website all the time sadly!

kripton: Sure. If manufacturers tend to violate common-www-laws, one can still use https://web.archive.org for "permanent links" ;)
But reasonable explanation, let's merge this

kripton (approved):
LGTM

plugins/usbdmx/README.md
peternewman: I've opened #1839 to track this

PR #1830, opened 2023-03-05
Pack and unpack UIDSets

peternewman:
Reduces the differences to review later in #1612


kripton (approved):
Seems all reasonable to me, tests are PASSed on my machine

PR #1832, opened 2023-03-06
Update manufacturer names

peternewman:


data/rdm/manufacturer_names.proto
```diff
@@ -630,10 +1066,22 @@ manufacturer {
   manufacturer_id: 20781
   manufacturer_name: "DJPOWER ELECTRONIC STAGE LIGHTING FIXTURE FACTORY (GUANGZHOU)"
 }
+manufacturer {
+  manufacturer_id: 434
+  manufacturer_name: "DJSI Schinstad ANS (Northern Light)"
```
kripton, line 1071: should we codespell-ignore those somehow?

kripton (approved):
phew, a PR without description ;)
codespell complains about some but changes look good as they are

data/rdm/manufacturer_names.proto
peternewman: Yeah I forgot to revisit that, now done and will hopefully come out clean...

PR #1836, opened 2023-03-09
GitHub Actions Build: distcheck, check, verify-trees, coverage, coveralls

DaAwesomeP:
This is just a draft! Still lots to do, but it should go quickly once I fix this pesky GH Actions cancel issue.

Only problem right now is getting GitHub not to cancel the build after 45 minutes. It must be OOM or some other resource; GitHub usually allows jobs to run for a very, very long time.

Example cancelled build (not by me!):
 - Rerunning this one now, maybe it will work: https://github.com/DaAwesomeP/ola/actions/runs/4355751567
 - https://github.com/DaAwesomeP/ola/actions/runs/4339605290

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -183,6 +183,6 @@ jobs:
         run: tar -xvzf $GH_OLA_VERIFY_TREES_TARBALL
       - name: Verify trees
         shell: bash
-        # Always succeed for now until verify trees is fixed
+        # TODO(Perry): Always succeed for now until verify trees is fixed
         # run: ./scripts/verify_trees.py ./ $(echo $GH_OLA_VERIFY_TREES_TARBALL | sed 's/.tar.gz$//')
         run: "./scripts/verify_trees.py ./ $(echo $GH_OLA_VERIFY_TREES_TARBALL | sed 's/.tar.gz$//') || true"
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,7 @@ jobs:
     name: "distcheck ${{ matrix.id }}"
     runs-on: ubuntu-latest
     container: debian:stable
+    timeout-minutes: 360
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,8 +9,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { id: debian:stable-amd64-gcc,   compiler: { cc: "gcc",   cxx: "g++"     } }
-          - { id: debian:stable-amd64-clang, compiler: { cc: "clang", cxx: "clang++" } }
+          - { id: "debian:stable-amd64-gcc",   compiler: { cc: "gcc",   cxx: "g++"     } }
+          - { id: "debian:stable-amd64-clang", compiler: { cc: "clang", cxx: "clang++" } }
     env:
       CC: "{{ matrix.compiler.cc }}"
       CXX: "{{ matrix.compiler.cxx }}"
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -64,7 +64,7 @@ jobs:
         run: ./configure $GH_OLA_CONFIGURE_ARGS
       - name: check
         # TODO(Perry): Parallelization causes GH Actions to hang -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }}
-        run: make {{ matrix.task }} -j1 VERBOSE=1
+        run: make ${{ matrix.task }} -j1 VERBOSE=1
       - name: Display structure of the built files
         if: always() && env.ACTIONS_STEP_DEBUG == 'true'
         run: ls -alR
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: "${{ matrix.id }}"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     container: ${{ matrix.container }}
     timeout-minutes: 360
     strategy:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,10 +2,21 @@ name: build
 on: [push, pull_request]
 jobs:
   distcheck:
-    name: Distcheck
+    name: "distcheck {{ matrix.id }}"
     runs-on: ubuntu-latest
     container: debian:stable
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { id: debian:stable-amd64-gcc,   compiler: { cc: "gcc",   cxx: "g++"     }}
+          - { id: debian:stable-amd64-clang, compiler: { cc: "clang", cxx: "clang++" }}
+    env:
+      CC: "{{ matrix.compiler.cc }}"
+      CXX: "{{ matrix.compiler.cxx }}"
     steps:
+      - name: Get number of CPU cores
+        run: echo "NUM_CPU_CORES=$(grep -c processor /proc/cpuinfo)" >> $GITHUB_OUTPUT
       - name: Update package database
         run: apt-get update -y
       # See comments beginning at
@@ -30,7 +41,7 @@ jobs:
             libmicrohttpd-dev protobuf-compiler python3-protobuf \
             libprotobuf-dev libprotoc-dev zlib1g-dev libftdi-dev \
             libusb-1.0-0-dev liblo-dev libavahi-client-dev python3-numpy \
-            default-jdk-headless
+            default-jdk-headless maven
       - name: Autoreconf
         run: autoreconf -i
       - name: Set configure arguments
@@ -45,5 +56,30 @@ jobs:
         run: echo "./configure $GH_OLA_CONFIGURE_ARGS"
       - name: Configure
         run: ./configure $GH_OLA_CONFIGURE_ARGS
-      - name: Distcheck
-        run: make distcheck VERBOSE=1
+      - name: distcheck
+        run: make distcheck -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }} VERBOSE=1
+      - name: Display structure of the built files
+        if: always() && env.ACTIONS_STEP_DEBUG == 'true'
+        run: ls -alR
+      - name: Archive artifacts to speed up slow GH Actions upload/download
+        if: always()
+        shell: bash
+        # If the file does not exist when tar excludes it, then it will not
+        # actually exclude it, so it must first be touched
+        run: |
+          touch ola-{{ matrix.id }}-distcheck-source-tree.tar.gz
+          tar --exclude=ola-{{ matrix.id }}-distcheck-source-tree.tar.gz -cvzf ola-{{ matrix.id }}-distcheck-source-tree.tar.gz .
+      - name: SHA256 artifact archives
+        if: always()
+        run: sha256sum ola-*.tar.gz
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ola-{{ matrix.id }}-distcheck-source-tree
+          path: ola-{{ matrix.id }}-distcheck-source-tree.tar.gz
+      - uses: actions/upload-artifact@v3
+        with:
+          name: ola-{{ matrix.id }}-distcheck
+          path: |
+            ola-*.tar.gz
+            !ola-{{ matrix.id }}-distcheck-source-tree.tar.gz
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -126,7 +126,7 @@ jobs:
       - name: Update package database
         run: apt-get update -y
       - name: Install coverage tools
-        run: apt-get -y install curl gcov-10 python3-pip
+        run: apt-get -y install curl gcc-10 python3-pip
       - name: Install Python coverage tools
         run: python3 -m pip install --no-input gcovr
       - name: Run gcovr
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -62,7 +62,7 @@ jobs:
       - name: Setup Python venv
         shell: bash
         run: |
-          python3 -m venv ../venv
+          python3 -m venv --system-site-packages ../venv
           source ../venv/bin/activate
           echo "PATH=$PATH" >> $GITHUB_ENV
       - name: Install Python build tools
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -49,10 +49,8 @@ jobs:
       - name: Autoreconf
         run: autoreconf -i
       - name: Set configure arguments
-        # usbdmx may cause GH action build to hang and fail
-        # TODO(Perry): re-renable ja-rule, usbdmx, and usbpro
         run: |
-          echo "GH_OLA_CONFIGURE_ARGS=--enable-rdm-tests --enable-java-libs --enable-e133 --disable-usbdmx --disable-usbpro" >> $GITHUB_ENV
+          echo "GH_OLA_CONFIGURE_ARGS=--enable-rdm-tests --enable-java-libs --enable-ja-rule --enable-e133" >> $GITHUB_ENV
       - name: Set additional Linux configure arguments
         if: runner.os == 'Linux'
         # Silence all deprecated declarations on Linux due to auto_ptr making the build log too long
@@ -63,7 +61,9 @@ jobs:
       - name: Configure
         run: ./configure $GH_OLA_CONFIGURE_ARGS
       - name: check
-        run: make check -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }} VERBOSE=1
+        # TODO(Perry): Parallelization causes GH Actions to hang
+        # run: make check -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }} VERBOSE=1
+        run: make check -j1 VERBOSE=1
       - name: Display structure of the built files
         if: always() && env.ACTIONS_STEP_DEBUG == 'true'
         run: ls -alR
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,7 +29,8 @@ jobs:
           apt-get -y install libcppunit-dev uuid-dev libncurses5-dev \
             libmicrohttpd-dev protobuf-compiler python3-protobuf \
             libprotobuf-dev libprotoc-dev zlib1g-dev libftdi-dev \
-            libusb-1.0-0-dev liblo-dev libavahi-client-dev python3-numpy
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,100 @@
...
+        if: runner.os == 'Linux'
+        # Silence all deprecated declarations on Linux due to auto_ptr making the build log too long
+        run: |
+          echo "GH_OLA_CONFIGURE_ARGS=$GH_OLA_CONFIGURE_ARGS CPPFLAGS=-Wno-deprecated-declarations" >> $GITHUB_ENV
+      - name: Print configure command
+        run: echo "./configure $GH_OLA_CONFIGURE_ARGS"
+      - name: Configure
+        run: ./configure $GH_OLA_CONFIGURE_ARGS
+      - name: distcheck
+        run: make distcheck -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }} VERBOSE=1
```
peternewman, line 64: So have you tried this? Getting us just `make check` for now would be great if that is quick enough to not time out?
```suggestion
      - name: makecheck
        run: make check -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }} VERBOSE=1
```

There's also ccache which we've used successfully in the past and ought to speed things up if the cache is cached too.

DaAwesomeP: I'll try simply `make check`!

What's weird is that it isn't timing out per say...it's almost like a watchdog has decided there hasn't been any output for a while and that is why it is stopping or some OOM situation. I think I need to open an issue elsewhere about it.

Yeah I'll look into ccache.

DaAwesomeP:
OK, I tried `make check` and this is fascinating...it also gets stucking compiling somewhere around `libolausbdmx` and the `usbdmx` and `usbpro` plugins. Maybe I can try turning those off.

DaAwesomeP:
OK, I tried and I am having an issue where the following command does not disable `usbdmx`:
```bash
./configure --enable-rdm-tests --enable-java-libs --enable-ja-rule --enable-e133 --disable-usbdmx --disable-usbpro CPPFLAGS=-Wno-deprecated-declarations
```
```
OLA Version 0.10.9

Prefix: '/usr/local'
Compiler: 'g++ -g -O2 -std=gnu++11 -pthread '
Linker: '/usr/bin/ld -m elf_x86_64   -ldl '
Python: /usr/bin/python3

Python API: yes
Java API: yes
Enable HTTP Server: yes
RDM Responder Tests: yes
Ja Rule: yes
Enabled Plugins: artnet dummy e131 espnet ftdidmx gpio karate kinet milinst opendmx openpixelcontrol osc pathport renard sandnet shownet spi stageprofi uartdmx usbdmx
UUCP Lock Directory: /var/lock
```
I am able to reproduce this locally.

DaAwesomeP:
Hmmm, disabling these gets it further but not to completion. I'm thinking it must be running out of memory. I'll try disabling some of the parallelization.

DaAwesomeP:
Alright, finally! It took 20 minutes to run, but disabling parallelization (forcing `-j1`) causes it to get to the end of `make check`. This might fix `distcheck` too!

One check does fail (`CredentialsTest)` because the tests are running as root. What is this test testing? Why can't it run as root?

DaAwesomeP:
@peternewman @kripton Hey look at that! Fully working `make distcheck` and `make check`! I'm going to separate the verify trees into a separate job after `distcheck` and then this is good to go!

Looks like maybe nobody has run that verify trees check in a while.

I am no longer running `make dist` after `make distcheck` because it appears to be redundant: https://www.gnu.org/software/automake/manual/html_node/Preparing-Distributions.html

We still run `make check` separately because it also does gcov. It also has different configure flags that I copied from the Travis file.

peternewman:
> **UPDATE:** It appears `ja-rule` causes this. Disabling for now.

Yeah sorry I could have told you that, it's because ja-rule needs the usbdmx plugin to talk to the interfaces and send DMX.

> Alright, finally! It took 20 minutes to run, but disabling parallelization (forcing `-j1`) causes it to get to the end of `make check`.

That's the same sort of runtime as Travis used to be.

> This might fix `distcheck` too!
🤞 

> One check does fail (`CredentialsTest)` because the tests are running as root. What is this test testing? Why can't it run as root?

Because running as root is bad and means our code is at risk of being responsible for a security breach. There's a configure option to run as root (which I assume skips that test), but not running the check as root would be better!

> @peternewman @kripton Hey look at that! Fully working `make distcheck` and `make check`! I'm going to separate the verify trees into a separate job after `distcheck` and then this is good to go!

Amazing, just the minimally working check/distcheck would be great progress for now! Indeed would probably be better for the short term than even verify trees etc and having them in a later PR.

> Looks like maybe nobody has run that verify trees check in a while.

Yeah our CI has been broken for quite a while! I missed some bits in the release because it was AWOL (and I didn't run it manually). Some of that stuff hasn't changed though, so I'm slightly suspicious about other results.

> I am no longer running `make dist` after `make distcheck` because it appears to be redundant: https://www.gnu.org/software/automake/manual/html_node/Preparing-Distributions.html

That seems about as I vaguely understood.

> We still run `make check` separately because it also does gcov. It also has different configure flags that I copied from the Travis file.

That makes sense, but again leaving the gcov until later would be fine and probably makes sense too.

Is it worth updating this issue as it sounds like you've found the cause? https://github.com/actions/runner-images/issues/7004

I wonder if our Docker usage has made it worse? Can we tell Docker to offer more memory/use more cores?

DaAwesomeP:
> Because running as root is bad and means our code is at risk of being responsible for a security breach. There's a configure option to run as root (which I assume skips that test), but not running the check as root would be better!

Fixed, I now make a new user and run it the same way that the Debian build does.

> Amazing, just the minimally working check/distcheck would be great progress for now! Indeed would probably be better for the short term than even verify trees etc and having them in a later PR.

Separated it out already! It has the same results as before separating.

> Yeah our CI has been broken for quite a while! I missed some bits in the release because it was AWOL (and I didn't run it manually). Some of that stuff hasn't changed though, so I'm slightly suspicious about other results.

I think I'm not going to fix the source of these failures in this pull, especially because I (:sob:) need to rebase this to 0.10.

> That makes sense, but again leaving the gcov until later would be fine and probably makes sense too.

Nearly done with this too, just fixing the Coveralls upload bit. Should be done in an hour or so I think.

> Is it worth updating this issue as it sounds like you've found the cause? https://github.com/actions/runner-images/issues/7004

I know I should (and I will), though their lack of support with issues like these is not promising.

> I wonder if our Docker usage has made it worse? Can we tell Docker to offer more memory/use more cores?

Unsure. I can maybe try this out later.

DaAwesomeP:
Testing this is a funny process. I have a chain of different commits trying different things every 5 minutes, but because it takes 20 to 30 minutes to run I have to keep track of what was what in many tabs. Before I solved the OOM issue I had to keep these tabs open or I would lose the logs too.

DaAwesomeP:
OK, I think that I managed to re-target this to 0.10. This is good to go!

 - `distcheck` for both gcc and clang. This includes testing `make dist`, `make check`, `make install`, and `make clean`, and those artifacts are downloadable
 - `check` with gcov generation (`*.gcno` and `*.gcna` files), and those artifacts are downloadable
 - `coverage` for coverage reports with `gcovr`, and those reports are sent to Coveralls as well as nice HTML reports as artifacts are downloadable
 - Using `gcovr` means that coverage reports can be made offline and checked without needing Coveralls (or we can switch to other tools) in case they go the way of Travis (:-1:) or you want to check coverage before pushing
 - Verify trees to check `dist` output including a problem matcher for nice GitHub annotations
 - Not going to fix the verify trees issues in this pull; that is a larger feat and this already a large pull

@peternewman Please consider adding @coveralls as a collaborator to the repo and enabling the "LEAVE COMMENTS?" option so that it can report back coverage changes on pull requests (see https://github.com/coverallsapp/github-action/tree/release/v2#coveralls-comments-arent-added-to-my-pull-request).

.github/workflows/build.yaml
```diff
@@ -0,0 +1,204 @@
...
+            default-jdk-headless maven
+      - name: Generate coverage
+        shell: bash
+        run: make coverage VERBOSE=1
+      - name: Upload coverage to Coveralls
+        uses: coverallsapp/github-action@v2
+        with:
+          # Coveralls GitHub action does not support its own format
+          # see: https://github.com/coverallsapp/github-action/issues/104
+          # file: coverage.coveralls.json
```
peternewman, line 151:
```suggestion
          # file: coverage/coverage.coveralls.json
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,204 @@
...
+          format: cobertura
+          flag-name: ${{ matrix.id }}
+      - name: Upload coverage artifacts
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: ola-${{ matrix.id }}-coverage
+          path: coverage/
+  verify-trees:
+    name: 'Verify trees for ${{ matrix.id }}'
```
peternewman, line 162: I'd rather this didn't run while its returning what I suspect are false positives. Can we either comment it out or removed the commits to a future PR (but don't bother with the latter unless it's trivial).

I think the issue is either this build being more complete, or us not properly cleaning up the contents of nodist, but I'm not sure which is technically the problem.

configure.ac
```diff
@@ -1036,4 +1040,5 @@ Now type 'make @<:@<target>@:>@'
     check        - run the tests
     doxygen-doc  - generate the HTML documentation
     lint         - run the linters
+    coverage     - generate coverage post --enable-gcov
```
peternewman, line 1043: I think this should be able to force --enable-gcov like we do with say ja-rule and the USB stuff.

peternewman (changes requested):
I'd be quite happy to just get one or more of check or distcheck in first, and review the other stuff in due course.

peternewman:
> @peternewman Please consider adding @coveralls as a collaborator to the repo and enabling the "LEAVE COMMENTS?" option so that it can report back coverage changes on pull requests (see https://github.com/coverallsapp/github-action/tree/release/v2#coveralls-comments-arent-added-to-my-pull-request).

I think I turned off the coveralls comments last time as they were a bit spammy. You can still see the stats in the checks and we could presumably mandate a minimum level to pass...

.github/workflows/build.yaml
DaAwesomeP: Removing the commits will not be trivial; I'll see if I can mark as OK-to-fail or I'll comment it out. We did a similar thing for spellintian (still runs but issues only warnings and returns success).

configure.ac
DaAwesomeP: `--enable-gcov` is a `configure` flag, but running `make coverage` would happen after the fact. Are you saying that running `make coverage` should re-run `configure` with `--enable-gcov` added?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,204 @@
...
+    runs-on: ubuntu-20.04
+    container: ${{ matrix.container }}
+    timeout-minutes: 360
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { id: "coverage-debian-stable-amd64-gcc",    task: "check",     configure-args: "--enable-ja-rule --enable-e133 --enable-gcov",                         container: "debian:stable", compiler: { CC: "/usr/bin/gcc",   CXX: "/usr/bin/g++",     pkg: "gcc g++" } }
+          - { id: "distcheck-debian-stable-amd64-gcc",   task: "distcheck", configure-args: "--enable-ja-rule --enable-e133 --enable-rdm-tests --enable-java-libs", container: "debian:stable", compiler: { CC: "/usr/bin/gcc",   CXX: "/usr/bin/g++",     pkg: "gcc g++" } }
+          - { id: "distcheck-debian-stable-amd64-clang", task: "distcheck", configure-args: "--enable-ja-rule --enable-e133 --enable-rdm-tests --enable-java-libs", container: "debian:stable", compiler: { CC: "/usr/bin/clang", CXX: "/usr/bin/clang++", pkg: "clang"   } }
```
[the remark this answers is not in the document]
DaAwesomeP: Fixed.

.github/workflows/build.yaml
DaAwesomeP: Fixed.

DaAwesomeP: Build succeeds now!

peternewman: I quite liked the spellintian solution, or leaving the matcher but matching as info, but this is also fine.

.github/workflows/build.yaml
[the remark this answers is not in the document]
peternewman: I hadn't really clocked that, and it does odd wrapping for me on the GitHub diff (e.g. the space after CC). I think you can still see quite clearly e.g. which configure lines differ without too much trouble.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,204 @@
...
+        run: |
+          tar -xvzf ola-${{ matrix.id }}-source-tree.tar.gz .
+          rm ola-${{ matrix.id }}-source-tree.tar.gz
+      - name: Display structure of extracted files
+        if: env.ACTIONS_STEP_DEBUG == 'true'
+        run: ls -alR
+      - name: Update package database
+        run: apt-get update -y
+      - name: Install coverage tools
+        run: apt-get -y install curl gcc python3-pip
```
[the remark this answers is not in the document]
peternewman: Cool, I was just checking we shouldn't be using the compiler arg from the earlier matrix config.

configure.ac
peternewman: Ah gotcha. Yeah maybe not that, but presumably we could at least throw a different error if `--enable-gcov` wasn't passed to configure and we call `make coverage`, given it presumably won't work in this case.

DaAwesomeP: Ah yes, we can certainly catch that. I am honestly not very familiar with automake so I'm not certain how to do that.

DaAwesomeP: OK, I think I got this to work. Running `make coverage` should now build if it needs to and then run the coverage generation (similar to how `make check` builds and runs unit tests).

DaAwesomeP: OK, this is passing! Running `make check` runs the checks and then the coverage. I couldn't figure out how to get it to compile the checks without running them, but realistically this isn't very necessary.

DaAwesomeP:
Hey @peternewman anything else to change or are we good to go?

peternewman (commented):
Sorry a few more comments...

.github/workflows/build.yaml
```diff
@@ -0,0 +1,188 @@
...
+        # Silence all deprecated declarations on Linux due to auto_ptr making the build log too long
+        run: |
+          echo "GH_OLA_CONFIGURE_ARGS=$GH_OLA_CONFIGURE_ARGS CPPFLAGS=-Wno-deprecated-declarations" >> $GITHUB_ENV
+      - name: Print configure command
+        run: echo "./configure $GH_OLA_CONFIGURE_ARGS"
+      - name: Configure
+        run: sudo -u builduser ./configure $GH_OLA_CONFIGURE_ARGS
+      - name: ${{ matrix.task }}
+        # TODO(Perry): Parallelization causes GH Actions to hang -j${{ steps.num-cpu-cores.outputs.NUM_CPU_CORES }}
+        run: sudo -u builduser make ${{ matrix.task }} -j1 VERBOSE=1
```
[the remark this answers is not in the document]
DaAwesomeP: Fixed!

.github/workflows/build.yaml
```diff
@@ -0,0 +1,188 @@
...
+      - name: Find dist build tarball
+        run: |
+          echo "GH_OLA_VERIFY_TREES_TARBALL=$(ls -t --time=birth ola*.tar.gz| head -1)" >> $GITHUB_ENV
+      - name: Print dist build tarball name
+        run: echo "$GH_OLA_VERIFY_TREES_TARBALL"
+      - name: Extract dist build
+        run: tar -xvzf $GH_OLA_VERIFY_TREES_TARBALL
+      - name: Verify trees
+        shell: bash
+        # Always succeed for now until verify trees is fixed
```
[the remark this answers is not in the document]
DaAwesomeP: Fixed!

DaAwesomeP:
Debian 12 is now Debian stable, and it handles Python 3 system dependencies/pip differently, so I need to fix that. Working on that now!

DaAwesomeP:
OK, I think I fixed the Python3 venv/system-wide issue (all builds are passing the `configure` stage). We'll know for sure in about 25 minutes.

Here is what needed to change:

1. Debian 12 no longer allows pip to install packages system-wide alongside `python3-*` packages installed by APT.
2. We now create a clean `venv` for the build. This `venv` is set to also use system packages, so we can still use the Debian-provided packages as they would be used for building. Like before, we also install the newer tools that we want from pip.
3. When we run the build as `builduser` (since we can't run as `root` or we fail an OLA test), we must pass along the `$PATH` variable (which activates the `venv`) and also the `CC`/`CXX` variables. The `--preserve-env` flag won't pass `$PATH`, so we pass that with the `env` helper command.

DaAwesomeP:
Passing! I was simply missing `id:` on the number of cores step.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,199 @@
...
+          tar -xvzf ola-${{ matrix.id }}-source-tree.tar.gz .
+          rm ola-${{ matrix.id }}-source-tree.tar.gz
+      - name: Display structure of extracted files
+        if: env.ACTIONS_STEP_DEBUG == 'true'
+        run: ls -alR
+      - name: Update package database
+        run: apt-get update -y
+      - name: Install Python
+        run: apt-get -y install python3 python-is-python3
+      # Disable problem matcher for now until verify trees is fixed
```
peternewman, line 185:
```suggestion
      # TODO(Perry): Disable problem matcher for now until verify trees is fixed
```

peternewman (approved):
Just the one TODO comment which I'll auto merge.

Thanks again for this @DaAwesomeP ! It's awesome!

.github/workflows/build.yaml
DaAwesomeP: Committed!

peternewman: Thanks @DaAwesomeP . It wouldn't let me edit your code, maybe you've disabled contributor access?

peternewman:
Coverage wasn't in the required list, but I figure it probably doesn't need to be unless we're being really strict!

I guess we'll now see if it pushes our coverage status correctly...

DaAwesomeP:
> Coverage wasn't in the required list, but I figure it probably doesn't need to be unless we're being really strict!

It might be possible to setup rules that block merging is coverage dips by X percent in total? I'm not sure.

> I guess we'll now see if it pushes our coverage status correctly...

Passed! See https://coveralls.io/github/OpenLightingProject/ola?branch=0.10 and https://coveralls.io/jobs/123479842

peternewman:
> > Coverage wasn't in the required list, but I figure it probably doesn't need to be unless we're being really strict!
> 
> It might be possible to setup rules that block merging is coverage dips by X percent in total? I'm not sure.

Maybe if we can just get it to comment if it drops by say 5% or something? Then it will only be usefully spammy!

PR #1841, opened 2023-03-13
E1.33 controller

peternewman:
I realised you might not have got the @mention in my personal repo ( https://github.com/peternewman/ola/pull/43 ) so I'll stick it here too.

This PR would want to go in after #1612 as it's done on top of that.

Please don't actually start reviewing this @kripton , but if you fancy taking a look at it in due course, I was going to break it down a bit to reduce the overall changes of this and https://github.com/OpenLightingProject/ola/pull/1612/files .

Most of the commits add a new PDU each (across three files), do you think there's an optimal number of files/size of commit to split it into for review?

I'll need to cherry-pick to get them into master without just queuing them up behind the existing LLRP stuff, but they should work fine standalone functionally...


peternewman:
Assigned to you @kripton just for thoughts on my questions at the top, please don't bother actually reviewing it yet as above...

kripton:
Cherry-picked parts of it to #1939 for easier review and merging in small chunks

peternewman:
I've not fully committed and pushed, but I've just finally got the basics of an OLAd E1.33 plugin talking to the ETC RDMNet implementation and even running the RDM Responder Tests against it, so it would be great to get this in so others can start doing E1.33 testing too...

Issue #1848, opened 2023-03-31
Add Elegato light control


peternewman:
Some useful info:
https://web.archive.org/web/20210806013041/https://trevorsullivan.net/2020/08/17/automate-the-elgato-key-light-air-with-powershell/
https://github.com/pcgeek86/elgato/blob/master/elgato/elgato.psm1
https://github.com/NickParks/elgato-light-api

Basic functionality ought to be possible via ola_trigger, like so:
https://github.com/OpenLightingProject/ola/blob/master/tools/ola_trigger/contrib/philips_hue_osram_lightify.conf

But a plugin would offer nicer integration of e.g. discovery via Bonjour etc and generally simplify patching and controlling a whole load of them on different channel numbers.

peternewman:
@gobo-ws strikes again! https://github.com/gobo-ws/ola-trigger-elgato-key-light-dmx

Do you want to put this in the contrib folder too @gobo-ws ?

Or maybe we should just link to your projects list or something?

I'm also wondering with some of the rest of these, do we need to improve ola_trigger to make them a bit more useable for situations like this? E.g. as step one, if you could pass in some variables on the command line, you could control 2 or many more of them with one common config file by running multiple instances like so:
ola_trigger elgato_key_light.conf --offset 0 --config "hostname='192.168.0.1:9123'"
ola_trigger elgato_key_light.conf --offset 2 --config "hostname='192.168.0.2:9123'"

I think that could potentially be as simple as just prepending the config lines to the beginning of the parsed config file, with maybe multiple lines being passed in via multiple --config options.

I guess the next extension would be a meta config, where you don't need to run two copies of ola_trigger, you can pass in one config file which sets all those options for multiple internal instances, but that's perhaps one for the future.

gobo-ws:
Hehe!
Certainly I can do that @peternewman but unfortunately I haven't updated the trigger configuration files that I have already contributed to OLA.
So maybe it's better to link to my other trigger files such as Elgato Key Light in my repositorie?

Your ideas regarding improvement of ola_trigger sound very good. I look forward to the further development of that part.

peternewman:
Adding a scale feature would also be useful:


              I think we should have the dimmer work over the full range rather than just part of it which will be rather confusing for users.

I suspect separately this means some sort of scale command would simplify repetitive config a lot.

_Originally posted by @peternewman in https://github.com/OpenLightingProject/ola/pull/1849#discussion_r1164224261_

[...]

PR #1849, opened 2023-04-03
ola_trigger config for Elgato Key Light

gobo-ws:
ola_trigger config for Elgato Key Light or Elgato Key Light Air. 
See #1848

### DIFF
--- a/tools/ola_trigger/Makefile.mk
+++ b/tools/ola_trigger/Makefile.mk
@@ -4,6 +4,7 @@ EXTRA_DIST += \
 
 dist_noinst_DATA += \
     tools/ola_trigger/contrib/crelay.conf \
+    tools/ola_trigger/contrib/elgato_key_light.conf \
     tools/ola_trigger/contrib/mac_itunes.conf \
     tools/ola_trigger/contrib/mac_volume.conf \
     tools/ola_trigger/contrib/philips_hue_osram_lightify.conf \
@@ -76,7 +77,7 @@ tools_ola_trigger_ActionTester_LDADD = $(COMMON_TESTING_LIBS) \
 test_scripts += tools/ola_trigger/FileValidateTest.sh
 
 tools/ola_trigger/FileValidateTest.sh: tools/ola_trigger/Makefile.mk
-	echo "for FILE in ${srcdir}/tools/ola_trigger/example.conf ${srcdir}/tools/ola_trigger/test_file.conf ${srcdir}/tools/ola_trigger/test_file_falling.conf ${srcdir}/tools/ola_trigger/test_file_rising.conf ${srcdir}/tools/ola_trigger/contrib/crelay.conf ${srcdir}/tools/ola_trigger/contrib/mac_volume.conf ${srcdir}/tools/ola_trigger/contrib/mac_itunes.conf ${srcdir}/tools/ola_trigger/contrib/philips_hue_osram_lightify.conf; do echo \"Checking \$$FILE\"; ${top_builddir}/tools/ola_trigger/ola_trigger${EXEEXT} --validate \$$FILE; STATUS=\$$?; if [ \$$STATUS -ne 0 ]; then echo \"FAIL: \$$FILE caused ola_trigger to exit with status \$$STATUS\"; exit \$$STATUS; fi; done; exit 0" > $(top_builddir)/tools/ola_trigger/FileValidateTest.sh
+	echo "for FILE in ${srcdir}/tools/ola_trigger/example.conf ${srcdir}/tools/ola_trigger/test_file.conf ${srcdir}/tools/ola_trigger/test_file_falling.conf ${srcdir}/tools/ola_trigger/test_file_rising.conf ${srcdir}/tools/ola_trigger/contrib/crelay.conf ${srcdir}/tools/ola_trigger/contrib/elgato_key_light.conf ${srcdir}/tools/ola_trigger/contrib/mac_volume.conf ${srcdir}/tools/ola_trigger/contrib/mac_itunes.conf ${srcdir}/tools/ola_trigger/contrib/philips_hue_osram_lightify.conf; do echo \"Checking \$$FILE\"; ${top_builddir}/tools/ola_trigger/ola_trigger${EXEEXT} --validate \$$FILE; STATUS=\$$?; if [ \$$STATUS -ne 0 ]; then echo \"FAIL: \$$FILE caused ola_trigger to exit with status \$$STATUS\"; exit \$$STATUS; fi; done; exit 0" > $(top_builddir)/tools/ola_trigger/FileValidateTest.sh
 	chmod +x $(top_builddir)/tools/ola_trigger/FileValidateTest.sh
 
 CLEANFILES += tools/ola_trigger/FileValidateTest.sh


gobo-ws:
@peternewman please review if you are not too busy

tools/ola_trigger/contrib/elgato_key_light.conf
```diff
@@ -0,0 +1,127 @@
...
+2       220-225  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":154}]}' 'http://${hostname}/elgato/lights'`
+2       226-231  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":152}]}' 'http://${hostname}/elgato/lights'`
+2       232-237  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":149}]}' 'http://${hostname}/elgato/lights'`
+2       238-243  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":147}]}' 'http://${hostname}/elgato/lights'`
+2       244-249  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":145}]}' 'http://${hostname}/elgato/lights'`
+2       250-255  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":143}]}' 'http://${hostname}/elgato/lights'`
```
peternewman, line 127: I think you should be able to simplify/de-duplicate this, if you look at:
https://github.com/OpenLightingProject/ola/blob/master/tools/ola_trigger/contrib/mac_itunes.conf

Something like this (entirely untested):
```suggestion
2       4-9      ct='344'
2       10-15    ct='335'
2       16-21    ct='323'
...
2       250-255    ct='143'
2       4-255  `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"temperature":${ct}}]}' 'http://${hostname}/elgato/lights'`
```

I think the fact you're already doing it for hostname should mean it just works! :crossed_fingers: Unless it doesn't like the duplicate entry for the channel and values?

You could probably set both and then fire once to avoid two curl calls too, which might help responsiveness?

peternewman (changes requested):
Thanks as always @gobo-ws . A few ways I think we could improve the config slightly...

tools/ola_trigger/contrib/elgato_key_light.conf
peternewman: The only thing to consider with one curl call would be if you don't send enough channels, it might not do anything, but that probably matches lots of real lights TBH.

tools/ola_trigger/contrib/elgato_key_light.conf
```diff
@@ -0,0 +1,127 @@
...
+
+# DMX chart
+
+###############################################################################
+
+# +---------+---------+----------------------------+
+# | Channel | Value   | Function                   |
+# +---------+---------+----------------------------+
+# | 1       | 0-100   | Dimmer (0-100%)            |
+# |         | 101-255 | Unused range               |
```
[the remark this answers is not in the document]
gobo-ws: > I think we should have the dimmer work over the full range rather than just part of it which will be rather confusing for users.
> 
> I suspect separately this means some sort of scale command would simplify repetitive config a lot.
> 
> I see @nomis52 did this originally with the Mac Volume example, I'm not sure I agree with it personally.

I totally agree with your opinion, but I don't see how it can be done in the configuration script since it does not support any scale commands.
Well the only thing to do it is to hard code the input range and the scaled range, something like this:

| Input range 0-255 | Output range 0-100 |
|-------------------|--------------------|
| 0                 | 0                  |
| 1                 | 0                  |
| 2                 | 1                  |
| 3                 | 1                  |
| 4                 | 2                  |
| 5                 | 2                  |
| 6                 | 2                  |
| 7                 | 3                  |
| 8                 | 3                  |
| 9                 | 4                  |
| 10                | 4                  |
| 11                | 4                  |
| 12                | 5                  |
| 13                | 5                  |
| 14                | 5                  |
| 15                | 6                  |
| 16                | 6                  |
| 17                | 7                  |
| 18                | 7                  |
| 19                | 7                  |
| 20                | 8                  |
| 21                | 8                  |
| 22                | 9                  |
| 23                | 9                  |
| 24                | 9                  |
| 25                | 10                 |
| 26                | 10                 |
| 27                | 11                 |
| 28                | 11                 |
| 29                | 11                 |
| 30                | 12                 |
| 31                | 12                 |
| 32                | 13                 |
| 33                | 13                 |
| 34                | 13                 |
| 35                | 14                 |
| 36                | 14                 |
| 37                | 15                 |
| 38                | 15                 |
| 39                | 15                 |
| 40                | 16                 |
| 41                | 16                 |
| 42                | 16                 |
| 43                | 17                 |
| 44                | 17                 |
| 45                | 18                 |
| 46                | 18                 |
| 47                | 18                 |
| 48                | 19                 |
| 49                | 19                 |
| 50                | 20                 |
| 51                | 20                 |
| 52                | 20                 |
| 53                | 21                 |
| 54                | 21                 |
| 55                | 22                 |
| 56                | 22                 |
| 57                | 22                 |
| 58                | 23                 |
| 59                | 23                 |
| 60                | 24                 |
| 61                | 24                 |
| 62                | 24                 |
| 63                | 25                 |
| 64                | 25                 |
| 65                | 25                 |
| 66                | 26                 |
| 67                | 26                 |
| 68                | 27                 |
| 69                | 27                 |
| 70                | 27                 |
| 71                | 28                 |
| 72                | 28                 |
| 73                | 29                 |
| 74                | 29                 |
| 75                | 29                 |
| 76                | 30                 |
| 77                | 30                 |
| 78                | 31                 |
| 79                | 31                 |
| 80                | 31                 |
| 81                | 32                 |
| 82                | 32                 |
| 83                | 33                 |
| 84                | 33                 |
| 85                | 33                 |
| 86                | 34                 |
| 87                | 34                 |
| 88                | 35                 |
| 89                | 35                 |
| 90                | 35                 |
| 91                | 36                 |
| 92                | 36                 |
| 93                | 36                 |
| 94                | 37                 |
| 95                | 37                 |
| 96                | 38                 |
| 97                | 38                 |
| 98                | 38                 |
| 99                | 39                 |
| 100               | 39                 |
| 101               | 40                 |
| 102               | 40                 |
| 103               | 40                 |
| 104               | 41                 |
| 105               | 41                 |
| 106               | 42                 |
| 107               | 42                 |
| 108               | 42                 |
| 109               | 43                 |
| 110               | 43                 |
| 111               | 44                 |
| 112               | 44                 |
| 113               | 44                 |
| 114               | 45                 |
| 115               | 45                 |
| 116               | 45                 |
| 117               | 46                 |
| 118               | 46                 |
| 119               | 47                 |
| 120               | 47                 |
| 121               | 47                 |
| 122               | 48                 |
| 123               | 48                 |
| 124               | 49                 |
| 125               | 49                 |
| 126               | 49                 |
| 127               | 50                 |
| 128               | 50                 |
| 129               | 51                 |
| 130               | 51                 |
| 131               | 51                 |
| 132               | 52                 |
| 133               | 52                 |
| 134               | 53                 |
| 135               | 53                 |
| 136               | 53                 |
| 137               | 54                 |
| 138               | 54                 |
| 139               | 55                 |
| 140               | 55                 |
| 141               | 55                 |
| 142               | 56                 |
| 143               | 56                 |
| 144               | 57                 |
| 145               | 57                 |
| 146               | 57                 |
| 147               | 58                 |
| 148               | 58                 |
| 149               | 59                 |
| 150               | 59                 |
| 151               | 59                 |
| 152               | 60                 |
| 153               | 60                 |
| 154               | 61                 |
| 155               | 61                 |
| 156               | 61                 |
| 157               | 62                 |
| 158               | 62                 |
| 159               | 63                 |
| 160               | 63                 |
| 161               | 63                 |
| 162               | 64                 |
| 163               | 64                 |
| 164               | 65                 |
| 165               | 65                 |
| 166               | 65                 |
| 167               | 66                 |
| 168               | 66                 |
| 169               | 67                 |
| 170               | 67                 |
| 171               | 67                 |
| 172               | 68                 |
| 173               | 68                 |
| 174               | 69                 |
| 175               | 69                 |
| 176               | 69                 |
| 177               | 70                 |
| 178               | 70                 |
| 179               | 71                 |
| 180               | 71                 |
| 181               | 71                 |
| 182               | 72                 |
| 183               | 72                 |
| 184               | 73                 |
| 185               | 73                 |
| 186               | 73                 |
| 187               | 74                 |
| 188               | 74                 |
| 189               | 75                 |
| 190               | 74                 |
| 191               | 75                 |
| 192               | 75                 |
| 193               | 76                 |
| 194               | 76                 |
| 195               | 76                 |
| 196               | 77                 |
| 197               | 77                 |
| 198               | 78                 |
| 199               | 78                 |
| 200               | 78                 |
| 201               | 79                 |
| 202               | 79                 |
| 203               | 80                 |
| 204               | 80                 |
| 205               | 80                 |
| 206               | 81                 |
| 207               | 81                 |
| 208               | 82                 |
| 209               | 82                 |
| 210               | 82                 |
| 211               | 83                 |
| 212               | 83                 |
| 213               | 84                 |
| 214               | 84                 |
| 215               | 84                 |
| 216               | 85                 |
| 217               | 85                 |
| 218               | 86                 |
| 219               | 86                 |
| 220               | 86                 |
| 221               | 87                 |
| 222               | 87                 |
| 223               | 88                 |
| 224               | 88                 |
| 225               | 88                 |
| 226               | 89                 |
| 227               | 89                 |
| 228               | 90                 |
| 229               | 90                 |
| 230               | 90                 |
| 231               | 91                 |
| 232               | 91                 |
| 233               | 92                 |
| 234               | 92                 |
| 235               | 92                 |
| 236               | 93                 |
| 237               | 93                 |
| 238               | 94                 |
| 239               | 94                 |
| 240               | 94                 |
| 241               | 95                 |
| 242               | 95                 |
| 243               | 96                 |
| 244               | 96                 |
| 245               | 96                 |
| 246               | 97                 |
| 247               | 97                 |
| 248               | 98                 |
| 249               | 98                 |
| 250               | 98                 |
| 251               | 99                 |
| 252               | 99                 |
| 253               | 99                 |
| 254               | 100                |
| 255               | 100                |


Otherwise this can be solved by adding a shell script but in this case I wanted to do a simple script without the need for any shell script.
For more advanced uses I use Node-RED and the Range node wich works fine for this kind of scaling.

tools/ola_trigger/contrib/elgato_key_light.conf
```diff
@@ -0,0 +1,127 @@
...
+# +---------+---------+----------------------------+
+# | 2       |         | Color temperature (kelvin) |
+# |         | 0-3     | Unused range               |
```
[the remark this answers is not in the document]
gobo-ws: > What's the reason for missing this bit of the range, just to make the numbers nicer?

I wrote the original script some years ago and I think that my idea was that if you set the fader on that channel to 0 then the color temperature should not be changed, but we can definitely start at 0 instead of 4.

tools/ola_trigger/contrib/elgato_key_light.conf
gobo-ws: I probably hadn't really thought of that, will try this solution :thumbsup:

tools/ola_trigger/contrib/elgato_key_light.conf
```diff
@@ -0,0 +1,127 @@
...
+hostname='192.168.0.1:9123'  # Elgato Key light hostname/IP:port (default 9123)
+
+
+
+# Triggers
+
+###############################################################################
+# Slot  Values   Action
+
+1       0-100    `curl '-s' '-o /dev/null' '-k' '-H' '"Accept: application/json"' '-X' 'PUT' '--data' '{"numberOfLights":1,"lights":[{"on":1,"brightness":${slot_value}}]}' 'http://${hostname}/elgato/lights'`
```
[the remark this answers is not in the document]
gobo-ws: When I tested it back in 2020 I can't remember that the lighting fixture was switched off below brightness of 3. Maybe something has changed in the API. I need to get an Elgato Key Light again to verify it.

tools/ola_trigger/contrib/elgato_key_light.conf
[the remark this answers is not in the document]
peternewman: > I totally agree with your opinion, but I don't see how it can be done in the configuration script since it does not support any scale commands. Well the only thing to do it is to hard code the input range and the scaled range, something like this:

Yeah exactly, just like you're already doing with the colour temperature. Although as I mentioned with that if you map to a variable then you only need to write the actual command once. So it should only be 101-102 rows of config!

But yes, clearly a built in scale would be really helpful.

> Otherwise this can be solved by adding a shell script but in this case I wanted to do a simple script without the need for any shell script. For more advanced uses I use Node-RED and the Range node wich works fine for this kind of scaling.

Ah right. Is that the main missing feature from ola_trigger? Or are you generally doing more complicated logic when you're using Node Red?

tools/ola_trigger/contrib/elgato_key_light.conf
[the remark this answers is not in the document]
peternewman: Ah, yeah I'd not considered wanting to leave it as is. That feels like a bad behaviour as a DMX fixture, but perhaps a good workaround if you're trying to control it with a simpler lighting desk.

tools/ola_trigger/contrib/elgato_key_light.conf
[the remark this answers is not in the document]
peternewman: Okay, it was just something I noticed in some other code. I'll try and post a link if I rediscover it. Well that and the fact they've got an on/off parameter as well as brightness...

peternewman:
> Thanks for your feedback @peternewman. I will take a look on this and get back to you.

Thanks for the contribution!

> I don't have any Elgato Key Light at the moment so I might need to get one again to verify that the API still works the same way as in 2020. There are also some new lighting fixtures from Elgato which I assume shares the same API.

I might be able to get my hands on one in a few weeks too or get someone to test changes for me. Yeah I think I found some docs/code suggesting it was generally the same. Maybe the LED strip was a bit different.

tools/ola_trigger/contrib/elgato_key_light.conf
gobo-ws: It does not like the duplicate entry for the channel and values:

```
# test.conf
hostname='192.168.0.1:9123'
ct=''

2       4-9      ct='344'
2       10-15    ct='335'
2       16-21    ct='323'

2       4-21     `echo 'test ct: ${ct} hostname: ${hostname}'`
```
```
$ ola_trigger --validate test.conf
tools/ola_trigger/Action.cpp:405: Interval [4, 21] overlaps [4, 9]
tools/ola_trigger/ParserActions.cpp:156: Line 19: value [4, 21] collides with existing values.
```

peternewman: > It does not like the duplicate entry for the channel and values:

Try this instead:
```
# test.conf
hostname='192.168.0.1:9123'
ct=''

2       4-9      ct='344'
2       10-15    ct='335'
2       16-21    ct='323'

2       %        `echo 'test ct: ${ct} hostname: ${hostname}'`
```

I'm not sure what will happen when you send values 1-3 as currently set...

I think this should work, it currently passes the validation at least, so we should probably either fix the behaviour or the validation if it doesn't...

gobo-ws: Blank ct output on value 4-21. Value 1-3, 22-255 response ct 323 on channel 2

peternewman: So if you see this:
```
1       4-9      `echo 'Value 4-9'`
1       10-15    `echo 'Value 10-15'`
1       16-21    `echo 'Value 16-21'`
1       %        `echo 'Generic value'`
```

I think it's working in a sensible and effective way, although not very useful for this particular purpose.

Unfortunately I just tried this:
```
# test.conf
1       4-9      `echo 'Value 4-9'`
1       10-15    `echo 'Value 10-15'`
1       16-21    `echo 'Value 16-21'`
1       %        `echo 'Generic value'`
2       %        `echo 'Always send'`
```

But always send is only fired when slot two changes value.

So for this PR/without other changes, I think what you had is pretty good and we can't change much (aside from possibly the on/off thing mentioned elsewhere).

I guess this is another thing for #1848 though, there could be some sort of pre and post processing options where it would run the given command before or after and slot based processing if that slot had changed. So say:
```
# test.conf
1       pre      `echo 'Run before'`
1       4-9      `echo 'Value 4-9'`
1       10-15    `echo 'Value 10-15'`
1       16-21    `echo 'Value 16-21'`
1       %        `echo 'Generic value'`
1       post     `echo 'Run after'`
2       %        `echo 'Always send'`
```

gobo-ws: Sorry for the late reply but I've been sick off and on lately. I will continue with this when I recover. Will also order a Key Light to verify it works properly.

peternewman: > Sorry for the late reply but I've been sick off and on lately. I will continue with this when I recover.

Sorry to hear that @gobo-ws . Obviously no rush whatsoever, I hope you're feeling better soon!

> Will also order a Key Light to verify it works properly.

I'm pretty certain I can still get access to one, so please don't buy one just for testing purposes (although they do look fun)!

gobo-ws: I ordered an Elgato Key Light (will not keep it though) and made some tests.


Elgato Key Light dimming range starts at 3 % in the Elgato Control Center app but when using the API the light lights up already at 2 percent.

Sending 1 % does not generate any error from the API but I can't see any visible light output.
If you set the value to zero by using the API the light will turn off (it will change the light state on to zero)
So I think we keep the 0-100 % range.

Don't know if it applies to all models though.

I have a problem to get a soft dimmer level with OLA when using the fader in the web interface with this trigger file.
Don't remember having this issue before. But I think it works a little bit better when using the 0-100 range instead of scaling the input from 0-255 to 0-100.
Also made some tests by removing sending the on command on every brightness request but it made no difference.

gobo-ws: Any comments @peternewman?

tools/ola_trigger/contrib/elgato_key_light.conf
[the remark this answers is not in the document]
peternewman: So on/off seems to be mentioned here for example:
https://github.com/bitfocus/companion-module-elgato-keylight/blob/master/src/actions.js

tools/ola_trigger/contrib/elgato_key_light.conf
peternewman: > Elgato Key Light dimming range starts at 3 % in the Elgato Control Center app but when using the API the light lights up already at 2 percent.

So 2% is when it first turns on?

> Sending 1 % does not generate any error from the API but I can't see any visible light output. If you set the value to zero by using the API the light will turn off (it will change the light state on to zero) So I think we keep the 0-100 % range.

Yeah that makes sense in terms of the final output.

> Don't know if it applies to all models though.

Without a published API I think we just have to go with what works for one unit and wait for people to find issues!

> I have a problem to get a soft dimmer level with OLA when using the fader in the web interface with this trigger file. Don't remember having this issue before. But I think it works a little bit better when using the 0-100 range instead of scaling the input from 0-255 to 0-100. Also made some tests by removing sending the on command on every brightness request but it made no difference.

Have you tried logging what its sending to check there's nothing wrong with the maths or similar? TBH this feels more like its an issue with the web interface's fader than anything else to me.

Have you tried using say ola_recorder or one of the examples to play back a fade from 0-255 over a certain amount of time? Then you could switch between the two configs and see how it behaves (aside from speeding up when going over a shorter range of inputs).

My gut feeling is still that the benefit of it behaving just like any other fixture and running over the whole range cancels out any potential minor issues with our web UI.

> Running firmware: 1.0.3 (217) Did not want to update the firmware since I am not keeping the device.

Yeah fair enough.

PR #1855, opened 2023-04-21
Update Debian Package Services and Config Directory

DaAwesomeP:
Now that I am using the Debian build from `master` in production I have found some outdated parts.

This updates the Debian package to use a modern Systemd service and moves the config folder. It also removes the reliance on Debconf to enable/disable which definitely not needed for Systemd and is redundant on Debian for init scripts (and especially redundant now that Systemd wraps init scripts by default).

Per Debian package recommendations both the init script and Systemd service will be installed, but users running `systemctl start|enable|etc olad` will now trigger the Systemd service instead of the init script Systemd wrapper. This fixes a big bug I was experiencing where Systemd does not properly trigger the init script and so `systemctl restart olad` did not work with the init script but `start` and `stop` did.

Systemd users can now override options with Systemd overrides instead of an environment file `/etc/default` (new preferred method).

Per Debian package recommendations the configuration files should go in `/etc/ola` since they are user editable and not 100% runtime-controlled. This is also where I think most users expect to find the configuration.

This creates some breaking changes, but I think this is OK for the following reasons:
1. Until #1812 the Debian build in this repo was broken, so surely the raw Debian build from this repo has not been used in actual production in some time.
2. Due to the above I am 99% sure that anyone using OLA on a Debian-based system is either:
  a. Building from source, installing with `make install`, and creating their own config directory and service scheme
  b. Installing from the Debian package repository which already uses `/etc/ola` for configs and has removed the Debconf bits
3. I am submitting this against `master` instead of `0.10`.

These changes should make it much smoother to move between the Debian repo packages and the ones created here. As new features and bugfixes seem to have accelerated recently, the ability to use the pre-packaged master build is quite nice. Personally I experienced this because I needed KiNETv2 support (#1787).

The main remaining difference in the packages is in the way that the web assets are handled, but most users won't perceive a difference between these builds with that or would possibly try this build if that one breaks. I don't really see a need to update that part here.

Hopefully this eventually makes its way downstream to Debian and brings the Systemd service to everyone.


DaAwesomeP:
Requesting review from @peternewman

DaAwesomeP:
OK, I have this running on a production machine and it works like a charm!

DaAwesomeP:
@peternewman polite ping!

debian/changelog
```diff
@@ -1,3 +1,15 @@
+ola (0.11.0-1) UNRELEASED; urgency=medium
```
yoe, line 1: The version parser in dpkg sorts ~ before anything, even an empty part. Therefore, according to dpkg,

    0~0 < 0~1 < 0

With that background, it makes sense to render this version number as `0.11.0~git-1`, so that anyone who installs a git snapshot doesn't have to wait for an "official" `0.11.0-2` (or above) before their system is upgraded.

Alternatively, a script could be run that uses the `dch` command from the `devscripts` package to generate a new changelog entry from CI configuration, which would use the output from `git describe` to generate a relevant version number; e.g., something like `0.11.0~git<git describe output>-1` could work. If this is combined with uploading the resulting package into one of the many package repository services out there, then this will allow people to install a git snapshot, get automatic upgrades, and then eventually upgrade to the "proper" released package, once it exists.

debian/control
```diff
@@ -2,7 +2,7 @@ Source: ola
 Priority: optional
 Maintainer: Wouter Verhelst <wouter@debian.org>
 Uploaders: RenZO <renzo@imaginux.com>
-Build-Depends: debhelper (>= 13), autotools-dev, dh-autoreconf, dh-python, bash-completion, libcppunit-dev, bison, flex, pkg-config, uuid-dev, python3, python3-protobuf, libprotobuf-dev, protobuf-compiler, libprotoc-dev, libusb-1.0-0-dev, libftdi1-dev, liblo-dev, libmicrohttpd-dev, libncurses5-dev, libavahi-client-dev, python3-numpy
+Build-Depends: debhelper-compat (= 12), autotools-dev, dh-autoreconf, dh-python, bash-completion, libcppunit-dev, bison, flex, pkg-config, uuid-dev, python3, python3-protobuf, libprotobuf-dev, protobuf-compiler, libprotoc-dev, libusb-1.0-0-dev, libftdi1-dev, liblo-dev, libmicrohttpd-dev, libncurses5-dev, libavahi-client-dev, python3-numpy
```
yoe, line 5: According to [debhelper-compat-upgrade-checklist(7)](https://manpages.debian.org/bullseye-backports/debhelper/debhelper-compat-upgrade-checklist.7.en.html), compat level 13 is currently the recommended mode. We effectively already required debhelper 13 before, so that isn't a regression; I would therefore recommend making the dependency be `debhelper-compat (=13)`. If the concern is that it won't be installable on buster (where the available version is limited to 12), there *is* a debhelper 13 in buster-backports.

yoe (commented):
I know you didn't ask for my review, but I'm giving it anyway ;-P

Although I made some suggestions for improvement, none of them are really critical, so from my POV things can go in as-is and that would be better than the status quo; but I'll leave actual approval up to @peternewman.

debian/changelog
DaAwesomeP: > With that background, it makes sense to render this version number as 0.11.0~git-1, so that anyone who installs a git snapshot doesn't have to wait for an "official" 0.11.0-2 (or above) before their system is upgraded.

I really like this idea, though it is ultimately up to @peternewman.

> Alternatively, a script could be run that uses the `dch` command from the `devscripts` package to generate a new changelog entry from CI configuration, which would use the output from `git describe` to generate a relevant version number; e.g., something like `0.11.0~git<git describe output>-1` could work.

This is definitely more involved and could go in a later pull (manual versioning is probably fine for now). This mirrors the way that `setuptools_scm` works for Python/Pypi projects that I have used before and it is fantastic (you get unique build numbers even by commit hash...so useful!). This would play directly into dumping these builds into a nightly or dev build Debian repo!

DaAwesomeP:
> I know you didn't ask for my review, but I'm giving it anyway ;-P

I didn't ask but you're an infinitely more experienced package maintainer than me and I am _excited_ that you left a review! :D

debian/control
yoe: Oh, wow. I completely forgot about that.

The thing is that in compat level 13, we need to account for every file. If `make install` creates a file but it's not installed in any package and also not explicitly listed as something we don't install, you get an error. Between it being short to release time last time around, and having had to remove the rdm tester which produced a lot of those errors, I decided it wasn't the time for that.

When Debian 12 comes out, I'll be able to re-enable the rdm tester, so that might be a good point to try that again. I'll see if I can do a pull request at that point.

DaAwesomeP: > Oh, wow. I completely forgot about that.

Heh, don't worry I really dug through _all_ of the commits on Salsa!

> When Debian 12 comes out, I'll be able to re-enable the rdm tester, so that might be a good point to try that again.

I fixed building the RDM tester in #1812 and it works with Python 3 now. Between that and removing debconf/moving to Systemd I think those should be the last bits removed in the Debian downstream from here.

> I'll see if I can do a pull request at that point.

Amazing!

yoe: > > Oh, wow. I completely forgot about that.
> 
> Heh, don't worry I really dug through _all_ of the commits on Salsa!

Apparently :-D

> > When Debian 12 comes out, I'll be able to re-enable the rdm tester, so that might be a good point to try that again.
> 
> I fixed building the RDM tester in #1812 and it works with Python 3 now. Between that and removing debconf/moving to Systemd I think those should be the last bits removed in the Debian downstream from here.

I know. The problem though is that I couldn't introduce a new package into Debian anymore at the time, because the freeze had started and Debian's policies don't allow for that anymore then. Once the release has happened, this will not be a problem and it's just a matter of adding the right configuration then.

debian/changelog
DaAwesomeP: I'm going to work on this soon. I think of all of the pulls I have worked on recently this is the last one blocking me from using my own forked builds vs official builds in production.

peternewman:
Note from the init/systemd side there's also #1444 which I suspect is stuck on me looking at some stuff (as always)...

debian/changelog
peternewman: Yeah I love @yoe 's idea, even more so if we can somehow add it into bits of https://github.com/OpenLightingProject/ola/blob/master/common/base/Version.cpp and similar, so it shows up at the top of the log (and maybe on the webserver), given currently someone running master looks the same as someone running 0.10 branch. The challenge will be not forcing it to recompile each time, so it might just be a case of including the branch unless we give it some special flag for when its doing a package build.

I hadn't even considered pushing it to a repo, just rolling our own, is there one you'd recommend/suggest @yoe ?

DaAwesomeP: > The challenge will be not forcing it to recompile each time, so it might just be a case of including the branch unless we give it some special flag for when its doing a package build.

I have some web applications that check for the presence of `.git/FETCH_HEAD` at runtime to get the last commit hash (this doesn't require forking/executing the Git command during runtime). Though given that OLA is a project centered around Make/Autoconf, I think the best way would be to bake this into `Version.h` during the build. This way there is no dependency on Git during runtime.

Here is what I'm thinking:

1. When arbitrarily building, use the output of `git describe` or `.git/FETCH_HEAD` to generate a unique build identifier/version.
2. If the current directory is not a Git repo (missing `.git`), then fallback to the current method. This is useful if someone downloads a release source tarball instead of cloning and checking out. In this case they should still see the correct release number.
3. We add a flag to `./configure` that _forces_ the build to use the current versioning method so that when we build on a new tag/release we get a more official `0.11~git-1` build instead of `0.11~git-aabbcc-1` like we would want for nightlies and pull request builds. It's possible that `git describe` will also just work in this situation, but I need to read the docs on that.

Separately, we need to see what is up with `git describe` on this repo. I'm getting `0.10.2-3549-ge398f6d9b` on the `master` branch of my fork and I don't know what is causing that.

> I hadn't even considered pushing it to a repo, just rolling our own, is there one you'd recommend/suggest @yoe ?

I have seen a few repos that build in GitHub Actions and then push to the `gh-pages` branch and it gets automatically served by GitHub! It could push to this repo or we could have a separate repo for the builds. I would be really excited to have a nightly build APT repo as well as a release APT repo.

Other than that there are of course things like PPA, OBS, etc., though I am not familiar with the Debian-specific flavor (interestingly OBS can build most platforms).

DaAwesomeP: OK, part one is done: baking a build name into the Autoconf build. I left all of the existing versioning bit alone and simply added a new "build name" flag. I have it showing up on `./configure`, in `olad --version`, and all three web interfaces.

It uses `git describe` (this is all behavior from Git; I did not come up with this scheme except for the unknowns):

1. If Git is missing,the build name will be `unknown`.
2. If the folder is not a Git repo (we can't infer where it came from), then the build name will be `unknown-out-of-tree`. The existing versioning system still applies; this is just the build name.
3. If the Git repo is checked out to a tag or there are no new commits since the last tag, then the build name will simply be `0.10.9`. This means that when simply checking out a release it will reflect the untouched build in the build name.
4. If there have been any commits since the last tag, then Git describe will append `numberOfCommits-commitHash` to the end of the tag name. In this case the tag name is the most recent tag traced the current branch. So building in `master` would result in for example `0.10.9-9-gaabbccd` if there have been 9 commits since 0.10.9.
5. If there are changed but uncommited files in the repo, then Git will append `-dirty` to the build name.
6. Please don't name a tag `unknown` or `unknown-out-of-tree` :pleading_face: 

Examples:

- `0.10.9`: Checked out to the 0.10.9 tag
- `0.10.9-dirty`: Same as above but some files have changed and not been commited
- `0.10.9-2073-g0bf40e7`: Checked out to a branch that the last traceable tag as 0.10.9 and there have been 2073 commits since that
- `0.10.9-2073-g0bf40e7-dirty`: Same as above but some files have changed and not been commited
- `unknown-out-of-tree`: Folder is not a Git repository
- `unknown`: Git command is missing

I will now work on getting this into the Debian build!

DaAwesomeP: OK, this is working! The CI generates files like:
```
libola-dev_0.11.0~git-8482527-1_amd64.deb
libola1-dbgsym_0.11.0~git-8482527-1_amd64.deb
libola1_0.11.0~git-8482527-1_amd64.deb
ola-dbgsym_0.11.0~git-8482527-1_amd64.deb
ola-python_0.11.0~git-8482527-1_all.deb
ola-rdm-tests_0.11.0~git-8482527-1_all.deb
ola_0.11.0~git-8482527-1_amd64.buildinfo
ola_0.11.0~git-8482527-1_amd64.changes
ola_0.11.0~git-8482527-1_amd64.deb
```
I can also confirm that `debchange` indeed thinks this is a lower version than `0.11.0`.

It is only including the commit hash because the checkout action only checks out one commit at a time by default (it does not clone the tags to compare with). We can change this easily if we want the full/longer `git describe` output.

DaAwesomeP: UPDATE: OK, I enabled checking out all tags so we get builds like `0.11.0~git-0.10.9-2088-g69dcfdc-1`.

I realized the number of commits since last tag is useful because it will (usually) facilitate version increases between builds. The commit hash will never increment cleanly, but the number of commits always will due to Git's linearity. 

Example: `0.11.0~git-g1122334-1 < 0.11.0~git-g2233445-1` even if `0.11.0~git-g1122334-1` came later, but `0.11.0~git-0.10.9-124-g1122334-1 > 0.11.0~git-0.10.9-123-g2233445-1`.

DaAwesomeP: OK, I have added the version to the GitHub Actions build artifact names.

DaAwesomeP:
Polite ping! @peternewman @yoe

DaAwesomeP:
@peternewman @yoe resynced and ready for review!

DaAwesomeP:
Hey @peternewman, would appreciate a review when you have a spare moment. I want to push this out to some production machines soon if I can.

DaAwesomeP:
@peternewman one final ping!

DaAwesomeP:
@peternewman @yoe I know that this is a large pull request, but I would really appreciate your feedback!

peternewman:
Hi @DaAwesomeP ,

Sorry for the lack of reply. Could you do a new PR which just cherry picks the build version stuff and not all the Systemd bits please initially, so we can get that in before we have to consider your changes versus:

> Note from the init/systemd side there's also #1444 which I suspect is stuck on me looking at some stuff (as always)...

DaAwesomeP:
> so we can get that in before we have to consider your changes versus:
> 
> > Note from the init/systemd side there's also #1444 which I suspect is stuck on me looking at some stuff (as always)...

I don't think that these are mutually exclusive/conflict. I'm only adding the Systemd service to the Debian install; I'm not adding notify support. When that pull is merged (or as a part of merging it) we can modify this service to include the notify support. Right now this is just a drop-in replacement for the initd scripts.

DaAwesomeP:
@peternewman Please let me know! I am a bit hesitant to work on other things while this and other pulls have been open for so long. I completely understand that this is open source work, everyone is working on their available time, and what gets merged is up to you, but I am a bit less likely to contribute when pulls stagnate amidst other active development.

This doesn't conflict with the notify implementation. When notify support is implemented only a few lines in the Systemd service will be changed. This pull simply replaces the init.d implentation and continues to treat OLA as a simple start/stop service. I should also note that #1444 does not touch any of the service files in the `debian/` directory.

debian/changelog
yoe: > I hadn't even considered pushing it to a repo, just rolling our own, is there one you'd recommend/suggest @yoe ?

I tend to use [reprepro](https://packages.debian.org/reprepro), but setting that up in a CI system is involved, as it requires state be retained on the filesystem.

I have not that much experience with any of the repository-as-a-service thingies out there, so can't really help with that.

debian/ola-rdm-tests.rdm_test_server.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Open Lighting Architecture RDM Test Server
+Documentation=man:olad(1)
+After=network.target remote-fs.target olad.service
+Wants=olad.service
+
+[Service]
+User=olad
+Environment=RDM_TEST_SERVER_OPTS="--world-writable"
```
yoe, line 9: It's usually slightly more user friendly to keep this in an EnvironmentFile (which could live under /etc/default), as it is easier to edit that than it is to edit the systemd unit.

See systemd.exec(5) for details on that one.

debian/rules
```diff
@@ -11,7 +11,7 @@
 export VERBOSE=1
 
 %:
-	dh $@ --parallel --with autotools_dev,autoreconf,bash_completion,python3
```
yoe, line 14: Why are we removing the autotools_dev and autoreconf addons here?

debian/ola-rdm-tests.rdm_test_server.service
DaAwesomeP: Ah, yeah I see. The init script also read `/etc/default/ola-rdm-tests`. It looks like if I do the following then Systemd will use a default and override it if it is found in the environment file (which matches the init script behavior):
```service
Environment=RDM_TEST_SERVER_OPTS="--world-writable"
EnvironmentFile=/etc/default/ola-rdm-tests
```

debian/rules
DaAwesomeP: Oof, it has been some months and now I don't remember. I would guess that they weren't necessary or caused some error? I will re-stest this and let you know.

debian/ola-rdm-tests.rdm_test_server.service
DaAwesomeP: Fixed (see next line) in file

debian/rules
DaAwesomeP: @yoe Aha! It was to match your version here, so I would assume they are not necessary:

https://salsa.debian.org/wouter/ola/-/blob/3f36e8f3bab30f4cf631f5503894049cedcba201/debian/rules

yoe: *blush* okay, that explains that. I guess they're not needed anymore then :)

yoe (approved):

kripton:
I saw your ping. However, since I neither use Debian nor systemd, I'm probably not really qualified to comment this one. However, when it improves the situation, I'd rather merge it (and fix later in case of problems) instead of not merging it. Ping @peternewman  ;)

aroffringa (approved):
I went over the changes and it looks great to me. It's hard to tell everything is robust to the future but I think that's fine, we can deal with that later. Shall I merge it?

peternewman (changes requested):
Thanks again @DaAwesomeP . A few relatively minor comments from me...

olad/www/ola.js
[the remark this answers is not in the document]
DaAwesomeP: Well, it's been 3 years so I have almost no memory of it now hah.

debian/ola.olad.init
```diff
@@ -17,16 +17,13 @@ PIDFILE=/var/run/$NAME.pid
 DESC="OLA daemon"
 USER=olad
 LOG_LEVEL=3
-CONFIG_DIR="/var/lib/ola/conf"
+CONFIG_DIR="/etc/ola"
```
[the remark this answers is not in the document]
DaAwesomeP: Same applies here. I am quite confident that nobody is using our deb package over the Debian one.

debian/ola.olad.service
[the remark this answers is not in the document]
DaAwesomeP: Also discussed this here in this PR: https://github.com/OpenLightingProject/ola/pull/1855#issuecomment-1775515581

debian/ola.olad.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Open Lighting Architecture daemon
+Documentation=man:olad(1)
```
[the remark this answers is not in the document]
peternewman: > Can we add these in a followup PR? I want to avoid tacking on anything else to this PR since it has been open so long.

Yeah fair point.

> Separately, I think these links should go in the man page instead of in the Systemd service. That way you see them no matter how you get to the docs on Debian. Also means fewer places to keep them updated.

I think they should perhaps go in both (assuming people actually put web links in man pages, although it feels a bit like defeating the point of man. I'd hope we never need to update them anywhere regardless, as we'd just put redirects in.

peternewman (approved):
LGTM, lets finally get this in...

debian/ola-rdm-tests.rdm_test_server.init
```diff
@@ -17,16 +17,11 @@ DAEMON=/usr/bin/$NAME
 PIDFILE=/var/run/$CMD.pid
 DESC="OLA RDM Test Server"
 USER=olad
+DAEMON_ARGS="--world-writable"
 
 # Reads config file (will override defaults above)
 [ -r /etc/default/ola-rdm-tests ] && . /etc/default/ola-rdm-tests
 
-if [ "$RUN_DAEMON" = "true" ] || [ "$RUN_DAEMON" = "yes" ] ; then
-  DAEMON_ARGS="--world-writeable"
```
peternewman, line 25: After a bit of headbanging at Plugfest (not helped by SystemD silently stealing control of the init script code, I eventually spotted we'd changed to the other spelling of the argument (without the e).

For consistence (for now) I've corrected it in:
https://github.com/OpenLightingProject/ola/pull/2046/commits/b860a754ef2c50fca7b28b3d3b3df0d5c36915a1

I guess we could think about a breaking change to drop the e at some point in the future @DaAwesomeP ?

ghost:
SHUT Up!!! Please unsubscribe me!!!

peternewman:
> SHUT Up!!! Please unsubscribe me!!!

We can't @Calebn0102 . You need to click this link:

peternewman:
> OK, I have rebased and the build is working for Debian bullseye, bookworm, and trixie. Debian forky (testing) and sid (unstable) are showing C++ compilation errors to be fixed in other PRs:
> 
> 1. [`./include/ola/rdm/ResponderTagSet.h:191:15: error: 'uint8_t' has not been declared`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24085)
> 2. [`./include/ola/rdm/ResponderTagSet.h:195:7: error: 'uint8_t' was not declared in this scope`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24093)
> 3. [`./include/ola/rdm/ResponderTagSet.h:195:16: error: 'ptr' was not declared in this scope`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24097)
> 4. [`common/rdm/ResponderTagSetTest.cpp:56:29: error: cannot convert 'uint8_t*' {aka 'unsigned char*'} to 'int*'`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24105)
> 5. [`common/rdm/ResponderTagSetTest.cpp:83:30: error: cannot convert 'uint8_t*' {aka 'unsigned char*'} to 'int*'`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24116)
> 6. [`common/rdm/ResponderTagSetTest.cpp:84:29: error: cannot convert 'uint8_t*' {aka 'unsigned char*'} to 'int*'`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24127)
> 7. [`common/rdm/ResponderTagSetTest.cpp:100:29: error: cannot convert 'uint8_t*' {aka 'unsigned char*'} to 'int*'`](https://github.com/OpenLightingProject/ola/actions/runs/23385001971/job/68030465129?pr=1855#step:12:24138)

It would be good to know what's still broken after various other stuff has gone in now @DaAwesomeP ?

debian/ola-rdm-tests.rdm_test_server.init
DaAwesomeP: > I eventually spotted we'd changed to the other spelling of the argument (without the e).
> I guess we could think about a breaking change to drop the e at some point in the future @DaAwesomeP ?

Oh wow huh. I probably did this without realizing. I definitely did not mean to change the spelling of any existing args.

> After a bit of headbanging at Plugfest (not helped by SystemD silently stealing control of the init script code

I will say, this has been the default behavior for a very long time now. I was going to suggest fully removing the init script in v0.12.0 or later.

Is there a reason you were still using the init script over the Systemd one? You should be able to trivially set `OLAD_OPTS` in a Systemd override. As simple as: `systemctl edit olad` and `Environment=OLAD_OPTS=myopts`.

DaAwesomeP:
> It would be good to know what's still broken after various other stuff has gone in now @DaAwesomeP ?

It looks like tests are building and passing in `master` so we must be good?

peternewman:
> It looks like tests are building and passing in `master` so we must be good?

Yeah, I'm just a bit confused given I don't think I did anything to fix the last few...

debian/ola-rdm-tests.rdm_test_server.init
peternewman: > > I eventually spotted we'd changed to the other spelling of the argument (without the e).
> > I guess we could think about a breaking change to drop the e at some point in the future @DaAwesomeP ?
> 
> Oh wow huh. I probably did this without realizing. I definitely did not mean to change the spelling of any existing args.

No need to apologise, all three of us reviewing it missed it too!

> > After a bit of headbanging at Plugfest (not helped by SystemD silently stealing control of the init script code
> 
> I will say, this has been the default behavior for a very long time now. I was going to suggest fully removing the init script in v0.12.0 or later.

Ah right, I didn't know. Wouldn't that break things for people choosing not to use SystemD (I've no idea how many people that actually is).

> Is there a reason you were still using the init script over the Systemd one? You should be able to trivially set `OLAD_OPTS` in a Systemd override. As simple as: `systemctl edit olad` and `Environment=OLAD_OPTS=myopts`.

My issue was SystemD wasn't telling me what was wrong. It was only when I ran with `set -x` and stopped the include function dragging in SystemD so it really ran my bash code, I actually saw the error onscreen to realise what needed fixing. Also I've rarely had to override things at all, let alone in SystemD (does that also mean it won't load the /etc/defaults file too anymore)? Anyway I don't know if that's an us issue, our SystemD config, or SystemD itself, but I never saw the error in journalctl, just that it had exited.

PR #1862, opened 2023-06-18
Run CI Daily

DaAwesomeP:
This pull causes the CI to run everyday at 12am UTC. This should hopefully help catch issues before they end up failing in unrelated pull requests. This should definitely help catch the frequent new codespell matches.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,10 @@
 name: build
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * *'
+
 jobs:
   build:
     name: "${{ matrix.id }}"
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -2,7 +2,12 @@
 # https://salsa.debian.org/wouter/ola/-/blob/a38a396f6994b2b1af8efec9e208aee4e67f77aa/.gitlab-ci.yml
 
 name: debian
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * *'
+
 jobs:
   debian-build:
     name: 'Debian Build ${{ matrix.image_tag }} ${{ matrix.architecture }}'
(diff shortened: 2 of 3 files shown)


DaAwesomeP:
There seems to be a race condition with `python/ola/ClientWrapperTest.sh`. It failed on bullseye last run and then this run in sid.

Fitting that this is discovered in this pull request of all pull requests.

peternewman:
Yeah this has unfortunately been kicking around for a while, it's supposed to respond within a 750 uS window and its taken 2990.
> microseconds=750) delta (datetime.timedelta(microseconds=2990

Although now writing this I'm slightly suspicious that's its micro and I wonder/assumed it was supposed to be milli (without rechecking the code). It does inherently need to be fairly tight unfortunately, due to the nature of the check it's running.

But either way its definitely unfortunately worse on our various VMs/images.

peternewman (commented):
I can see the general sense, I wonder if we should stagger things a bit though, I guess it rather depends on the frequency of PRs coming in, but is daily an appropriate frequency for doing all the full builds? I wonder if they should be less often given only external package changes can break them.

I'm more onboard with doing the spelling CI regularly given as you say that will often get tripped up by external factors.

The other thing is unless these raise an issue or something, will anyone actually spot the CI issues?

DaAwesomeP:
> I can see the general sense, I wonder if we should stagger things a bit though, I guess it rather depends on the frequency of PRs coming in, but is daily an appropriate frequency for doing all the full builds? I wonder if they should be less often given only external package changes can break them.

My thinking is that daily/nightly allows for us to find those external breaking factors before they make it into an unrelated PR. We could switch to weekly if you would prefer. We could run lint daily and builds weekly.

> I'm more onboard with doing the spelling CI regularly given as you say that will often get tripped up by external factors.

I think running all of it on some frequency is nice because you find changes that you really aren't expecting (i.e. maybe Debian changes how Python libraries are installed or Autoconf behavior changes in a new version). Importantly, it's much nicer to find these changes one at a time and not many at once after a longer period of inactivity on the repo.

> The other thing is unless these raise an issue or something, will anyone actually spot the CI issues?

IIRC repo watchers will get a notification that a run failed.

PR #1863, opened 2023-06-18
GitHub Actions Lint: weblint

DaAwesomeP:
This one ended up being pretty simple!

@kripton I had to add a `package-lock.json`. Please let me know if what I generated is acceptable or if you have a different version from your own machine. If the latter, then please push it to branch `0.10 or suggest/modify the file in this pull.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -267,7 +267,7 @@ jobs:
           node-version: 18 # LTS
           cache: 'npm'
           cache-dependency-path: './javascript/new-src/package-lock.json'
-      - name: Install Node.js lint tools and dependencies
+      - name: Install npm dependencies
         run: npm ci
         working-directory: ./javascript/new-src
       - name: Run weblint
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -256,3 +256,36 @@ jobs:
         uses: codespell-project/codespell-problem-matcher@v1
       - name: codespell
         run: ./scripts/spelling.sh codespell
+  weblint:
+    name: weblint
+    runs-on: ubuntu-latest
+    container: debian:stable
+    steps:
+      - name: Update package database
+        run: apt-get update -y
+      - name: Install Git
+        run: apt-get -y install git
+      - uses: actions/checkout@v3
+      - name: Cache NPM
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-npm
+        with:
+          # This is the correct directory to cache
+          # See https://docs.npmjs.com/cli/v8/commands/npm-ci
+          path: ~/.npm
+          key:  debian_stable-lint-weblint-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            debian_stable-lint-weblint-${{ env.cache-name }}-
+            debian_stable-lint-weblint-
+      - name: Setup Node.js APT repo
+        shell: bash
+        run: curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
+      - name: Install lint tools
+        run: apt-get install -y nodejs
+      - name: Install Node.js lint tools
+        run: npm ci
+        working-directory: ./javascript/new-src
+      - name: Run NPM tests
+        run: npm test
+        working-directory: ./javascript/new-src
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -271,6 +271,7 @@ jobs:
         with:
           node-version: 18 # LTS
           cache: 'npm'
+          cache-dependency-path: './javascript/new-src/package-lock.json'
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -256,3 +256,38 @@ jobs:
...
+          restore-keys: |
+            debian_stable-lint-weblint-${{ env.cache-name }}-
+            debian_stable-lint-weblint-
+      - name: Setup Node.js APT repo
+        shell: bash
+        run: |
+          apt-get -y install curl
+          curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
+      - name: Install lint tools
+        run: apt-get install -y nodejs
```
FloEdelmann, line 287: This can probably be simplified with `actions/setup-node`, see e.g. https://github.com/OpenLightingProject/open-fixture-library/blob/a809729f90eaf635f2aac66ad70bb93c8247c4e1/.github/workflows/test.yaml#L13-L17

.github/workflows/lint.yaml
```diff
@@ -256,3 +256,38 @@ jobs:
...
+        env:
+          cache-name: cache-npm
+        with:
+          # This is the correct directory to cache
+          # See https://docs.npmjs.com/cli/v8/commands/npm-ci
+          path: ~/.npm
+          key:  debian_stable-lint-weblint-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            debian_stable-lint-weblint-${{ env.cache-name }}-
+            debian_stable-lint-weblint-
```
peternewman, line 280: Is this a match to search for, and then it finds the matching hash? Why the variants with and without the cache-name?

peternewman (changes requested):
This should either be new web lint, or we should add a todo to potentially check the old Google Closure one too at some point...

Plus @FloEdelmann actions comment

.github/workflows/lint.yaml
DaAwesomeP: Fixed!

.github/workflows/lint.yaml
DaAwesomeP: Fixed by switching to `setup-node`.

Using caches with fallback priorities is recommended here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows. I guess you can have multiple levels of caching (more specific or less specific) or caching in multiple steps for different things and try all of them if there isn't an immediate match.

DaAwesomeP:
@peternewman Annotation check is failing because Python 2.7 is no longer supported by `setup-python`. See https://github.com/actions/setup-python/issues/672.

.github/workflows/lint.yaml
```diff
@@ -256,3 +256,25 @@ jobs:
...
+      - name: Install Git
+        run: apt-get -y install git
+      - uses: actions/checkout@v3
+      - name: Setup Node.js v18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18 # LTS
+          cache: 'npm'
+          cache-dependency-path: './javascript/new-src/package-lock.json'
+      - name: Install Node.js lint tools and dependencies
```
FloEdelmann, line 275: Technically, this step installs "npm dependencies", not "Node.js" dependencies.

```suggestion
      - name: Install npm dependencies
```

.github/workflows/lint.yaml
```diff
@@ -256,3 +256,25 @@ jobs:
...
         run: ./scripts/spelling.sh codespell
+  weblint:
+    name: weblint
+    runs-on: ubuntu-latest
+    container: debian:stable
+    steps:
+      - name: Update package database
+        run: apt-get update -y
+      - name: Install Git
+        run: apt-get -y install git
```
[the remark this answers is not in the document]
DaAwesomeP: Fixed! I had it there for consistency with the others.

FloEdelmann (approved):

DaAwesomeP:
> This should either be new web lint, or we should add a todo to potentially check the old Google Closure one too at some point...

I couldn't find this in the old Travis CI version. Was this ever tested/ran with the CI previously?

peternewman (approved):
LGTM, thanks for the fixes @DaAwesomeP !

peternewman:
> I couldn't find this in the old Travis CI version. Was this ever tested/ran with the CI previously?

I think this can definitely be a TODO/future issue/item in the main CI TODO, but it seems this was actually done on the RDM App, by @FloEdelmann as it happens:
https://github.com/OpenLightingProject/rdm-app/pull/128/files

DaAwesomeP:
> > I couldn't find this in the old Travis CI version. Was this ever tested/ran with the CI previously?
> 
> I think this can definitely be a TODO/future issue/item in the main CI TODO, but it seems this was actually done on the RDM App, by @FloEdelmann as it happens: https://github.com/OpenLightingProject/rdm-app/pull/128/files

Added to the list in #1815!

PR #1866, opened 2023-06-19
Github Actions Debian trixie

DaAwesomeP:
The next Debian release (current testing) after bookworm is trixie: https://www.debian.org/releases/testing/


peternewman (approved):
Looks pretty trivial!

PR #1868, opened 2023-06-22
Master Resync

DaAwesomeP:
For #1801

PR #1869, opened 2023-06-22
Codespell Ignore package-lock.json

DaAwesomeP:
Found in #1868. Technically the `0.10` branch passes lint without this, but this file should never be checked regardless (machine generated).


peternewman (approved):
LGTM thanks

PR #1870, opened 2023-06-23
Switch to kind file Spelllintian problem matcher

peternewman:


DaAwesomeP:
For some unrelated reason on this run Spellintian found way fewer errors: https://github.com/OpenLightingProject/ola/actions/runs/5352911556/jobs/9708319367

GitHub is still treating it as line 1 it seems:
image

I don't see the `kind` option here (but maybe this is not the right documentation?): https://github.com/actions/toolkit/blob/91d3933eb52b351f437151400a88ba7d57442a9b/docs/problem-matchers.md

PR #1887, opened 2023-07-27
Support added for Nicolaudie SIUDI

CreaValix:
Hello to the ola project,

this patch adds support for Nicolaudie Sunlite intelligent USB DMX interfaces.

I have an old American DJ MyDMX (1.0) interface, which is a rebranded SIUDI-6. These patches work fine for me.

I chose not to extend the existing Sunlite USBDMX2, as the USB protocol is different (and much simpler). Also, there are many SIUDI devices where these patches could work for (due to lack of hardware I could not test this).

SIUDI interfaces require a constant stream of the complete DMX data, 512 bytes every ~20 ms (the original software uses 40, but 20 seems to work well).
If the frequency is too low, the interface will go into some of standby mode, which will result in flickering DMX devices if the next package arrives.
I used `ThreadedUsbSender` to provide this constant data stream, even for the async interface.

### DIFF
--- a/plugins/usbdmx/Siudi.cpp
+++ b/plugins/usbdmx/Siudi.cpp
@@ -1,0 +1,155 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Siudi.cpp
+ * The synchronous SIUDI widgets.
+ * Copyright (C) 2023 Alexander Simon
+ */
+
+#include "plugins/usbdmx/Siudi.h"
+
+#include <string.h>
+#include <unistd.h>
+
+#include "libs/usb/LibUsbAdaptor.h"
+#include "ola/Constants.h"
+#include "ola/Logging.h"
+#include "plugins/usbdmx/AsyncUsbSender.h"
+#include "plugins/usbdmx/ThreadedUsbSender.h"
+
+namespace ola {
+namespace plugin {
+namespace usbdmx {
+
+using ola::usb::LibUsbAdaptor;
+
+namespace {
+
+static const uint8_t ENDPOINT = 2;
+static const unsigned int TIMEOUT = 50; // 50ms is ok
+enum {SIUDI_PACKET_SIZE = 512};
+
+}  // namespace
+
+// SiudiThreadedSender
+// -----------------------------------------------------------------------------
+
+/*
+ * Sends messages to a SIUDI device in a separate thread.
+ */
+class SiudiThreadedSender: public ThreadedUsbSender {
+ public:
+  SiudiThreadedSender(LibUsbAdaptor *adaptor,
+                        libusb_device *usb_device,
+                        libusb_device_handle *handle);
+
+  bool Start();
+
+private:
+  LibUsbAdaptor* const m_adaptor;
+  uint8_t m_packet[SIUDI_PACKET_SIZE];
+  libusb_device_handle* const m_usb_handle;
+
+  bool TransmitBuffer(libusb_device_handle *handle,
+                      const DmxBuffer &buffer);
+};
+
+SiudiThreadedSender::SiudiThreadedSender(
+    LibUsbAdaptor *adaptor,
+    libusb_device *usb_device,
+    libusb_device_handle *usb_handle)
+    : ThreadedUsbSender(usb_device, usb_handle),
+      m_adaptor(adaptor), m_usb_handle(usb_handle) {
+  memset(m_packet, 0x00, SIUDI_PACKET_SIZE);
+}
+
+bool SiudiThreadedSender::Start() {
+  if (!ThreadedUsbSender::Start()) {
+    return false;
+  }
+
+  // Read device info. This call takes about 270 ms.
+  // Discard the buffer as the format is currently unknown.
+  uint8_t buf[64];
+  int ret = libusb_control_transfer(m_usb_handle,
+                                LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_ENDPOINT_IN,
+                                0x3f, 0x00c4, 1, buf, 64, 500);
+  if (ret != 64) {
+    OLA_WARN << "Failed to read SIUDI information: "
+            << (ret < 0 ? LibUsbAdaptor::ErrorCodeToString(ret) : "Short read");
+    return false;
+  }
+
+  // Unstall the endpoint. The original software seems to call this regularly.
+  ret = libusb_clear_halt(m_usb_handle, ENDPOINT);
+  if (ret != 0) {
+    OLA_WARN << "Failed to reset SIUDI endpoint: "
+            << (ret < 0 ? LibUsbAdaptor::ErrorCodeToString(ret) : "Unknown");
+    return false;
+  }
+  usleep(10000);
+
+  return true;
+}
+
+bool SiudiThreadedSender::TransmitBuffer(libusb_device_handle *handle,
+                                           const DmxBuffer &buffer) {
+  for (unsigned int i = 0; i < buffer.Size(); i++) {
+    m_packet[i] = buffer.Get(i);
+  }
+  int transferred;
+  int r = m_adaptor->BulkTransfer(
+    handle, ENDPOINT, (unsigned char*) m_packet,
+    SIUDI_PACKET_SIZE, &transferred, TIMEOUT);
+  if (transferred != SIUDI_PACKET_SIZE) {
+    // not sure if this is fatal or not
+    OLA_WARN << "SIUDI driver failed to transfer all data";
+  }
+  usleep(20000);
+  return r == 0;
+}
+
+// SynchronousSiudi
+// -----------------------------------------------------------------------------
+
+SynchronousSiudi::SynchronousSiudi(LibUsbAdaptor *adaptor,
+                                                   libusb_device *usb_device)
+    : Siudi(adaptor, usb_device) {
+}
+
+bool SynchronousSiudi::Init() {
+  libusb_device_handle *usb_handle;
+
+  bool ok = m_adaptor->OpenDeviceAndClaimInterface(
+      m_usb_device, 0, &usb_handle);
+  if (!ok) {
+    return false;
+  }
+
+  std::auto_ptr<SiudiThreadedSender> sender(
+      new SiudiThreadedSender(m_adaptor, m_usb_device, usb_handle));
+  if (!sender->Start()) {
+    return false;
+  }
+  m_sender.reset(sender.release());
+  return true;
+}
+
+bool SynchronousSiudi::SendDMX(const DmxBuffer &buffer) {
+  return m_sender.get() ? m_sender->SendDMX(buffer) : false;
+}
+}  // namespace usbdmx
+}  // namespace plugin
+}  // namespace ola
--- a/plugins/usbdmx/SiudiFactory.cpp
+++ b/plugins/usbdmx/SiudiFactory.cpp
@@ -1,0 +1,51 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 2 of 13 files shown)


plugins/usbdmx/UsbDmxPlugin.h
```diff
@@ -43,7 +43,7 @@ namespace usbdmx {
  *   - Eurolite DMX USB Pro.
  *   - Eurolite DMX USB Pro MK2.
  *   - Scanlime's Fadecandy.
- *   - Sunlite.
+ *   - Sunlite USBDMX2 and SIUDI.
```
peternewman, line 46: Minor styling nit:
```suggestion
 *   - Sunlite SIUDI.
 *   - Sunlite USBDMX2.
```

plugins/usbdmx/SyncPluginImpl.cpp
```diff
@@ -50,6 +50,8 @@
 #include "plugins/usbdmx/GenericDevice.h"
 #include "plugins/usbdmx/ShowJockeyDMXU1.h"
 #include "plugins/usbdmx/ShowJockeyDMXU1Factory.h"
+#include "plugins/usbdmx/Siudi.h"
+#include "plugins/usbdmx/SiudiFactory.h"
```
peternewman, line 54: Are they Sunlite officially and just rebadged as others? If so we should probably call this SunliteSIUDI.*

plugins/usbdmx/Siudi.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+bool SiudiThreadedSender::TransmitBuffer(libusb_device_handle *handle,
+                                           const DmxBuffer &buffer) {
+  for (unsigned int i = 0; i < buffer.Size(); i++) {
+    m_packet[i] = buffer.Get(i);
+  }
+  int transferred;
+  int r = m_adaptor->BulkTransfer(
+    handle, ENDPOINT, (unsigned char*) m_packet,
```
peternewman, line 114: But given you're just writing it untouched, you could probably just get away with GetRaw here:
http://docs.openlighting.org/ola/doc/latest/classola_1_1_dmx_buffer.html#a50eff7fa1c6a8d21ffa5ccb14a781de9

plugins/usbdmx/Siudi.cpp
```diff
@@ -0,0 +1,155 @@
...
+namespace plugin {
+namespace usbdmx {
+
+using ola::usb::LibUsbAdaptor;
+
+namespace {
+
+static const uint8_t ENDPOINT = 2;
+static const unsigned int TIMEOUT = 50; // 50ms is ok
+enum {SIUDI_PACKET_SIZE = 512};
```
peternewman, line 42: This could be assigned to be our DMX frame size constant, or might even become redundant as per later comments.

plugins/usbdmx/SiudiFactory.cpp
```diff
@@ -0,0 +1,51 @@
...
+    WidgetObserver *observer,
+    libusb_device *usb_device,
+    const struct libusb_device_descriptor &descriptor) {
+  if (descriptor.idVendor == VENDOR_ID &&
+      descriptor.idProduct == HOT_PRODUCT_ID) {
+    OLA_INFO << "Found a new Nicoleaudie SIUDI-6 device";
+    Siudi *widget = NULL;
+    widget = new SynchronousSiudi(m_adaptor, usb_device);
+    return AddWidget(observer, widget);
+  }
```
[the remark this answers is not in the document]
CreaValix: My device doesn't need a firmware download. VendorID is 0x6244 and the DevID is 0x0301. Checking `http://www.linux-usb.org/usb.ids`, there seems to be a cold device.

Anyway, I'll add the warning.

plugins/usbdmx/UsbDmxPlugin.h
CreaValix: Ok

plugins/usbdmx/SyncPluginImpl.cpp
CreaValix: Here is what I found: Nicolaudie in France seems to be the ventor. They have this brand called sunlite. See here for example: `https://pdf.archiexpo.com/pdf/nicolaudie-europe/catalog-developer-kits-siudi-6c/54341-176345.html` and here `https://www.audiolight.sg/product-brand/sunlite/`.

It seems that firmware development was done by `LightingSoft AG` in Switzerland. I think this an incorrect entry in the usb ID list.

Long story short: It is a „Sunlite Intelligent Usb Dmx Interface“, short SIUDI made by Nicolaudie.

SunliteSIUDI would be bad as the S in SIUDI already stands for Sunlite. Better NicolaudieSIUDI.

plugins/usbdmx/Siudi.cpp
```diff
@@ -0,0 +1,155 @@
...
+    return false;
+  }
+
+  // Read device info. This call takes about 270 ms.
+  // Discard the buffer as the format is currently unknown.
+  uint8_t buf[64];
+  int ret = libusb_control_transfer(m_usb_handle,
+                                LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_ENDPOINT_IN,
+                                0x3f, 0x00c4, 1, buf, 64, 500);
+  if (ret != 64) {
```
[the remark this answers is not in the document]
CreaValix: Of course. I'll add a clean Wireshark USB monitor file with the original software communication later.

plugins/usbdmx/Siudi.cpp
```diff
@@ -0,0 +1,155 @@
...
+  if (!ThreadedUsbSender::Start()) {
+    return false;
+  }
+
+  // Read device info. This call takes about 270 ms.
+  // Discard the buffer as the format is currently unknown.
+  uint8_t buf[64];
+  int ret = libusb_control_transfer(m_usb_handle,
+                                LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_ENDPOINT_IN,
+                                0x3f, 0x00c4, 1, buf, 64, 500);
```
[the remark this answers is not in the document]
CreaValix: Ok

CreaValix:
> I think we want about 22ms for 44Hz.
> 
> Do you know if the device is just sending the data directly (although there's no start code), or does it retransmit on its own if you don't send a USB packet?
> 
> Approximately what sort of frequency is too low, and is the behaviour identical with different types of lights?

The device has a built-in LED. I suppose it resembles TX. The device starts sending as soon as I power it on, the LED starts flickering, similar to a Ethernet link with active traffic.

If I start the original software, it continues to flicker, but a little bit slower. If I stop it, the LED stops flickering for maybe half a second, until it flickers again, just like before the software start. DMX channel stay at the last set value.

I made some experiments with Python. When I send a single package, the LED will turn off and effects stay off (or will go off). Roughly 400 ms later, it will start flickering and effects get the values of this package.

When I send the same package twice, the LED will stay off or on (random, but not flicker), but effects get the values immediately. Again ~400 ms later, the LED will flicker and lights will stay on.

I am pretty sure now that one USB write will result in a single DMX TX. The firmware enters a USB streaming mode after two endpoint writes. After missing USB traffic for ~400 ms it will revert to a failure state and repeat the last package.

I also played with the `usleep()` parameter in `TransmitBuffer()`. Below 200 µS effects will start flickering. The USB response time is 32 ms then. I could also measure 32 pkt/s.
If I `usleep()` 30 ms, I get the same frequency, but only 2 ms USB response time. So `usleep(30000)` seems to be a good choice.

It looks like the firmware delays USB responses until an ongoing DMX TX is finished. Probably it just disables USB interrupts. To be sure, I set one effect to channel 512 to assure that no DMX packet is „cut“ by a following USB transmission.

So it seems that this device needs 32 ms for a full DMX package, resulting in 32 Hz frequency at best.

CPU load is not an issue, ola is somewhere at 1 % at this configuration.

I found a Windows API [here](https://github.com/acarreras/ofxDmxSuidi) with a [linked developer kit to download](https://storage.googleapis.com/nicolaudie-eu-software/dmxsoft/siudi_usb_developerkit.exe). Here is the [Developper Kit.pdf](https://github.com/OpenLightingProject/ola/files/12185877/Developper.Kit.pdf) from inside this package.
There, they also mention that the device will go to sleep after 5 seconds. However, I found different values.

By changing some parameters in the Windows API, we could find out some of the USB values...

plugins/usbdmx/Siudi.cpp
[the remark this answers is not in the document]
CreaValix: [usb_sniff_SUIDI-6_slmini.pcapng.gz](https://github.com/OpenLightingProject/ola/files/12188001/usb_sniff_SUIDI-6_slmini.pcapng.gz)
Here is the USB sniff using the demo application slmini.exe from the developer kit. I left comments on packages #1, 23 and 65.
Package 67 is the only message sent to endpoint 0.

I then made a 2nd sniff where I started the MyDMX and slmini.exe applications many times. Sometimes, wIndex is 1, but mostly it is 0.
wValue changes between Windows restarts and partly random values. slmini starts with 0x55 or 0x56 whereas MyDMX always starts with 0x00. I could just be uninitialized memory of the library.
[usb_sniff_SIUDI-6_application_starts.pcapng.gz](https://github.com/OpenLightingProject/ola/files/12188136/usb_sniff_SIUDI-6_application_starts.pcapng.gz)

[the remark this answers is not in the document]
CreaValix: image
image
Here are two screenshots of MyDMX and that `slmini.exe`. Both programs show a firmware version of 200 or 2.00. MyDMX also detects an empty serial number.

I assume these come from the API `DHC_VERSION` and `DHC_SERIAL` calls.

I could not find the version number (0xc8 or 0x02) in the URB_CONTROL call. But `lsusb` shows me this:
`  bcdDevice            0.c8`
[lsusb_siudi-6.txt](https://github.com/OpenLightingProject/ola/files/12188231/lsusb_siudi-6.txt)

About the serial, I have no idea where it comes from. Perhaps it is the zeroes (in ASCII) at the end of the URB_CONTROL response.

CreaValix:
> Amazing thanks @CreaValix !
> 
> Please add yourself to Authors: https://github.com/OpenLightingProject/ola/blob/master/AUTHORS
> 
> And the device to udev (assuming that's required): https://github.com/OpenLightingProject/ola/blob/master/debian/ola.udev

Done [abcddc1](https://github.com/OpenLightingProject/ola/pull/1887/commits/abcddc15de2aca4c0915825edea2f7bd61e3c9d2)

plugins/usbdmx/SiudiFactory.cpp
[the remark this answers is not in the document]
CreaValix: I changed the code to also accept the other SIUDI6 variants (C, A and D, see USB IDs) here [75d5f56](https://github.com/OpenLightingProject/ola/pull/1887/commits/75d5f5670e5e2db74cdd1aaf9c363b46a0920054). 

In the udev rules, I captured the whole 030* range. Without the udev rule for the cold device, the factory would not even be called due to missing permissions.

plugins/usbdmx/Siudi.cpp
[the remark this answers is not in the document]
CreaValix: I used `ola::strings::FormatData` to log out this package in debug level here fd056b5b64dc0be25e685c0a7e8f319062648ae4. I introduced constants and changed `wIndex` (see above) to 0x0000 here 768def2d844b9cd0739ed302310aa48e207d9a17. I still get the same result.

plugins/usbdmx/Siudi.cpp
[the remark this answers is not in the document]
CreaValix: Done 768def2d844b9cd0739ed302310aa48e207d9a17

plugins/usbdmx/Siudi.cpp
CreaValix: Due to the fact that I always have to send 512 bytes, I could only use `GetRaw` when `DmxBuffer.Size()` was 512. If it differs, I used `DmxBuffer.GetRange()` to copy 512 or less bytes. If it is less than 512, the remaining bytes are `memset` to 0x00.

See 96e10f70e7d02ce70151973234a2cd4a310aeebf and a fixup in 42b23c745c3ef089f3543c4a098e51c9cd791af1.

CreaValix:
> So it seems that this device needs 32 ms for a full DMX package, resulting in 32 Hz frequency at best.
> 
Long story short: I set the inter-bulk-transfer delay to 30 ms and the USB timeout to 10 ms. This should be fine, assuming that the USB response comes in after ~2 ms. See 8e48a0a3971ef78e1a04bdbb8fc70ce934a76840

plugins/usbdmx/SiudiFactory.cpp
```diff
@@ -0,0 +1,68 @@
...
+    return false;
+  }
+  if (descriptor.idProduct == SIUDI6C_HOT_ID ||
+      descriptor.idProduct == SIUDI6A_HOT_ID ||
+      descriptor.idProduct == SIUDI6D_HOT_ID) {
+    if (descriptor.idProduct == SIUDI6C_HOT_ID) {
+      OLA_INFO << "Found a new Nicoleaudie SIUDI-6C device";
+    } else if (descriptor.idProduct == SIUDI6A_HOT_ID) {
+      OLA_INFO << "Found a new Nicoleaudie SIUDI-6A device";
+    } else if (descriptor.idProduct == SIUDI6A_HOT_ID) {
```
kengruven, line 57: This should check for 6D, right?

CreaValix: Yes. Fixed in 60bc8febe

plugins/usbdmx/Siudi.cpp
CreaValix: Now using `ola::DMX_UNIVERSE_SIZE` in 96e10f70e7d02ce70151973234a2cd4a310aeebf

plugins/usbdmx/SyncPluginImpl.cpp
CreaValix: I left it as `Siudi.cpp` as other drivers seem to have only the product name in their file names.

PR #1888, opened 2023-08-03
Add a Eurolite serial number check

aroffringa:
This MR adds an option 'eurolite_serials' to the usb dmx plugin. This option accepts a comma separated list of serial numbers that are assumed to be Eurolite mk2 devices (as long as their product and vendor ids also match).

This allows using the Eurolite MK2 without enabling the 'enable_eurolite_mk2' option. The advantage is that is avoids the conflicts that 'enable_eurolite_mk2' produces with other plugins. These
conflicts make it impossible to connect the mk2 simultaneously together with most other interfaces to have multiple universes and/or input universes.

I have tested this by connecting an Enttec pro together with the Eurolite Mk2, and this works.

I still need to update documentation -- happy to finish that too if this approach is considered acceptable.


plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -46,12 +48,16 @@ const uint16_t EuroliteProFactory::VENDOR_ID_MK2 = 0x0403;
...
+const char EuroliteProFactory::EUROLITE_SERIALS_KEY[] =
+    "eurolite_serials";
 
 EuroliteProFactory::EuroliteProFactory(ola::usb::LibUsbAdaptor *adaptor,
                                        Preferences *preferences)
   : BaseWidgetFactory<class EurolitePro>("EuroliteProFactory"),
     m_adaptor(adaptor),
     m_enable_eurolite_mk2(IsEuroliteMk2Enabled(preferences)) {
+  StringSplit(preferences->GetValue(EUROLITE_SERIALS_KEY),
+              m_expected_eurolite_serials, ",");
```
peternewman, line 60: I'd probably go for multiple eurolite_serial lines, like this:
https://docs.openlighting.org/ola/conf/ola-nanoleaf.conf.html

As it's easier to manage each individual one or comment them out or whatever, and we don't actually care about the order of them (unlike we might elsewhere).

Then you'll need:
https://docs.openlighting.org/ola/doc/latest/classola_1_1_preferences.html#aa4bfa8a5218cfd23c573d23a52bdc88b

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -46,12 +48,16 @@ const uint16_t EuroliteProFactory::VENDOR_ID_MK2 = 0x0403;
 
 const char EuroliteProFactory::ENABLE_EUROLITE_MK2_KEY[] =
     "enable_eurolite_mk2";
+const char EuroliteProFactory::EUROLITE_SERIALS_KEY[] =
+    "eurolite_serials";
```
peternewman, line 52: At which point this would be singular:
```suggestion
const char EuroliteProFactory::EUROLITE_SERIAL_KEY[] =
    "eurolite_serial";
```

But I think should also mention the mk2:
```suggestion
const char EuroliteProFactory::EUROLITE_MK2_SERIAL_KEY[] =
    "eurolite_mk2_serial";
```

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -104,9 +116,12 @@ bool EuroliteProFactory::DeviceAdded(
...
     } else {
-      OLA_INFO << "Connected FTDI device could be a Eurolite "
-               << "USB-DMX512-PRO MK2 but was ignored, because "
-               << ENABLE_EUROLITE_MK2_KEY << " was false.";
+      OLA_INFO << "Connected FTDI device with serial " << info.serial
+               << " could be a Eurolite USB-DMX512-PRO MK2 but was "
+               << "ignored, because "
+               << ENABLE_EUROLITE_MK2_KEY << " was false and "
+               << "its serial number was not in the "
+               << "configuration.";
```
peternewman, line 124: Maybe something like this is a bit clearer?
```suggestion
               << "its serial number was not listed "
               << "specifically in the configuration.";
```

Or better yet tell them which key they can add it under!

plugins/usbdmx/EuroliteProFactory.h
```diff
@@ -46,10 +49,12 @@ class EuroliteProFactory : public BaseWidgetFactory<class EurolitePro> {
   static bool IsEuroliteMk2Enabled(Preferences *preferences);
 
   static const char ENABLE_EUROLITE_MK2_KEY[];
+  static const char EUROLITE_SERIALS_KEY[];
```
peternewman, line 52:
```suggestion
  static const char EUROLITE_MK2_SERIAL_KEY[];
```

plugins/usbdmx/EuroliteProFactory.h
```diff
@@ -46,10 +49,12 @@ class EuroliteProFactory : public BaseWidgetFactory<class EurolitePro> {
...
 
  private:
   ola::usb::LibUsbAdaptor *m_adaptor;
   bool m_enable_eurolite_mk2;
+  std::vector<std::string> m_expected_eurolite_serials;
```
peternewman, line 57:
```suggestion
  std::vector<std::string> m_expected_eurolite_mk2_serials;
```

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -88,12 +94,18 @@ bool EuroliteProFactory::DeviceAdded(
...
+    LibUsbAdaptor::DeviceInformation info;
+    if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
+      return false;
+    }
+
+    const bool serial_matches = std::find(m_expected_eurolite_serials.begin(),
+      m_expected_eurolite_serials.end(), info.serial) !=
+      m_expected_eurolite_serials.end();
+
+    if (m_enable_eurolite_mk2 || serial_matches) {
```
peternewman, line 106: You can just use our STLContains function here:
https://docs.openlighting.org/ola/doc/latest/group__stl.html#ga6bc94dfb62cab831bfa2f0fc7e4daa04

peternewman (changes requested):
Thanks @aroffringa .

Yeah I think this is an acceptable compromise, its just a bit niche to have bothered doing it before unless you're in the position of having both (the benefits of open source).

I think there's just the one main suggested change, and then just look at the logging and readme but you've covered most of that to make it nice and easy,

plugins/usbdmx/EuroliteProFactory.h
aroffringa: Changed to a std::set and renamed.

plugins/usbdmx/EuroliteProFactory.cpp
aroffringa: Thanks for pointing that out.

aroffringa:
@peternewman I've adopted your comments and added the config key to the usbdmx readme.

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -46,12 +48,27 @@ const uint16_t EuroliteProFactory::VENDOR_ID_MK2 = 0x0403;
...
+const char EuroliteProFactory::EUROLITE_MK2_SERIAL_KEY[] =
+    "eurolite_mk2_serial";
 
 EuroliteProFactory::EuroliteProFactory(ola::usb::LibUsbAdaptor *adaptor,
                                        Preferences *preferences)
   : BaseWidgetFactory<class EurolitePro>("EuroliteProFactory"),
     m_adaptor(adaptor),
     m_enable_eurolite_mk2(IsEuroliteMk2Enabled(preferences)) {
+  const std::vector<std::string> serials =
+    preferences->GetMultipleValue(EUROLITE_MK2_SERIAL_KEY);
```
peternewman, line 60: Indent further as its not an if statement or similar
```suggestion
      preferences->GetMultipleValue(EUROLITE_MK2_SERIAL_KEY);
```

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -88,12 +105,17 @@ bool EuroliteProFactory::DeviceAdded(
...
+    if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
+      return false;
+    }
+
+    const bool serial_matches =
+      STLContains(m_expected_eurolite_mk2_serials, info.serial);
+
+    if (m_enable_eurolite_mk2 || serial_matches) {
+      OLA_INFO << "Found a possible new Eurolite USB-DMX512-PRO MK2 device "
+                  "with serial " << info.serial;
```
peternewman, line 118: I don't know if it's worth logging two different messages here (especially given we've already done the "hard" work of checking for a match), then we could see from the log how it had matched (general or specific serial).

peternewman (changes requested):
Some minor wording tweaks (and a possible extension idea).

It might be good to set a default (blank) serial key here:
https://github.com/OpenLightingProject/ola/blob/96aa95e1ced41ab8093ffaff619ec3b875f2384e/plugins/usbdmx/UsbDmxPlugin.cpp#L95-L115

Although it would then throw the empty line error each time...

I don't know what you think about this (maybe as a future enhancement for another PR):
1. Start up OLAd in enable_eurolite_mk2
2. If no (non-blank) eurolite_mk2_serial set (at startup), it adds each found one to the list
3. So workflow wise you could plug in your Eurolite dongles, turn on the setting, run up OLA. Turn off the setting and plug in your other devices.
4. Upsides you wouldn't have to read logs or find serial numbers via other methods
5. Downsides some potential confusion if you don't follow the right workflow
6. Alternative suggestion is do the same thing but put them in another key (e.g. discovered_eurolite_mk2_serial) which you could rename if required.

FYI you can merge the suggestions straight in via the web UI if you want to:

plugins/usbdmx/EuroliteProFactory.cpp
aroffringa: Sure, done, have a check if this is what you mean...

aroffringa:
I've added the blank default and made sure that if a single blank serial number is given no warning is produced. I'm not sure I understand your suggestion, do you want ola to rewrite the config file and add the serial numbers? As a user that's really not what I would expect ola to do, to be honest, but I might be misunderstanding you.

aroffringa:
@peternewman is this MR ready to merge?

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -104,9 +137,12 @@ bool EuroliteProFactory::DeviceAdded(
...
-               << "USB-DMX512-PRO MK2 but was ignored, because "
-               << ENABLE_EUROLITE_MK2_KEY << " was false.";
+      OLA_INFO << "Connected FTDI device with serial " << info.serial
+               << " could be a Eurolite USB-DMX512-PRO MK2 but was "
+               << "ignored, because "
+               << ENABLE_EUROLITE_MK2_KEY << " was false and "
+               << "its serial number was not specifically specified with "
+               << EUROLITE_MK2_SERIAL_KEY;
       return false;
     }
```
peternewman, line 147: We need to fix the code a few lines below here, where it then ignores our special serial number and uses the device and bus number, for the MK2 at least, as that should mean the patch persists even when you plug it into a different USB port, which it won't do currently.

You should be able to see that either way as the serial number should show in the device name/description in OLA.

Is it true that ALL MK2's have a serial number (as they're using an FTDI chip?)?

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -46,12 +48,27 @@ const uint16_t EuroliteProFactory::VENDOR_ID_MK2 = 0x0403;
...
+    "eurolite_mk2_serial";
 
 EuroliteProFactory::EuroliteProFactory(ola::usb::LibUsbAdaptor *adaptor,
                                        Preferences *preferences)
   : BaseWidgetFactory<class EurolitePro>("EuroliteProFactory"),
     m_adaptor(adaptor),
     m_enable_eurolite_mk2(IsEuroliteMk2Enabled(preferences)) {
+  const std::vector<std::string> serials =
+    preferences->GetMultipleValue(EUROLITE_MK2_SERIAL_KEY);
+  for (const std::string& serial : serials) {
```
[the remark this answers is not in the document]
peternewman: Yeah noted, I'd probably rather do it en-mass than piecemeal

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -46,12 +48,34 @@ const uint16_t EuroliteProFactory::VENDOR_ID_MK2 = 0x0403;
...
+  // A single empty string is considered the same as specifying
+  // no serial numbers. This is useful as a default value.
+  const bool has_default_value =
+      serials.size() == 1 && serials[0].empty();
+  if (!has_default_value) {
+    for (std::vector<std::string>::const_iterator iter = serials.begin();
+         iter != serials.end(); ++iter) {
+      if (iter->empty()) {
+        OLA_WARN << EUROLITE_MK2_SERIAL_KEY
+                 << " requires a serial key string, but it is empty.";
```
peternewman, line 70: Key is just our internal term for the key part of the key-value setting.
```suggestion
                 << " requires a serial number, but it is empty.";
```

plugins/usbdmx/README.md
```diff
@@ -20,25 +20,36 @@ This plugin supports various USB DMX devices including:
...
 0 = No logging, 4 = Verbose debug.
 
 `enable_eurolite_mk2 = {false,true}`
 Whether to enable detection of the Eurolite USB-DMX512 PRO MK2.
-Default = false. This device is indistinguishable from other devices
+Default = `false`. This device is indistinguishable from other devices
 with an FTDI chip, and is therefore disabled by default. When enabled,
 this plugin will conflict with the usbserial, StageProfi and FTDI USB DMX
-plugins.
+plugins. If this is undesirable, the `eurolite_mk2_serial` key can be
```
peternewman, line 32: Again we can hide the complex terminology
```suggestion
plugins. If this is undesirable, the `eurolite_mk2_serial` setting can be
```

plugins/usbdmx/README.md
```diff
@@ -20,25 +20,36 @@ This plugin supports various USB DMX devices including:
...
+PRO MK2.
 
-`nodle-<serial>-mode = {0,1,2,3,4,5,6,7}`  
+`eurolite_mk2_serial = <serial>`
+Claim the usb device with the given serial number as a Eurolite USB-DMX512
+PRO MK2 even when `enable_eurolite_mk2 = false`. This makes it possible
+to use the Eurolite USB-DMX512 PRO MK2 together with other devices that
+can not be distinguished otherwise. This key has no effect when
+`enable_eurolite_mk2 = true` or no device is connected with the given
+serial. The key may be specified multiple times to use multiple Eurolite
```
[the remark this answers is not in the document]
peternewman: Updated or pending elsewhere

plugins/usbdmx/README.md
```diff
@@ -20,25 +20,36 @@ This plugin supports various USB DMX devices including:
...
 with an FTDI chip, and is therefore disabled by default. When enabled,
 this plugin will conflict with the usbserial, StageProfi and FTDI USB DMX
-plugins.
+plugins. If this is undesirable, the `eurolite_mk2_serial` key can be
+used instead, which manually associates a usb device as Eurolite USB-DMX512
+PRO MK2.
 
-`nodle-<serial>-mode = {0,1,2,3,4,5,6,7}`  
+`eurolite_mk2_serial = <serial>`
+Claim the usb device with the given serial number as a Eurolite USB-DMX512
```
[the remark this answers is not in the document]
peternewman: Done.

plugins/usbdmx/README.md
```diff
@@ -20,25 +20,36 @@ This plugin supports various USB DMX devices including:
...
 
 `enable_eurolite_mk2 = {false,true}`
 Whether to enable detection of the Eurolite USB-DMX512 PRO MK2.
-Default = false. This device is indistinguishable from other devices
+Default = `false`. This device is indistinguishable from other devices
 with an FTDI chip, and is therefore disabled by default. When enabled,
 this plugin will conflict with the usbserial, StageProfi and FTDI USB DMX
-plugins.
+plugins. If this is undesirable, the `eurolite_mk2_serial` key can be
+used instead, which manually associates a usb device as Eurolite USB-DMX512
```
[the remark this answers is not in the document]
peternewman: Done.

plugins/usbdmx/EuroliteProFactory.cpp
peternewman: Looks good thanks

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -46,12 +48,27 @@ const uint16_t EuroliteProFactory::VENDOR_ID_MK2 = 0x0403;
...
                                        Preferences *preferences)
   : BaseWidgetFactory<class EurolitePro>("EuroliteProFactory"),
     m_adaptor(adaptor),
     m_enable_eurolite_mk2(IsEuroliteMk2Enabled(preferences)) {
+  const std::vector<std::string> serials =
+    preferences->GetMultipleValue(EUROLITE_MK2_SERIAL_KEY);
+  for (const std::string& serial : serials) {
+    if (serial.empty()) {
+      OLA_WARN << EUROLITE_MK2_SERIAL_KEY
+               << " requires a serial key string, but is empty.\n";
```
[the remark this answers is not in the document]
peternewman: Done

plugins/usbdmx/EuroliteProFactory.cpp
peternewman: Done thanks

plugins/usbdmx/EuroliteProFactory.h
peternewman: Done thanks

plugins/usbdmx/EuroliteProFactory.cpp
peternewman: Done thanks

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -88,13 +112,22 @@ bool EuroliteProFactory::DeviceAdded(
...
+    if (!m_adaptor->GetDeviceInfo(usb_device, descriptor, &info)) {
+      return false;
+    }
+
+    const bool serial_matches =
+      STLContains(m_expected_eurolite_mk2_serials, info.serial);
 
+    if (m_enable_eurolite_mk2 || serial_matches) {
+      if (serial_matches) {
+        OLA_INFO << "Enabling new Eurolite USB-DMX512-PRO MK2 device "
```
peternewman, line 125: We should probably be slightly pessimistic here, as the manufacturer and product name checks could still fail, e.g. "Found a probable..."

peternewman (changes requested):
Sorry for the big delay in getting back to you! This wasn't raised from an issue was it @aroffringa that we ought to link it to?

Quite a bit of this is just a few minor style things, feel free to just merge my suggestions as is for them. Please can you also resync with master.

Just to check, when you specify by serial and have one or more of FTDI, Stageprofi or USB Pro enabled, does it all play nicely, or does ConflictsWith need tweaking still? USB Pro would be fine, as there's two way comms and the Eurolite won't behave as a USB Pro (plus you said you've tested it), I can't remember about Stageprofi (so I probably need to check the code), but won't FTDI just happily start spewing DMX data at it when it finds it, which would just make it a race condition which plugin detects it first?

peternewman:
> I've added the blank default and made sure that if a single blank serial number is given no warning is produced.

Great thanks!

> I'm not sure I understand your suggestion, do you want ola to rewrite the config file and add the serial numbers? As a user that's really not what I would expect ola to do, to be honest, but I might be misunderstanding you.

I was just trying to think through how someone will use it. If like you I've got one Eurolite Mk2 and one Enttec, I find I can't use them both together unless I enable the serial number mode, but how do I find the serial number? If what's printed on the bottom of the unit is exactly what I need to type in, then perfect, but otherwise I was thinking OLA could effectively pull either any serials it finds, or the serials it found in the last run, into the config file, for you to copy and paste. Otherwise you need to either use lsusb or trawl through all the OLA logs or similar. But you're the one using the kit, does that make sense as a workflow or seem redundant? It can definitely be a future PR/open issue and certainly isn't a blocker to getting this merged!

plugins/usbdmx/EuroliteProFactory.cpp
aroffringa: I'm not sure what you mean, can you propose the fix?

aroffringa: BTW I suspect all MK2's have a different serial number indeed, they indeed use an FTDI chip.

aroffringa:
> I was just trying to think through how someone will use it. If like you I've got one Eurolite Mk2 and one Enttec, I find I can't use them both together unless I enable the serial number mode, but how do I find the serial number? If what's printed on the bottom of the unit is exactly what I need to type in, then perfect, but otherwise I was thinking OLA could effectively pull either any serials it finds, or the serials it found in the last run, into the config file, for you to copy and paste. Otherwise you need to either use lsusb or trawl through all the OLA logs or similar. But you're the one using the kit, does that make sense as a workflow or seem redundant? It can definitely be a future PR/open issue and certainly isn't a blocker to getting this merged!

Someone would now indeed have to look at the OLA output or lsusb to find the serial number -- this is indeed not super user friendly. What would be really useful is if this could be configured on the web interface (including that the interface would show the serial numbers), but that's a whole different scope ;). What's currently implemented at least demonstrates the use of the serial number to solve the conflict. In fact, given that quite some modules conflict with each other, a future idea could even be to incorporate this in a more generic way so that the user would be shown the possible serial ids and select the right device for it.

For now I wouldn't want to enlarge the scope of this pull request — as it is it seems to work.

I again applied your suggestions and updated the branch to master, so I think it is ready to be merged.

plugins/usbdmx/EuroliteProFactory.cpp
```diff
@@ -104,9 +137,12 @@ bool EuroliteProFactory::DeviceAdded(
...
       is_mk2 = true;
     } else {
-      OLA_INFO << "Connected FTDI device could be a Eurolite "
-               << "USB-DMX512-PRO MK2 but was ignored, because "
-               << ENABLE_EUROLITE_MK2_KEY << " was false.";
+      OLA_INFO << "Connected FTDI device with serial " << info.serial
+               << " could be a Eurolite USB-DMX512-PRO MK2 but was "
+               << "ignored, because "
+               << ENABLE_EUROLITE_MK2_KEY << " was false and "
+               << "its serial number was not listed specifically with "
```
peternewman, line 144: Minor grammar tweak:
```suggestion
               << "its serial number was not listed specifically in "
```

plugins/usbdmx/EuroliteProFactory.cpp
peternewman: I'll do another PR for your to test. It just improves what this has, so this can be merged as is.

peternewman (approved):
LGTM thanks for all the hard work @aroffringa and sorry it's taken me quite so long to get this reviewed and in!

peternewman:
I think outstanding tasks are as follows:

For you to test:
> Just to check, when you specify by serial and have one or more of FTDI, Stageprofi or USB Pro enabled, does it all play nicely, or does ConflictsWith need tweaking still? USB Pro would be fine, as there's two way comms and the Eurolite won't behave as a USB Pro (plus you said you've tested it), I can't remember about Stageprofi (so I probably need to check the code), but won't FTDI just happily start spewing DMX data at it when it finds it, which would just make it a race condition which plugin detects it first?

For me to do and then for you to test:
> We need to fix the code a few lines below here, where it then ignores our special serial number and uses the device and bus number, for the MK2 at least, as that should mean the patch persists even when you plug it into a different USB port, which it won't do currently.
> 
> You should be able to see that either way as the serial number should show in the device name/description in OLA.

I'll open a new issue to track potential simplification for the user of finding serial numbers to put in the config.

plugins/usbdmx/EuroliteProFactory.cpp
peternewman: Opened #1932 for this so will track it there.

peternewman:
> @peternewman I enabled the stageprofi, opendmx and ftdi plugins one by one and checked if the Eurolite continues to work. The serial plugin (that's what you mean with usb pro enabled?) I have indeed already tested extensively.

Yes sorry, serial=usbpro.

> Conclusion: the Stageprofi and opendmx plugins are not an issue, but enabling the FTDI device causes the Eurolite to no longer work. It's somewhat weird because the Eurolite is still detected and patched, but it seems that the FTDI device also tries to claim it and that causes issues.

Yeah as I feared, they discover in different ways (via libusb and libftdi) so I don't think they can share the file locking stuff, so it just becomes a free for all, and the FTDI plugin will just pump DMX packets at any DMX interface it finds. Opendmx works similar, but needs a kernel driver so is more niche and I think if the device got detected as that it wouldn't be found via libusb and vice versa so we're probably okay there (but it may be safe to disable it too)...

I assume the StageProfi plugin did detect but then ignored the device? Looking at the code, there is some two way comms to establish it really is a StageProfi so I think we're safe there.

Anyway would you mind doing another minor PR please that makes it so that only StageProfi and serial/usbpro can avoid conflicting when the specific serial setting is on (i.e. ensure the FTDI and OpenDMX continue to conflict and they can't break as your test just found)?

So some other lines around here:
https://github.com/aroffringa/ola/blob/ee8796219dd55a50cbbb6360a9804797fbc0b07d/plugins/usbdmx/UsbDmxPlugin.cpp#L122-L129

PR #1890, opened 2023-08-03
Replace a few more deprecated pre-C++11 styles

aroffringa:
The functions std::binary_function and std::mem_fun have been deprecated in C++11, and are removed in C++20. This pull request replaces them with more modern code. Together with #1889 this removes all warnings on recent compilers. This one is fairly small.


DaAwesomeP (commented):
I'm not really familiar with `std::binary_function` so I can't speak to those. I suppose we would see compile errors if it broke any typing.

Well I guess there isn't any more fun in the code! (sarcasm)

kripton:
Wow, the syntax in StringUtils.cpp needs a bit getting used to. On the other hand, what we had before also was not that easy to read. I'd say: As long as this doesn't break stuff, reducing compiler warnings is a good thing!

PR #1898, opened 2023-09-07
Add the ability to set a universe to blackout instead via ola_set_dmx

peternewman:


kripton (approved):
Thanks, looking good!

PR #1901, opened 2023-09-18
Initial IPv6 classes

peternewman:


common/network/IPV6Address.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+using std::string;
+
+IPV6Address::IPV6Address(const uint8_t *address) {
+  // TODO(Peter): Deal with any network byte order conversion?
+  memcpy(&m_address.s6_addr[0], address, sizeof (struct in6_addr));
+}
+
+bool IPV6Address::operator<(const IPV6Address &other) const {
+  // TODO(Peter): Deal with any network byte order conversion?
```
kripton, line 57: can this TODO comment be removed? The next line seems to clarify it?

common/network/IPV6Address.cpp
```diff
@@ -0,0 +1,172 @@
...
+  // Stored in network byte order, so convert to sort appropriately
+  return (memcmp(&m_address.s6_addr[0],
+          &other.m_address.s6_addr[0],
+          sizeof (struct in6_addr)) < 0);
+}
+
+bool IPV6Address::operator>(const IPV6Address &other) const {
+  // TODO(Peter): Deal with any network byte order conversion?
```
kripton, line 65: can this TODO comment be removed? The next line seems to clarify it?

kripton (approved):
Looks good to me, thanks!

common/network/IPV6Address.cpp
peternewman: TBH the next line is a copy from the IPv4 function. Maybe I should make it more clear the answer is unknown.

Currently these classes are entirely standalone, I've only shuffled bytes into them in the order I've chosen and got the results I expected, the real test (which needs more code) will be to actually read something from a machine (or over the network) and confirm we still get the right result. And then repeat on a BE machine...

common/network/IPV6Address.cpp
peternewman: As above

common/network/IPV6Address.cpp
peternewman: Actually it's worse than that, I'll just drop the invalid comment on the line below for now:
https://github.com/OpenLightingProject/ola/blob/87b1318ded6947d78be62ee8fda429305558f94b/common/network/IPV4Address.cpp#L58-L66

peternewman: Okay false comment now removed, I'll get this one merged as there are related bits backing up behind it, can/will review/reconsider when it does something more than just shuffling bytes in and out of other functions...

common/network/IPV6Address.cpp
peternewman: As above again.

PR #1904, opened 2023-09-19
Do some minor formatting tweaks

peternewman:
To reduce the diff before lots more significant changes appear...

View it here to simplify it further:
https://github.com/OpenLightingProject/ola/pull/1904/files?w=1


kripton (approved):
Very good. I dislike leaving out the curly braces on one-instruction conditions. Took me a while to review since when I first looked at it, I had the impression you changed the logic at one location. But looking at it closer now, it's all harmless changes :+1:

peternewman:
> Very good. I dislike leaving out the curly braces on one-instruction conditions.

Yeah I've been caught out with it a few times before hence trying to remove them.

> Took me a while to review since when I first looked at it, I had the impression you changed the logic at one location. But looking at it closer now, it's all harmless changes 👍

Ah that's odd. Anyway glad you're happy!

PR #1914, opened 2023-10-19
Fix spelling/grammar in RPC error message

kengruven:


peternewman (approved):
Good spot thanks @kengruven !

PR #1920, opened 2023-10-31
Add protobuf syntax version

kengruven:
The protocol buffer spec suggests that this is required. I've tried two protobuf libraries, and one errors without this, and the other prints a warning.

There's no documentation in OLA I can find that it's proto2, but it works this way, and it makes sense.  Both files use "required", which was removed from proto3.


kripton (approved):
Looks reasonable, thanks! @peternewman ?

PR #1926, opened 2023-12-03
Add support for the IPv6 data type to the RDM messaging

peternewman:
Plus the additional E1.33 and E1.37-7 PIDs that enables

### DIFF
--- a/common/messaging/SchemaPrinterTest.cpp
+++ b/common/messaging/SchemaPrinterTest.cpp
@@ -33,6 +33,7 @@ using std::vector;
 
 using ola::messaging::BoolFieldDescriptor;
 using ola::messaging::IPV4FieldDescriptor;
+using ola::messaging::IPV6FieldDescriptor;
 using ola::messaging::MACFieldDescriptor;
 using ola::messaging::Descriptor;
 using ola::messaging::FieldDescriptor;
@@ -85,6 +86,8 @@ void SchemaPrinterTest::testPrinter() {
       "Count", false, 10);
   IPV4FieldDescriptor *ipv4_descriptor = new IPV4FieldDescriptor(
       "Address");
+  IPV6FieldDescriptor *ipv6_descriptor = new IPV6FieldDescriptor(
+      "v6 Address");
   MACFieldDescriptor *mac_descriptor = new MACFieldDescriptor(
       "MAC Address");
   UIDFieldDescriptor *uid_descriptor = new UIDFieldDescriptor("Device");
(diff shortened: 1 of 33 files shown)


common/messaging/SchemaPrinterTest.cpp
```diff
@@ -85,6 +86,8 @@ void SchemaPrinterTest::testPrinter() {
       "Count", false, 10);
   IPV4FieldDescriptor *ipv4_descriptor = new IPV4FieldDescriptor(
       "Address");
+  IPV6FieldDescriptor *ipv6_descriptor = new IPV6FieldDescriptor(
+      "v6 Address");
```
kripton, line 90: "v6" or "IPv6"?

include/ola/messaging/MessagePrinter.h
```diff
@@ -40,6 +40,7 @@ class MessagePrinter: public MessageVisitor {
 
     virtual void Visit(const BoolMessageField*) {}
     virtual void Visit(const IPV4MessageField*) {}
+    virtual void Visit(const IPV6MessageField*) {}
```
kripton, line 43: Any reason to have "IPV6MessageField" instead of "IPv6MessageField"? I mean, sure, IPV4 has been here before but IPv4 and IPv6 are easier to read then IPV4 and IPV6

kripton (approved):
Thanks, looking good. The only thing is IPV4Message and IPV6Message vs. IPv4Message and IPv6Message for better readability. But perfectly fine otherwise!

common/messaging/SchemaPrinterTest.cpp
peternewman: It's just a test file, I stuck with the ambiguous one mirrored from the line before. You'll see further down it's how it renders those fields (this is effectively the name) so also good to confirm we've not got name and type mixed up.

include/ola/messaging/MessagePrinter.h
peternewman: Just legacy/compatibility for now really, we've generally gone for either CamelCase or being consistently upper-cased, and IPVx tracks through a lot of other code too, e.g. IPV4SocketAddress

I find DmxBuffer more jarring personally!

Also we should probably prioritise fixing this, which renders out user-facing code before worrying too much about our internal names:
https://github.com/OpenLightingProject/ola/blob/653449120b2d5630edddda123c480880fe324d4f/common/utils/StringUtils.cpp#L437-L453

Although at least C++ has it, the Python doesn't have anything from memory. Obviously ideally there would be one place, that could ideally handle the more complicated cases, for this to be managed. probably via a shell script that generates an include file or something.

kripton: Sure, IPV4SocketAddress is fine for me (as is DmxBuffer :D). It was just for better readability since IPv4 and IPv6 are much more commonly used.

PR #1932, opened 2024-02-03
Track Eurolite MK2 by actual serial number where available

peternewman:
@aroffringa can you give this a test as mentioned in #1888 please.

This should mean the patch persists even when you plug it into a different USB port, which it won't do currently.

You should be able to see that either way as the serial number should show in the device name/description in OLA.

_Originally posted by @peternewman in https://github.com/OpenLightingProject/ola/pull/1888#discussion_r1404340274_

### DIFF
--- a/plugins/usbdmx/AVLdiyD512Factory.cpp
+++ b/plugins/usbdmx/AVLdiyD512Factory.cpp
@@ -63,6 +63,9 @@ bool AVLdiyD512Factory::DeviceAdded(
   // Some AVLdiy devices don't have serial numbers. Since there isn't another
   // good way to uniquely identify a USB device, we only support one of these
   // types of devices per host.
+  // TODO(Peter): We could instead use the device & bus number (like the
+  // Eurolite plugin). You could use more than one device, but the patch
+  // wouldn't follow if you plugged it into a different port
   if (info.serial.empty()) {
     if (m_missing_serial_number) {
       OLA_WARN << "Failed to read serial number or serial number empty. "
--- a/plugins/usbdmx/AnymauDMXFactory.cpp
+++ b/plugins/usbdmx/AnymauDMXFactory.cpp
@@ -63,6 +63,9 @@ bool AnymauDMXFactory::DeviceAdded(
   // Some Anyma devices don't have serial numbers. Since there isn't another
   // good way to uniquely identify a USB device, we only support one of these
   // types of devices per host.
+  // TODO(Peter): We could instead use the device & bus number (like the
+  // Eurolite plugin). You could use more than one device, but the patch
+  // wouldn't follow if you plugged it into a different port
   if (info.serial.empty()) {
     if (m_missing_serial_number) {
       OLA_WARN << "Failed to read serial number or serial number empty. "
--- a/plugins/usbdmx/DMXCreator512BasicFactory.cpp
+++ b/plugins/usbdmx/DMXCreator512BasicFactory.cpp
@@ -55,6 +55,9 @@ bool DMXCreator512BasicFactory::DeviceAdded(
   // vendor and product ids. Also, since DMXCreator 512 Basic devices don't have
   // serial numbers and there is no other good way to uniquely identify a USB
   // device, we only support one of these types of devices per host.
+  // TODO(Peter): We could instead use the device & bus number (like the
+  // Eurolite plugin). You could use more than one device, but the patch
+  // wouldn't follow if you plugged it into a different port
   if (info.serial.empty()) {
     if (m_missing_serial_number) {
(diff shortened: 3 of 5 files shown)


aroffringa:
Hi @peternewman , I've compiled and tested this branch. Indeed with this branch I can insert the Eurolite MK2 in a different USB port and OLA will still patch a universe to it if it existed. With the master branch this is indeed not the case; whenever I change the USB port, the universe is not patched.

As a side note, I've discovered that the Eurolite only works when it is already present when ola starts. If I insert it later, I get this log:
```
libs/usb/HotplugAgent.cpp:172: USB hotplug event: 2:8 @0x7f4038002e40 [add]
libs/usb/LibUsbAdaptor.cpp:53: libusb_get_string_descriptor_ascii failed: LIBUSB_ERROR_BUSY
libs/usb/LibUsbAdaptor.cpp:129: Failed to get manufacturer name
libs/usb/LibUsbAdaptor.cpp:53: libusb_get_string_descriptor_ascii failed: LIBUSB_ERROR_BUSY
libs/usb/LibUsbAdaptor.cpp:134: Failed to get product name
libs/usb/LibUsbAdaptor.cpp:53: libusb_get_string_descriptor_ascii failed: LIBUSB_ERROR_BUSY
libs/usb/LibUsbAdaptor.cpp:139: Failed to read serial number, the device probably doesn't have one
plugins/usbdmx/EuroliteProFactory.cpp:140: Connected FTDI device with serial  could be a Eurolite USB-DMX512-PRO MK2 but was ignored, because enable_eurolite_mk2 was false and its serial number was not listed specifically in eurolite_mk2_serial
plugins/usbpro/WidgetDetectorThread.cpp:215: Found potential USB Serial device at /dev/ttyUSB0
common/io/Serial.cpp:199: Acquired /var/lock/LCK..ttyUSB0
plugins/usbpro/WidgetDetectorThread.cpp:389: trying stage 0 for 0x7f4040001780
plugins/usbpro/UsbProWidgetDetector.cpp:323: USB Widget didn't respond to messages, esta id 0, device id 0
plugins/usbpro/UsbProWidgetDetector.cpp:326: Is device in USB Controller mode if it's a Goddard?
plugins/usbpro/WidgetDetectorThread.cpp:389: trying stage 1 for 0x7f4040001780
plugins/usbpro/WidgetDetectorThread.cpp:386: no more detectors to try for  0x7f4040001780
common/io/Serial.cpp:247: Released /var/lock/LCK..ttyUSB0
```
So somehow the `libusb_get_string_descriptor_ascii` fails and the device information including serial is therefore not known to OLA, causing it to not be recognized. This is not a new issue caused by the serial id check; it also doesn't work with the old method of setting `enable_eurolite_mk2=true`, as that results in:
```
libs/usb/HotplugAgent.cpp:172: USB hotplug event: 2:12 @0x7f6580002e40 [add]
libs/usb/LibUsbAdaptor.cpp:53: libusb_get_string_descriptor_ascii failed: LIBUSB_ERROR_BUSY
libs/usb/LibUsbAdaptor.cpp:129: Failed to get manufacturer name
libs/usb/LibUsbAdaptor.cpp:53: libusb_get_string_descriptor_ascii failed: LIBUSB_ERROR_BUSY
libs/usb/LibUsbAdaptor.cpp:134: Failed to get product name
libs/usb/LibUsbAdaptor.cpp:53: libusb_get_string_descriptor_ascii failed: LIBUSB_ERROR_BUSY
libs/usb/LibUsbAdaptor.cpp:139: Failed to read serial number, the device probably doesn't have one
plugins/usbdmx/EuroliteProFactory.cpp:127: Found a probable new Eurolite USB-DMX512-PRO MK2 device with serial 
libs/usb/LibUsbAdaptor.cpp:150: Manufacturer mismatch: FTDI != 
```
All works fine when the usb device is connected beforehand. I've therefore thus also not tested your fix by hotplugging, but using these steps:

- Connect Eurolite
- Start OLA and patch the Eurolite
- Stop OLA
- Connect Eurolite to another port
- Restart OLA and check if the universe is still patched.

So before your fix the universe isn't patched, afterwards it is.

peternewman:
> Hi @peternewman , I've compiled and tested this branch. Indeed with this branch I can insert the Eurolite MK2 in a different USB port and OLA will still patch a universe to it if it existed. With the master branch this is indeed not the case; whenever I change the USB port, the universe is not patched.

> So before your fix the universe isn't patched, afterwards it is.

Great thanks. I'll merge this then, as it clearly makes things better!

> As a side note, I've discovered that the Eurolite only works when it is already present when ola starts. If I insert it later, I get this log:

:cry: That's a shame, lets open a new issue for that given you say its unrelated to this PR (or indeed your previous one).

PR #1936, opened 2024-02-26
Ignore more minified files that might contain spelling false positives

peternewman:
Fixes our broken CI and hence resolves the issues in #1935 !


kripton (approved):
Indeed, not the best idea to spellcheck minified gibberish

peternewman:
Although it's actually picked up what it thinks is a genuine typo, it just happens to be a false positive as it's a special Javascript word!

PR #1937, opened 2024-02-27
Workaround gcovr parse race condition bug

DaAwesomeP:
This bug is causing the coverage run to fail every so often. Using an officially documented workaround from here: https://gcovr.com/en/stable/guide/gcov_parser.html#negative-hit-counts

See also a similar bug in another project: https://github.com/zephyrproject-rtos/zephyr/issues/64396

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -272,9 +272,9 @@ COVERAGE_OUTPUTS = --txt coverage/coverage.txt \
 	--html-details coverage/details.html/coverage.details.html \
 	--coveralls coverage/coverage.coveralls.json
 # See https://gcovr.com/en/stable/guide/gcov_parser.html#negative-hit-counts
-COVERAGE_FLAGS=--gcov-ignore-parse-errors=negative_hits.warn_once_per_file
-COVERAGE_GCOV_EXE=--gcov-executable /usr/bin/gcov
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -271,6 +271,8 @@ COVERAGE_OUTPUTS = --txt coverage/coverage.txt \
 	--cobertura coverage/coverage.cobertura.xml \
 	--html-details coverage/details.html/coverage.details.html \
 	--coveralls coverage/coverage.coveralls.json
+# See https://gcovr.com/en/stable/guide/gcov_parser.html#negative-hit-counts
+COVERAGE_FLAGS=--gcov-ignore-parse-errors=negative_hits.warn_once_per_file
 COVERAGE_GCOV_EXE=--gcov-executable /usr/bin/gcov
 COVERAGE_FILTERS=-e '.*Test\.cpp$$' \
```
peternewman, line 277: We could be a bit more consistent with spacing to match the rest of the file:
```suggestion
COVERAGE_FLAGS = --gcov-ignore-parse-errors=negative_hits.warn_once_per_file
COVERAGE_GCOV_EXE = --gcov-executable /usr/bin/gcov
COVERAGE_FILTERS = -e '.*Test\.cpp$$' \
```

kripton (approved):

PR #1945, opened 2024-03-16
E1.33 cherry pick the second

peternewman:
The next sub-part of #1841


libs/acn/RootPDU.h
```diff
@@ -33,8 +33,8 @@ namespace acn {
 
 class RootPDU: public PDU {
  public:
-  explicit RootPDU(unsigned int vector):
-    PDU(vector),
+  explicit RootPDU(unsigned int vector, bool force_length_flag = false):
+    PDU(vector, FOUR_BYTES, force_length_flag),
```
kripton, line 37: Just a question: What are the "FOUR_BYTES"? Some kind of magic/FourCC?

tools/e133/E133StatusHelper.cpp
```diff
@@ -96,5 +96,51 @@ string StatusCodeToString(E133StatusCode status_code) {
   }
   return "Unknown E1.33 Status Code";
 }
+
+
+bool IntToConnectStatusCode(uint16_t input,
+                            E133ConnectStatusCode *connect_status_code) {
+  switch (input) {
+    case ola::e133::CONNECT_OK:
```
kripton, line 104: how about:
```
 case ola::e133::CONNECT_OK:
 case ola::e133::CONNECT_SCOPE_MISMATCH:
.......
      *connect_status_code = input;
      return true;
```
?

kripton (approved):
See inline comments. Some might be stale due to changes pushed during my review :)

include/ola/acn/ACNPort.h
```diff
@@ -61,6 +61,11 @@ const uint16_t ACN_PORT = 5568;
  */
 const uint16_t E133_PORT = 5569;
 
+/**
+ * @brief The port used for E1.33 LLRP communication.
+ */
+const uint16_t LLRP_PORT = 5569;
```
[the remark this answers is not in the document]
peternewman: I think E133_PORT becomes redundant looking at where it's used, and they essentially reused the port they'd chosen/had allocated.

I'm inclined to leave them separate for now and I've added a task to the primary PR to remove all the redundant code from the pre-standards version.

libs/acn/RootPDU.h
peternewman: Not really, it's just an internal enum for the width of the vector, defined here:
https://github.com/OpenLightingProject/ola/blob/e7b403927ad462d6ba4eed7d72cdd3e1fb1fa4e8/libs/acn/PDU.h#L39-L43

Just a bit more user-friendly than magic numbers like 4 appearing all over the code with no context.

tools/e133/E133StatusHelper.cpp
peternewman: That's a very elegant solution.

Unfortunately that code doesn't solely exist in E1.33 land. Do you fancy a clean-up of this (once merged) and the other similar code (e.g. there's some in https://github.com/OpenLightingProject/ola/blob/master/common/rdm/RDMHelper.cpp )?

We'd also need to add OLA_FALLTHROUGH to cover us from other compiler warnings, but looks neat.

peternewman:
> See inline comments.

Thanks.

> Some might be stale due to changes pushed during my review :)

Yeah sorry it was a bit lumpy! Do you want to check my replies and double-check nothing else snuck through whilst you **were** looking?

libs/acn/RootPDU.h
kripton: Okay, thanks for explaining :)

tools/e133/E133StatusHelper.cpp
kripton: I can have a look, yes. Ping me once you merged it!

kripton:
> > Some might be stale due to changes pushed during my review :)
> 
> Yeah sorry it was a bit lumpy! Do you want to check my replies and double-check nothing else snuck through whilst you **were** looking?

No problem. It should be fine, github makes it quite easy to track what changed after I started reviewing so go ahead merging it :+1:

tools/e133/E133StatusHelper.cpp
peternewman: It's in now @kripton but I assume you got an automatic notification anyway...

PR #1950, opened 2024-03-28
E1.33 cherry pick part 4

peternewman:


libs/acn/RPTPDU.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+
+/*
+ * Pack the header portion.
+ */
+bool RPTPDU::PackHeader(uint8_t *data, unsigned int *length) const {
+  unsigned int header_size = HeaderSize();
+
+  if (*length < header_size) {
```
kripton, line 55: would this be a null-dereference if someone forgets to pass length? Should check for `!= nullptr` before?

libs/acn/RPTPDU.cpp
```diff
@@ -0,0 +1,132 @@
...
+  memcpy(data, &header, *length);
+  return true;
+}
+
+
+/*
+ * Pack the data portion.
+ */
+bool RPTPDU::PackData(uint8_t *data, unsigned int *length) const {
+  if (m_pdu)
```
kripton, line 79: add curly braces

libs/acn/RPTPDU.cpp
```diff
@@ -0,0 +1,132 @@
...
+ */
+void RPTPDU::PackHeader(OutputStream *stream) const {
+  RPTHeader::rpt_pdu_header header;
+  m_header.SourceUID().Pack(header.source_uid, sizeof(header.source_uid));
+  header.source_endpoint = HostToNetwork(m_header.SourceEndpoint());
+  m_header.DestinationUID().Pack(header.destination_uid,
+                                 sizeof(header.destination_uid));
+  header.destination_endpoint = HostToNetwork(m_header.DestinationEndpoint());
+  header.sequence = HostToNetwork(m_header.Sequence());
+  stream->Write(reinterpret_cast<uint8_t*>(&header),
```
kripton, line 97: possible null-dereference here as well?

libs/acn/RPTPDU.cpp
```diff
@@ -0,0 +1,132 @@
...
+                sizeof(RPTHeader::rpt_pdu_header));
+}
+
+
+/*
+ * Pack the data into a buffer
+ */
+void RPTPDU::PackData(OutputStream *stream) const {
+  if (m_pdu)
```
kripton, line 106: curly braces missing

libs/acn/RPTPDU.cpp
```diff
@@ -0,0 +1,132 @@
...
+  source_uid.Pack(header.source_uid, sizeof(header.source_uid));
+  header.source_endpoint = HostToNetwork(source_endpoint);
+  destination_uid.Pack(header.destination_uid, sizeof(header.destination_uid));
+  header.destination_endpoint = HostToNetwork(destination_endpoint);
+  header.sequence = HostToNetwork(sequence_number);
+  stack->Write(reinterpret_cast<uint8_t*>(&header),
+               sizeof(RPTHeader::rpt_pdu_header));
+
+  vector = HostToNetwork(vector);
+  stack->Write(reinterpret_cast<uint8_t*>(&vector), sizeof(vector));
```
kripton, line 128: null-dereference? I seem to be quite sensitive to them today :D

libs/acn/RPTRequestPDU.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace ola {
+namespace acn {
+
+using ola::io::OutputStream;
+using ola::network::HostToNetwork;
+
+void RPTRequestPDU::PrependPDU(ola::io::IOStack *stack) {
+  uint32_t vector = HostToNetwork(
+      static_cast<uint32_t>(VECTOR_REQUEST_RDM_CMD));
+  stack->Write(reinterpret_cast<uint8_t*>(&vector), sizeof(vector));
```
kripton, line 35: possible null dereference?

kripton (approved):
See inline comments :)

libs/acn/RPTPDU.cpp
peternewman: Added.

libs/acn/RPTPDU.cpp
peternewman: Added.

libs/acn/RPTPDU.cpp
peternewman: It does indeed segfault. I've now added something a bit higher up in the stack which covers us for everything apart from someone calling PackHeader/PackData directly themselves, but means we don't need to tweak every single PDU...

kripton: Where's the commit for those changes "higher up in the stack"? I've seen your changes but it looks like they are per PDU?

kripton (approved):
Nice that you added tests for the methods when NULL is passed as input :+1:

peternewman:
> Nice that you added tests for the methods when NULL is passed as input 👍

Yeah I like a bit of test driven development; make it segfault, then stop it doing so!

libs/acn/RPTPDU.cpp
peternewman: https://github.com/OpenLightingProject/ola/pull/1950/commits/d6a13dd9218c4f335905318b200462ac2218ed2d

There's basically no uses of PackHeader/PackData directly, only via Pack, hence just doing it there for now.

I've also added a proper fix to the backlog in #1841 as there are a lot of other bits to cover too.

libs/acn/RPTPDU.cpp
peternewman: I've now trapped this one directly above.

libs/acn/RPTPDU.cpp
peternewman: This one is fixed upstream in the PDU::Pack call.

libs/acn/RPTRequestPDU.cpp
peternewman: I've now trapped this one directly above.

peternewman:
Okay I think that's the outstanding comments fixed in one way or another @kripton .

kripton (approved):
Re-reviewed :heavy_check_mark:

PR #1953, opened 2024-03-28
Add initial E1.33 RDM tests

peternewman:
After #1952 .

### DIFF
--- a/tools/rdm/TestDefinitions.py
+++ b/tools/rdm/TestDefinitions.py
@@ -8441,60 +8441,60 @@ class SetEndpointToUniverseWithExtraData(TestMixins.SetWithDataMixin,
   DATA = b'foobar'
 
 
-class AllSubDevicesGetRdmTrafficEnable(TestMixins.AllSubDevicesGetMixin,
+class AllSubDevicesGetRDMTrafficEnable(TestMixins.AllSubDevicesGetMixin,
                                        OptionalParameterTestFixture):
   """Send a get RDM_TRAFFIC_ENABLE to ALL_SUB_DEVICES."""
   PID = 'RDM_TRAFFIC_ENABLE'
   DATA = [0x0001]
 
 
-# class GetRdmTrafficEnable(TestMixins.,
+# class GetRDMTrafficEnable(TestMixins.,
 #                           OptionalParameterTestFixture):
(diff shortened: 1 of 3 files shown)


tools/rdm/TestDefinitions.py
```diff
@@ -7923,6 +7923,791 @@ class SetInterfaceHardwareAddressType1WithData(
   PID = 'INTERFACE_HARDWARE_ADDRESS_TYPE1'
 
 
+# E1.33/E1.37-7 PIDS
```
kripton, line 7926: Is this auto-generated code? I would have gotten mad writing that by hand ...

kripton (approved):

tools/rdm/TestDefinitions.py
peternewman: Yeah list_rdm_tests.py ( https://github.com/OpenLightingProject/ola/blob/af3b2a1e3934d529b1a5b878d7423f990ab81f96/tools/rdm/list_rdm_tests.py ) does as much as it can, which generally covers writing most of the easiest failing tests and the stub of the working ones, you just have to supply the valid data and then validate the results (at the simplest level).

It also (hence the name) lists all test types that should exist (e.g. AllSubDevicesGetFoo) so we can compare that against what we've currently got and work out what's missing.

peternewman: An eyeball would still help though, as e.g. I noticed we've no magic casing so the RDM ones were FooRdmBar at first until I manually fixed them.

kripton: Eases my mind that you didn't have to type it all by hand.
I did read it all and didn't find any glaring discrepancies. The only thing I saw multiple times is the naming "withExtraData" vs. "with more than 2 bytes of data"

peternewman: So the class of test named ...WithExtraData, have at least one additional byte on the end of what should ideally otherwise be a valid request, the responder should then NAck them for sending too much data. The "with more than 2 bytes" value is calculated based on how much data they should have, I think most of these will probably be things like endpoints which are two bytes long, hence the repetition.

In comparison say GetDMXPersonalityDescriptionWithExtraData says with more than 1 byte of data.

PR #1959, opened 2024-04-14
E1.33 cherry pick number 6

peternewman:


libs/acn/LLRPProbeReplyPDU.cpp
```diff
@@ -32,6 +32,27 @@ using ola::network::HostToNetwork;
...
+  *length = sizeof(llrp_probe_reply_pdu_data);
+  memcpy(data, &pdu_data, *length);
+  return true;
+}
+
+void LLRPProbeReplyPDU::PackData(ola::io::OutputStream *stream) const {
+  llrp_probe_reply_pdu_data data;
+  m_target_uid.Pack(data.target_uid, sizeof(data.target_uid));
+  m_hardware_address.Pack(data.hardware_address, sizeof(data.hardware_address));
+  data.type = HostToNetwork(static_cast<uint8_t>(m_type));
```
kripton, line 51: since `type` is an uint8_t, the `HostToNetwork` call will most probably not be necessary. However, if we have it in more places, we can leave it here for consistency

kripton: I don't know if this was the case in other PDUs as well, I just realized here for the first time

libs/acn/LLRPProbeReplyPDU.h
```diff
@@ -53,6 +74,11 @@ class LLRPProbeReplyPDU : private PDU {
     uint8_t type;
```
kripton, line 74: `type` (mentioned above) is defined here

kripton (approved):
Some TODOs inside but the changes themselves are fine

libs/acn/LLRPProbeReplyPDU.cpp
peternewman: We've done our best to make it a No Op if its used:
https://github.com/OpenLightingProject/ola/blob/0fff67114a6c25e0fb19598c69f284355383867f/include/ola/network/NetworkUtils.h#L33-L37

I'm not sure we've been very consistent, but given how much it can be inlined, there's possibly an argument to keep it as the risk of people missing a HostToNetwork on a bigger value causes a lot more issues than some noise which hopefully the compiler will optimise away on a uint8_t.

kripton: Okay, that looks like a very nice solution and I agree that the trouble of missing it once is way larger than the possible performance penalty of using it everywhere. Well done :)

peternewman: Just standing on the shoulders of giants!

PR #1960, opened 2024-04-21
E1.33 cherry pick vii

peternewman:


libs/acn/LLRPProbeReplyPDUTest.cpp
```diff
@@ -176,6 +176,14 @@ void LLRPProbeReplyPDUTest::testPrepend() {
     0xff
   };
   OLA_ASSERT_DATA_EQUALS(expected_data, sizeof(expected_data), buffer, length);
+
+  // test null stack
+  LLRPProbeReplyPDU::PrependPDU(
```
kripton, line 181: does this test actually test something? shouldn't it check the return value?

libs/acn/LLRPProbeRequestPDUTest.cpp
```diff
@@ -191,6 +191,15 @@ void LLRPProbeRequestPDUTest::testPrepend() {
     0x56, 0x78, 0, 0, 0, 2,
   };
   OLA_ASSERT_DATA_EQUALS(expected_data, sizeof(expected_data), buffer, length);
+
+  // test null stack
+  LLRPProbeRequestPDU::PrependPDU(NULL,
```
kripton, line 196: does this test actually test something? shouldn't it check the return value?

tools/e133/ManagementEndpoint.h
```diff
@@ -30,7 +30,7 @@ using ola::rdm::RDMResponse;
 using ola::rdm::RDMCallback;
 
 /**
- * The ManagementEndpoint handles RDMCommands directed at this E1.33 device. It
+ * The ManagementEndpoint handles RDMCommands directed at this E1.33 Component. It
```
kripton, line 33: capital C is on purpose?

kripton (approved):
see inline comments

libs/acn/LLRPProbeReplyPDUTest.cpp
peternewman: There is no return value, it's a void. It only modifies stack, which we've passed in as null anyway, so no change there. I guess we could prove the stack was still null, but I'm not sure it's very beneficial.

Without the null fix in the code though, the test itself segfaults too, so we'll know its broken...

libs/acn/LLRPProbeRequestPDUTest.cpp
peternewman: As above

tools/e133/ManagementEndpoint.h
peternewman: Yeah, I think from the spec that both Controllers and Devices now have management endpoints, I believe Component (or RPT Component) is the overarching term for Controller and/or Device.

peternewman: "thing" would also be acceptable! ;)

kripton: I rather meant that the previous "device" was lower-case and the new "Component" is upper-case. No issue with the wording itself

peternewman: I think that was device in the generic sense, whereas I was capitalising the official term from the standard to match the standard.

peternewman:
Merging given @kripton was happy and I ought to try and get the rest merged in...

PR #1975, opened 2024-09-07
Changes from building cleanly on Fedora with GCC 14.2

kiilerix:


DaAwesomeP:
Related to #1889 and #1890

kiilerix:
> Related to #1889 and #1890

Right, thanks. I had not checked for existing pending PRs. It seems like things stopped moving :-(

DaAwesomeP:
@peternewman I understand your concerns for backward compatibility, but we really need to address these couple of PRs (#1889 and #1890). `auto_ptr` was deprecated in C++11 and removed in C++17. Surely we could at least merge them to the `master` branch instead of `0.10`?

kripton (approved):

PR #1976, opened 2024-09-19
Add support for the uint64 and int64 data types to the RDM messaging

peternewman:
Plus some string and endian handling to enable this


kripton (approved):
Seems all reasonable to me :)

PR #1977, opened 2024-09-23
Fix template usage with % in OSC device listing

kiilerix:
ola_dev_info was showing un unexpanded '%d':

    Device 8: OSC Device
      port 0, IN /dmx/universe/%d, priority 100
      port 1, IN /dmx/universe/%d, priority 100
      port 2, IN /dmx/universe/%d, priority 100

### DIFF
--- a/plugins/osc/OSCPlugin.cpp
+++ b/plugins/osc/OSCPlugin.cpp
@@ -158,8 +158,8 @@ bool OSCPlugin::SetDefaultPreferences() {
 
   for (unsigned int i = 0; i < GetPortCount(INPUT_PORT_COUNT_KEY); i++) {
     const string key = ExpandTemplate(PORT_ADDRESS_TEMPLATE, i);
-    save |= m_preferences->SetDefaultValue(key, StringValidator(),
-                                           DEFAULT_ADDRESS_TEMPLATE);
+    const string value = ExpandTemplate(DEFAULT_ADDRESS_TEMPLATE, i);
+    save |= m_preferences->SetDefaultValue(key, StringValidator(), value);
   }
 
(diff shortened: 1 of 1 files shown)


DaAwesomeP:
We should maybe target the `0.10` branch with this.

kripton:
> We should maybe target the `0.10` branch with this.

Indeed, that would make sense. @kiilerix thanks for the fix!

peternewman:
Thanks for the PR @kiilerix . However did you try patching port to a universe with the orginal code?

As per the plugin's docs, available locally within OLAd but also:
https://docs.openlighting.org/ola/conf/ola-osc.conf.html

> port_N_address = /address
> The OSC address to listen on for port N. If the address contains %d it's
> replaced by the universe number for port N.

So if you patch a port using the original code, you should find it looks like so:

> ```
> Device 8: OSC Device
>   port 0, IN /dmx/universe/42, priority 100, patched to universe 42
>   port 1, IN /dmx/universe/%d, priority 100
>   port 2, IN /dmx/universe/%d, priority 100
> ```

So it dynamically updates to represent the universe you're sending DMX to. Without it patched to a universe, you can't use it anyway, so there's not really an edge case there as such.

With your suggested PR, the input OSC paths are permanently fixed to `/dmx/universe/0` to `/dmx/universe/3` by default, then you need to know that you patched port 2 to universe 42 or whatever.

Personally that doesn't seem as user friendly to me (and I didn't write the original code), but let me know if I've missed something? Having explained it, do we just need to try and improve out docs somehow, or should we add some ports of each flavour? The user can always do that themselves if required by tweaking the relevant config file.

PR #1978, opened 2024-09-29
Add DMXControl Projects e.V. Nodle R4S

StefanKrupop:
Our "new" readymade USB DMX Interface "Nodle R4S" has a different PID than the Nodle U1, but otherwise is compatible.
So I just added the additional PID and list the R4S along with the U1.


peternewman (approved):
Perfect thanks @StefanKrupop !

peternewman:
I've stuck it on the website now too ( https://www.openlighting.org/ola/ ). I forgot to ask, which operating systems had you tested it on, still just Linux for TX and RX?

PR #1979, opened 2024-10-02
Added plugin-related functions to the Java client

moritzvieli:
- Add OlaClient.reloadPlugins()
- Add OlaClient.getPluginState()
- Add OlaClient.setPluginState()
- Changed evaluation of PROTOBUF_VERSION, because pkg-config adds trailing ".0" to the version, which is not compatible with the available Java dependencies

The last change needs to be checked carefully. I tried building it on a Mac and it failed with the pkg-config version, because of the trailing ".0". Using "protoc --version" works, but might introduce other issues.

### DIFF
--- a/config/ola.m4
+++ b/config/ola.m4
@@ -25,10 +25,10 @@ AC_REQUIRE_CPP()
 PKG_CHECK_MODULES(libprotobuf, [protobuf >= $1])
 
 AC_MSG_CHECKING([protobuf library version])
-PROTOBUF_VERSION=`pkg-config --modversion protobuf`;
+PROTOBUF_VERSION=`protoc --version | awk '{print [$]2}'`;
+AC_MSG_NOTICE([PROTOBUF VERSION $PROTOBUF_VERSION])
 AC_MSG_RESULT([$PROTOBUF_VERSION])
 AC_SUBST([PROTOBUF_VERSION])
-
 AC_SUBST([libprotobuf_CFLAGS])
 
 AC_ARG_WITH([protoc],
--- a/java/ChangeLog
+++ b/java/ChangeLog
@@ -1,3 +1,9 @@
+02/10/2024 Moritz Vieli <moritz.vieli@gmail.com>
+ * v0.0.23
+  - Add OlaClient.reloadPlugins()
+  - Add OlaClient.getPluginState()
+  - Add OlaClient.setPluginState()
+
 04/22/2017 Erez Makavy <hepi.yellow@gmail.com>
  * v0.0.2 
   - Add OlaClient.getUniverseList()
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>ola</groupId>
   <artifactId>ola-java-client</artifactId>
-  <version>0.1.0</version>
+  <version>0.0.3</version>
   <description>Java implementation of OLA RPC</description>
   <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -36,8 +36,8 @@
        <artifactId>maven-compiler-plugin</artifactId>
        <version>2.5.1</version>
        <configuration>
-         <source>1.6</source>
-         <target>1.6</target>
+         <source>1.7</source>
+         <target>1.7</target>
        </configuration>
      </plugin>
      <plugin>
--- a/java/src/main/java/ola/OlaClient.java
+++ b/java/src/main/java/ola/OlaClient.java
@@ -18,6 +18,8 @@ package ola;
 
 import java.util.logging.Logger;
 
+import ola.proto.Ola;
+import ola.proto.Ola.Ack;
 import ola.proto.Ola.DeviceConfigReply;
 import ola.proto.Ola.DeviceConfigRequest;
 import ola.proto.Ola.DeviceInfoReply;
@@ -30,10 +32,14 @@ import ola.proto.Ola.OlaServerService;
 import ola.proto.Ola.OptionalUniverseRequest;
 import ola.proto.Ola.PatchAction;
 import ola.proto.Ola.PatchPortRequest;
+import ola.proto.Ola.PluginReloadRequest;
 import ola.proto.Ola.PluginDescriptionReply;
 import ola.proto.Ola.PluginDescriptionRequest;
 import ola.proto.Ola.PluginListReply;
 import ola.proto.Ola.PluginListRequest;
+import ola.proto.Ola.PluginStateRequest;
+import ola.proto.Ola.PluginStateReply;
+import ola.proto.Ola.PluginStateChangeRequest;
 import ola.proto.Ola.PortPriorityRequest;
 import ola.proto.Ola.RDMRequest;
 import ola.proto.Ola.RDMResponse;
@@ -67,7 +73,6 @@ public class OlaClient {
 
 
     public OlaClient() throws Exception {
-
         channel = new StreamRpcChannel();
         controller = new SimpleRpcController();
         serverService = OlaServerService.Stub.newStub(channel);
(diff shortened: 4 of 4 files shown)


java/pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>ola</groupId>
   <artifactId>ola-java-client</artifactId>
-  <version>0.1.0</version>
+  <version>0.0.3</version>
```
peternewman, line 5: We should never go backwards in our versions, even if I failed to update the separate Java changelog last time around.

I think this should be 0.2.0 given:
> MINOR version when you add functionality in a backward compatible manner

java/ChangeLog
```diff
@@ -1,3 +1,9 @@
+02/10/2024 Moritz Vieli <moritz.vieli@gmail.com>
+ * v0.0.23
```
peternewman, line 2: I think you mean v0.0.3 in your versioning, but as mentioned this should be >0.1.0, I believe 0.2.0.

java/pom.xml
```diff
@@ -36,8 +36,8 @@
        <artifactId>maven-compiler-plugin</artifactId>
        <version>2.5.1</version>
        <configuration>
-         <source>1.6</source>
-         <target>1.6</target>
+         <source>1.7</source>
+         <target>1.7</target>
```
peternewman, line 40: Is this just because 1.6 is out of support?

java/src/main/java/ola/OlaClient.java
peternewman, line 1: I'm less keen on all the whitespace changes (particularly alignment within Javadoc) given we don't currently have a linting scheme and the risk of someone else's future edits just conflicting with what you've done.

java/src/main/java/ola/OlaClient.java
```diff
@@ -130,6 +142,38 @@ public PluginDescriptionReply getPluginDescription(int pluginId) {
...
+        PluginStateRequest request = Ola.PluginStateRequest.newBuilder()
+                .setPluginId(pluginId)
+                .build();
+
+        return (PluginStateReply) callRpcMethod("GetPluginState", request);
+    }
+
+
+    /**
+     * Change the state of plugins.
```
peternewman, line 161:
```suggestion
     * Set the state of a plugin.
```

java/src/main/java/ola/OlaClient.java
```diff
@@ -114,14 +117,23 @@ public PluginListReply getPlugins() {
     }
 
 
+    /**
+     * Reload the plugins.
+     *
+     * @return Acknowledgement.
+     */
+    public Ack reloadPlugins() {
+        return (Ack) callRpcMethod("ReloadPlugins", PluginReloadRequest.newBuilder().build());
```
peternewman, line 126: I think the behaviour of patchPort is more useful:
https://github.com/OpenLightingProject/ola/blob/c4adb5034feddeaea32cb4554ef00837dbbcbec9/java/src/main/java/ola/OlaClient.java#L240
https://github.com/OpenLightingProject/ola/blob/c4adb5034feddeaea32cb4554ef00837dbbcbec9/java/src/main/java/ola/OlaClient.java#L252

java/src/main/java/ola/OlaClient.java
```diff
@@ -130,6 +142,38 @@ public PluginDescriptionReply getPluginDescription(int pluginId) {
...
+     * @param enabled  whether the plugin should be enabled or not
+     * @return Acknowledgement.
+     */
+    public Ack setPluginState(int pluginId, boolean enabled) {
+        PluginStateChangeRequest request = Ola.PluginStateChangeRequest.newBuilder()
+                .setPluginId(pluginId)
+                .setEnabled(enabled)
+                .build();
+
+        return (Ack) callRpcMethod("SetPluginState", request);
```
peternewman, line 173: I think the behaviour of patchPort is more useful:
https://github.com/OpenLightingProject/ola/blob/c4adb5034feddeaea32cb4554ef00837dbbcbec9/java/src/main/java/ola/OlaClient.java#L240
https://github.com/OpenLightingProject/ola/blob/c4adb5034feddeaea32cb4554ef00837dbbcbec9/java/src/main/java/ola/OlaClient.java#L252

java/src/main/java/ola/OlaClient.java
```diff
@@ -130,6 +142,38 @@ public PluginDescriptionReply getPluginDescription(int pluginId) {
     }
 
 
+    /**
+     * Return the state for a plugin.
```
peternewman, line 146: I think this should be:
```suggestion
     * Return the state of a plugin.
```

java/src/main/java/ola/OlaClient.java
```diff
@@ -130,6 +142,38 @@ public PluginDescriptionReply getPluginDescription(int pluginId) {
...
+     *
+     * @param pluginId number of the plugin for which to receive the state
+     * @return The list of plugings.
```
peternewman, line 149:
```suggestion
     * @return The state of the plugin.
```

java/src/main/java/ola/OlaClient.java
```diff
@@ -130,6 +142,38 @@ public PluginDescriptionReply getPluginDescription(int pluginId) {
     }
 
 
+    /**
+     * Return the state for a plugin.
+     *
+     * @param pluginId number of the plugin for which to receive the state
```
peternewman, line 148:
```suggestion
     * @param pluginId number of the plugin to fetch the state of
```

config/ola.m4
```diff
@@ -25,10 +25,10 @@ AC_REQUIRE_CPP()
 PKG_CHECK_MODULES(libprotobuf, [protobuf >= $1])
 
 AC_MSG_CHECKING([protobuf library version])
-PROTOBUF_VERSION=`pkg-config --modversion protobuf`;
+PROTOBUF_VERSION=`protoc --version | awk '{print [$]2}'`;
+AC_MSG_NOTICE([PROTOBUF VERSION $PROTOBUF_VERSION])
```
peternewman, line 29: I'm not hugely keen on adding another means of processing into the mix and we've already got a load of Protobuf versioning stuff in via https://github.com/OpenLightingProject/ola/commit/d4414feefd95adf9c91d7eaf1e94380296c35f7a so hopefully we can just reuse some of that to make a Java version that's compatible.

peternewman (changes requested):
Thanks @moritzvieli .

I don't know whether you want to split this in two, as apart from some minor styling, the client changes are ready to go, whereas the Protobuf Java versioning may need a bit more testing and thought.

It would also be good to add some tests too please:
https://github.com/OpenLightingProject/ola/blob/master/java/src/test/java/ola/OlaClientTest.java

java/pom.xml
moritzvieli: Fixed with commit 79630929f9cde69e0dbff67f7e1cbdfa4e3acbb6

java/ChangeLog
moritzvieli: Fixed with commit 79630929f9cde69e0dbff67f7e1cbdfa4e3acbb6

java/pom.xml
moritzvieli: Yes. I was unable to compile this Java version with my locally installed Java version 21. 1.7 is still outdated, but a little bit less worse imo.

config/ola.m4
moritzvieli: As suggested, took it out of the PR with commit e792f5255f4ddf1a35be0cef7e1cb1ab16f0c853

java/src/main/java/ola/OlaClient.java
moritzvieli: @peternewman Sorry, not sure, what you mean. With patchPort, no reloadPlugins would be required? I thought I have to use it after activating/deactivating plugins in my client, but seems like it's not required. I can remove it, if you think it's not necessary.

moritzvieli: @peternewman Ah, you mean the return value? Changed it with commit b5be0b94de51153921ab8c051f000487f9d3db9f

java/src/main/java/ola/OlaClient.java
moritzvieli: Fixed with commit b5be0b94de51153921ab8c051f000487f9d3db9f

java/src/main/java/ola/OlaClient.java
moritzvieli: I removed the whitespace changes.

moritzvieli:
@peternewman I incorporated all changes. The conversation about the Java source version is not yet resolved. Please let me know, if I missed something. Thanks a lot!

PR #1983, opened 2024-11-13
Fix build for macOS where endian.h is not available

cho-m:
It is possible other platforms were impacted as the header/functions aren't that portable.

Specific errors:
```
common/network/NetworkUtils.cpp:166:2: error: "No be64toh for NetworkToHost, please report this."
  166 | #error "No be64toh for NetworkToHost, please report this."
      |  ^
common/network/NetworkUtils.cpp:182:2: error: "No be64toh for NetworkToHost, please report this."
  182 | #error "No be64toh for NetworkToHost, please report this."
      |  ^
common/network/NetworkUtils.cpp:206:2: error: "No htobe64 for HostToNetwork, please report this."
  206 | #error "No htobe64 for HostToNetwork, please report this."
      |  ^
common/network/NetworkUtils.cpp:214:2: error: "No htobe64 for HostToNetwork, please report this."
  214 | #error "No htobe64 for HostToNetwork, please report this."
      |  ^
4 errors generated.
```

### DIFF
--- a/common/network/NetworkUtils.cpp
+++ b/common/network/NetworkUtils.cpp
@@ -60,6 +60,10 @@ typedef uint32_t in_addr_t;
 
 #ifdef HAVE_ENDIAN_H
 #include <endian.h>
+#elif defined(__APPLE__)
+#include <libkern/OSByteOrder.h>
+#define be64toh(x) OSSwapBigToHostInt64(x)
+#define htobe64(x) OSSwapHostToBigInt64(x)
 #endif  // HAVE_ENDIAN_H
 #include <errno.h>
 #include <limits.h>
@@ -160,11 +164,11 @@ uint32_t NetworkToHost(uint32_t value) {
 }
 
 uint64_t NetworkToHost(uint64_t value) {
-#ifdef HAVE_ENDIAN_H
+#if defined(HAVE_ENDIAN_H) || defined(__APPLE__)
   return be64toh(value);
 #else
 #error "No be64toh for NetworkToHost, please report this."
-#endif  // HAVE_ENDIAN_H
+#endif  // defined(HAVE_ENDIAN_H) || defined(__APPLE__)
 }
 
 int16_t NetworkToHost(int16_t value) {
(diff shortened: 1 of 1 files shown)


common/network/NetworkUtils.cpp
```diff
@@ -60,6 +60,10 @@ typedef uint32_t in_addr_t;
 
 #ifdef HAVE_ENDIAN_H
 #include <endian.h>
+#elif defined(__APPLE__)
+#include <libkern/OSByteOrder.h>
+#define be64toh(x) OSSwapBigToHostInt64(x)
+#define htobe64(x) OSSwapHostToBigInt64(x)
```
peternewman, line 66: So remove this:
```suggestion
```

common/network/NetworkUtils.cpp
```diff
@@ -60,6 +60,10 @@ typedef uint32_t in_addr_t;
 
 #ifdef HAVE_ENDIAN_H
 #include <endian.h>
+#elif defined(__APPLE__)
+#include <libkern/OSByteOrder.h>
```
peternewman, line 64: Switch this to including OSByteOrder.h if its present.

common/network/NetworkUtils.cpp
```diff
@@ -160,11 +164,11 @@ uint32_t NetworkToHost(uint32_t value) {
 }
 
 uint64_t NetworkToHost(uint64_t value) {
-#ifdef HAVE_ENDIAN_H
+#if defined(HAVE_ENDIAN_H) || defined(__APPLE__)
   return be64toh(value);
 #else
 #error "No be64toh for NetworkToHost, please report this."
-#endif  // HAVE_ENDIAN_H
+#endif  // defined(HAVE_ENDIAN_H) || defined(__APPLE__)
```
peternewman, line 171: Then switch this, and the following cases, to something more like https://github.com/OpenLightingProject/ola/blob/e8c755cdb1aba608785cce501ff5cc86f866e6d6/common/network/NetworkUtils.cpp#L684-L712 (but I don't think we need the extra wrapper functions here).

peternewman (changes requested):
Thanks for this @cho-m  and sorry for breaking things for you.

Rather than redefining be64toh and htobe64, we prefer to do ifdefs to handle the different cases. We should probably also add a header check for libkern/OSByteOrder.h and use that rather than __APPLE__ to switch the behaviour.

peternewman:
I've done these fixes in #2014 if you want to take a look @cho-m . They resolve the build for the Mac I've tried it on.

PR #1986, opened 2024-12-01
Fixed a minor typo

ijansch:
Just a tiny typo, but wanting to go through the pull request process once before I start proposing bigger changes.

### DIFF
--- a/python/examples/rdm_compare.py
+++ b/python/examples/rdm_compare.py
@@ -26,7 +26,7 @@ import webbrowser
 
 from ola.UID import UID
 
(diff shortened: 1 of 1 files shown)


peternewman (approved):
Great work thanks @ijansch and a good spot! Obviously this is the kind of thing that unfortunately tools like Codespell and spellintian won't notice either.

There's an Authors file you can add yourself to in future as well:
https://github.com/OpenLightingProject/ola/blob/master/AUTHORS

What sort of future changes were you thinking about?

Issue #2005, opened 2025-04-08
Next target versions


peternewman:
| OS/Component | Linux - Ubuntu LTS | Linux - Debian | Linux - Arch | BSD | OS X | Windows -MinGW | Windows - ? |
| --- | --- | --- | --- | --- | --- | --- | --- |
| C++ |  |  |  |  |  |  |  |
| Protobuf |  |  |  |  |  |  |  |
| Libmicrohttpd |  |  |  |  |  |  |  |
| Python |  |  |  |  |  |  |  |
| Java |  |  |  |  |  |  |  |

Assumptions:

1. Python 3 only
2. Protobuf 3 only

Proposal, from the above:

nomis52:
Yeah, I think it's fine to say Jammy Jellyfish (22) / gcc 11 is the cut off.

peternewman:
> * Providing "best effort" support for C++11 and C++14 (i.e. supporting modern C++ may slowly break those)

Personally I'm slightly uneasy about this, again partly for the CI reason, and also from a bug perspective, either we're supporting it (and running CI against that version) or we're not (and it might work if you're lucky). But there had also been desire to add some of the shiny new features (that go rather over my head), again either we should let people use them (and perhaps rewrite the whole codebase to use them consistently) or ask them not to (and keep the compatibility).

> Rather than arbitrarily setting new targets like this every few years, it might be better to align with one or more OSes and matching their timeline. For example, we should always aim to support stable Debian (and possible oldstable as well) and Ubuntu LTS. This can help ensure we are frequently remaining up-to-date without accumulating technical debt.

I only feel we need a target that everyone is working to, because we need to decide what our CI is running (and hence what gets through that gate), otherwise someone running Arch or whatever will open a PR to merge something that breaks someone running on a slightly older OS, and it would get merged unnoticed. I'd agree that defining our oldest target(s) in a more formulaic way (e.g. based on Debian/Ubuntu) makes that whole process a lot simpler, as we just update our CI builds when the n-2 or whatever LTS changes.

> BTW are people very attached to the autotools? If things are moving again, another idea could be to look into changing it to cmake, which imho is a bit more modern and easier. I'm happy to do that work, if people feel this is useful.

[...]

PR #2012, opened 2025-06-25
Remove some duplicate test category config

peternewman:


peternewman:
> @peternewman may you please take at the other C++20 related PRs as well (#1889, #1890, #1989, possibly #1975)? Thanks!

Thanks @DaAwesomeP . I'd like to address the general C++20 etcetera PRs via making a decision in #2005 first, I just pulled this bit in, as I noticed we'd had the same issue in the two newest CI builds and it doesn't upset the older stuff.

peternewman:
Removing the retrospective and the review request given #2013 essentially mirrors this and got reviewed.

PR #2014, opened 2025-07-28
Tidy the Mac OS Endian behaviour

peternewman:
On top of @cho-m 's fixes in #1983

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -149,6 +149,7 @@ AC_CHECK_HEADERS([asm/termbits.h asm/termios.h assert.h dlfcn.h endian.h \
                   execinfo.h linux/if_packet.h math.h net/ethernet.h \
                   stropts.h sys/ioctl.h sys/param.h sys/types.h sys/uio.h \
                   sysexits.h])
+AC_CHECK_HEADERS([libkern/OSByteOrder.h])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -149,6 +149,7 @@ AC_CHECK_HEADERS([asm/termbits.h asm/termios.h assert.h dlfcn.h endian.h \
                   execinfo.h linux/if_packet.h math.h net/ethernet.h \
                   stropts.h sys/ioctl.h sys/param.h sys/types.h sys/uio.h \
                   sysexits.h])
+AC_CHECK_HEADERS([libkern/OSByteOrder.h])
```
kripton, line 152: Is this a thing that exists on MacOS? Automatically or does it need to be installed?
Or do we introduce a new dependency here?

kripton (commented):
Changes look good, please see inline comments.
Do we have any tests that do actually verify the system's correct behaviour?

kripton (approved):
sry, meant to approve in my previous comment

configure.ac
peternewman: On the Mac system I tested on it just worked, so I assume so... but I don't know for certain TBH.

Seems to be sort of agreed here too:
https://stackoverflow.com/questions/20813028/endian-h-not-found-on-mac-osx

Issue #2035, opened 2026-03-21
Ola seems unmaintained: Can we help with maintenance?


aroffringa:
I've mailed to the ola mailing list about this before but didn't get a reply, so I'll try here... I'm noticing that Ola isn't getting maintained anymore. As a result, it is getting dropped from distributions for issues that seem relatively easy to fix, and compatibility with dependencies and newer compilers is becoming an issue (e.g. Debian doesn't ship with Ola because of [this issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1039294), which will hit Ubuntu probably also soon, whereas other distro's seem to have issues with an older protobuf dependency). Without packages, people must compile OLA from source, but there are various (though small) reasons that break compilation of the master branch of ola on a few recent distributions. If nothing is done, I suspect OLA will become irrelevant, which I would hate to see.

I've opened pull requests to fix some of the issues (#1989, #1890, #1889), but none of those have been merged into master. @peternewman hasn't answered in any of those pull requests why they aren't merged, which leaves me guessing. I would be happy to improve OLA, but with no way of merging things to master this seems a bit pointless.

My preferred solution would be to have multiple people with merge rights. I'd be happy to be one of them, if just to keep OLA work with a modern setup. Would this be an idea? I think that's better than making a fork, which makes it confusing what version is what, though if that's really necessary I would be happy to contribute to it. Are there more people that feel similar or with ideas?

DaAwesomeP:
@peternewman is there anything else we need for v0.11.0 other than: #1855, #2039, #2040? I see you're working on some PRs but would be great to hear your thoughts in this process.

DaAwesomeP:
> Sounds good, do you want to do that after 0.11.0, or open a PR to get it into there too? I guess if we do the latter we can confirm it covered everything/we didn't need to add new steps.

Yeah, let's do it after v0.11.0. It won't really affect anyone but us. For v0.12.0 (or later) I'd also like to make the CI do a bunch of the build steps for us (we're most of the way there with that already).

> There shouldn't be too many conflicts, but yes, this seems useful. I see you've already found my milestone (and given your rename it looks like we've settled on 0.12.0?) so I think as you say we should triage any open PRs (and probably issues too) into one of those or the 0.future/Future milestones. Assuming 0.11.0 gets released when it's milestone is empty (by either merging or moving outstanding stuff).

[...]

PR #2044, opened 2026-04-18
Revert "Replace a few more deprecated pre-C++11 styles"

peternewman:
Reverts OpenLightingProject/ola#1890

Please discuss in one place here:
https://github.com/OpenLightingProject/ola/issues/2035#issuecomment-4273996888

### DIFF
--- a/include/ola/ExportMap.h
+++ b/include/ola/ExportMap.h
@@ -77,7 +77,8 @@ class BaseVariable {
   std::string m_name;
 };
 
-struct VariableLessThan {
+struct VariableLessThan: public std::binary_function<BaseVariable*,
+                                                     BaseVariable*, bool> {
   bool operator()(BaseVariable *x, BaseVariable *y) {
     return x->Name() < y->Name();
   }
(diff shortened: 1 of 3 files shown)


DaAwesomeP (approved):

include/ola/ExportMap.h
```diff
@@ -77,7 +77,8 @@ class BaseVariable {
   std::string m_name;
 };
 
-struct VariableLessThan {
+struct VariableLessThan: public std::binary_function<BaseVariable*,
+                                                     BaseVariable*, bool> {
```
aroffringa, line 81: Technically you don't have to revert this line as the fix was just to remove the (deprecated) base class, which doesn't affect the code otherwise. However, it might be useful to keep the PR as one whole..

peternewman: Yeah, and I did find I only needed to undo a subset of the changes across both to fix it, but that feels like more pain than it's worth IMHO, the reverts always seem to end up a nightmare either way...

aroffringa: :+1:

PR #2052, opened 2026-05-30
Modify compiler flags to suppress variable length array error

peternewman:
Added -Wno-error=vla to COMMON_CXXFLAGS and COMMON_TESTING_FLAGS to prevent fatal warnings for variable-length arrays.

Plus other minor tidying.


DaAwesomeP:
In the future I think we should move to `std::vector` or similar. VLAs are not supported by some platforms (I think Windows is one of these?).

peternewman:
I've created #2053 to track that. TBH I'd assumed extracting from argc/argv would be a bit of a pain, but it doesn't look too bad thankfully.

plugins/spidmx/SPIDMXWidget.cpp
```diff
@@ -100,8 +100,9 @@ bool SPIDMXWidget::ReadWrite(uint8_t *tx_buf, uint8_t *rx_buf,
   struct spi_ioc_transfer tr;
   memset(&tr, 0, sizeof(spi_ioc_transfer));
 
-  tr.tx_buf        = (uint64_t) tx_buf;
-  tr.rx_buf        = (uint64_t) rx_buf;
+  // Changing these to a static_cast breaks the tests at least
+  tr.tx_buf        = (uint64_t) tx_buf;   // NOLINT(readability/casting)
+  tr.rx_buf        = (uint64_t) rx_buf;   // NOLINT(readability/casting)
```
aroffringa: I don't think a static_cast should even compile here; the C cast will behave like a reinterpret_cast in this case (but is more permissive). I think best would be to do an explicit reinterpret_cast:
```suggestion
  tr.tx_buf        = reinterpret_cast<uint64_t>(tx_buf);
  tr.rx_buf        = reinterpret_cast<uint64_t>(rx_buf);
```

aroffringa (commented):
Flag changes look fine. I added one suggestion for the cast for which you use NOLINT.

plugins/spidmx/SPIDMXWidget.cpp
peternewman: My memory is clearly rubbish:

1. There aren't any tests
2. It did indeed fail to compile

I tried your reinterpret_cast locally and it does at least compile, but there aren't any tests so I'll have to take it on faith...

@FloEdelmann if you've still got a test setup for this, please feel free to give it a go and let us know, but we'll merge as is.

peternewman: Applied, but just re-opened pending any possible testing.

aroffringa (approved):

PR #2077, opened 2026-09-18
Fix MemoryPreferences::GetMultipleValue

cho-m:
There is no guarantee on ordering of std::multimap::find https://en.cppreference.com/cpp/container/multimap/find:
> If there are several elements with the requested key in the container,
> any of them may be returned.

In recent LLVM / Apple Clang, the PreferencesTest fails as LLVM changed their implementation so the assumption of exact returned element is no longer true:
- https://github.com/llvm/llvm-project/issues/193311

Instead, switch to an interface that provides ordering guarantee.


peternewman:
Hi @cho-m ,

Thanks for this. Hopefully it won't make a difference, but please can you also try for you with the additional tests in #2080 .

peternewman (commented):
I was a bit worried about ordering requirements for lower_bound, but it seems like that doesn't apply to multimap just other containers. Should we also check for upper_bound to be more efficient (potentially).

If it's insertion order, then:
```
a=1
b=0
a=5
c=42
d=11
```

We could avoid inspecting c or d presumably if we stop at the upper_bound of a?

It would definitely be good to add a slightly more complex test to cover that if so.

